PR #2, opened 2019-05-12
Update requirements.txt to match setup

Bharat123rox:


seanlaw:
Have you had a chance to run the unit tests with older versions of Numba?

seanlaw:
Actually, on second thought, your PR is correct and the requirements.txt should really match the setup.py.

Thank you!

PR #4, opened 2019-05-13
Include LICENSE.txt in sdist and wheel

xhochy:
See https://github.com/conda-forge/staged-recipes/pull/8369


seanlaw:
@xhochy Thank you for helping with this! I'm still a beginner when it comes to packaging so I'll take all of the help that I can get.

PR #15, opened 2019-05-19
Fix #9 Add CI pipeline in Azure 

### LINKED ISSUE #9: Add Continuous Integration (CI)
Currently, there is no continuous integration for pull requests and everything is performed manually. Sadly, I have no experience here and would appreciate some help and/guidance.

- [ ] Use Azure Pipelines
- [ ] flake8
- [ ] mention code style requirement in CONTRIBUTING.md
- [ ] run unit tests (Dask is an additional dependency)
- [ ] run coverage tests (Dask is an additional dependency)
- [ ] Add CI badge(s) to README.rst (builds passing, test coverage, etc)

cuichenli:
Add Azure pipeline yaml file.
Some other changes:
- Change `test.sh` to `test.ps1` file, for better cross-platform performance. Since PowerShell core is baked in all azure agents. Otherwise, there may need duplicated files for the same logic while for different platform. 
- Change `requirements.txt` filename to `pipeline-requirements.txt`. Since the requirements for the package's installation has been well documented in the `setup.py` file. 

NOTE: [Azure pipeline](https://dev.azure.com/stumpy-dev/stumpy/_build?definitionId=1). May still need to connect this pipeline to the origin repo. It currently connects to my forked repo.


seanlaw:
Thank you. I will review this

seanlaw:
Again, thank you for all of your help. I am learning a lot!

I understand that we are using Azure/Microsoft so there is an obvious case for leveraging PowerShell. Unfortunately, we are primarily MacOSX/RHEL and so using `test.ps1` instead of a more generic `test.sh` will be, in my opinion, less maintainable in the long term. I want to make sure that our scripts are easily portable to any CI platform in the future and I want to avoid overcommitting to a given CI tool or painting ourselves into a corner. Again, just thinking out loud and reacting on (possibly irrational) instinct here as I don’t have much experience in this space and thinking through who will be maintaining this in the long run. Any scripts should able run in a local environment as well as on Azure and we should try to avoid Azure specific scripts (except for a yaml file) where possible. 

Can you elaborate on what you mean by “better cross-platform performance”? The code base is relatively small and compact and so I’d value ease of maintainability over marginal performance gains at this point and look updating it when the need arises. I don’t expect a ton of PRs to come through but I could be wrong.

Similarly, I don’t mind adding a yaml file for Azure pipelines but renaming of the requirements.txt file feels like the wrong approach since that is something that a lot of Python users would expect to be present. Based on [dask-ml](https://dask-ml.readthedocs.io/en/latest/contributing.html) and other Python projects, it looks like we could alleviate some of this tension between having a user installation (which should only have three dependencies) and a dev installation (with more dependencies like flake8 and black). If I understand correctly, this can be added to the setup.py.

Perhaps you have experience with this already but I’d like to look at a few popular Python packages that currently use Azure Pipelines to see how they’ve approached this for their projects. We probably aren’t the first ones to encounter this and our case might be simpler since it our code isn’t complex.

Additionally, maybe you’ve done this already, but if there a way to ensure that I am ultimately the owner of the stumpy_dev org on Azure even though you created it? I will need to relinquish this over to my employer.

cuichenli:
Since it is bedtime here, I will just make a quick response 😅 .

For the PowerShell scripts, my concern was just that if we are targeting running the pipeline in multiple platforms, we may need to maintain two copies of the `test` scripts. One is bash file, which can be used in MacOs and Linux and the other one will be a `ps1` file, which will be used in Windows system. This situation would cause some duplicated of works. Assume you want to make some changes in the test scripts, you will need to make changes in both files. Considered that all Azures's agents come with Powershell Core, I changed it to `ps1` file (In the meantime, `ps1` file can also run on MacOSX/RHEL system. But you may need to install Powershell Core in local machine.) Have said that, I personally don't think this is a perfect solution as well. You will need to install some other packages in your local machines. So there was another solution, which is to run the pytest from a python file. But here comes another issue, I not sure how to disable the JIT of numby from python scripts. That is why in the end I chose to change it to `ps1` file. (But actually, I would prefer to run it through a python file, if we can figure out how to disable the JIT issue.) 

For the `requirements` file, you are absolutely correct, we should maybe move it to `setup.py` file, I can handle this later.

For the Azure organization, I have already changed you to the owner of the organization. One email should have been sent to your Gmail and now I can see you are the owner of the org. You can check it by go to `organization settings` -> `overview` -> `Organization owner`. More detail about how to check it can be found [here](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/change-organization-ownership?view=azure-devops) 

(Hope I did not make too much grammar issues here.... 😅 )

seanlaw:
Awesome! It is good to know that we are on the same page.

For the tests, it looks like we actually call the non-JIT compiled version of a function by accessing its `.py_func` attribute according to this [SO answer](https://stackoverflow.com/a/53948648). So, we can avoid turning Numba on and off and, instead, we can add some extra tests that call the underlying non-JIT compiled versions of the  code. I am in a bunch of meetings over the next two days but I can work on adding those tests. Once done then we can figure out how to run the unit tests and coverage testing through a Python script instead. How does that sound?

Great. Let’s revert back to keeping a `requirements.txt` and then figure out how to have a different dev installation in the `setup.py`.

Okay, I see that I am the owner now. Thank you for doing that.

Your grammar is great!

cuichenli:
Was busy yesterday, I will also have a look at how to run the test.

cuichenli:
Just did some search and found that bash is also installed in Azure's windows agents. So I just reverted the `test.sh` back. The pipeline seems to be happy with that change. 

For the `requirements`, I moved the requirements into `setup.py`. I added `extra_vars` in the configuration. Hence, if we are going to install all these dependencies, we can use `pip install -e .[ci]`. Accordingly, I removed the first line of `setup.sh`, since there is no `requirement.txt` file now.

seanlaw:
This is great. I am traveling back home this evening but I can try to find some time to dive into this in more detail.

Currently, we reference the `requirements.txt` in the README and contrast it with `conda install numpy scipy numba`. Not sure if we should still keep the `requirements.txt` file since some might expect it. I have seen both having it and and not having it in many repos so I am cool with either in this case.

seanlaw:
Okay, I've reviewed the changes and I think it is ready to be accepted. Any last thoughts?

cuichenli:
No, should be good to go

seanlaw:
@cuichenli What do we need to do to connect the pipeline to the origin repo?

cuichenli:
@seanlaw I am investigating it.

cuichenli:
@seanlaw  Seems like I dont have access to this repo, so you may need to do it by yourself.

[build](https://dev.azure.com/stumpy-dev/stumpy/_build?definitionId=1) In this page, click the `New` button -> `New pipeline` -> `Github` -> select this repo -> `Existing pipeline YAML file` -> select the path to `azure-pipeline.yml`.

seanlaw:
@cuichenli I was able to follow your instructions and also successfully ran the pipeline on this PR.

Does this imply that everything is automated for when the next PR comes in? Or is there something extra that I'll need to do?

seanlaw:
Looks like the build badge needs to get updated. I tried updating the README but it doesn't seem like it's enough

seanlaw:
Nvm, I was able to change the badge it but I can't figure out how to get the linting/formatting/testing to automatically run when there is a new PR and then post an update to the PR.

cuichenli:
You don't need to do anything from my understanding. It will run the pipeline automatically when there is a new PR. And the PR's status can be viewed in the PR itself. [link](https://github.com/cuichenli/stumpy/pull/1) here is an example from my repo.

seanlaw:
ohh perfect. I see it now!

seanlaw:
@cuichenli I was starting to think whether or not `black` is actually doing anything in the pipeline. Unlike `flake8` which will error out, it seems that `black` will modify incorrectly formatted files in place before running the test suite. However, I just realized that this isn't actually useful from a maintenance perspective as the code that is merged may not actually adhere to `black` formatting. I actually want `black` to error out and notify the user if `black` has not been run on the code ahead of time and to force the user to run it on their end. 

Right now, `black` is run ahead of `flake8` within the pipeline and, to some extent, this is hiding the fact that the PR may contain incorrectly formatted code. So, while `flake8` is passing (thanks to `black`), we are still required to run `black` after the PR is merged. Am I missing something in my logic?

cuichenli:
Yes, you are right, the black will correct all the format issues in the project. I also thought about this issue but forgot to mention that.

seanlaw:
Alright, I did a little research and realized that `black` has a `--check` option that does the following:

    --check                         Don't write the files back, just return the
                                    status.  Return code 0 means nothing would
                                    change.  Return code 1 means some files
                                    would be reformatted.  Return code 123 means
                                    there was an internal error.

So, I've updated the pipeline to `black --check ./` instead. I think this should work.

Do you know if there's a way to have azure pipelines automatically post a comment in the PR if either `black` has a return code of 0?  I'd like it to remind the user in a PR comment to run `black` on their code if reformatting is necessary.

cuichenli:
It is possible but it would be a lot of works (you may need to set up a GitHub application). As a tradeoff, maybe you want to have a look at [GitHub Check API](https://github.blog/2018-05-07-introducing-checks-api/)? It won't comment directly but open a check panel in the PR.

Issue #7, opened 2019-05-18
Add additional documentation to readthedocs.io


seanlaw:
Currently, this repo is associated with [https://stumpy.readthedocs.io/en/latest/](https://stumpy.readthedocs.io/en/latest/) and all of the docstrings are in restructured text format. It would be a nice good first issue for anybody who'd like to contribute better documentation for the API. 

- [x] Convert docstrings to documentation
- [x] Remove install-from-source from the README and add to readthedocs
- [x] Embed the tutorial notebooks into the Sphinx documentation on ReadTheDocs. (Example: [freud](https://freud.readthedocs.io/en/stable/examples/module_intros/Density-RDF-AccumulateFluid.html))

- [x] Add badge to README.rst

PR #22, opened 2019-05-26
Add first draft of documentation.

bdice:
This PR tackles #7, with a rough draft of sphinx documentation for ReadTheDocs. Hope this is a helpful starting point!

- [x] Convert docstrings to documentation
- [x] Remove install-from-source from the README and add to readthedocs
- [x] Embed the tutorial notebooks into the Sphinx documentation on ReadTheDocs. (Example: [freud](https://freud.readthedocs.io/en/stable/examples/module_intros/Density-RDF-AccumulateFluid.html))
- [x] Add badge to README.rst

### Testing on ReadTheDocs
I'm not sure if RTD will require special configuration for the tools needed to build the docs, like `nbsphinx` and its dependencies. Maintainers would need to [add a version](https://readthedocs.org/projects/stumpy/versions/) for this branch to test it before merging, but I'm not sure if/how that works with external contributors' pull requests. If special configuration is needed, it might be easiest to add a custom conda environment (see [this list of ReadTheDocs settings](https://docs.readthedocs.io/en/stable/config-file/v2.html), also [freud has an example](https://github.com/glotzerlab/freud/blob/master/.readthedocs.yml) that works with nbsphinx).
- [ ] Test this PR with ReadTheDocs

### Tutorials
Every heading in each notebook is creating a separate table of contents entry, which makes it hard to tell where one notebook ends and the next begins.
- [ ] Rewrite the Jupyter notebook tutorials to have only one top-level heading with the notebook title. All other headings should be second-level or lower.

### Docstring summaries
Some core functions' docstrings first lines are long, which makes the summary at the top of the API page difficult to read.
- [ ] Rewrite first line of the core functions' docstrings to concisely describe the function

### Add requirements.txt back
- [ ] Add this file back since it was removed during CI integration


bdice:
Screenshot: 
image

seanlaw:
Thanks, @bdice! It might take me a few days but I will take a look

seanlaw:
It took a couple of tries but I got the docs to build and updated a bunch of things as per your suggestion. Thanks for your help @bdice!

seanlaw:
@bdice Any idea why the images are not rendering in the [notebooks](https://stumpy.readthedocs.io/en/latest/notebooks/Tutorial_1.html)?

bdice:
@seanlaw I am guessing it’s because the notebooks are in a higher directory - you might try creating a symlink to the tutorials directory that is inside the docs directory.

seanlaw:
> I am guessing it’s because the notebooks are in a higher directory - you might try creating a symlink to the tutorials directory that is inside the docs directory.

@bdice If I'm not mistaken, I think you had already done that. Currently, inside the`docs` directory, there is a symlink to the `notebooks` directory that contains the tutorials. Is this what you mean?

seanlaw:
Okay, I was able to build it locally and it looks like it doesn't like having the symlinks. I've cleaned things up now after building the docs locally (instead of on RTD). Hopefully, this fixes things

PR #41, opened 2019-06-20
Added TDA requirement for language in code of conduct

bfattori-TDA:
I've added the language the the code of conduct at the tail of the **Enforcement** section.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.idea
 *csv
 *checkout
 *pyc
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-/.idea
+.idea
 *csv
 *checkout
 *pyc
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -65,6 +65,9 @@ Project maintainers who do not follow or enforce the Code of Conduct in good
 faith may face temporary or permanent repercussions as determined by other
 members of the project's leadership.
 
+In addition to this Contributor Code of Conduct, TD Ameritrade Associates remain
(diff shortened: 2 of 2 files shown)


seanlaw:
@bfattori-TDA Can you please remove the /.idea from the .gitignore. I believe that this is specific to your setup and should not be in the .gitignore

bfattori-TDA:
I can change the commit message, however I would argue that anyone that is using IntelliJ Idea for development will appreciate that their IDE configuration isn't getting stored to the repo. I use IntelliJ exclusively because it supports so many languages. 

Honestly it doesn't hurt anyone to have it (you have one for vscode in there) and keeps the code free of influences from an IDE. Anyone with Idea that checked out the code would inherit the IDE preferences of whomever checked them in, and the next time that person that checked it in pulls the code, their settings will be overwritten.

Thoughts?

seanlaw:
The `.vscode` was added due to azure pipelines integration but, after checking other repos, I concede and accept adding `.idea` to the `.gitignore` file. Can you please remove the forward slash?

seanlaw:
Final question, was the additional coc line provided for you? I only ask because it sounds like "company policy" should be pluralized to "company policies". However, I'll leave that decision up to you. We can merge first and create a new issue/PR fix in the future if it needs changing.

bfattori-TDA:
I received an internal response to an email that contained that text. Whether it was reviewed by Legal and provided from them, I can't be certain. The final response came in a thread that included Legal and was not challenged by them so I assume it's what they intended. I will make sure and update if necessary with another PR.

bfattori-TDA:
I received a response: "either is fine" so if you'd like me to update to "policies" let me know and I will.

PR #70, opened 2019-10-02
MPDist Initial Code

Deepakgthomas:
Hello Sean,

Can you please have a quick look at my code for MPDist? For some reason, my test case fails when I'm trying to use Stumpy.

Thanks,
Deepak Thomas

### DIFF
--- a/MPDist Tutorial  (1).ipynb
+++ b/MPDist Tutorial  (1).ipynb
@@ -1,0 +1,199 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Tutorial 4 - MPdist (A distance measure for real-world time series datasets)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div style=\"text-align: justify\">\n",
+    "MPdist is a time series similarity measure developed by UCR researchers as an alternative to the widely used Eucledian distance and Dynamic Time Warping.<br><br>\n",
+    "It can be optimized for speed as well as be implemented on real-world datasets that contain \"spikes, dropouts, wandering baseline and missing values$^1$.\"<br><br>\n",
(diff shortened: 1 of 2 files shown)


seanlaw:
@Deepakgthomas Thank you for the PR. Yes, let me take a look!

Deepakgthomas:
There is a lot to be done from my end. At the moment, I'm just trying to
figure out the error.

seanlaw:
Okay, so there a few things/suggestions:

1. I know you are still working through things so I'm sure you'll get to it later but the code base and tutorials are using four spaces for each tab. I recommend throwing the code into a .py file and running it through `black` to check the code formatting. Again, let's get it working first (just writing this down here so I don't forget).

2. This step is essentially not needed:

```
#Data Preprocessing Step
def MPdist(t1,t2):
  #t1 -> Time series 1
  #t2 -> Time series 2
  #sub -> subsequence length
  if t1.ndim!=1 or t2.ndim!=1:
    raise Exception ('t1 & t2 should be univariate series with dimensions equalling 1')
  t1=t1.reshape(-1,1)
  t2=t2.reshape(-1,1)
  return t1,t2
```

Specifically, the reshaping is not necessary. `stumpy.stump` assumes that your one-dimensional input should be a row of data. So, this data input is expected as is:

```
t1=np.array([1,2,3.0,4,5,6,6])
t2=np.array([1,2.0,3,4,5,6,6])
stumpy.stump(t1, 5)
```

3. Finally, assuming that your input array are as they are shown above, you must define a window size, `m`, that is less than or equal to the length of your input array. Currently, you have `m=50` but your input array of, say, length 7.

```
def mat_join(t1,t2):
    m=50  # This is your problem! Try setting this to 5 
    #t1,t2=MPdist(t1,t2)
    AB=stumpy.stump(t1, m, T_B=t2, ignore_trivial=True)
    BA=stumpy.stump(t2, m, T_B=t1, ignore_trivial=True)
    PABBA=np.concatenate((AB, BA), axis=0)
    return t1,t2,PABBA
``` 

Let me know if this helps. The next error that you'll hit is that the variable, `thr`, is not defined.

PR #72, opened 2019-10-08
Updated MPdist docs

Deepakgthomas:
Hi Sean,

Looks like I hit upon another error. I would highly appreciate it if you could look into it. I will also require some help to generate the time series, Keogh and his team used to illustrate MPdist. It looks like they converted certain sentences to time series.

### DIFF
--- a/MPDist Tutorial  (1).ipynb
+++ b/MPDist Tutorial  (1).ipynb
@@ -1,0 +1,199 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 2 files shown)


seanlaw:
@Deepakgthomas I'm not seeing any contents in the last commit.

Regarding the data, I would like us to stay away from the names data as it is too hard for people to understand. There is no real way to calculate z-normalized Euclidean distance for words.

I believe that one would have to do a bit of digging in order to get to the original data. I'd look at the references

PR #73, opened 2019-10-08
Updated MPDist docs

Deepakgthomas:


seanlaw:
@Deepakgthomas Is there something specific that you wanted me to look at? The last few commits don't seem to have any changes in them and I see that there are 2 files that were added?

Deepakgthomas:
Got it. For the last 2 commits, I was focusing on the documentation portion. However, while calculating the MPdist, I run across a typing error in the last kernel. It seems like a Numba issue. I'd appreciate it if you could please look into it.

Thanks!

seanlaw:
@Deepakgthomas After removing the two `reshape` calls, I wasn't able to get the same error as you. Can you remove those two lines and see if it clears up the error?

Is there a reason why there are 2 notebooks? Also, can you move them to the `docs` directory instead of having them in the root directory please?

Deepakgthomas:
I think the uploaded the same notebook twice by mistake :)

Deepakgthomas:
Done!

PR #89, opened 2019-12-06
Fixed #81 Add Python 3.8 to Azure Pipelines

### LINKED ISSUE #81: Add Python 3.8 to Azure Pipelines
This is still not fully available yet but should be ready soon on Azure Pipelines

seanlaw:


seanlaw:
It looks like NumPy is not compatible with Python 3.8 yet and we need to wait for the next available (NumPy) release for things to work.

Issue #92, opened 2019-12-14
Add Type Hints or Annotations to Code


seanlaw:
This may be a good idea for future development as it brings some nice things to editors.

Here is a nice summary how-to:

https://inventwithpython.com/blog/2019/11/24/type-hints-for-busy-python-programmers/

cydave:
hello :wave: I could take a stab at this.
I'm not aquatinted with your code base and would rely on the doc strings if that's fine with you.

seanlaw:
@cydave That would be super helpful and we’d welcome a PR! The docstrings should be up to date and I am always here to answer any questions that you may have. I have no experience with Type Hints and whether there are:

1. Compatibility concerns (check our dependencies)
2. Things we should add for continuous integration with Azure Pipelines

Please fork this repo and we’ll go from there. Do you have experience with Type Hints? Is there anything we should discuss before proceeding? Is this project to small to warrant it? I am totally open to not adding it as well.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cydave:
Type hints convey information about what you'd expect the user (or developer) of your project to pass. So I'd argue adding type hints is a good thing. 

About the compatibility concerns:  
Type hints themselves were introduced in Python 3.6, but since your project clearly states older versions are not supported, this shouldn't be an issue. If you use the run of the mill python (without mypy installed) the type hints are simply ignored during execution. So it's up to the developer to decide if they want to *only* convey additional information about their code, or if they enforce type checking by having mypy installed.

Personally I've only ever used type hints to convey information for developers on what to pass in.
This sufficed for my use cases for quite a while now.

Important to note is that numpy does not have stable type hints. This means you can't enforce strict type checking (without excluding most of the dependencies you pull in) via mypy.

seanlaw:
This sounds great and I am on board! Thanks again for contributing to our library

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cydave:
This PR has moved to #96

PR #93, opened 2019-12-14
Add some type hints

cydave:
Work in progress for #92

### DIFF
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -2,13 +2,15 @@
 # Copyright 2019 TD Ameritrade. Released under the terms of the 3-Clause BSD license.  # noqa: E501
 # STUMPY is a trademark of TD Ameritrade IP Company, Inc. All rights reserved.
 
+from typing import Optional, Tuple, Iterator
+
 import numpy as np
 import scipy.stats
 from . import core
 import copy
 
 
-def _nnmark(I):
+def _nnmark(I: np.ndarray) -> np.ndarray:
     """
     Count the number of nearest neighbor overhead crossings or arcs.
 
@@ -20,7 +22,7 @@ def _nnmark(I):
     Returns
     -------
     nnmark : ndarray
-        Counts of nearest neigbor overheard crossings or arcs.
+        Counts of nearest neighbor overheard crossings or arcs.
 
     Notes
     -----
@@ -46,8 +48,12 @@ def _nnmark(I):
 
 
 def _iac(
-    width, bidirectional=True, n_iter=1000, n_samples=1000, seed=0
-):  # pragma: no cover
+    width: int,
+    bidirectional: bool = True,
+    n_iter: int = 1000,
+    n_samples: int = 1000,
+    seed: int = 0,
+) -> np.ndarray:  # pragma: no cover
     """
     Compute the bidirectional idealized arc curve (IAC). This is based
     on a beta distribution that is scaled with a width that is identical
@@ -72,6 +78,9 @@ def _iac(
 
     n_samples : int
         Number of distribution samples to draw during each iteration
+
+    seed : int
+        TODO
 
     Returns
     -------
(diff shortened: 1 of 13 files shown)


stumpy/floss.py
```diff
@@ -139,6 +155,9 @@ def _cac(I, L, bidirectional=True, excl_factor=5, custom_iac=None, seed=0):
         A custom idealized arc curve (IAC) that will used for correcting the
         arc curve
 
+    seed : int
+        TODO
```
cydave: I wasn't sure on what might be a good description for this.
But I added it just for you :)

stumpy/floss.py
```diff
@@ -73,6 +79,9 @@ def _iac(
     n_samples : int
         Number of distribution samples to draw during each iteration
 
+    seed : int
+        TODO
```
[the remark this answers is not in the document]
seanlaw: Great observation! I can add the description afterward

stumpy/floss.py
seanlaw: Great observation! I can add the description afterward

seanlaw:
@cydave Is this ready for review or should I hold off?

cydave:
This should be ready to go.

stumpy/stump.py
```diff
@@ -53,6 +65,8 @@ def _get_first_stump_profile(start, T_A, T_B, m, excl_zone, M_T, Σ_T, ignore_tr
     I : int64
         Matrix profile index for the window with index equal to `start`
     """
+    # TODO: return value of I is not simply an int64, looks like it's a tuple.
+    # it's either: (np.ndarray, int, int) or (np.ndarray, np.ndarray, np.ndarray)
```
seanlaw: You are right. The docstring is out of date. The return value of `I` should be a tuple of 3x `int64` (e.g., (3, 19, 20) or (10, -1, -1) ). The docstring should probably say something like:

```
I : Tuple[int64, int64, int64]
        Matrix profile index, left matrix profile index, and right matrix profile index for the window with index equal to `start`. The left and right matrix profile indices are automatically set to `-1` for self-joins (i.e., when `ignore_trivial` is set to `True`.
```

@cydave Would you mind updating the docstring (it's not obvious to me how the tuple should be represented there) and the type hint?

stumpy/gpu_stump.py
```diff
@@ -154,23 +155,23 @@ def _compute_and_update_PI_kernel(
...
+    Σ_T: np.ndarray,
+    QT: np.ndarray,
+    QT_first: np.ndarray,
+    μ_Q: np.ndarray,
+    σ_Q: np.ndarray,
+    k: int,
+    ignore_trivial: bool = True,
+    range_start: int = 1,
+    threads_per_block: int = THREADS_PER_BLOCK,
+    device_id: int = 0,
```
seanlaw: I have some yet-to-be-committed updates where `device_id` can either be a single `int` (as it is shown here) or a list of `int`. Can you please tell me what this line should be if `device_id` can be either a single `int` or a list of `int`?  This way, I can update this on my own after your PR.

cydave: device_id: Union[int, List[int]] = 0
I think. Union => https://mypy.readthedocs.io/en/latest/kinds_of_types.html#union-types

stumpy/stump.py
cydave: This looks to me like:
```
Tuple[np.int64, np.int64, np.int64] (if you require a numpy.int64, if you don't care too much about the specific type you could use Tuple[int, int, int]  :)

Try to find the actual type of the variables you're using and use those as the types you expect them to be in.

cydave: Although on second thought (after searching for a bit) the type might not need to be prefixed with `np.`. So the one you came up with appears to be the correct way to go about it.

seanlaw: It’s actually just a single element so I think int is perfectly fine

cydave: @seanlaw  I think the change at https://github.com/TDAmeritrade/stumpy/pull/93/commits/c37e1d7877e3e78fc6c47a88606b159c1e1fc606  should do it.

seanlaw:
It looks like the entire `stumpy/chains.py` file was deleted?

cydave:
@seanlaw  sorry about the mess of commits.

seanlaw:
@cydave This may be too much to ask so feel free to push back and say, “no” but would you be open to just closing this PR and then taking the final version of your branch and creating a new PR with a single commit (with type hints and typos)?  As you pointed out, it would greatly help to reduce the number of commits to the main branch and keep the history more compact and searchable.

PR #96, opened 2019-12-15
Add typehints and update docstrings

cydave:
> @cydave This may be too much to ask so feel free to push back and say, “no” but would you be open to just closing this PR and then taking the final version of your branch and creating a new PR with a single commit (with type hints and typos)? As you pointed out, it would greatly help to reduce the number of commits to the main branch and keep the history more compact and searchable.

@seanlaw  As requested, a new pull request :+1:

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -2,6 +2,8 @@
 # Copyright 2019 TD Ameritrade. Released under the terms of the 3-Clause BSD license.  # noqa: E501
 # STUMPY is a trademark of TD Ameritrade IP Company, Inc. All rights reserved.
 
+from typing import Optional, Tuple, Any
+
 import numpy as np
 import scipy.signal
 
@@ -11,7 +13,7 @@ except ImportError:
     pass
 
 
-def driver_not_found(*args, **kwargs):  # pragma: no cover
+def driver_not_found(*args: Any, **kwargs: Any) -> None:  # pragma: no cover
     """
     Helper function to raise CudaSupportError driver not found error
     """
@@ -19,7 +21,7 @@ def driver_not_found(*args, **kwargs):  # pragma: no cover
     _raise_driver_not_found()
 
 
-def get_pkg_name():  # pragma: no cover
+def get_pkg_name() -> str:  # pragma: no cover
     """
     Return package name
     """
@@ -27,7 +29,7 @@ def get_pkg_name():  # pragma: no cover
     return __name__.split(".")[0]
 
 
-def rolling_window(a, window):
+def rolling_window(a: np.ndarray, window: int) -> np.ndarray:
     """
     Use strides to generate rolling/sliding windows for a numpy array
 
@@ -51,7 +53,7 @@ def rolling_window(a, window):
     return np.lib.stride_tricks.as_strided(a, shape=shape, strides=strides)
 
 
-def z_norm(a, axis=0):
+def z_norm(a: np.ndarray, axis: int = 0) -> np.ndarray:
     """
     Calculate the z-normalized input array `a` by subtracting the mean and
     dividing by the standard deviation along a given axis.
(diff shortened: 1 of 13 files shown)


seanlaw:
@cydave Thank you for doing this (both the new PR and taking the time to contribute). I need to take the kids to swim class but I will review this ASAP so please don't think that I am ignoring you! :)

P.S. I rarely get to know more about people who cross paths with STUMPY and I am super curious about:

1. How did you came across this library?
2. Are you currently using it for interesting use cases?
3. Do you have a Twitter handle (so I can thank you publicly for our future release)?

cydave:
1. I searched the internet for issues that need help, and yours looked quite easy for me to help you out with.
2. I'm not using it personally.
3. No need to thank me on other places, you're welcome :+1:

stumpy/floss.py
```diff
@@ -2,13 +2,16 @@
...
 # STUMPY is a trademark of TD Ameritrade IP Company, Inc. All rights reserved.
 
+import copy
+from typing import Optional, Tuple, Iterator
+
 import numpy as np
 import scipy.stats
+
 from . import core
-import copy
```
seanlaw: I think this `import copy` is required since we need to perform a `copy.deepcopy` later in the file. I'm kind of surprised that the tests didn't fail :(

seanlaw: Nevermind, I mis-read this. I see that you are moving it higher up in the imports

stumpy/core.py
```diff
@@ -107,7 +109,7 @@ def check_dtype(a, dtype=np.floating):  # pragma: no cover
     return True
 
 
-def transpose_dataframe(a):  # pragma: no cover
+def transpose_dataframe(a: np.ndarray) -> np.ndarray:  # pragma: no cover
```
blester125: This type annotation seems to not capture the full range of inputs, It seems like it should be more like

```python
def transpose_dataframe(a: Union[np.ndarray, "DataFrame", "Series"]) -> Union[np.ndarray, "DataFrame", "Series"]:
```

I don't know if that will cause mypy problems though

seanlaw: I guess we really only want a Pandas DataFrame (not Series or ndarray) as input to this function so the docstring needs to be changed to a Pandas DataFrame.

seanlaw: All other inputs except for DataFrame is essentially a no-op

seanlaw: Hmm, perhaps this should be `Any` type

seanlaw: @cydave Feel free to chime in. 

@blester Are there other type hints that aren't correct or could be improved? I'm a total novice when it comes to this so I'd appreciate your input too!

cydave: Because I tried strictly following the docs that were present and don't know the code base too well I'm not aware of all the inputs. My advice would be to check all references to this function and then decide upon a type that is acceptable for the invocation of this function.

I can take a look and try and see if @blester125's suggestion would work without complains from mypy.

seanlaw: Hmmm, I'm struggling a bit here. We certainly do not want to import Pandas as it should not and is not dependency of STUMPY. I think I'll just merge this PR as is and deal with improving this later. Thank you for your input!

seanlaw:
@cydave Thank you for your contributions! I appreciate your taking the time to work on this

stumpy/core.py
blester125: Right, I knew pandas wasn't a dependency which is why i was suggesting to use the string names instead of the actual objects (i still don't know if that plays nicely with mypy but it is how you type a method on a class that takes a member of that class so it seems like it is supported) that way STUMPY wouldn't have a dep on pandas but someone who is typechecking their code that passes in a data frame wouldn't get errors.

cydave: mypy must be able to resolve the type you define it. When you wrap it in quotes it still tries to resolve the type after having imported everything. This means mypy would complain:

```
$ mypy stumpy       
stumpy/core.py:112: error: Name 'DataFrame' is not defined
stumpy/core.py:112: error: Name 'Series' is not defined
Found 2 errors in 1 file (checked 11 source files)
```

An ugly workaround would be to wrap the pandas import like so:
```
if False:
    import pandas as pd
```

This way you can use the type annotation but do not depend on pandas directly (as it will never be imported).

Issue #115, opened 2020-01-14
Handling NaN and inf Values for 1-D Matrix Profile Algorithms


mexxexx:
When implementing the SDTS algorithm from the Matrix Profile IV Paper, the matrix profile of a time series containing inf values has to be computed. Such a case is not yet handled in stumpy.

In line with the paper I would propose to ignore all subsequences, where at least one value is nan or inf. The tsmp R project does this, see [here on line 98] (https://github.com/cran/tsmp/blob/master/R/stomp.R). Such a change would require (at least, as for as I have seen) three changes: 
1. Changing the calculate_mean_std function to use the np.strided rolling window approach. This should be done anyways, because then the numpy std and mean functions can be used that should be more numerically stable and optimized.
2. All algorithms that calculate the MP need to keep track of subsequences that have to be skipped and need to replace nan and inf values in the time series (see the implementation in tsmp)
3. Adding a test that contains such values. We could i.e. use the time series [1, 0, 0, inf, 1, 0, 0] with m=3 that should result in the profile [0, inf, inf, inf, 0]

mexxexx:
I implemented this functionality and some tests for STAMP and STOMP and am working on STUMP right now. 
@seanlaw Are you interested in this functionality? Then I could issue a pull request when I'm a bit more advanced

seanlaw:
@mexxexx Yes, we would be interested in taking a look. As you pointed out, the problem is that STUMPY doesn't work when NaN/inf exists within the time series. My hesitation is largely attributed to the fact that I haven't thought about this much but I'm open to it.

> Changing the calculate_mean_std function to use the np.strided rolling window approach. This should be done anyways, because then the numpy std and mean functions can be used that should be more numerically stable and optimized.

Would you mind elaborating a little more on this? We are using stride tricks elsewhere within the code (say, `core.rolling_window`) and I'd like to confirm this statement regarding "numerically stable and optimized". Since the standard deviation/mean are only computed once before a matrix profile is computed, I'm not sure the juice is really worth the squeeze here. Perhaps, the requirement or use case is different for SDTS? Or are you saying that the NumPy functions will handle NaN/infs?

> All algorithms that calculate the MP need to keep track of subsequences that have to be skipped and need to replace nan and inf values in the time series (see the implementation in tsmp)

[...]

PR #122, opened 2020-01-20
Fixed Issue #115

mexxexx:
This pull request fixes issue #115. 

The fix itself is simple. We set the mean value of every sub sequence that contains at least one inf/NaN value to infinity. Afterwards, we replace the values in the sequences with 0 so to not get any computational problems. The marked sub sequences are then considered at two points: When calculating the distance profile of Q with T, we have to set the distance from Q to every illegal sequence in T to infinity, and if Q itself is an illegal sequence, the whole distance profile needs to be set to infinity.


seanlaw:
@mexxexx It looks like the CI is failing on the code formatting. Can you please install and run `black` (for code formatting) and `flake8` (for linting). To run the tests locally, you’ll need to install `pytest-cov` and then you should be able to run the `test.sh` script. This is all done in the CI but handling this on your end will help eliminate minor commits. Let me know if you need any help.

mexxexx:
@seanlaw Sorry, I didn't know I had to run this script. It passes everything now on my machine, with 100% coverage. What is the procedure now?

seanlaw:
Please push the newly formatted (by `black`) code and the tests will be run again above. Right now, it says that all checks have failed because the code is not currently formatted correctly according to the `black` specifications. If you click on “Details” above then you can dig into the error that is causing the check/tests to fail. Let me know if this makes sense or else I am happy to step you through it.

I should add a PR template

mexxexx:
It is still failing because stump.py is not formatted correctly, but if I run the test.sh file on my local machine everything is working fine and black says that no file would be reformatted. Is there anything I am overlooking?
I can gladly open a clean pull request with only one commit once everything works.

seanlaw:
@mexxexx Thank you for the quick turnaround! Indeed, I see the same thing as you (it is fine locally but failing on Azure Pipelines). I've added an extra `--diff` flag to `black` and, hopefully, it'll give us a bit more information. But this seems strange indeed.

Yes, let's get all of the kinks ironed out fully in this PR and then we can move it over to a new PR as a single commit. I've made a couple of minor changes in the code base and so you'll need to merge with the latest version of the code anyways. But let's wait until we figure things out here first.

seanlaw:
@mexxexx Okay, so `black` seems to be inconsistent locally vs on Azure Pipelines. On Azure, even though the `black` version is the same as my local version (19.3b0 on both Mac OSX and Ubuntu), it's complaining about [these lines of code](https://dev.azure.com/stumpy-dev/stumpy/_build/results?buildId=388&view=logs&j=bc65f802-b6e4-57a5-b2d4-87d04d7c1e8a&t=ec05a641-a7e1-516a-589c-80bbb52671b8&l=34) in `stump.py`.

If you wouldn't mind, can you change the following:

```
T_B[
        np.isnan(T_B)
    ] = (
        0
    )  # Only after calculating the first profile we can remove all nan values from T_B
```

into:

```
T_B[
        np.isnan(T_B)
    ] = 0  # Remove all nan values from T_B only after first profile is calculated
```

Let's see if this solves our problem.

seanlaw:
@mexxexx Yes, we've made some progress! It is now failing on one of the [test cases that you've added](https://dev.azure.com/stumpy-dev/stumpy/_build/results?buildId=389&view=logs&j=92a4a009-755e-5cdc-87aa-eb6d35b44c3c&t=f482398c-bb8d-5e44-26d0-845cce732938). Where the results of `naive_mass` doesn't quite match the output of `stump`

mexxexx:
I moved the comment up a line and now at least black is reporting no errors. Now, the tests are failing, what I don't understand, because locally everything seems fine. The fail occurred in one deterministic case, so I really don't understand what is going on... I will try printing out the full results

mexxexx:
Okay, apparently something goes wrong when calculating the matrix profile index. I will have a look if I can find out what, but I'm still confused that it isn't reproducible on my machine...

seanlaw:
@mexxexx I've encountered something like this before and we **_might_** be running up against numerical precision. By the way, this is a fantastic and simple test case that you've put together! Please do not be tempted to change it so that it matches the `stump` function output (even I've been guilty  of being tempted before). I recommend going up a level and printing out the full unmodified distance profile (from `naive_mass` and from `stump`).

mexxexx:
@seanlaw Yes that might be the problem. I printed out the distance profile in stump, first without and then with the exclusion zone applied, and afterward the Matrix Profile Index.

We see the following output 
```
1 [           inf 1.33226763e-15 9.00000000e+00 9.00000000e+00
 1.33226763e-15]
1 [           inf            inf 9.00000000e+00 9.00000000e+00
 1.33226763e-15]
1 4
2 [           inf 9.00000000e+00 1.33226763e-15 9.00000000e+00
 9.00000000e+00]
2 [inf inf inf  9.  9.]
2 3
3 [           inf 9.00000000e+00 9.00000000e+00 1.33226763e-15
 9.00000000e+00]
3 [inf  9. inf inf  9.]
3 4
4 [           inf 1.33226763e-15 9.00000000e+00 9.00000000e+00
 1.33226763e-15]
4 [           inf 1.33226763e-15 9.00000000e+00            inf
            inf]
4 1
```
The weird part is for index i=3 (the lines leading with a three). After applying the exclusion zone we get the distance profile [inf  9. inf inf  9.]. Now the index is the argmin of this array. According to numpy specifications, when having duplicates, it is always the first occurrence, so the result should be 1. This would be in line with the naive_mass result. 

Do you have any ideas why it is 4 and what to do?

seanlaw:
@mexxexx In that particular case (i=3), can you try adding a print statement like:

```
if i == 3:  # you may need to adjust this for zero-based indexing
    if D[1] < D[4]:
        print("Index 1 is smaller")
    elif D[4] < D[1]:
        print("index 4 is smaller")
    else:
        print("Both numbers are identical!")
```

mexxexx:
Good idea! Index 4 actually is smaller... Do you have any idea how to fix this?

seanlaw:
So, for these trickier cases where we have multiple exact matches (which is somewhat rare in the real world) and numerical precision is the culprit, we don't really have much leverage aside from:

1.  looking for approaches that improve the precision.
2. making sure that we perform these types of checks (as we've done above) and then only compare the matrix profile and ignore the indices

seanlaw:
Okay, so I thought about this a little more. What if we choose a different input time series where only one exact match can be found?

stumpy/core.py
```diff
@@ -471,19 +469,27 @@ def mass(Q, T, M_T=None, Σ_T=None):
 
     Q = np.asarray(Q)
     check_dtype(Q)
-    check_nan(Q)
```
seanlaw: Should we remove the `check_nan` function if it isn't being used? I guess it is still being used in `gpu_stump` so, perhaps, we should create a Github issue to add the `np.nan`/`np.inf` stuff to `gpu_stump` and then delete the function afterward

stumpy/core.py
```diff
@@ -471,19 +469,27 @@ def mass(Q, T, M_T=None, Σ_T=None):
...
 
     if Q.ndim != 1:  # pragma: no cover
         raise ValueError(f"Q is {Q.ndim}-dimensional and must be 1-dimensional. ")
+    m = Q.shape[0]
```
seanlaw: I see, you've moved this line up higher. Cool

stumpy/core.py
```diff
@@ -471,19 +469,27 @@ def mass(Q, T, M_T=None, Σ_T=None):
...
     check_dtype(T)
     check_nan(T)
 
     if T.ndim != 1:  # pragma: no cover
         raise ValueError(f"T is {T.ndim}-dimensional and must be 1-dimensional. ")
+    n = T.shape[0]
+
+    if np.any(np.isnan(Q)):
+        distance_profile = np.empty(n - m + 1)
+        distance_profile[
```
seanlaw: We should squash these three lines into a single line and remove the comment. I think this is pretty self-explanatory and is consistent with the rest of the code base

stumpy/core.py
```diff
@@ -471,19 +469,27 @@ def mass(Q, T, M_T=None, Σ_T=None):
...
+            :
+        ] = np.inf  # Setting the Profile to inf if it cannot be calculated
+        return distance_profile
```
seanlaw: This is a personal preference but I tend to like having a single return statement within a function (especially if the function is long) rather than having it short circuit and return halfway through. It could simply be an `if`/`else` statement where we store the output into `distance_profile` in both cases (i.e., set `distance_profile = calculate_distance_profile(m, QT, μ_Q.item(0), σ_Q.item(0), M_T, Σ_T)`) and return `distance_profile` only once at the end. The only thing we are really saving in your version is the intermediate storage of `distance_profile` but I think "future Sean" would thank us for being more explicit. :) 

Again, this is just a personal preference

stumpy/stamp.py
```diff
@@ -59,20 +61,22 @@ def mass(Q, T, M_T, Σ_T, trivial_idx=None, excl_zone=0, left=False, right=False
         if D[:trivial_idx].size:
             IL = np.argmin(D[:trivial_idx])
             PL = D[IL]
-        if zone_start <= IL < zone_stop:
-            IL = -1
+            if PL == np.inf or zone_start <= IL < zone_stop:
```
seanlaw: Maybe this is just how it is showing up on Github but this looks over-indented.

stumpy/stamp.py
```diff
@@ -59,20 +61,22 @@ def mass(Q, T, M_T, Σ_T, trivial_idx=None, excl_zone=0, left=False, right=False
...
+                IL = -1
 
         IR = -1
         PR = np.inf
-        if D[trivial_idx:].size:
-            IR = trivial_idx + np.argmin(D[trivial_idx:])
+        if D[trivial_idx + 1 :].size:
```
seanlaw: What is this "+1" for? Same question below

stumpy/stamp.py
```diff
@@ -59,20 +61,22 @@ def mass(Q, T, M_T, Σ_T, trivial_idx=None, excl_zone=0, left=False, right=False
...
+            IR = trivial_idx + 1 + np.argmin(D[trivial_idx + 1 :])
             PR = D[IR]
-        if zone_start <= IR < zone_stop:
-            IR = -1
+            if PR == np.inf or zone_start <= IR < zone_stop:
```
seanlaw: Over-indented?

stumpy/stomp.py
```diff
@@ -101,36 +116,49 @@ def stomp(T_A, m, T_B=None, ignore_trivial=True):
...
-            m, QT, μ_Q[i].item(0), σ_Q[i].item(0), M_T, Σ_T
-        )
+
+        D = core.calculate_distance_profile(m, QT, μ_Q[i], σ_Q[i], M_T, Σ_T)
         if ignore_trivial:
             zone_start = max(0, i - excl_zone)
             zone_stop = min(k, i + excl_zone)
             D[zone_start:zone_stop] = np.inf
+
+        # If the mean of the query is nan that mean this subsequence should be ignored
```
seanlaw: Change this comment to "Ignore distance profile if query mean is np.nan"

stumpy/stump.py
```diff
@@ -154,11 +154,15 @@ def _calculate_squared_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
     denom = m * σ_Q * Σ_T
     denom[denom == 0] = 1e-10  # Avoid divide by zero
     D_squared = np.abs(2 * m * (1.0 - (QT - m * μ_Q * M_T) / denom))
+    # In fastmath mode checking for np.nan is not possible, so we use infs instead
+    D_squared_inf = np.isinf(D_squared)
```
seanlaw: Change to `inf_mask` please

stumpy/stump.py
```diff
@@ -154,11 +154,15 @@ def _calculate_squared_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
...
+
     threshold = 1e-7
     if σ_Q < threshold:  # pragma: no cover
         D_squared[:] = m
     D_squared[Σ_T < threshold] = m
     D_squared[(Σ_T < threshold) & (σ_Q < threshold)] = 0
+    D_squared[D_squared_inf] = np.inf
```
seanlaw: Change to `inf_mask` please

stumpy/stump.py
```diff
@@ -154,11 +154,15 @@ def _calculate_squared_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
     denom = m * σ_Q * Σ_T
     denom[denom == 0] = 1e-10  # Avoid divide by zero
     D_squared = np.abs(2 * m * (1.0 - (QT - m * μ_Q * M_T) / denom))
+    # In fastmath mode checking for np.nan is not possible, so we use infs instead
```
seanlaw: Can you remind me where else we are checking for `np.nan`? Should this comment be here?

stumpy/stump.py
```diff
@@ -312,28 +316,47 @@ def _stump(
             QT_odd[0] = QT_first[i]
             D = _calculate_squared_distance_profile(m, QT_odd, μ_Q[i], σ_Q[i], M_T, Σ_T)
 
+        # If the mean of the query is inf that mean this subsequence should be ignored
```
seanlaw: Please update this to be consistent with the similar suggestion above.

stumpy/stump.py
```diff
@@ -312,28 +316,47 @@ def _stump(
...
+        if np.isinf(μ_Q[i]):
+            D[:] = np.inf
+        print(i, D)
         if ignore_trivial:
             zone_start = max(0, i - excl_zone)
             zone_stop = min(k, i + excl_zone)
             D[zone_start:zone_stop] = np.inf
+        print(i, D)
```
seanlaw: Please remove

stumpy/stump.py
```diff
@@ -312,28 +316,47 @@ def _stump(
             QT_odd[0] = QT_first[i]
             D = _calculate_squared_distance_profile(m, QT_odd, μ_Q[i], σ_Q[i], M_T, Σ_T)
 
+        # If the mean of the query is inf that mean this subsequence should be ignored
+        if np.isinf(μ_Q[i]):
+            D[:] = np.inf
+        print(i, D)
```
seanlaw: Please remove

stumpy/stump.py
```diff
@@ -312,28 +316,47 @@ def _stump(
...
         if ignore_trivial:
             zone_start = max(0, i - excl_zone)
             zone_stop = min(k, i + excl_zone)
             D[zone_start:zone_stop] = np.inf
+        print(i, D)
 
         I = np.argmin(D)
+        print(i, I)
```
seanlaw: Please remove

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
     """
 
     T_A = np.asarray(T_A)
-    core.check_dtype(T_A)
-    core.check_nan(T_A)
-
     if T_A.ndim != 1:  # pragma: no cover
-        raise ValueError(
-            f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. "
-            "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
```
seanlaw: Is there a reason why this line is removed?

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
-        )
+        raise ValueError(f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. ")
+    n = T_A.shape[0]
+
+    T_A = T_A.copy()
+    # Treat inf values the same as nan values,
```
seanlaw: Remove comment

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
+    # because z normalization is undefined in this case
```
seanlaw: Remove comment

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
-
     if T_A.ndim != 1:  # pragma: no cover
-        raise ValueError(
-            f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. "
-            "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
-        )
+        raise ValueError(f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. ")
+    n = T_A.shape[0]
+
+    T_A = T_A.copy()
```
seanlaw: Is this copy necessary? Is it to avoid altering the original time series outside of this function?

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
+    T_A[np.isinf(T_A)] = np.nan
+    core.check_dtype(T_A)
 
-    if T_B is None:  # Self join!
+    if T_B is None:
         T_B = T_A
         ignore_trivial = True
+
     T_B = np.asarray(T_B)
+    T_B = T_B.copy()
```
seanlaw: Is this copy necessary? Is it to avoid altering the original time series outside of this function?

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
+    # Treat inf values the same as nan values,
```
seanlaw: Remove comment

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
+    # because z normalization is undefined in this case
```
seanlaw: Remove comment

stumpy/stump.py
```diff
@@ -404,26 +427,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
+    T_B[np.isinf(T_B)] = np.nan
     core.check_dtype(T_B)
-    core.check_nan(T_B)
```
seanlaw: Why are we deleting this dimensionality check?

stumpy/stump.py
```diff
@@ -454,6 +479,9 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
         start, T_A, T_B, m, excl_zone, M_T, Σ_T, ignore_trivial
     )
 
+    # Only after calculating the first profile we can remove all nan values from T_B
```
seanlaw: Move this comment to the next line

stumpy/stump.py
```diff
@@ -454,6 +479,9 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
+    T_B[np.isnan(T_B)] = 0
```
seanlaw: Append the following comment to the end of the line `  # Remove all nan values from T_B only after first profile is calculated`

stumpy/stumped.py
```diff
@@ -84,27 +84,34 @@ def stumped(dask_client, T_A, m, T_B=None, ignore_trivial=True):
...
-
     if T_A.ndim != 1:  # pragma: no cover
         raise ValueError(
             f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
+    n = T_A.shape[0]
+
+    T_A = T_A.copy()
+    # Treat inf values the same as nan values,
```
seanlaw: Remove comment

stumpy/stumped.py
```diff
@@ -84,27 +84,34 @@ def stumped(dask_client, T_A, m, T_B=None, ignore_trivial=True):
...
+    # because z normalization is undefined in this case
```
seanlaw: Remove comment

stumpy/stumped.py
```diff
@@ -84,27 +84,34 @@ def stumped(dask_client, T_A, m, T_B=None, ignore_trivial=True):
...
+        ignore_trivial = True
 
+    T_B = np.asarray(T_B)
     if T_B.ndim != 1:  # pragma: no cover
         raise ValueError(
             f"T_B is {T_B.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
+    T_B = T_B.copy()
+    # Treat inf values the same as nan values,
```
seanlaw: Remove comment

stumpy/stumped.py
```diff
@@ -84,27 +84,34 @@ def stumped(dask_client, T_A, m, T_B=None, ignore_trivial=True):
...
+    # because z normalization is undefined in this case
```
seanlaw: Remove comment

stumpy/stumped.py
```diff
@@ -124,13 +131,27 @@ def stumped(dask_client, T_A, m, T_B=None, ignore_trivial=True):
     M_T, Σ_T = core.compute_mean_std(T_A, m)
     μ_Q, σ_Q = core.compute_mean_std(T_B, m)
 
+    T_A[np.isnan(T_A)] = 0
```
seanlaw: Remind me, what is the logic here again?

stumpy/stump.py
```diff
@@ -443,6 +466,8 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
     M_T, Σ_T = core.compute_mean_std(T_A, m)
     μ_Q, σ_Q = core.compute_mean_std(T_B, m)
 
+    T_A[np.isnan(T_A)] = 0
```
seanlaw: Remind me, what is the logic here again?

stumpy/stumped.py
```diff
@@ -124,13 +131,27 @@ def stumped(dask_client, T_A, m, T_B=None, ignore_trivial=True):
...
+    QT_futures = []
+    QT_first_futures = []
+    for i, start in enumerate(range(0, l, step)):
+        profile[start], indices[start, :] = _get_first_stump_profile(
+            start, T_A, T_B, m, excl_zone, M_T, Σ_T, ignore_trivial
+        )
+
+    T_B[
+        np.isnan(T_B)
+    ] = 0  # After having calculated all first profiles, we can remove nans from T_B
```
seanlaw: Replace comment with `  # Remove all nan values from T_B only after first profile is calculated`

tests/test_stamp.py
```diff
@@ -40,7 +40,7 @@ def test_stamp_self_join(T_A, T_B):
     zone = int(np.ceil(m / 2))
     left = np.array(
         [
-            naive_mass(Q, T_B, m, i, zone)
+            utils.naive_mass(Q, T_B, m, i, zone, ignore_trivial=True)[0:2]
```
seanlaw: I don't like the indexing `[0:2]` here. `left` should be raw and untouched. Move the indexing to the assertion so that it is more explicit/obvious as to what we are comparing. Since most other places are comparing all of matrix profille and indices, it would be more consistent to have it in the assertion

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
 def test_stamp_A_B_join(T_A, T_B):
     m = 3
     left = np.array(
-        [naive_mass(Q, T_A, m) for Q in core.rolling_window(T_B, m)], dtype=object
+        [utils.naive_mass(Q, T_A, m)[0:2] for Q in core.rolling_window(T_B, m)],
```
seanlaw: Please move index `[0:2]` to assertion

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
...
+        dtype=object,
     )
     right = stamp.stamp(T_A, T_B, m)
     utils.replace_inf(left)
     utils.replace_inf(right)
     npt.assert_almost_equal(left, right)
+
+
+def test_stamp_nan_selfjoin_beginning():
```
seanlaw: There should be an underscore `_` between `self` and `join`

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
...
+    m = 3
+    T = np.array([np.nan, 1, 0, 0, 1, 0, 0])
+
+    zone = int(np.ceil(m / 2))
+    left = np.array(
+        [
+            utils.naive_mass(Q, T, m, i, zone, ignore_trivial=True)[0:2]
```
seanlaw: Please move index `[0:2]` to assertion

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
...
+    )
+
+    right = stamp.stamp(T, T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+def test_stamp_inf_selfjoin_beginning():
```
seanlaw: There should be an underscore `_` between `self` and `join`

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
...
+    m = 3
+    T = np.array([np.inf, 1, 0, 0, 1, 0, 0])
+
+    zone = int(np.ceil(m / 2))
+    left = np.array(
+        [
+            utils.naive_mass(Q, T, m, i, zone, ignore_trivial=True)[0:2]
```
seanlaw: Please move index `[0:2]` to assertion

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
...
+def test_stamp_nan_inf_selfjoin(T_A, T_B):
+    m = 3
+    T_B_nan_inf = np.random.uniform(size=len(T_B))
+    T_B[T_B_nan_inf > 0.90] = np.nan
+    T_B[T_B_nan_inf > 0.95] = np.inf
+
+    zone = int(np.ceil(m / 2))
+    left = np.array(
+        [
+            utils.naive_mass(Q, T_B, m, i, zone, ignore_trivial=True)[0:2]
```
seanlaw: Please move index `[0:2]` to assertion

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
     utils.replace_inf(left)
     utils.replace_inf(right)
     npt.assert_almost_equal(left, right)
+
+
+def test_stomp_nan_selfjoin_beginning():
```
seanlaw: Need to add underscore between self and join

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
+    )
+
+    right = stomp(T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+def test_stomp_inf_selfjoin_beginning():
```
seanlaw: Need to add underscore between self and join

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
+
+    right = stomp(T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stomp_nan_inf_selfjoin(T_A, T_B):
```
seanlaw: Need to add underscore to self and join

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
...
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stamp_nan_inf_selfjoin(T_A, T_B):
+    m = 3
+    T_B_nan_inf = np.random.uniform(size=len(T_B))
+    T_B[T_B_nan_inf > 0.90] = np.nan
```
seanlaw: It is possible that no values are above 0.9 and so you may not have anything set to `np.nan`. This is more likely when the sequences are short.

tests/test_stamp.py
```diff
@@ -55,9 +55,95 @@ def test_stamp_self_join(T_A, T_B):
...
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stamp_nan_inf_A_B_join(T_A, T_B):
+    m = 3
+    T_A_nan_inf = np.random.uniform(size=len(T_A))
+    T_B_nan_inf = np.random.uniform(size=len(T_B))
+    T_A[T_A_nan_inf > 0.90] = np.nan
```
seanlaw: Same as comment from above

tests/test_stomp.py
seanlaw: Need underscore between self and join

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stomp_nan_inf_A_B_join(T_A, T_B):
+    m = 3
+    T_A_nan_inf = np.random.uniform(size=len(T_A))
```
seanlaw: Same comment as above

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
     right = stump(pd.Series(T_B), m, pd.Series(T_A), ignore_trivial=False)
     utils.replace_inf(right)
     npt.assert_almost_equal(left[:, 0], right[:, 0])  # ignore indices
+
+
+def test_stump_nan_selfjoin_beginning():
```
seanlaw: Need underscore

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+
+    right = stump(T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stump_nan_inf_selfjoin(T_A, T_B):
```
seanlaw: Need underscore

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+    m = 3
+    T_B_nan_inf = np.random.uniform(size=len(T_B))
```
seanlaw: Same comment as above

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stump_nan_inf_A_B_join(T_A, T_B):
+    m = 3
+    T_A_nan_inf = np.random.uniform(size=len(T_A))
```
seanlaw: Same comment as above

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
     utils.replace_inf(left)
     utils.replace_inf(right)
     npt.assert_almost_equal(left, right)
+
+
+def test_stumped_nan_selfjoin_beginning(dask_client):
```
seanlaw: Need underscore

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+    )
+
+    right = stumped(dask_client, T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+def test_stumped_inf_selfjoin_beginning(dask_client):
```
seanlaw: Need underscore

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stumped_nan_inf_selfjoin(T_A, T_B, dask_client):
+    m = 3
+    T_B_nan_inf = np.random.uniform(size=len(T_B))
```
seanlaw: Same comment as above

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stumped_nan_inf_A_B_join(T_A, T_B, dask_client):
+    m = 3
+    T_A_nan_inf = np.random.uniform(size=len(T_A))
```
seanlaw: Same comment as above

tests/utils.py
```diff
@@ -4,17 +4,21 @@
 
 def z_norm(a, axis=0, threshold=1e-7):
     std = np.std(a, axis, keepdims=True)
-    std[std < threshold] = 1.0
+    std[np.less(std, threshold, where=~np.isnan(std))] = 1.0
```
seanlaw: This is some serious NumPy-fu! Very impressive. Thank you, I've learned something new!

seanlaw:
@mexxexx I can tell that you've spent a lot of time on this so thank you! Your changes are thorough and thoughtful and I really, really appreciate it.

mexxexx:
> Okay, so I thought about this a little more. What if we choose a different input time series that still satisfies the NaN/inf components that we are trying to test but where only one exact match can be found for each sequence? Technically, the rest of the sequence could be random. Perhaps, you can use the original `test_data` and replace parts of those sequences with `np.nan`/`np.inf` at the time of testing each array?

I would say ignoring the index in this particular case would be the way to go, because if the matrix profile is computed correctly, the index will also be 'correct', even if it is not the same one as computed by naive mass.

> I can tell that you've spent a lot of time on this so thank you! Your changes are thorough and thoughtful and I really, really appreciate it. I'd added a couple of more changes to the code base so you'll need to make sure to move things over accordingly without simply replacing new files.

Thank you! I will make the changes today and hopefully create a version that passes all test. Then I'd create a new, clean pull request. I'm going on a weeks vacation tomorrow, so I'm not sure I'll have time to finish everything until next week though.

stumpy/core.py
```diff
@@ -296,21 +299,10 @@ def compute_mean_std(T, m):
...
-    cumsum_T_squared[0] = 0
-
-    subseq_sum_T = cumsum_T[m:] - cumsum_T[: n - m + 1]
-    subseq_sum_T_squared = cumsum_T_squared[m:] - cumsum_T_squared[: n - m + 1]
-    M_T = subseq_sum_T / m
-    Σ_T = np.abs((subseq_sum_T_squared / m) - np.square(M_T))
-    Σ_T = np.sqrt(Σ_T)
+    M_T = np.mean(rolling_window(T, m), axis=1)
+    M_T[np.isnan(M_T)] = np.inf
+    Σ_T = np.nanstd(rolling_window(T, m), axis=1)
```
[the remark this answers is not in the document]
mexxexx: I don't know much about matlab, but I can imagine that there is no convenient method out of the box. I think it makes sense to use numpy wherever possible, for speed and readability alike.

stumpy/core.py
```diff
@@ -357,11 +349,17 @@ def calculate_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
     denom = np.asarray(denom)
     denom[denom == 0] = 1e-10  # Avoid divide by zero
     D_squared = np.abs(2 * m * (1.0 - (QT - m * μ_Q * M_T) / denom))
+    D_squared_inf = np.isinf(D_squared)
+
     threshold = 1e-7
     if σ_Q < threshold:  # pragma: no cover
         D_squared[:] = m
-    D_squared[Σ_T < threshold] = m
+    D_squared[(Σ_T < threshold)] = m
```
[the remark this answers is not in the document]
mexxexx: Only a mistake

stumpy/stamp.py
mexxexx: Actually this indentation should be ok, because if we are outside the `if D[:trivial_idx].size` bracket, then IL is -1 anyways. But we can remove one indent if you say it improves readability.

stumpy/stamp.py
mexxexx: Because the right profile index is the lowest index to the right of the trivial index, and D[trivial_idx] is always the lowest value (=0) in D[trivial_idx:]

stumpy/stamp.py
mexxexx: See above

stumpy/stump.py
mexxexx: In my first implementation I was working with NaN instead of inf until I realized that we run into issues in fastmath mode. I removed the comment

stumpy/core.py
mexxexx: Yes, as long as it is being used we should keep in in core

stumpy/stump.py
mexxexx: Exactly, because this might produce some undesired results on the user side and it is only one single copy at the beginning

stumpy/stump.py
mexxexx: See above

stumpy/stump.py
mexxexx: Mistake, I added it back in

stumpy/stump.py
mexxexx: We want to have the mean for every sub sequence that contains a inf/nan value set to inf, because this is how unwanted sequences are being filtered out in the end. But our time series T_A is not allowed to have nans anywhere, because of how QT is computed (incrementally). If it were to have nans, QT would be nan for all values after a certain point. Thats why we set T_A to zero, to allow computations, and the artificially change only the unwanted sequences back to nan using the mean. That is actually exactly the job of the inf_mask.

tests/test_stamp.py
mexxexx: That is true. I will slightly change the logic here.

tests/test_stamp.py
mexxexx: See above

tests/utils.py
mexxexx: Rather thank the people from stack overflow :D

stumpy/core.py
mexxexx: I changed it the way you suggested. I prefer to have short circuit return statements as you call them, because then I can directly see when the function is over for a certain branching.

stumpy/stump.py
mexxexx: By mistake

stumpy/stumped.py
mexxexx: See below

tests/test_stamp.py
mexxexx: I thought that by having [0:2] there we would have more readability. naive mass also returns the left and right indices, something that STAMP does not. Before there was a dedicated naive_mass function in test_stamp, but I don't like having the same function in two different places

seanlaw:
@mexxexx That sounds great. Do what you can but quality over speed. If you are able to get a PR in that passes then that would be awesome and I can take it from there. Otherwise, don’t sweat it and please enjoy your vacation!

mexxexx:
I implemented all changes now, take a look if you agree with them. One test still failed, but only on one machine, I guess this is still the same floating point precision error, and I haven't seen this happening before.

seanlaw:
@mexxexx I think this is a legitimate failure. The issue is that since you are using `permutation` the tests are necessarily different each time you run it (and on different machines). Notice that we try to test deterministic inputs (hardcoded numbers) and then we test random inputs separately. I think we should do the same thing and start with deterministic test cases where the inputs and outputs are known as this will cover 80%+ of the cases.

Additionally, based on the outputs of the failed test:

```
E       Mismatched elements: 3 / 248 (1.21%)
E       Max absolute difference: 17
E       Max relative difference: 0.5
E        x: array([[0.01754829871880146, 51, -1, 51],
E              [0.03592144994497408, 55, -1, 55],
E              [0, -1, -1, -1],...
E        y: array([[nan, 34, -1, 34],
E              [0.03592144994499201, 55, -1, 55],
E              [0, -1, -1, -1],...
```

The output of the first row suggests that we aren't handling NaNs correctly in `utils.naive_mass`?

The test cases

mexxexx:
You are right, I didn't see it. I think we see a failure of stumped, because there should be no nan values in the matrix profile (y=stumped(...)). I never saw that problem before and I don't know how to reproduce it. Maybe there actually is a problem in core.mass, because this is how stumped calculates the very first matrix profile value...

Sorry for the chaotic PR :grimacing:

stumpy/core.py
seanlaw: Added this issue for future tracking https://github.com/TDAmeritrade/stumpy/issues/125

stumpy/stamp.py
seanlaw: Yes, please un-indent. This will help reduce the complexity of the nested if conditions (two deep is okay, three deep likely means we should re-think the logic)

stumpy/stamp.py
seanlaw: IMHO, I think the original logic is cleaner. Can you think of a scenario where, given the same input, the output would be different between the two versions of the code? If not, then I recommend we stick with the original version as the `+1` requires some mental gymnastics and isn't obvious (at least for "future Sean").

Also, if we change it here, I think there are other places within the code that we'll need to update as well.

stumpy/stamp.py
seanlaw: Yes, please un-indent in order to reduce the nested `if` statements and reduce code complexity

stumpy/stomp.py
```diff
@@ -101,36 +117,49 @@ def stomp(T_A, m, T_B=None, ignore_trivial=True):
...
     for i in range(1, l):
         QT[1:] = (
             QT[: k - 1] - T_B[i - 1] * T_A[: k - 1] + T_B[i - 1 + m] * T_A[-(k - 1) :]
         )
         QT[0] = QT_first[i]
-        D = core.calculate_distance_profile(
-            m, QT, μ_Q[i].item(0), σ_Q[i].item(0), M_T, Σ_T
-        )
+
+        D = core.calculate_distance_profile(m, QT, μ_Q[i], σ_Q[i], M_T, Σ_T)
```
seanlaw: Would you mind adding the `.item(0)` back as this helps to ensure that the type being passed over is a scalar and not an array?

stumpy/stump.py
```diff
@@ -312,28 +315,35 @@ def _stump(
             QT_odd[0] = QT_first[i]
             D = _calculate_squared_distance_profile(m, QT_odd, μ_Q[i], σ_Q[i], M_T, Σ_T)
 
+        # Ignore distance profile if query mean is np.nan
+        if np.isinf(μ_Q[i]):
+            D[:] = np.inf
```
seanlaw: I think there should be a space between this line and the next. This is minor but the two `if` statements seem independent so it would be nice to have visual separation. Again, minor.

stumpy/stump.py
seanlaw: Perhaps this comment should be added here? 

`  # Remove all nan values from T_B only after first profile is calculated`

stumpy/stumped.py
seanlaw: Perhaps this comment should be added here?

# Remove all nan values from T_A only after first profile is calculated

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+            utils.naive_mass(Q, T_B, m, i, zone, ignore_trivial=True)
             for i, Q in enumerate(core.rolling_window(T_B, m))
         ],
         dtype=object,
     )
     right = stamp.stamp(T_B, T_B, m, ignore_trivial=True)
     utils.replace_inf(left)
     utils.replace_inf(right)
-    npt.assert_almost_equal(left, right)
+    npt.assert_almost_equal(left[:, [0, 1]], right)
```
seanlaw: For consistency with other tests, can you please change this to:

```
left[:, :2]
```

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
     m = 3
     left = np.array(
-        [naive_mass(Q, T_A, m) for Q in core.rolling_window(T_B, m)], dtype=object
+        [utils.naive_mass(Q, T_A, m) for Q in core.rolling_window(T_B, m)], dtype=object
     )
     right = stamp.stamp(T_A, T_B, m)
     utils.replace_inf(left)
     utils.replace_inf(right)
-    npt.assert_almost_equal(left, right)
+    npt.assert_almost_equal(left[:, [0, 1]], right)
```
seanlaw: Please change this to `left[:, :2]`

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+
+
+def test_stamp_nan_self_join_beginning():
+    m = 3
+    T = np.array([np.nan, 1, 0, 0, 1, 0, 0])
```
seanlaw: Is there a reason why the rest of the sequence is `[1, 0, 0, 1, 0, 0]`? Could we use the `test_data` and replace the first element with `np.nan`?

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+    right = stamp.stamp(T, T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
+
+
+def test_stamp_inf_self_join_beginning():
+    m = 3
+    T = np.array([np.inf, 1, 0, 0, 1, 0, 0])
```
seanlaw: Is there a reason why the rest of the sequence is `[1, 0, 0, 1, 0, 0]`? Could we use the `test_data` and replace the first element with `np.inf`?

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+            for i, Q in enumerate(core.rolling_window(T, m))
+        ],
+        dtype=object,
+    )
+
+    right = stamp.stamp(T, T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
```
seanlaw: Using the `test_data` might result in the matrix profile indices to be correct

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
```
seanlaw: Using the `test_data` may cause the matrix profile indices to be correct

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
     utils.replace_inf(left)
     utils.replace_inf(right)
     npt.assert_almost_equal(left, right)
+
+
+def test_stomp_nan_self_join_beginning():
+    m = 3
+    T = np.array([np.nan, 1, 0, 0, 1, 0, 0])
```
seanlaw: Is there a reason why the rest of the sequence is `[1, 0, 0, 1, 0, 0]`? Could we use the `test_data` and replace the first element with `np.nan`?

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
+    right = stomp(T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
+
+
+def test_stomp_inf_self_join_beginning():
+    m = 3
+    T = np.array([np.inf, 1, 0, 0, 1, 0, 0])
```
seanlaw: Is there a reason why the rest of the sequence is `[1, 0, 0, 1, 0, 0]`? Could we use the `test_data` and replace the first element with `np.inf`?

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+
+    left = np.array(
+        [utils.naive_mass(Q, T_A, m) for Q in core.rolling_window(T_B, m)], dtype=object
+    )
+
+    right = stamp.stamp(T_A, T_B, m)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, [0, 1]], right)
```
seanlaw: Change to left[:, [:2]]

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+            for i, Q in enumerate(core.rolling_window(T_B, m))
+        ],
+        dtype=object,
+    )
+
+    right = stamp.stamp(T_B, T_B, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, [0, 1]], right)
```
seanlaw: Change to `left[:, :2]`

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stamp_nan_inf_self_join(T_A, T_B):
+    m = 3
+    nan_inf_sample_B = np.random.permutation(np.arange(len(T_B)))
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

tests/test_stamp.py
```diff
@@ -40,24 +22,108 @@ def test_stamp_self_join(T_A, T_B):
...
+    right = stamp.stamp(T_B, T_B, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, [0, 1]], right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stamp_nan_inf_A_B_join(T_A, T_B):
+    m = 3
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
+            for i, Q in enumerate(core.rolling_window(T, m))
+        ],
+        dtype=object,
+    )
+
+    right = stomp(T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
```
seanlaw: The matrix index profiles may be the same if the `test_data` is used

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
```
seanlaw: The matrix profile indices may be fine if `test_data` is used

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stomp_nan_inf_self_join(T_A, T_B):
+    m = 3
+    nan_inf_sample_B = np.random.permutation(np.arange(len(T_B)))
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

tests/test_stomp.py
```diff
@@ -63,3 +63,87 @@ def test_stomp_A_B_join(T_A, T_B):
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stomp_nan_inf_A_B_join(T_A, T_B):
+    m = 3
+    nan_inf_sample_A = np.random.permutation(np.arange(len(T_A)))
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
     right = stump(pd.Series(T_B), m, pd.Series(T_A), ignore_trivial=False)
     utils.replace_inf(right)
     npt.assert_almost_equal(left[:, 0], right[:, 0])  # ignore indices
```
seanlaw: We should check if the indices would resolve itself when the input is random. The precision problems really only occur when there are multiple subsequence matches

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+
+
+def test_stump_nan_self_join_beginning():
+    m = 3
+    T = np.array([np.nan, 1, 0, 0, 1, 0, 0])
```
seanlaw: Can we use `test_data` here?

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+
+    print(left)
+    print(right)
+
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
+
+
+def test_stump_inf_self_join_beginning():
+    m = 3
+    T = np.array([np.inf, 1, 0, 0, 1, 0, 0])
```
seanlaw: Can we use `test_data` here?

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+            for i, Q in enumerate(core.rolling_window(T, m))
+        ],
+        dtype=object,
+    )
+
+    right = stump(T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
```
seanlaw: This may be resolved if we use `test_data`

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stump_nan_inf_self_join(T_A, T_B):
+    m = 3
+    nan_inf_sample_B = np.random.permutation(np.arange(len(T_B)))
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

tests/test_stump.py
```diff
@@ -146,3 +146,91 @@ def test_two_constant_subsequences_A_B_join():
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stump_nan_inf_A_B_join(T_A, T_B):
+    m = 3
+    nan_inf_sample_A = np.random.permutation(np.arange(len(T_A)))
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
     utils.replace_inf(left)
     utils.replace_inf(right)
     npt.assert_almost_equal(left, right)
+
+
+def test_stumped_nan_self_join_beginning(dask_client):
+    m = 3
+    T = np.array([np.nan, 1, 0, 0, 1, 0, 0])
```
seanlaw: Can we try `test_data`?

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+            for i, Q in enumerate(core.rolling_window(T, m))
+        ],
+        dtype=object,
+    )
+
+    right = stumped(dask_client, T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
```
seanlaw: Matrix profile indices may be fine with `test_data`

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+
+
+def test_stumped_inf_self_join_beginning(dask_client):
+    m = 3
+    T = np.array([np.inf, 1, 0, 0, 1, 0, 0])
```
seanlaw: Try `test_data`

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+            for i, Q in enumerate(core.rolling_window(T, m))
+        ],
+        dtype=object,
+    )
+
+    right = stumped(dask_client, T, m, ignore_trivial=True)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left[:, 0], right[:, 0])
```
seanlaw: Matrix profile indices may be fine with `test_data`

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stumped_nan_inf_self_join(T_A, T_B, dask_client):
+    m = 3
+    nan_inf_sample_B = np.random.permutation(np.arange(len(T_B)))
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

tests/test_stumped.py
```diff
@@ -150,3 +150,87 @@ def test_stumped_A_B_join_df(T_A, T_B, dask_client):
...
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_almost_equal(left, right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+def test_stumped_nan_inf_A_B_join(T_A, T_B, dask_client):
+    m = 3
+    nan_inf_sample_A = np.random.permutation(np.arange(len(T_A)))
```
seanlaw: This seems overly complicated for a unit test. It is more important that these particular test cases have hard coded `np.nan` and `np.inf` positions.

stumpy/stump.py
seanlaw: I missed this comment that helps explain the logic as well. https://github.com/TDAmeritrade/stumpy/pull/122#issue-364812444

seanlaw:
I’ve been thinking about what should be default behavior. That is, whether we should have the user be explicit if their data contains NaN/inf and for the user to say whether or not to automatically handle it. I don’t have experience with dealing with NaN/inf in my time series so I don’t know what the expectation of a user would be. I like that the ease of use for a user but I don’t want them to shoot themselves in the foot either. 

Another option is to update the `check_nan` function to also check for inf but to have it warn the user that “NaN/int found. We are doing X to your time series data before and after the analysis”. Any thoughts?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mexxexx:
Hi @seanlaw I will work through your comments today. Would you suggest to use hard coded nan/inf positions instead of the random ones? 

Also, my take on handling nan/inf values: I think it is less complicated to have one default behavior and be consistent with it, rather that giving the user full control. I don't see where the user would benefit there. There is no way that we can know all possible use cases where nan/inf values occur, so trying to do anything rather than ignoring these values seems pointless to me. If the user really needs them in any way, he will have to do some preprocessing, like replacing them with zero, but I don't think this should be done by stumpy, because we cannot foresee every use case. As long as the behavior of time series without nans/infs doesn't change, I think doing it this way is the most universal approach.

Issuing a warning might be a good idea, I think numpy does so too. Maybe "NaN/Inf found, ignoring all sub sequences containing these values"?

seanlaw:
@mexxexx, welcome back! I hope that you had a restful time off.

> Would you suggest to use hard coded nan/inf positions instead of the random ones?

Perhaps, my earlier use of the phrase "hard coded" probably should've been read as "deliberate" but I think you understood the point. Firstly, I'm not an expert in test-driven-development (TDD) but, from what I've observed, people usually start with writing the unit test first and then writing the code to pass the test. In this case, I think you've identified the following potential sanity tests (I'm probably missing some obvious "deliberate" ones):

**One sequence self-join**
- [ ] One sequence (self-join) with NaN at the first position
- [ ] One sequence (self-join) with NaN at the last position
- [ ] One sequence (self-join) with NaN in the middle of the sequence
- [ ] One sequence (self-join) with multiple (maybe start with 2) non-consecutive NaN in the sequence
**One sequence AB-join**
- [ ] One sequence (AB-join) with multiple (maybe start with 2) consecutive NaN in the sequence
- [ ] One sequence (AB-join) with NaN at the first position
- [ ] One sequence (AB-join) with NaN at the last position
- [ ] One sequence (AB-join) with NaN in the middle of the sequence
- [ ] One sequence (AB-join) with multiple (maybe start with 2) non-adjacent NaN in the sequence
- [ ] One sequence (AB-join) with multiple (maybe start with 2) adjacent NaN in the sequence
**Two sequences AB-join**
- [ ] Two sequences (AB-join) with multiple (maybe start with 2) consecutive NaN in the sequence
- [ ] Two sequences (AB-join) with NaN at the first position
- [ ] Two sequences (AB-join) with NaN at the last position
- [ ] Two sequences (AB-join) with NaN in the middle of the sequence
- [ ] Two sequences (AB-join) with multiple (maybe start with 2) non-adjacent NaN in the sequence
- [ ] Two sequences (AB-join) with multiple (maybe start with 2) adjacent NaN in the sequence

And then, we can replace `NaN` with `inf`. Finally, we have the cases where there is a mixture of `NaN` and`inf` within the sequence. Btw, I'm not trying to dictate here. Just listing things out loud so feel free to do what you feel is right as I know it may be a lot of work. We might be able to wrap this into a `for-loop` for each of the three sections above and do something like (**not tested!!**):

```
test_data = [
    (
        np.array([9, 8100, -60, 7], dtype=np.float64),
        np.array([584, -11, 23, 79, 1001, 0, -19], dtype=np.float64),
    ),
    (
        np.random.uniform(-1000, 1000, [8]).astype(np.float64),
        np.random.uniform(-1000, 1000, [64]).astype(np.float64),
    ),
]

@pytest.mark.parametrize("T_A, T_B", test_data)
def test_stump_self_join_one_sequence_nan(T_A, T_B):
    m = 3
    T = T_B.copy()
    nan_positions = [(0), (-1), (1), (-2), (2:5)]
    for nan_pos in nan_positions:
        T[:] = T_B[:]  # reset T with the values from T_B
        T[nan_pos] = np.nan
        left = np.array(
            [
                utils.naive_mass(Q, T, m, i, zone, True)
                for i, Q in enumerate(core.rolling_window(T, m))
            ],
            dtype=object,
        )
        right = stump(T, m, ignore_trivial=True)
        utils.replace_inf(left)
        utils.replace_inf(right)
        npt.assert_almost_equal(left, right)
```

Again, the above is just conceptual and **not tested**!

Ultimately, I trust your judgement in terms of "what cases have you seen and what cases do you reasonably expect that users will most likely encounter"? And then implement those exact tests. Later, when there is a case that is causing a user errors then that specific case will be added as a new unit test and then we can improve the code to pass that particular test. Does that make sense?

mexxexx:
@seanlaw Thank you very much for taking the time, I am also very new to test driven development. I agree, your approach makes sense, and I will probably implement all cases you mentioned :)

seanlaw:
@mexxexx Awesome! I am here to help

mexxexx:
I implemented the changes in a similar manner that you suggested, but dask is causing some problems when the test_stumped_nan_inf_A_B_join tests are run. Have you seen something like this before?

seanlaw:
@mexxexx Let me take a look. Dask on Azure Pipelines is rather flaky. If things pass on your machine then it is a good sign.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

seanlaw:
@mexxexx Since this is happening for all operating systems then my suspicion is that there is something wrong with your tests as it relates to Dask. To isolate the problem, please do the following:

1. Run all tests on your local machine. Do they pass?
2. Similar to `test_stumped_one_constant_subsequence.py`
and `test_stumped_two_constant_subsequences.py`, please move all of your NaN/inf tests into two separate test files `test_stumped_one_subsequence_nan_inf.py` and
`test_stumped_two_subsequences_nan_inf.py`. The rationale for this is really stupid but we should do this regardless and it will likely make your tests pass.
3. A minor point/suggestion. I haven't really examined all of your changes but I noticed that your test has the `T_A.copy()` nested inside the `for-loops`. While the arrays are small and the copy is inexpensive, it's likely better to create a `copy` outside of the `for-loops` and then do `T_A_sub[:] = T_A[:]`. It's subtle but this avoids creating a brand new array and only copies over/replaces the array entries (since the size has not changed). Again, only a suggestion.
4. You need to be careful in that the first test_data may be too short in length for slicing.

Believe it or not number 2. above is the most important.

seanlaw:
@mexxexx It looks like the tests are getting stuck. I took a quick look at the new file and it looks like there are a ton of tests. When Dask is involved during a test, the goal is to keep the test file with about 10 tests. Otherwise, it can get stuck on Azure.

mexxexx:
@seanlaw I see.. I found that running stumped multiple times in for loops produced the issues from before (having dask reporting a CancelledError, even on a local machine). This is why there are so many tests now. Should we exclude some of the tests? Maybe only keep the longer time series and remove the tests with inf values? Because if it works for long serieses, it also works for slightly shorter ones, and the inf case is basically the same as the nan case

seanlaw:
@mexxexx I think it is important to test both `NaN` and `inf`. I might recommend splitting things up into (minimally) 7 separate test files:

1. one subsequence self-join (NaN)
2. one subsequence ab-join (NaN)
3. two subsequences ab-join (NaN, NaN)
4. one subsequence self-join (inf)
5. one subsequence ab-join (inf)
6. two subsequences ab-join (inf, inf)
7. two subsequences ab-join (NaN, inf)

Please take the last test `test_stumped_nan_inf_self_join` out of `test_stumped`. I think `test_stumped` is already oversaturated.

seanlaw:
@mexxexx Alright! I ran the test twice and it passed both times. Would you mind forking a fresh instance of the main repo (there have been a lot of changes since you've worked on it) and then move over the new changes? Close this one and I'll do a clean review.

stumpy/stamp.py
```diff
@@ -126,15 +130,30 @@ def stamp(T_A, T_B, m, ignore_trivial=False):
...
+        )
     core.check_dtype(T_B)
-    core.check_nan(T_B)
+
     core.check_window_size(m)
     subseq_T_B = core.rolling_window(T_B, m)
     excl_zone = int(np.ceil(m / 2))
     M_T, Σ_T = core.compute_mean_std(T_A, m)
 
+    T_A[np.isnan(T_A)] = 0
```
seanlaw: Should there be an equivalent `T_B[np.isnan(T_B)] = 0` line after this?

seanlaw: Please refresh my memory here. So:

1. We take in an array
2. We check to see if there are any `np.inf` values and, if so, then convert them to `np.nan` since `np.inf` cannot be handled properly by `np.mean`/`np.std` but np.nan is fine
3. After the mean and stddev are computed then, according to the [original comment by zpzim](https://github.com/TDAmeritrade/stumpy/issues/115#issuecomment-574500923) that says that:

> You also have to set the time series values which are NaN to zero when actually computing the MP so that your dot product doesn't end up NaN.

we set the `np.nan` values to `0` so that the dot product/distance profile can be computed by `core.mass`/`core.calculate_distance_profile`

4. However, even if a subsequence that has an `np.nan`/`np.inf` within it (and that has been converted temporarily to zero), its distance to any other subsequence will always be np.nan since its mean/stddev is `np.nan`. 

Is this the correct way of thinking about it? Am I missing any logic here that should be called out?

PR #126, opened 2020-02-04
Fixed Issue #115

mexxexx:
This pull request fixes Issue #115, the handling of NaN and inf values in the 1D Matrix Profile Algorithms.

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -260,22 +260,41 @@ def sliding_dot_product(Q, T):
 
 
 def rolling_chunked_std(T, m):
+    """
+    Compute the sliding standard deviation for the array `T` with
+    a window size of `m` by splitting T into smaller chunks if memory
+    errors arise.
+
+    Parameters
+    ----------
+    T : ndarray
+        Time series or sequence
+
+    m : int
+        Window size
+
+    Returns
+    -------
+    Σ_T : ndarray
+        Sliding standard deviation
+    """
+
     n_chunks = 1
     while True:
         try:
-            chunk_size = math.ceil((T.shape[0] + 1)/n_chunks)
+            chunk_size = math.ceil((T.shape[0] + 1) / n_chunks)
             std_chunks = []
             for chunk in range(n_chunks):
                 start = chunk * chunk_size
                 stop = min(start + chunk_size + m - 1, T.shape[0])
                 tmp_std = np.nanstd(rolling_window(T[start:stop], m), axis=1)
                 std_chunks.append(tmp_std)
-            chunked_std = np.hstack(std_chunks)  # Note the use of hstack instead of concatenate
+            Σ_T = np.hstack(std_chunks)  # Note the use of hstack instead of concatenate
             break
 
         except MemoryError:
             n_chunks *= 2
-    return chunked_std
+    return Σ_T
 
 
 def compute_mean_std(T, m):
@@ -322,7 +341,6 @@ def compute_mean_std(T, m):
 
     M_T = np.mean(rolling_window(T, m), axis=1)
     M_T[np.isnan(M_T)] = np.inf
-    #Σ_T = np.nanstd(rolling_window(T, m), axis=1)
     Σ_T = rolling_chunked_std(T, m)
 
     return M_T, Σ_T
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -259,7 +259,7 @@ def sliding_dot_product(Q, T):
     return QT.real[m - 1 : n]
 
 
-def rolling_chunked_std(T, m):
+def rolling_nanstd_chunked(T, m):
     """
     Compute the sliding standard deviation for the array `T` with
     a window size of `m` by splitting T into smaller chunks if memory
@@ -289,10 +289,10 @@ def rolling_chunked_std(T, m):
                 stop = min(start + chunk_size + m - 1, T.shape[0])
                 tmp_std = np.nanstd(rolling_window(T[start:stop], m), axis=1)
                 std_chunks.append(tmp_std)
-            Σ_T = np.hstack(std_chunks)  # Note the use of hstack instead of concatenate
+            Σ_T = np.hstack(std_chunks)
             break
 
-        except MemoryError:
+        except MemoryError:  # pragma nocover
             n_chunks *= 2
     return Σ_T
 
@@ -341,9 +341,64 @@ def compute_mean_std(T, m):
 
     M_T = np.mean(rolling_window(T, m), axis=1)
     M_T[np.isnan(M_T)] = np.inf
-    Σ_T = rolling_chunked_std(T, m)
+    Σ_T = rolling_nanstd_chunked(T, m)
+    Σ_T[np.isnan(Σ_T)] = 0
 
     return M_T, Σ_T
+
+
+@njit(parallel=True, fastmath=True)
+def calculate_squared_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
+    """
+    Compute the distance profile
+
+    Parameters
+    ----------
+    m : int
+        Window size
+
+    QT : ndarray
+        Dot product between `Q` and `T`
+
+    μ_Q : ndarray
+        Mean of `Q`
+
+    σ_Q : ndarray
+        Standard deviation of `Q`
+
+    M_T : ndarray
+        Sliding mean of `T`
+
+    Σ_T : ndarray
+        Sliding standard deviation of `T`
+
+    Returns
+    -------
+    output : ndarray
+        Distance profile squared
+
+    Notes
+    -----
+    `DOI: 10.1109/ICDM.2016.0179 \
+    <https://www.cs.ucr.edu/~eamonn/PID4481997_extend_Matrix%20Profile_I.pdf>`__
+
+    See Equation on Page 4
+    """
+
+    denom = m * σ_Q * Σ_T
+    denom = np.asarray(denom)
+    denom[denom == 0] = 1e-10  # Avoid divide by zero
+    D_squared = np.abs(2 * m * (1.0 - (QT - m * μ_Q * M_T) / denom))
+    inf_mask = np.isinf(D_squared)
+
+    threshold = 1e-7
+    if σ_Q < threshold:  # pragma: no cover
+        D_squared[:] = m
+    D_squared[Σ_T < threshold] = m
+    D_squared[(Σ_T < threshold) & (σ_Q < threshold)] = 0
+    D_squared[inf_mask] = np.inf
+
+    return D_squared
 
 
 @njit(parallel=True, fastmath=True)
@@ -384,18 +439,7 @@ def calculate_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
     See Equation on Page 4
     """
 
-    denom = m * σ_Q * Σ_T
-    denom = np.asarray(denom)
-    denom[denom == 0] = 1e-10  # Avoid divide by zero
-    D_squared = np.abs(2 * m * (1.0 - (QT - m * μ_Q * M_T) / denom))
-    inf_mask = np.isinf(D_squared)
-
-    threshold = 1e-7
-    if σ_Q < threshold:  # pragma: no cover
-        D_squared[:] = m
-    D_squared[Σ_T < threshold] = m
-    D_squared[(Σ_T < threshold) & (σ_Q < threshold)] = 0
-    D_squared[inf_mask] = np.inf
+    D_squared = calculate_squared_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T)
 
     return np.sqrt(D_squared)
 
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -157,7 +157,10 @@ def are_arrays_equal(a, b):  # pragma: no cover
     if id(a) == id(b):
         return True
 
-    return np.array_equal(a, b)
+    if a.shape != b.shape:
+        return False
+
+    return ((a == b) | (np.isnan(a) & np.isnan(b))).all()
 
 
 def are_distances_too_small(a, threshold=10e-6):  # pragma: no cover
@@ -271,7 +274,7 @@ def compute_mean_std(T, m):
     Returns
     -------
     M_T : ndarray
-        Sliding mean
+        Sliding mean. All nan values are replaced with np.inf
 
     Σ_T : ndarray
         Sliding standard deviation
@@ -296,21 +299,10 @@ def compute_mean_std(T, m):
     Note that Mueen's algorithm has an off-by-one bug where the
     sum for the first subsequence is omitted and we fixed that!
     """
-    n = T.shape[0]
 
-    cumsum_T = np.empty(len(T) + 1)
-    np.cumsum(T, out=cumsum_T[1:])  # store output in cumsum_T[1:]
-    cumsum_T[0] = 0
-
-    cumsum_T_squared = np.empty(len(T) + 1)
-    np.cumsum(np.square(T), out=cumsum_T_squared[1:])
-    cumsum_T_squared[0] = 0
-
-    subseq_sum_T = cumsum_T[m:] - cumsum_T[: n - m + 1]
-    subseq_sum_T_squared = cumsum_T_squared[m:] - cumsum_T_squared[: n - m + 1]
-    M_T = subseq_sum_T / m
-    Σ_T = np.abs((subseq_sum_T_squared / m) - np.square(M_T))
-    Σ_T = np.sqrt(Σ_T)
+    M_T = np.mean(rolling_window(T, m), axis=1)
+    M_T[np.isnan(M_T)] = np.inf
+    Σ_T = np.nanstd(rolling_window(T, m), axis=1)
 
     return M_T, Σ_T
 
@@ -357,11 +349,15 @@ def calculate_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
     denom = np.asarray(denom)
     denom[denom == 0] = 1e-10  # Avoid divide by zero
     D_squared = np.abs(2 * m * (1.0 - (QT - m * μ_Q * M_T) / denom))
+    inf_mask = np.isinf(D_squared)
+
     threshold = 1e-7
     if σ_Q < threshold:  # pragma: no cover
         D_squared[:] = m
     D_squared[Σ_T < threshold] = m
     D_squared[(Σ_T < threshold) & (σ_Q < threshold)] = 0
+    D_squared[inf_mask] = np.inf
+
     return np.sqrt(D_squared)
 
 
@@ -471,24 +467,32 @@ def mass(Q, T, M_T=None, Σ_T=None):
 
     Q = np.asarray(Q)
     check_dtype(Q)
-    check_nan(Q)
 
     if Q.ndim != 1:  # pragma: no cover
         raise ValueError(f"Q is {Q.ndim}-dimensional and must be 1-dimensional. ")
+    m = Q.shape[0]
+
     T = np.asarray(T)
     check_dtype(T)
     check_nan(T)
 
     if T.ndim != 1:  # pragma: no cover
         raise ValueError(f"T is {T.ndim}-dimensional and must be 1-dimensional. ")
+    n = T.shape[0]
 
-    QT = sliding_dot_product(Q, T)
-    m = Q.shape[0]
-    μ_Q, σ_Q = compute_mean_std(Q, m)
-    if M_T is None or Σ_T is None:
-        M_T, Σ_T = compute_mean_std(T, m)
+    distance_profile = np.empty(n - m + 1)
+    if np.any(np.isnan(Q)):
+        distance_profile[:] = np.inf
+    else:
+        QT = sliding_dot_product(Q, T)
+        μ_Q, σ_Q = compute_mean_std(Q, m)
+        if M_T is None or Σ_T is None:
+            M_T, Σ_T = compute_mean_std(T, m)
+        distance_profile = calculate_distance_profile(
+            m, QT, μ_Q.item(0), σ_Q.item(0), M_T, Σ_T
+        )
 
-    return calculate_distance_profile(m, QT, μ_Q.item(0), σ_Q.item(0), M_T, Σ_T)
+    return distance_profile
 
 
 def array_to_temp_file(a):
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -6,6 +6,7 @@ import numpy as np
 from numba import njit
 import scipy.signal
 import tempfile
+import math
 
 try:
     from numba.cuda.cudadrv.driver import _raise_driver_not_found
@@ -258,6 +259,25 @@ def sliding_dot_product(Q, T):
     return QT.real[m - 1 : n]
 
 
+def rolling_chunked_std(T, m):
+    n_chunks = 1
+    while True:
+        try:
+            chunk_size = math.ceil((T.shape[0] + 1)/n_chunks)
+            std_chunks = []
+            for chunk in range(n_chunks):
+                start = chunk * chunk_size
+                stop = min(start + chunk_size + m - 1, T.shape[0])
+                tmp_std = np.nanstd(rolling_window(T[start:stop], m), axis=1)
+                std_chunks.append(tmp_std)
+            chunked_std = np.hstack(std_chunks)  # Note the use of hstack instead of concatenate
+            break
+
+        except MemoryError:
+            n_chunks *= 2
+    return chunked_std
+
+
 def compute_mean_std(T, m):
     """
     Compute the sliding mean and standard deviation for the array `T` with
@@ -302,7 +322,8 @@ def compute_mean_std(T, m):
 
     M_T = np.mean(rolling_window(T, m), axis=1)
     M_T[np.isnan(M_T)] = np.inf
-    Σ_T = np.nanstd(rolling_window(T, m), axis=1)
+    #Σ_T = np.nanstd(rolling_window(T, m), axis=1)
+    Σ_T = rolling_chunked_std(T, m)
 
     return M_T, Σ_T
 
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -259,48 +259,10 @@ def sliding_dot_product(Q, T):
     return QT.real[m - 1 : n]
 
 
-def rolling_nanstd_chunked(T, m):
-    """
-    Compute the sliding standard deviation for the array `T` with
-    a window size of `m` by splitting T into smaller chunks if memory
-    errors arise.
-
-    Parameters
-    ----------
-    T : ndarray
-        Time series or sequence
-
-    m : int
-        Window size
-
-    Returns
-    -------
-    Σ_T : ndarray
-        Sliding standard deviation
-    """
-
-    n_chunks = 1
-    while True:
-        try:
-            chunk_size = math.ceil((T.shape[0] + 1) / n_chunks)
-            std_chunks = []
-            for chunk in range(n_chunks):
-                start = chunk * chunk_size
-                stop = min(start + chunk_size + m - 1, T.shape[0])
-                tmp_std = np.nanstd(rolling_window(T[start:stop], m), axis=1)
-                std_chunks.append(tmp_std)
-            Σ_T = np.hstack(std_chunks)
-            break
-
-        except MemoryError:  # pragma nocover
-            n_chunks *= 2
-    return Σ_T
-
-
-def compute_mean_std(T, m):
+def compute_mean_std(T, m, num_chunks=1, max_iter=10):
     """
     Compute the sliding mean and standard deviation for the array `T` with
-    a window size of `m`
+    a window size of `m` by splitting up T in `n_chunks`
 
     Parameters
     ----------
@@ -309,6 +271,13 @@ def compute_mean_std(T, m):
 
     m : int
         Window size
+
+    num_chunks : int
+        Number of chunks to use for the first iteration. If a Memory Error is raised,
+        this number is doubled and the computation tried again.
+
+    max_iter : int
+        Maximum number of iterations
 
     Returns
     -------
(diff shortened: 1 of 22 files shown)


seanlaw:
@mexxexx Thank you for this PR. It looks like all of the tests have passed so I will no go over everything very carefully with a finer toothed comb.

seanlaw:
For posterity of information, this PR is a clean up version of #122

seanlaw:
@mexxexx while it is still fresh in your mind, would you mind compiling a list of things that still need to be done after this PR? A few items come to mind like adding NaN/inf support for `gpu_stump`, changing `check_nan` to `check_nan_inf` and giving a warning if either are detected, additional tests that we may have missed etc. This list will inform the creation of new issues after the PR is merged. Thanks in advance!

mexxexx:
@seanlaw Of course, I will keep it in this message and update it, if something else comes into my mind.

- [ ] Add nan/inf support to mstump + tests
- [ ] Add nan/inf support to mstumped + tests
- [ ] Add nan/inf support to gpu_stump + tests
- [ ] Change check_nan to check_nan_inf and giving a warning
- [ ] Doing this nan/inf check in the 1D algorithms

mexxexx:
I noticed that the new implementation of calculating the mean and standard deviation might lead to memory errors for very long time series and window lengths. The reason is that apparently numpys np.std() function does a copy of the array that is being passed. So while the rolling_window function only creates a view, this gets converted to a huge array inside np.std.

I will try out if this indeed leads to a problem, and if it does we will probably have to change the way the std is computed again.

seanlaw:
> I noticed that the new implementation of calculating the mean and standard deviation might lead to memory errors for very long time series and window lengths. The reason is that apparently numpys np.std() function does a copy of the array that is being passed. So while the rolling_window function only creates a view, this gets converted to a huge array inside np.std.

> I will try out if this indeed leads to a problem, and if it does we will probably have to change the way the std is computed again.

@mexxexx Sounds good! I hope that it isn't a problem. In case it matters, currently, our largest supported time series length is 100,000,000 (100 million) data points.

tests/test_stomp.py
```diff
@@ -63,3 +66,59 @@ def test_stomp_A_B_join(T_A, T_B):
...
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize("substitute_A", substitution_values)
+@pytest.mark.parametrize("substitute_B", substitution_values)
+def test_stomp_nan_inf_A_B_join(T_A, T_B, substitute_A, substitute_B):
+    m = 3
+
+    T_A_sub = T_A.copy()
+    T_B_sub = T_B.copy()
+
+    substitution_locations = [slice(0, 0), 0, -1, slice(1, 3), [0, 3]]
```
seanlaw: Move out and replace with pytest.mark

tests/test_stomp.py
```diff
@@ -63,3 +66,59 @@ def test_stomp_A_B_join(T_A, T_B):
...
+            [
+                utils.naive_mass(Q, T_B_sub, m, i, zone, True)
+                for i, Q in enumerate(core.rolling_window(T_B_sub, m))
+            ],
+            dtype=object,
+        )
+        right = stomp(T_B_sub, m, ignore_trivial=True)
+        utils.replace_inf(left)
+        utils.replace_inf(right)
+        npt.assert_almost_equal(left, right)
```
seanlaw: This is great. This implies that our tests are deterministic and that the matrix profile indices are correct!

tests/test_stump.py
```diff
@@ -146,3 +149,69 @@ def test_two_constant_subsequences_A_B_join():
...
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize("substitute_B", substitution_values)
+def test_stump_nan_inf_self_join(T_A, T_B, substitute_B):
+    m = 3
+
+    T_B_sub = T_B.copy()
+
+    substitution_locations = [slice(0, 0), 0, -1, slice(1, 3), [0, 3]]
```
seanlaw: Move to pytest.mark

tests/test_stump.py
```diff
@@ -146,3 +149,69 @@ def test_two_constant_subsequences_A_B_join():
...
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize("substitute_A", substitution_values)
+@pytest.mark.parametrize("substitute_B", substitution_values)
+def test_stump_nan_inf_A_B_join(T_A, T_B, substitute_A, substitute_B):
+    m = 3
+
+    T_A_sub = T_A.copy()
+    T_B_sub = T_B.copy()
+
+    substitution_locations = [slice(0, 0), 0, -1, slice(1, 3), [0, 3]]
```
seanlaw: Move to pytest.mark

tests/test_stumped.py
```diff
@@ -29,6 +29,8 @@ def dask_client():
     ),
 ]
 
+substitution_values = [np.nan, np.inf]
```
seanlaw: I don't think these three lines are needed since `np.nan`/`np.inf` testing is no longer performed in this test file

tests/test_stumped_two_subsequences_inf_A_B_join.py
```diff
@@ -0,0 +1,60 @@
...
+
+substitution_locations = [(0, 1), (-1, slice(1, 3)), (slice(1, 3), 1), ([0, 3], 1)]
+
+
+@pytest.mark.filterwarnings("ignore:numpy.dtype size changed")
+@pytest.mark.filterwarnings("ignore:numpy.ufunc size changed")
+@pytest.mark.filterwarnings("ignore:numpy.ndarray size changed")
+@pytest.mark.filterwarnings("ignore:\\s+Port 8787 is already in use:UserWarning")
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize(
```
seanlaw: This syntax looks wrong.

tests/test_stumped_two_subsequences_nan_A_B_join.py
```diff
@@ -0,0 +1,60 @@
...
+
+substitution_locations = [(0, 1), (-1, slice(1, 3)), (slice(1, 3), 1), ([0, 3], 1)]
+
+
+@pytest.mark.filterwarnings("ignore:numpy.dtype size changed")
+@pytest.mark.filterwarnings("ignore:numpy.ufunc size changed")
+@pytest.mark.filterwarnings("ignore:numpy.ndarray size changed")
+@pytest.mark.filterwarnings("ignore:\\s+Port 8787 is already in use:UserWarning")
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize(
```
seanlaw: This syntax looks wrong

tests/test_stumped_two_subsequences_nan_inf_A_B_join.py
```diff
@@ -0,0 +1,60 @@
...
+
+substitution_locations = [(0, 1), (-1, slice(1, 3)), (slice(1, 3), 1), ([0, 3], 1)]
+
+
+@pytest.mark.filterwarnings("ignore:numpy.dtype size changed")
+@pytest.mark.filterwarnings("ignore:numpy.ufunc size changed")
+@pytest.mark.filterwarnings("ignore:numpy.ndarray size changed")
+@pytest.mark.filterwarnings("ignore:\\s+Port 8787 is already in use:UserWarning")
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize(
```
seanlaw: This syntax looks wrong

tests/test_stumped_two_subsequences_nan_inf_A_B_join.py
```diff
@@ -0,0 +1,60 @@
...
+    "substitution_location_A, substitution_location_B", substitution_locations
+)
+def test_stumped_two_subsequences_nan_inf_A_B_join(
```
seanlaw: In this particular case, would you mind adding an additional test where the`T_A_sub` is `np.nan` and `T_B_sub` is `np.inf`? Basically, I'm just swapping "where" `np.nan`/`np.inf` exists

mexxexx:
The following code produces a Memory Error on my machine, because numpy.std internally creates a copy of the rolling window.

```
import stumpy
import numpy as np
T = np.random.random(size=10000000)
m = 1000
stumpy.core.compute_mean_std(T, m)
```

The easiest fix I see is using the pandas.rolling functionality. This way we get a numerically stable, fast and optimized rolling standard deviation. The only problem is, that pandas is not an installation requirement and I understand if it should not become one.
The alternative is to adapt the old code, maybe in a fashion like below:

```
def compute_mean_std_nansupport(T, m):
    n = len(T)
    
    # Calculate the mean using a rolling window
    mean = np.mean(stumpy.core.rolling_window(T, m), axis = 1)
    
    # Remove nan values from T so cumsum produces correct values
    T = T.copy()
    T[np.isnan(T)] = 0

    # Calculate the std and variance using the formula Var[X] = E[X^2]-E[X]^2
    cumsum_T_squared = np.empty(len(T) + 1)
    np.cumsum(np.square(T), out=cumsum_T_squared[1:])
    cumsum_T_squared[0] = 0

    subseq_sum_T_squared = cumsum_T_squared[m:] - cumsum_T_squared[: n - m + 1]
    std = np.abs((subseq_sum_T_squared / m) - np.square(mean))
    std = np.sqrt(std)
    
    # Remove nan values from std and set the means to inf
    mean[np.isnan(mean)] = np.inf
    std[np.isnan(std)] = 0
    return mean, std
```

There is now a further copy of T and we will have to see if that is necessary. But since the calculation for big time series's happens only once in the algorithms, this might be ok.
On word on numerical stability: The approach from the original code calculates Var[X] = E[X^2]-E[X]^2 which can lead to precision errors (see https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance), and if you compare the results with the pandas version for a random series, the results are only equal up to three digits. I'm not sure if this might cause trouble for some users.

stumpy/core.py
```diff
@@ -471,24 +467,32 @@ def mass(Q, T, M_T=None, Σ_T=None):
...
     check_dtype(Q)
-    check_nan(Q)
 
     if Q.ndim != 1:  # pragma: no cover
         raise ValueError(f"Q is {Q.ndim}-dimensional and must be 1-dimensional. ")
+    m = Q.shape[0]
+
     T = np.asarray(T)
     check_dtype(T)
     check_nan(T)
```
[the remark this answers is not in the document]
mexxexx: Yes, I think this can be removed

stumpy/stamp.py
```diff
@@ -59,20 +61,22 @@ def mass(Q, T, M_T, Σ_T, trivial_idx=None, excl_zone=0, left=False, right=False
         if D[:trivial_idx].size:
             IL = np.argmin(D[:trivial_idx])
             PL = D[IL]
-        if zone_start <= IL < zone_stop:
-            IL = -1
+            if PL == np.inf or zone_start <= IL < zone_stop:
```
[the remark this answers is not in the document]
mexxexx: Of course, I forgot that

stumpy/stamp.py
```diff
@@ -126,15 +130,30 @@ def stamp(T_A, T_B, m, ignore_trivial=False):
...
+        )
     core.check_dtype(T_B)
-    core.check_nan(T_B)
+
     core.check_window_size(m)
     subseq_T_B = core.rolling_window(T_B, m)
     excl_zone = int(np.ceil(m / 2))
     M_T, Σ_T = core.compute_mean_std(T_A, m)
 
+    T_A[np.isnan(T_A)] = 0
```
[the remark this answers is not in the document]
mexxexx: No there should be no line like this. If a sub-sequence in T_B is nan (or inf, since inf gets replaced with nan in line 142), then the distance profile of this whole sub-sequence has to be set to np.inf. This happens in core.mass, line 485, and that why it is important to keep this information.

stumpy/stamp.py
```diff
@@ -126,15 +130,30 @@ def stamp(T_A, T_B, m, ignore_trivial=False):
...
+
```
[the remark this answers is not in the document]
mexxexx: 2. is almost correct, but we convert to NaNs because of the `np.nanstd` function that ignores NaNs, as we do not want NaNs in our final Σ_T, because of warings raised by numpy when dividing by NaN in the distance profile calculation. But, I just saw this when I was trying to fix the Memory Error, `np.nanstd([np.nan, np.nan]) = np.nan`, so if T contains more NaNs in a row than the window size, this still happens. So at the end of compute_mean_std there should be something like `Σ_T[np.isnan(Σ_T)] = 0`
3. correct. We can set an arbitrary value there, because we do not care about these subsequences and the matrix profile is going to be set to `np.inf` regardless.
4. The distance will be `np.inf` since its mean will be 'np.inf` (see stumpy.core line 304). Its stddev will be 0, as noted in point three. There are two possibilities that can happen:
Say we want to calculate the distance profile of a subsequence Q to T.
1. If Q contains any nan value, then the distance profile will be `np.inf` everywhere. If the mean of Q is infinity, then Q contained a nan value *before* we replaced them all by 0 (Step 3).
2. If Q does not contain any nan value, then we calculate the distance profile. However, T can have nan values, so we might have to set the distance profile at some points to infinity (for all sub-sequences in T that contain a nan value, i.e whose mean is np.inf). 

Both cases are covered in core line 352/359, in the calculation of the distance profile, because `D_squared[i] = np.inf` if and only if `M_T[i] = np.inf` or `μ_Q = np.inf`. So we maybe should remove line 141 from stomp (and similar lines in stump and stumped).

stumpy/stomp.py
```diff
@@ -101,36 +117,51 @@ def stomp(T_A, m, T_B=None, ignore_trivial=True):
...
         D = core.calculate_distance_profile(
             m, QT, μ_Q[i].item(0), σ_Q[i].item(0), M_T, Σ_T
         )
         if ignore_trivial:
             zone_start = max(0, i - excl_zone)
             zone_stop = min(k, i + excl_zone)
             D[zone_start:zone_stop] = np.inf
+
+        # Ignore distance profile if query mean is np.nan
+        if np.isinf(μ_Q[i]):
```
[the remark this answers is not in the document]
mexxexx: `np.isinf` is correct, see the comment above, but we might remove this line all together as this case is already handled in the calculation of the distance profile.

stumpy/stump.py
```diff
@@ -454,6 +471,10 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
         start, T_A, T_B, m, excl_zone, M_T, Σ_T, ignore_trivial
     )
 
+    T_B[
+        np.isnan(T_B)
+    ] = 0  # Remove all nan values from T_B only after first profile is calculated
```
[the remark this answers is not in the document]
mexxexx: Because the distance profile for the first sub-sequence is computed using core.mass. If the first sub-sequence contains a `np.nan` value and we replace it, mass has no way of knowing this.

tests/test_stamp.py
```diff
@@ -33,31 +15,89 @@ def naive_mass(Q, T, m, trivial_idx=None, excl_zone=0):
...
+        [utils.naive_mass(Q, T_A, m) for Q in core.rolling_window(T_B, m)], dtype=object
     )
     right = stamp.stamp(T_A, T_B, m)
     utils.replace_inf(left)
     utils.replace_inf(right)
-    npt.assert_almost_equal(left, right)
+    npt.assert_almost_equal(left[:, :2], right)
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
```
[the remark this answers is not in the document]
mexxexx: Yes, that is a pretty hand feature, see https://github.com/pytest-dev/pytest/issues/815

tests/test_stamp.py
```diff
@@ -33,31 +15,89 @@ def naive_mass(Q, T, m, trivial_idx=None, excl_zone=0):
     ),
 ]
 
+substitution_values = [np.nan, np.inf]
```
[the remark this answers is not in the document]
mexxexx: I was not sure how to do it. Do we then need to pass it with parametrize somehow?

tests/test_stomp.py
```diff
@@ -63,3 +66,59 @@ def test_stomp_A_B_join(T_A, T_B):
...
+
+
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize("substitute_B", substitution_values)
+def test_stomp_nan_inf_self_join(T_A, T_B, substitute_B):
+    m = 3
+
+    T_B_sub = T_B.copy()
+
+    substitution_locations = [slice(0, 0), 0, -1, slice(1, 3), [0, 3]]
```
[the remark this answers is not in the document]
mexxexx: Should we pass the whole array as a single argument? I put it inside of the function to bundle it all as a single test. If we parametrize it we get way more single tests.

tests/test_stumped_two_subsequences_inf_A_B_join.py
mexxexx: What exactly do you mean?

tests/test_stumped_two_subsequences_nan_inf_A_B_join.py
mexxexx: One single test or all eight tests again, just reversed?

seanlaw:
> The following code produces a Memory Error on my machine, because numpy.std internally creates a copy of the rolling window.

> import stumpy
> import numpy as np
> T = np.random.random(size=10000000)
> m = 1000
> a, b = stumpy.core.compute_mean_std(T, m)

So, I tried this piece of code (10M) on my laptop (a 2017 Macbook Pro with 16GB of RAM) and it ran fine within seconds. The total memory used at the end was around 315MB (note that I store the mean and std in `a` and `b`) . I then increased the `size` to `100,000,000` (100M) and it was also fine but it did require close to 5GB of memory (in total). This also completed within seconds. Frankly, on a laptop, it would be impossible to generate the matrix profile and so the `size` that would be supported for a laptop would probably be around `1,000,000` or so. 

May I ask what your memory (RAM) resources are? I'm surprised that you were able to hit a memory error with 10M data points but maybe I'm only seeing the final memory that was consumed and not the intermediate/temporary memory. 

I'm inclined to just leave it as you've implemented it (with NumPy only) and then deal with it later if it really becomes a problem. For 10M data points, I think requiring 8-16GB of RAM is reasonable in terms of minimum requirements.

mexxexx:
I have 8GB of RAM installed on my machine, and got the message that (with the code I posted) a memory allocation of 5.9 GB failed.

I agree that for the MP computation more RAM should be required anyways, but I still think that resources should not be wasted. Since the std calculations in all algorithms happen on one machine, there might still be trouble if trying to calculate the MP on lets say a Cluster.

stumpy/stamp.py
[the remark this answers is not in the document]
seanlaw: Maybe I'm missing something but wasn't the point of setting `np.nan` to zero (after the mean and std are computed) so that the dot product wouldn't be NaN? What makes T_A so special/different from T_B that only T_A `np.nan` values are set to zero (after computing the mean and std) but not T_B?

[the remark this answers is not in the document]
mexxexx: Because of the internal workings of `mass`.
It takes a sub-sequence (`Q`, here a subsequence of `T_B`) and calculates the distance profile to whole `T` (here `T_A`). If `Q` is none somewhere, no distance profile is calculated (if statement in core line 485). If T would be none anywhere, then the whole dot product `QT` fails to be computed, so we have to replace any nan values in T. If we would however replace a `nan` in T_B, and would call `mass` with a sub-sequence that should be ignored, we would have no way of telling this to mass, because no other parameter (like mu_Q) is passed to the function.

seanlaw:
> I agree that for the MP computation more RAM should be required anyways, but I still think that resources should not be wasted. Since the std calculations in all algorithms happen on one machine, there might still be trouble if trying to calculate the MP on lets say a Cluster.

I agree with you but I think we can approach this pragmatically. Fundamentally, the memory error for stddev computation is separate from what this PR solves. I think we should go with the NumPy version since it is computationally more stable (at the cost of some memory inefficiency). Improvements to the memory usage for computing the stddev/mean should be submitted as a new issue and addressed in a separate PR. Let's add it to your list above and file an issue.

seanlaw:
@mexxexx Hopefully, I understand the `np.nanstd` problem a little better and I think that there's an easy fix (at least, for 1D inputs). Since it's a rolling window calculation, there's no reason why we can't split large arrays into multiple parts, compute the rolling `np.nanstd` for each part, and then concatenate the parts back into one. 

Here is some example code:

```
import numpy as np
import stumpy
import math

# Compute all stddevs at once
T = np.random.rand(23)
m = 5
std = np.nanstd(stumpy.core.rolling_window(T, m), axis=1)
print(std)

# Compute stddevs in three chunks
# Not tested for multi-dimensional arrays
chunks = 3
chunk_size = math.ceil((T.shape[0] + 1)/chunks)
std_chunks = []
for chunk in range(chunks):
    start = chunk * chunk_size
    stop = min(start + chunk_size + m - 1, T.shape[0])
    tmp_std = np.nanstd(stumpy.core.rolling_window(T[start:stop], m), axis=1)
    std_chunks.append(tmp_std)
print(np.concatenate(std_chunks))  # We need to take care for multi-dimensional arrays!
```

So, in your case, we'd do some sort of `try`/`except` where we could continually double the chunk size until the `MemoryError` error doesn't happen anymore (i.e., when the size of the chunks are small enough to process).

```
power = 0
while True:
    n_chunks = 2**power
    try:
        chunked_std(T, m, n_chunks)  # This function represents the code above
        power = power * 2
        break
    except MemoryError:    
        pass
```

mexxexx:
@seanlaw I like this idea! For the multidimensional case it will be a bit trickier to adapt this rolling window, I'm not sure if it will work out of the box. But then I think that the same approach should work.

seanlaw:
> For the multidimensional case it will be a bit trickier to adapt this rolling window, I'm not sure if it will work out of the box. But then I think that the same approach should work.

@mexxexx I played around with the multi-dimensional case and it appears that the `core.rolling_window` works out of the box. I might be overlooking things but I think the only thing is needed is to specify the axes for `np.nanstd` and to reference the correct `shape`. Technically, EVERYTHING is 2D in shape (in NumPy-land) so we should be able to hardcode these in. Please see my example below and let me know if I've made any errors.

```
import numpy as np
import stumpy
import math
import numpy.testing as npt

T = np.random.rand(3, 26)
m = 5
full_std = np.nanstd(stumpy.core.rolling_window(T, m), axis=2)  # Note axis=2

chunks = 3
chunk_size = math.ceil((T.shape[1] + 1)/chunks)  # Note T.shape[1] instead of T.shape[0]
stds = []
for chunk in range(chunks):
    start = chunk * chunk_size
    stop = min(start + chunk_size + m - 1, T.shape[1])  # Note T.shape[1] instead of T.shape[0]
    tmp_std = np.nanstd(stumpy.core.rolling_window(T[:, start:stop], m), axis=2)  # Note axis=2
    stds.append(tmp_std)
chunked_std = np.hstack(stds)  # Note the use of hstack instead of concatenate

npt.assert_almost_equal(full_std, chunked_std)
```

mexxexx:
Very cool! That does look correct to me :)  I'll give it a try to see if it fixes the problem I had.

seanlaw:
> import numpy as np
import stumpy
import math
import numpy.testing as npt

> T = np.random.rand(3, 26)
m = 5
full_std = np.nanstd(stumpy.core.rolling_window(T, m), axis=2)  # Note axis=2

>chunks = 3
chunk_size = math.ceil((T.shape[1] + 1)/chunks)  # Note T.shape[1] instead of T.shape[0]
stds = []
for chunk in range(chunks):
    start = chunk * chunk_size
    stop = min(start + chunk_size + m - 1, T.shape[1])  # Note T.shape[1] instead of T.shape[0]
    tmp_std = np.nanstd(stumpy.core.rolling_window(T[:, start:stop], m), axis=2)  # Note axis=2
    stds.append(tmp_std)
chunked_std = np.hstack(stds)  # Note the use of hstack instead of concatenate

> npt.assert_almost_equal(full_std, chunked_std)

@mexxexx FYI, my last code comment is only for the multi-dimensional case. `MSTUMP`/`MSTUMPED` actually has its own `_multi_compute_mean_std` (that is separate from `compute_mean_std`) and it can be found in `mstump.py`. Does that make sense?

mexxexx:
Thanks for the clarification, I saw that before. I slightly modified your implementation of the 1D stddev (power shouldn't be starting at 0, better to start with n_chunks=1 and then to n_chunks *=2) and it is working without any issues now. Should we implement this directly into this PR? 

I also commented on a few of the changes you requested to clarify things, I'm not sure if you have seen it.

seanlaw:
@mexxexx I missed your comments. I will take a look this morning.

Yes, let’s add it to this PR 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

seanlaw:
@mexxexx I reviewed the comments for the requested changes but I don't see any new comments from you (I had responded to most/all of them a few days ago). Can you take a look and point me to anything that I may have missed?

seanlaw:
>  I slightly modified your implementation of the 1D stddev (power shouldn't be starting at 0, better to start with n_chunks=1 and then to n_chunks*=2) and it is working without any issues now

I think I see what you mean. I got a little bit too cute with it :)

mexxexx:
@seanlaw No problem :)
There are a few small points that are still open for me, or where I don't completely understand your suggestion:

1. You want to move the `substitution_locations` outside of the tests. How should I do it? Just move it outside and leave the array as a global variable?
2. In `test_stumped_two_subsequences_inf_A_B_join ` you commented that the syntax looks wrong. Could you explain what you mean exactly?
3. In the case of `test_stumped_two_subsequences_nan_inf_A_B_join` you wanted me to add a new test that swaps the location of `nan` and `inf`. Should I do this for all substitution locations creating eight new tests?

Thank you :)

seanlaw:
> 1. You want to move the substitution_locations outside of the tests. How should I do it? Just move it outside and leave the array as a global variable?

Yes, I was thinking that you could do something like:

```
substitution_locations = [(slice(0, 0), 0, -1, slice(1, 3), [0, 3])]  # Note the tuple parentheses

@pytest.mark.parametrize("substitution_locations", substitution_locations)
@pytest.mark.parametrize("T_A, T_B", test_data)
def some_test(T_A, T_B, substitution_location):
    # Some setup code
    for substitution_location_B in substitution_locations:
        for substitution_location_A in substitution_locations:
            T_A_sub[:] = T_A[:]
            T_B_sub[:] = T_B[:]
            T_A_sub[substitution_location_A] = substitute_A
            T_B_sub[substitution_location_B] = substitute_B
```

So, fundamentally, the test itself still iterates through a `substitution_locations` for-loop but this information is passed into the test function as a tuple. The benefit now is that the `substitution_locations` is defined once at the top of the file and reused everywhere (in all other tests). 

> 2. In test_stumped_two_subsequences_inf_A_B_join you commented that the syntax looks wrong. Could you explain what you mean exactly?

I actually commented on this but no worries if you missed it. I'll repeat it here:

I see what you are doing now. I didn't realize that the substitution_locations list had changed from

```
substitution_locations = [0, -1, slice(1, 3), [0, 3]]
```
to

```
substitution_locations = [(0, 1), (-1, slice(1, 3)), (slice(1, 3), 1), ([0, 3], 1)]
```

Your code is fine. Can you explain why the list changed? I would've expected that you would use the same list for both sequences. In other words, I would've expected:

```
substitution_locations = [0, -1, slice(1, 3), [0, 3]]  # Note that these are not tuples
@pytest.mark.parametrize("substitution_location_A", substitution_locations)
@pytest.mark.parametrize("substitution_location_B, substitution_locations)
```

But it looks like you are using different pairings for the locations. I guess that's fine. I just want to make sure there's nothing special going on that I'm missing. Keeping the list the same would make it easier for maintenance since I'm sure that I'll assume that it's the same as other substitution locations in the other files.

> 3. In the case of test_stumped_two_subsequences_nan_inf_A_B_join you wanted me to add a new test that swaps the location of nan and inf. Should I do this for all substitution locations creating eight new tests?

Also answered in the comments but will re-state here:  

I would do all eight. You could write a second test function and call it `test_stumped_two_subsequences_nan_inf_A_B_join_swap` or something like that

-----

Are you able to see my comments [here](https://github.com/TDAmeritrade/stumpy/pull/126/files). I wonder if there is a button that I need to press in order to release my comments to be viewed by you. I'm still learning how to more efficiently use these Github tools. The workflow is still a bit clunky for me. :)

mexxexx:
> Are you able to see my comments here. I wonder if there is a button that I need to press in order to release my comments to be viewed by you. I'm still learning how to more efficiently use these Github tools. The workflow is still a bit clunky for me. :)

I can only see the first comments you did, but not the one you made on point two for example...


> So, fundamentally, the test itself still iterates through a substitution_locations for-loop but this information is passed into the test function as a tuple. The benefit now is that the substitution_locations is defined once at the top of the file and reused everywhere (in all other tests).

This makes sense, I will implement it that way


> Your code is fine. Can you explain why the list changed? I would've expected that you would use the same list for both sequences

I did this to reduce the number of tests. I was using what you suggested, and stumped was called 16 times more often this way. Somehow this caused Dask to crash when running the tests so I reduced the number of test cases while maintaining all important test cases.

seanlaw:
@mexxexx I hit a button that I had neglected. Are you able to see my comments now?

mexxexx:
@seanlaw Yes, perfect :)
I'm working an solving all changes and will hopefully push something later that day.

tests/test_stump.py
```diff
@@ -17,19 +17,8 @@
     ),
 ]
 
-
-@pytest.mark.parametrize("Q, T", test_data)
-def test_calculate_squared_distance_profile(Q, T):
```
seanlaw: I think this test should be moved to `test_core.py`

seanlaw: Maybe I'm just not seeing `test_core.py` as a changed file but the test coverage should've brought it up as a function that hasn't been tested

seanlaw:
@mexxexx I have some questions for you:

1. I was wondering where you are working from? I take it from the commit times that you are working somewhere in Europe? 
2. What applications do you have for computing the matrix profile?
3. Do you have a Twitter handle? I'd like to thank you publicly for our next release! 
4. What do you do for your day job?

stumpy/core.py
```diff
@@ -357,11 +389,58 @@ def calculate_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
...
         D_squared[:] = m
     D_squared[Σ_T < threshold] = m
     D_squared[(Σ_T < threshold) & (σ_Q < threshold)] = 0
+    D_squared[inf_mask] = np.inf
+
+    return D_squared
+
+
+@njit(parallel=True, fastmath=True)
+def calculate_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T):
```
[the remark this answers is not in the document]
mexxexx: I agree, that sounds reasonable.

mexxexx:
@ seanlaw Thanks for the review! I decided to do the whole mean std calculation chunked.

> I was wondering where you are working from? I take it from the commit times that you are working somewhere in Europe?

Exactly, I'm working from Switzerland at the moment :)

> Do you have a Twitter handle? I'd like to thank you publicly for our next release!

I don't use twitter, but that you for your consideration!

> What applications do you have for computing the matrix profile?

I am working on an ECR-Ion Source that is used to create lead Ions. We create a stream of these Ions, a current, and we want to keep it as intense and stable as possible, because these Ions are later on used for physics experiments.You can imagine the machine as a very sensitive microwave, that requires constant tuning by an operator. The goal is to support the operator in the tuning process, because it is very complex and not well understood. We try to analyze the data stream from our diagnostic tools to search for recurring patterns from which we hope to deduce insights about the inner workings of our machine and see if we can use any signals for predicting upcoming failure of the system.

> What do you do for your day job?

Basically this. I explore various methods to reach this goal and right now work intensely to see if the Matrix Profile can serve our needs.

seanlaw:
> @seanlaw Of course, I will keep it in this message and update it, if something else comes into my mind.
> 
> * [ ]  Add nan/inf support to mstump + tests
> * [ ]  Add nan/inf support to mstumped + tests
> * [ ]  Add nan/inf support to gpu_stump + tests
> * [ ]  Change check_nan to check_nan_inf and giving a warning
> * [ ]  Doing this nan/inf check in the 1D algorithms
> * [x]  Implement the rolling chunked stddev for 1D
> * [ ]  Implement the rolling chunked stddev for multi dimensional timeseries

@mexxexx Thanks again for supporting this PR and congratulations on the successful merge! Do you mind updating the above checklist (I think the last item is done but feel free to add more) and then I can go ahead and create new issues for any outstanding items. I'll attempt to work on them in the coming months but if would be really great if you had the bandwidth to help (absolutely no pressure) or at least assist me if/when I get stuck.

Also, I'd appreciate any and all feedback and criticism as it relates to the PR process, the overall code base (i.e., ways that we can improve readability and maintainability, etc), or anything you can think of to improve STUMPY and the community we are trying to foster. I'm all ears! Thanks in advance

mexxexx:
@seanlaw Finally :smile: 
The rolling chunked std is still not implemented, even if the code you posted should work. I didn't want to overload this PR.

I'll definitely assist with these points if I can. Depending on how promising the results I can achieve using the 1D Algorithms, I might want to do the same thing in a few more dimensions. So I might have a look at these points in the near future.

Overall, I think that the code base is really clear. I didn't have a lot of problems navigating around and I like the comments for all functions. I don't know a lot about software development and I'm sure that there are some ways to improve (eg merging to the master branch only when releasing), but I think it is fine, as stumpy comparatively small.

Issue #129, opened 2020-02-28
Move MSTUMP / MSTUMPED to Using Rolling Chunked STDDEV


seanlaw:
Currently, `core.compute_mean_std` might be able to handle chunked STDDEV calculation for multi-dimensional times series data. However, `mstump`/`mstumped` are not currently using this function to compute its STDDEV. Instead, `mstump`/`mstumped` uses `mstump._multi_compute_mean_std`.

mexxexx:
I'm taking care of this issue in my implementation for the handling of nan/inf values

seanlaw:
I just wanted to break this out so that each PR is small instead of one big PR. It makes it a lot easier to review too

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

seanlaw:
It just occurred to me that maybe we should move `mstump._multi_compute_mean_std` into `test_mstump.py` and call it `naive_multi_compute_mean_std` since the current test is using NumPy directly. So, our (future) chunked implementation (NumPy) and test (NumPy) are nearly identical and this is bad. We should test the future chunked implementation against a modified version of `mstump._multi_compute_mean_std`.

The same thing is true regarding `core.test_compute_mean_std`. It is currently testing against NumPy. We should probably replace the naive implementation with the original `core.compute_mean_std`. 

To be clear, I don't think we need to test `np.inf`/`np.nan` in these cases. This should be handled by NumPy's own testing. For the most part, we should trust that NumPy is correctly computing mean/stddev and what we should be testing is whether our function returns back the promised mean/stddev pair. Just thinking out loud here.

mexxexx:
> I just wanted to break this out so that each PR is small instead of one big PR. It makes it a lot easier to review too

I agree. But right now I have everything almost ready in one big pull pull request. Any suggestions?

mexxexx:
I modified `core.compute_mean_std` in the following way

```
def compute_mean_std(T, m, num_chunks=1, max_iter=10):
    if len(T.shape) == 1:
        shape = 0
    elif len(T.shape) == 2:
        shape = 1
    else:  # pragma nocover
        raise ValueError("T has to be one or two dimensional!")

    for iteration in range(max_iter):
        try:
            chunk_size = math.ceil((T.shape[shape] + 1) / num_chunks)

            mean_chunks = []
            std_chunks = []
            for chunk in range(num_chunks):
                start = chunk * chunk_size
[...]

PR #134, opened 2020-03-02
Fixed Issue #129 Rolling stddev in MSTUMP/MSTUMPED

mexxexx:
This PR fixes issue #129 and improves the tests of `core.compute_mean_std`.

### DIFF
--- a/stumpy/__init__.py
+++ b/stumpy/__init__.py
@@ -8,7 +8,6 @@ from .mstump import (  # noqa: F401
     _mstump,
     _get_first_mstump_profile,
     _get_multi_QT,
-    _multi_compute_mean_std,
     _multi_mass,
 )
 from .mstumped import mstumped  # noqa: F401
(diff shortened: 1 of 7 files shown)


seanlaw:
@mexxexx Everything looks good! 

I only have one main request: Since `naive_compute_mean_std` and `naive_compute_mean_std_multidimensional` are really only called in `test_core.py` and nowhere else, we should just move those functions into the top of `test_core.py`. The only reason why I moved `naive_mass` into `utils.py` is because it was being used in multiple test files.

mexxexx:
I see! I thought `utils.py` was meant as a module collecting all functions that are not tests. I will move it

seanlaw:
Yeah, I use to have `naive_mass` in every major test file and realized that I needed to update all of them if there was one change. So, I refactored it and put everything in one file so that it would be easier to maintain. 

With that being said, if we find that `naive_compute_mean_std` and `naive_compute_mean_std_multidimensional` are being called in other test files (in the future) then it would make sense to move them into `utils.py` as well.

Issue #136, opened 2020-03-02
Refactor _multi_mass to return multidimensional distance profile


mexxexx:
As mentioned in issue #135 the `mstump._multi_mass` function has to be refactored to return the multi dimensional distance profile.

#### _multi_mass 
Throughout the code we use the mass algorithm in two different ways: In `core.mass` it is used to calculate the distance profile between a query `Q` and a time series `T`. Then we have `stamp.mass` which first calculates the distance profile `D` between `Q` and a time series `T`, then applies an exclusion zone and in the end returns the minimum value of `D` as well as its position. This is then used to calculate a *single* value of the matrix profile.

So far so good. But now we also have `mstump._multi_mass`. Here we not only calculate a single value of the Matrix profile, but we return a whole array `P`. `P` is initialized with `np.inf` and later we find the lines
 
```
for i in range(d):
        D_prime = D_prime + D[i]
        D_prime_prime = D_prime / (i + 1)
        # Element-wise Min
        col_mask = P[i] > D_prime_prime #######
        P[i, col_mask] = D_prime_prime[col_mask]
        I[i, col_mask] = trivial_idx
```
So the line I marked with ####### will always return an array that is True everywhere (exept where `D_prime_prime=np.inf` but this case does not change anything), so effectively we are returning the d-dimensional distance profile between `Q` and `T` here. 
But at the same time we apply an exclusion zone
```
zone_start = max(0, trivial_idx - excl_zone)
zone_stop = min(k, trivial_idx + excl_zone)
D[:, zone_start : zone_stop + 1] = np.inf
```
which gives us something in between `core.mass` and `stamp.mass`. For better readability and extensibility I would suggest dropping the application of an exclusion zone and the calculation of the indices in the last line, so that `mstump._multi_mass` returns the actual d-dimensional distance profile and so that it is in line with its one dimensional brother in `core`. (Of course, then some changes have to be made, in particular in `_calculate_first_profile_index`)

Issue #135, opened 2020-03-02
Refactor MSTUMP


mexxexx:
While working on issue #117 I realized that the MSTUMP module has some opportunities for improving code readability and consistency, as well as significantly improving performance. I want to discuss these points here:

#### _multi_mass 
Throughout the code we use the mass algorithm in two different ways: In `core.mass` it is used to calculate the distance profile between a query `Q` and a time series `T`. Then we have `stamp.mass` which first calculates the distance profile `D` between `Q` and a time series `T`, then applies an exclusion zone and in the end returns the minimum value of `D` as well as its position. This is then used to calculate a *single* value of the matrix profile.

So far so good. But now we also have `mstump._multi_mass`. Here we not only calculate a single value of the Matrix profile, but we return a whole array `P`. `P` is initialized with `np.inf` and later we find the lines
 
```
for i in range(d):
        D_prime = D_prime + D[i]
        D_prime_prime = D_prime / (i + 1)
        # Element-wise Min
        col_mask = P[i] > D_prime_prime #######
        P[i, col_mask] = D_prime_prime[col_mask]
        I[i, col_mask] = trivial_idx
```
So the line I marked with ####### will always return an array that is True everywhere (exept where `D_prime_prime=np.inf` but this case does not change anything), so effectively we are returning the d-dimensional distance profile between `Q` and `T` here. 
But at the same time we apply an exclusion zone
```
zone_start = max(0, trivial_idx - excl_zone)
zone_stop = min(k, trivial_idx + excl_zone)
D[:, zone_start : zone_stop + 1] = np.inf
```
which gives us something in between `core.mass` and `stamp.mass`. For better readability and extensibility I would suggest dropping the application of an exclusion zone and the calculation of the indices in the last line, so that `mstump._multi_mass` returns the actual d-dimensional distance profile and so that it is in line with its one dimensional brother in `core`. (Of course, then some changes have to be made, in particular in `_calculate_first_profile_index`)

#### sorting

The most time consuming part of `mstump` (that I saw using cProfile) is the sorting in lines 378-381
```
# Column-wise sort
for col in range(k):
    # row_idx[:, col] = np.argsort(D[:, col])
    # D[:, col] = D[row_idx[:, col], col]
    D[:, col] = np.sort(D[:, col])
```
[...]

PR #137, opened 2020-03-03
Fixed Issue #136 and improved mstump tests

mexxexx:
I tried to stick as much to the issue #136 as I could, but I had to make some changes that touch issue #135 .

- The main change is that I converted `_multi_mass` to return the multidimensional distance profile as discussed. 
- I also changed `_get_first_mstump_profile` in a way so that it only returns the profile values and indices at index `start` so that this function is consistent with `stump._get_first_stump_profile`. I made this change in this PR because I had to move things to this function, e.g. the application of an exclusion zone, and it made sense to do it here. This does break the anytime ability of mstump though (because the distance profile of `T[start:start+m]` does not influence the whole matrix profile anymore). This is why I had to slightly change the logic of `_mstump`. I tried to leave it as untouched as possible. Also minor changes to `mstumped` had to be made.
- Some tests were improved to actually use the defined functions. I also moved the `naive_mstump` function to `utils.py` to be able to reuse it in `test_mstumped.py`.

Let me know what you think.

### DIFF
--- a/stumpy/mstump.py
+++ b/stumpy/mstump.py
@@ -12,10 +12,10 @@ from . import core
 logger = logging.getLogger(__name__)
 
 
-def _multi_mass(Q, T, m, M_T, Σ_T, trivial_idx, excl_zone):
+def _multi_mass(Q, T, m, M_T, Σ_T):
     """
     A multi-dimensional wrapper around "Mueen's Algorithm for Similarity Search"
-    (MASS) to compute multi-dimensional MASS.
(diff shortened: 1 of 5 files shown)


seanlaw:
@mexxexx Great! I will review this today

seanlaw:
I did a quick scan I think everything looks good. Just a heads up that I am overhauling the way that the `dask_client` unit tests are being run so you'll have to pull the new tests into your branch soon. I should be done in about an hour so please be aware so that you don't stay out of sync.

Issue #130, opened 2020-03-01
Handling NaN and inf Values for MSTUMPED


seanlaw:
This is a follow up to #115 to handle NaN/inf values in `mstumped`

seanlaw:
Note that this is the Dask Distributed version. The single server version `mstump` can be found in #117

seanlaw:
@mexxexx I've broken it up so that we can have smaller PRs instead of one giant one if possible.

PR #142, opened 2020-03-13
Fixed #117 and #130

### LINKED ISSUE #117: Handling NaN and inf Values for MSTUMP
This is a follow up to #115 to handle NaN/inf values in `mstump`

mexxexx:
This PR fixes the handling of nans in mstump and mstumped (issues #117  and #130 ). It also contains a bit of refactoring in order to clean the code (issue #135 )

The refactoring was done to improve readability and to have the internals more or less identical to `stump`.

### DIFF
--- a/stumpy/mstump.py
+++ b/stumpy/mstump.py
@@ -373,7 +373,7 @@ def mstump(T, m):
         start, T_A, T_B, m, excl_zone, M_T, Σ_T
     )
 
-    T_B[np.isnan(T_A)] = 0
+    T_B[np.isnan(T_B)] = 0
 
     QT, QT_first = _get_multi_QT(start, T_A, m)
 
--- a/stumpy/mstump.py
+++ b/stumpy/mstump.py
@@ -59,7 +59,7 @@ def _multi_mass(Q, T, m, M_T, Σ_T):
     return D
 
 
-def _get_first_mstump_profile(start, T, m, excl_zone, M_T, Σ_T):
+def _get_first_mstump_profile(start, T_A, T_B, m, excl_zone, M_T, Σ_T):
     """
     Multi-dimensional wrapper to compute the multi-dimensional matrix profile
     and multi-dimensional matrix profile index for a given window within the
@@ -72,9 +72,12 @@ def _get_first_mstump_profile(start, T, m, excl_zone, M_T, Σ_T):
         The window index to calculate the first matrix profile, matrix profile
         index, left matrix profile index, and right matrix profile index for.
 
-    T : ndarray
+    T_A : ndarray
         The time series or sequence for which the matrix profile index will
         be returned
+
+    T_B : ndarray
+        The time series or sequence that contains your query subsequences
 
     m : int
         Window size
@@ -98,8 +101,8 @@ def _get_first_mstump_profile(start, T, m, excl_zone, M_T, Σ_T):
         equal to `start`
     """
 
-    d, n = T.shape
-    D = _multi_mass(T[:, start : start + m], T, m, M_T, Σ_T)
+    d, n = T_A.shape
+    D = _multi_mass(T_B[:, start : start + m], T_A, m, M_T, Σ_T)
 
     zone_start = max(0, start - excl_zone)
     zone_stop = min(n - m + 1, start + excl_zone)
@@ -159,22 +162,7 @@ def _get_multi_QT(start, T, m):
 
 @njit(parallel=True, fastmath=True)
 def _mstump(
-    T,
-    m,
-    P,
-    I,
-    D,
-    D_prime,
-    range_stop,
-    excl_zone,
-    M_T,
-    Σ_T,
-    QT,
-    QT_first,
-    μ_Q,
-    σ_Q,
-    k,
-    range_start=1,
+    T, m, range_stop, excl_zone, M_T, Σ_T, QT, QT_first, μ_Q, σ_Q, k, range_start=1
 ):
     """
     A Numba JIT-compiled version of mSTOMP, a variant of mSTAMP, for parallel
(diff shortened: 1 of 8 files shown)


seanlaw:
I'm not sure why the tests aren't being ran here.

stumpy/mstump.py
```diff
@@ -59,7 +59,7 @@ def _multi_mass(Q, T, m, M_T, Σ_T):
     return D
 
 
-def _get_first_mstump_profile(start, T, m, excl_zone, M_T, Σ_T):
+def _get_first_mstump_profile(start, T_A, T_B, m, excl_zone, M_T, Σ_T):
```
seanlaw: This implies that we are prepared to perform AB-joins in addition to self-joins?

stumpy/mstump.py
```diff
@@ -98,8 +101,8 @@ def _get_first_mstump_profile(start, T, m, excl_zone, M_T, Σ_T):
         equal to `start`
     """
 
-    d, n = T.shape
-    D = _multi_mass(T[:, start : start + m], T, m, M_T, Σ_T)
+    d, n = T_A.shape
+    D = _multi_mass(T_B[:, start : start + m], T_A, m, M_T, Σ_T)
 
     zone_start = max(0, start - excl_zone)
     zone_stop = min(n - m + 1, start + excl_zone)
```
seanlaw: For AB-joins, this is not necessary. This is only needed for self-joins

stumpy/mstumped.py
```diff
@@ -58,63 +58,55 @@ def mstumped(dask_client, T, m):
     See mSTAMP Algorithm
     """
 
-    hosts = list(dask_client.ncores().keys())
-    nworkers = len(hosts)
+    T_A = np.asarray(core.transpose_dataframe(T)).copy()
+    T_B = T_A.copy()
```
seanlaw: I am a bit confused as to why we need T_B? It doesn't look like its being used for anything

tests/test_mstump.py
```diff
@@ -50,7 +55,7 @@ def test_get_first_mstump_profile(T, m):
     left_I = left_I[start, :]
 
     M_T, Σ_T = core.compute_mean_std(T, m)
-    right_P, right_I = _get_first_mstump_profile(start, T, m, excl_zone, M_T, Σ_T)
+    right_P, right_I = _get_first_mstump_profile(start, T, T, m, excl_zone, M_T, Σ_T)
```
seanlaw: This seems weird. I feel like `_get_first_mstump_profile` should detect whether one or two time series are being passed to it

seanlaw:
When I run the tests locally with the master branch (prior to this PR), I notice that the tests are consistently getting stuck on `tests/test_stumped_two_constant_subsequences.py`. This happens on Azure Pipelines as well. We may need to break this down further again.

stumpy/mstump.py
mexxexx: For the first mstump profile yes, then we can directly use this function.

Currently I implemented it here to implement `nan`/`inf` handling the same way as in `stump`. There (for the _first_ subsequence) all `nans` in `T_A` are replaced with zeros, while `T_B` still contains them, because otherwise there is no way to pass this information to `core.mass`.

To actually support AB-joins, there also have to be done some changes in `_stump` itself that are currently a bit out of range.

stumpy/mstump.py
mexxexx: Yes, but I did not plan to support AB-joins at the moment.

tests/test_mstump.py
mexxexx: 'stump._get_first_stump_profile` doesn't, so I kept it this way in `mstump`. The detection is done in the main `stump` function.

stumpy/mstumped.py
mexxexx: I wanted to implement nan/inf handling the same way as we did in `stump`. There you first calculate the `mean/stddev` and for each subsequence to ignore you set the mean to `inf`. Then you set `T_A=0` everywhere, where it was `nan`. Then you calculate the first matrix profile value, and here `T_B` can still be `nan`, because this information has to be passed to `core.mass`. Afterwards you set `T_B` to zero everywhere where it was `nan` and pass those two altered timeserieses to `_stump`, where the decision if one subsequence should be ignored is made only by looking at the mean and checking if it is `inf`.

So actually to support AB-joins we only need to pass `T_B` to `_mstump` and handle it from there, but I skipped this as this is a completely different issue.

Basically, at the moment `T_B` is only used to correctly calculate the first matrix profile value.

stumpy/mstump.py
seanlaw: Frankly, I'm not sure what an AB-join for multi-dimensional matrix profile actually means. I don't remember the original authors discussing it and I am concerned with supporting this for users if we haven't thought through it.

> To actually support AB-joins, there also have to be done some changes in _stump itself that are currently a bit out of range.

I'm not sure I understand, technically, `_stump` already supports AB-joins. Or, do you mean AB-joins that contain `nan`/`inf` in both sequences?

seanlaw: Alright, I think I am okay with this since it is a private function

tests/test_mstump.py
seanlaw: Got it! I overlooked the fact that:

1. This is a private function
2. This sets us up in case we decide to allow AB-joins in the future (currently, this is blocked/omitted in the public function)

seanlaw:
@mexxexx Thank you for completing this PR! This is another great step in getting STUMPY to the next level. I really appreciate it.

Additionally, given the current global situation, I hope that you and your loved ones are staying safe out there.

mexxexx:
Thank you very much, I really appreciate it. I wish you all the best too!

seanlaw:
Btw, there was indeed an issue with Azure Pipelines (they changed some things that caused CI to not get triggered). I ended up adding a couple of new lines to the `azure-pipelines.yml` file. So, when this finishes testing and gets merged, please remember to pull the latest code.

PR #151, opened 2020-03-31
[Work in Progress] Fix #143 mstump sort

### LINKED ISSUE #143: MSTUMP columnwise sort performance
Some profiling showed that the current approach to the necessary column-wise sort in MSTUMP is far from efficient.

I tried `mstump` using the two different sorting techniques and I got the following results (`m=10`):
|   | `n=10000, d=5` | `n=50000, d=5` | `n=10000, d=20` | CPU usage |
| ------------- | ------------- | ------------- | ------------- | ------------- | 
| sort_numba | 69s  | 80s | 1741s |~390% |
| sort_numpy  | 9s | 49s | 268s |~210-220% |

I did a single run with every configuration, precompiled mstump and tried to leave my computer running without doing anything else. 

So while the numpy version does not parallelize well, it is still faster by a large margin. We should check if somehow the numpy sorting algorithm can be leveraged to numba, for example in parallel chunks.

_Originally posted by @mexxexx in https://github.com/TDAmeritrade/stumpy/issues/135#issuecomment-597058818_

mexxexx:
So this here is a first draft for improving performance in `mstump` using parallel numpy sorting instead of numba. `mstumped` uses single core numpy sorting, but we should look for ways to use multiprocessing.

I did a really small benchmark to get a feeling (times always excluding compilation): 

```
>>> T = numpy.random.uniform(size=(5, 10000))
>>> m = 10

>>> stumpy.stump(T[0], m)
around 1.5s
>>> stumpy.mstump(T, m) # As in this PR
around 18s
>>> stumpy.mstump(T, m) # Sorting using single core, default numpy
around 12s
>>> stumpy.mstump(T, m) # stumpy v1.3.1
around 72s
```

```
>>> T = numpy.random.uniform(size=(5, 50000))
>>> m = 10

>>> stumpy.stump(T[0], m)
around 10s
>>> stumpy.mstump(T, m) # As in this PR, 8 Cores
around 212s
>>> stumpy.mstump(T, m) # Sorting using single core, default numpy
around 268s
```

```
>>> T = numpy.random.uniform(size=(15, 10000))
>>> m = 10

>>> stumpy.mstump(T, m) # As in this PR, 8 Cores
around 37s
>>> stumpy.mstump(T, m) # Sorting using single core, default numpy
around 45s
```

We can see that for small time series the single core variant outperforms the multi core one because of overhead, but this changes for larger `n`. Yet, using eight cores compared to one does not make amazing improvements. Still, it is way faster than the current implementation. `mstump` runtime should be `O(d log d n²)` according to the paper, but the jump from `stump` to `mstump` seams still really high.

I'm not sure if there are other improvements that can be made. I feel like it should be able to mitigate some overhead, because in this PR we have a numba function that spawns different processes (`_compute_multi_D`), than `Pool.map` spawning different processes, and then again a parallel numba function (`_compute_PI`)

### DIFF
--- a/stumpy/mstump.py
+++ b/stumpy/mstump.py
@@ -4,6 +4,9 @@
 
 import logging
 
+import multiprocessing
+import os
+import ctypes
 import numpy as np
 from numba import njit, prange
 
@@ -205,8 +208,11 @@ def _compute_multi_D(
     # D = np.sqrt(D)
 
 
-def _sort_columnwise(D):
-    D.sort(axis=0)
+_D = None
+
+
+def _sort_columnwise(s):
+    _D[:, s].sort(axis=0)
 
 
 @njit(parallel=True, fastmath=True)
@@ -223,8 +229,43 @@ def _compute_PI(d, idx, D, D_prime, range_start, P, I):
             I[i, pos] = -1
 
 
+def _prepare_parallel(d, k, num_cores):
+    shared_array_base = multiprocessing.Array(ctypes.c_double, d * k)
+
+    global _D
+    _D = np.frombuffer(shared_array_base.get_obj())
+    _D = _D.reshape(d, k)
+
+    if num_cores is None:
+        num_cores = os.cpu_count()
+    if num_cores is None or k < num_cores:
+        num_cores = 1
+
+    if num_cores == 1:
+        return None, None
+    else:
+        pool = multiprocessing.Pool(num_cores)
+
+        step = 1 + k // num_cores
+        slices = [slice(step * i, step * (i + 1)) for i in range(num_cores)]
+
+        return pool, slices
+
+
 def _mstump(
-    T, m, range_stop, excl_zone, M_T, Σ_T, QT, QT_first, μ_Q, σ_Q, k, range_start=1
+    T,
+    m,
+    range_stop,
+    excl_zone,
+    M_T,
+    Σ_T,
+    QT,
+    QT_first,
+    μ_Q,
+    σ_Q,
+    k,
+    range_start=1,
+    num_cores=None,
 ):
     """
     A Numba JIT-compiled version of mSTOMP, a variant of mSTAMP, for parallel
@@ -275,6 +316,10 @@ def _mstump(
         The starting index value along T_B for which to start the matrix
         profile calculation. Default is 1.
 
+    num_cores : int or None
+        The number of cores to be used for columnswise sorting. If one it defaults to
+        a pure numpy implementation. If None, os.cpu_count() cores are used.
+
     Returns
     -------
     P : ndarray
@@ -300,15 +345,34 @@ def _mstump(
 
     P = np.empty((d, range_stop - range_start))
     I = np.empty((d, range_stop - range_start))
-    D = np.empty((d, k))
     D_prime = np.empty(k)
+
+    pool, slices = _prepare_parallel(d, k, num_cores)
 
     for idx in range(range_start, range_stop):
         _compute_multi_D(
-            d, k, idx, D, T, m, excl_zone, M_T, Σ_T, QT_even, QT_odd, QT_first, μ_Q, σ_Q
+            d,
+            k,
+            idx,
+            _D,
+            T,
+            m,
+            excl_zone,
+            M_T,
+            Σ_T,
+            QT_even,
+            QT_odd,
+            QT_first,
+            μ_Q,
+            σ_Q,
         )
-        _sort_columnwise(D)
-        _compute_PI(d, idx, D, D_prime, range_start, P, I)
+
+        if pool is None:
+            _sort_columnwise(slices)
+        else:
+            pool.map(_sort_columnwise, slices)
+
+        _compute_PI(d, idx, _D, D_prime, range_start, P, I)
 
     return P, I
 
@@ -391,7 +455,7 @@ def mstump(T, m):
     QT, QT_first = _get_multi_QT(start, T_A, m)
 
     P[:, start + 1 : stop], I[:, start + 1 : stop] = _mstump(
-        T_A, m, stop, excl_zone, M_T, Σ_T, QT, QT_first, μ_Q, σ_Q, k, start + 1
+        T_A, m, stop, excl_zone, M_T, Σ_T, QT, QT_first, μ_Q, σ_Q, k, start + 1, None
     )
 
     return P.T, I.T
--- a/stumpy/mstump.py
+++ b/stumpy/mstump.py
@@ -161,6 +161,68 @@ def _get_multi_QT(start, T, m):
 
 
 @njit(parallel=True, fastmath=True)
+def _compute_multi_D(
+    d, k, idx, D, T, m, excl_zone, M_T, Σ_T, QT_even, QT_odd, QT_first, μ_Q, σ_Q
+):
+    for i in range(d):
+        # Numba's prange requires incrementing a range by 1 so replace
+        # `for j in range(k-1,0,-1)` with its incrementing compliment
+        for rev_j in prange(1, k):
+            j = k - rev_j
+            # GPU Stomp Parallel Implementation with Numba
+            # DOI: 10.1109/ICDM.2016.0085
+            # See Figure 5
+            if idx % 2 == 0:
+                # Even
+                QT_even[i, j] = (
+                    QT_odd[i, j - 1]
+                    - T[i, idx - 1] * T[i, j - 1]
+                    + T[i, idx + m - 1] * T[i, j + m - 1]
+                )
+            else:
+                # Odd
+                QT_odd[i, j] = (
+                    QT_even[i, j - 1]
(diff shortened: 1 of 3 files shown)


mexxexx:
Okay, something is going wrong on windows, and I don't know what...

seanlaw:
> Okay, something is going wrong on windows, and I don't know what...

@mexxexx I took a [look at the error](https://dev.azure.com/stumpy-dev/stumpy/_build/results?buildId=451&view=logs&j=8f7fd32b-564f-5020-75ec-897e474f4673&t=4ee92fed-b5d6-52bd-6b13-c23aa10ae57a) but it is hard to tell. Multiprocessing in Windows is notoriously bad/unreliable.

Looking at your numbers, I actually think that it is "better"/"fairer" to compare 2-core parallel with 1-core NumPy sort (because more cores will inevitably solve/speed up the process).  In other words, if we had 16 or 32 cores then we'd probably beat NumPy sort. This is just an aside comment as I found myself trying not let the timing distract me too much since the number of cores is 8x that of `np.sort` which uses a single core?

>  Sorting using single core, default numpy

What does this mean? In v1.3.1, we aren't actually using np.sort from NumPy but, instead, it is Numba JIT-compiled sorting. To me, this quote implies that Numba is able to call out to separate function that, in itself, calls `np.sort` from NumPy (which would be great!). Can you please clarify?

mexxexx:
> Looking at your numbers, I actually think that it is "better"/"fairer" to compare 2-core parallel with 1-core NumPy sort (because more cores will inevitably solve/speed up the process). In other words, if we had 16 or 32 cores then we'd probably beat NumPy sort. This is just an aside comment as I found myself trying not let the timing distract me too much since the number of cores is 8x that of np.sort which uses a single core?

Okay, I see your point. For this we can just set `num_cores` to 2 (in `mstump.py` line 458 change `None` to `2`)

> What does this mean? In v1.3.1, we aren't actually using np.sort from NumPy but, instead, it is Numba JIT-compiled sorting. To me, this quote implies that Numba is able to call out to separate function that, in itself, calls np.sort from NumPy (which would be great!). Can you please clarify?

Sure, I might have been a bit brief. So:

> `# As in this PR`
Here I used exactly the code I submitted in the PR, such that sorting will use all available cores.

> `# Sorting using single core, default numpy`
Here I limited sorting to one core, by changing `num_cores=None` in `mstumped.py` to `num_cores=1` This way the code falls back using the default numpy sort without even invoking any multiprocessing library (See lines `370-371` and `244-255`. Passing `num_cores=1` defaults to no `Pool` beeing created). This is also used for `mstumped` right now.

seanlaw:
Additionally, for testing purposes, you could also try submitting a new PR where you:

1. Add a new function(s) for strictly testing parallel sorting as you've shown above (maybe wrapped inside a Numba call)
2. Add a new unit test called `tests/test_parallel.py` that tests this function
3. Update `test.sh` to only test `test_parallel.py` (i.e., don't test any of the other test functions in the `tests/` directory
4. Submit the PR 

The goal here is to isolate and then be able to perform the incremental testing on all three platforms (windows, mac, linux) without having to run ALL of the tests. At least, not until you are able to understand the behavior.

I recognize that this is a lot of extra work so please don't take this as me pressuring you to do it. 😄Just thinking out loud here

stumpy/mstump.py
```diff
@@ -161,8 +164,108 @@ def _get_multi_QT(start, T, m):
...
+            )
+
+    zone_start = max(0, idx - excl_zone)
+    zone_stop = min(k, idx + excl_zone)
+    D[:, zone_start : zone_stop + 1] = np.inf
+
+    # D = np.sqrt(D)
+
+
+_D = None
```
seanlaw: Should this be de-indented here? It kinda looks like it's the middle of nowhere and is therefore a global variable?

seanlaw:
> Here I limited sorting to one core, by changing num_cores=None in mstumped.py to num_cores=1 This way the code falls back using the default numpy sort without even invoking any multiprocessing library (See lines 370-371 and 244-255. Passing num_cores=1 defaults to no Pool beeing created). This is also used for mstumped right now.

Thanks for the clarification! Okay, after quickly glancing through the code, it looks like you are able to call `np.sort` separately because you've split out the distance profile calculation (using Numba) from sorting the distance profile (using NumPy exclusively). This is SO interesting to me! I'm trying to wrap my head around whether there are any performance penalties (in both speed and memory efficiency - specifically, as long as we are never storing an entire `n-sqaured` distance matrix in memory then we are good) but, if not, I really love what you've done here! It's like we get the best of both worlds. Great work!

Honestly, considering that it takes 8 cores to get the same performance as a single core `np.sort`, I'm inclined to simply replace:

```
if pool is None:
    _sort_columnwise(slices)
else:
    pool.map(_sort_columnwise, slices)
```

with 

```
np.sort(slices)
```

And remove all of the multiprocessing parts and just call it a day. According to the numbers you showed above, `np.sort` is only ~20% slower than the 8-core version but the code is far, far cleaner and I think that is worth the tradeoff of not having to maintain `multiprocessing` code and having to solve the Windows problem.

stumpy/mstump.py
mexxexx: As far as I have understood multiprocessing with numpy, this has to be a global variable, which is why I made it "private".
The idea is to allocate shared memory using `multiprocessing.array` and then wrapping this chunk of memory with a global numpy array using `np.frombuffer`. This way all spawned processes get access to the same chunk of memory instead of everyone getting its own copy.

mexxexx:
I totally agree that the single threaded code would be easier to mentain, and right now I don't see any direct use case that would benefit from this small increase.

seanlaw:
> I totally agree that the single threaded code would be easier to mentain, and right now I don't see any direct use case that would benefit from this small increase.

Then, would you mind submitting a new PR with the necessary changes (minus multiprocessing) and closing this one? As always, I appreciate your contributions on this!

Btw, I saw from your previous PRs that you work/worked for CERN? I have a particle physics friend of mine here at the University of Michigan, [Claudio Ferretti](https://lsa.umich.edu/physics/people/staff/claudiof.html), visits CERN regularly for research. Do you happen to know him?

mexxexx:
Sure, I'll do it as soon as possible!

That's right, I'm working there as a student currently, but my work is not really connected to the particle physics part, but more on the accelerator complex itself. Sadly, I don't know him, no..

seanlaw:
@mexxexx It looks like the tests are passing. For cleanliness, would you mind closing this PR and creating a new PR with the final committed code and associate it with #143 ?

seanlaw:
@mexxexx It looks like I may have found a way to call NumPy functions within a Numba JIT function using [object mode](http://numba.pydata.org/numba-doc/latest/user/withobjmode.html?highlight=objmode#numba.objmode).  This might actually further simplify things

mexxexx:
Have you tried this mode and do you know anything about performance?

seanlaw:
> Have you tried this mode and do you know anything about performance?

No, I haven't tried it but it was recommended on the Numba Gitter channel. They confirmed that, indeed, "it's "experimental", as a lot of things, but it's been pretty stable since inception and IIRC it's had bugs fixed for it, which is a good sign"

Up to you. Just wanted to point it out but maybe there is no gain here

mexxexx:
> Sorry, I didn't mean to distract you. I say we just go with what you have!

No worries, I always want to learn new things :D
I think I will give it a try in the following days and see how the performance compares.

By the way, have you ever tried if it is possible to merge a request and squashing all the commit history? This way you don't have to recreate a PR if something gets changed and still keep a clean history in the main branch

seanlaw:
> By the way, have you ever tried if it is possible to merge a request and squashing all the commit history? This way you don't have to recreate a PR if something gets changed and still keep a clean history in the main branch

No, I haven't but it looks like I can do that. Should I squash and merge this one or would you like to play with "object mode` first?

seanlaw:
> No, I haven't but it looks like I can do that. Should I squash and merge this one or would you like to play with "object mode` first?

@mexxexx Just wanted to make sure you saw this comment.

mexxexx:
Hi @seanlaw, thanks for the reminder! I sadly didn't have much time to look into this lately. I'll try to do something tomorrow and will come back to you :smile:

mexxexx:
Okay, so using the object mode works and the tests are passing. I did some timings and there is no difference between using this mode or the implementation from this PR.

seanlaw:
@mexxexx My gut feeling is that we should go with your implementation since it is not "experimental" like object mode is (which could disappear in the future).

mexxexx:
Yes I agree, I don't think we will profit from object mode in its current state.

seanlaw:
@mexxexx Can this be merged? I can try to squash the commit into one as you recommended

mexxexx:
Yes, that would be great! Sorry for not replying earlier.

seanlaw:
Thanks, @mexxexx!

PR #167, opened 2020-05-11
Added SCRUMP tutorial

mexxexx:
Please take a look at this SCRUMP tutorial. I used the `steamgen` data set to show how the matrix profile can be approximated. It does not reproduce any figure from the paper, but I think it helps understanding how to use `stumpy.scrump`.


seanlaw:
Ohh, wow! This is actually really nice, @mexxexx! Well done! 

Providing feedback for a notebook is tricky since it won't be inline so I'll accept and I can go over it again after the PR is accepted.

mexxexx:
Thanks, I hope it helps the users, I think scrimp is really great if you don't have a clear idea yet as for what you are looking for.

Issue #113, opened 2020-01-13
Add SCRIMP++


seanlaw:
The original SCRIMP++ paper is [here](https://www.cs.ucr.edu/~eamonn/SCRIMP_ICDM_camera_ready_updated.pdf)

The original data, source code, and reference website for SCRIMP++ is [here](https://sites.google.com/site/scrimpplusplus/)

Additional [paper from another group](https://github.com/TDAmeritrade/stumpy/files/4477514/10.1007%40s11227-020-03199-w.pdf)

To compile the C++ code above:

```
conda install -y fftw
export LD_LIBRARY_PATH=/path/to/miniconda3/lib"
g++ -o prescrimp -I/path/to/miniconda3/include -L/path/to/miniconda3/lib -lfftw3 prescrimp.cpp
```

Then, create a time series input file with:

```
584.0
-11.0
23.0
79.0
1001.0
0.0
-19.0
```

And run the executable:

```
export LD_LIBRARY_PATH="/path/to/miniconda3/lib"
./prescrimp time_series 3 1
```

seanlaw:
According to Stuart Archibald (Numba dev), to force SCRIMP++ to complete within a user specified time, one could try something like:

```
from numba import njit
import numpy as np
import threading

@njit
def foo(m, n, FLAG):
    X = np.zeros((m, n))
    acc = 0
    while FLAG[0] == 0:
        print("In loop")
        acc += np.linalg.svd(X)[1][0]
    print("Got signal, exiting")
    return acc


m = n = 1000
FLAG = np.array([0])
runner_thread = threading.Thread(target=foo, args=(m, n, FLAG))

def SIGNAL_FLAG(flag):
    FLAG[0] = 1

# finish in permitted_time seconds or signal to stop
permitted_time = 5
timeout = threading.Timer(permitted_time, SIGNAL_FLAG, args=(FLAG,))
try:
    timeout.start()
    runner_thread.start()
    runner_thread.join()
finally:
    timeout.cancel()
```

seanlaw:
The example that was used in Figure 5, Figure 9, Figure 11, and Figure 12 was the typical steamgen data set:

```
import urllib
import ssl
import io
import os
import pandas as pd

colnames = ['drum pressure',
            'excess oxygen',
            'water level',
            'steam flow'
           ]

context = ssl.SSLContext()  # Ignore SSL certificate verification for simplicity
[...]

PR #170, opened 2020-05-13
Refactor scrump test

mexxexx:
This is still work in progress on some things from #113 .

First of all, I corrected a small indexing error in SCRUMP.  For now I removed the checks for the exclusion zone, because also for A_B joins this will not be necessary. I think I have an idea how they can be implemented. 

Then I refactored the SCRUMP tests and implemented naive methods, to also be able to check percentages. There are also two functions (`test_scrump_full`) that compare the final output to the matrix profile computed by stump, so that we know that the algorithm works. However, I did not implement a naive PRESCRUMP yet, so one of the tests (`test_scrump_plus_plus_self_join`) is failing.

Furthermore, I implemented a naive stamp function in `utils`, to remove some redundant code in all the other tests. The stumped tests still need to be updated.

Please take a look, if the direction this is going in seems right.


mexxexx:
I think we should focus on the docstrings in a separate issue, and for this issue only try to get the changes to the tests in. Maybe we want to ignore the prescrimp tests and open a new issue for them? Once this is done we can focus on the AB joins.

> Everything is on the right track. We probably want to replace all instances of calling naive_mass with naive_stamp in all of the unit tests

Definitely, but this should also be a separate issue I think.

seanlaw:
Yes, I agree with you. Let me know when this is ready to be merged. Thank you

mexxexx:
I ignored the test that is failing, so maybe its a good time to merge if there aren't any large issues. I will push some more changes to the PR that includes AB-joins and maybe then its a good opportunity to discuss some design choices.

Issue #171, opened 2020-05-14
Add AB joins for SCRIMP 


mexxexx:
A reference/discussion for adding AB joins for SCRIMP (`stumpy.scrump`). 

This can be done without changing much, we only have to keep in mind to calculate the whole distance matrix instead of the upper triangle.

seanlaw:
I'm really excited about this!

seanlaw:
[This PR](742babaefca827e898992663ac3d12bc74c7db8) fixes some of this to be consistent with having `T_B` containing the query subsequences

PR #176, opened 2020-05-15
Scrump A-B-Join

mexxexx:
An implementation of AB joins for scrump (#171 ).

This is probably a lot to digest, feel free to ask questions so I can explain the logic behind it.


stumpy/scrump.py
```diff
@@ -46,12 +49,19 @@ def _get_max_order_idx(m, n, orders, start, percentage):
     max_n_dist = 0
     for order_idx in range(orders.shape[0]):
         k = orders[order_idx]
-        max_n_dist = max_n_dist + (n - m + 1 - k)
+        if k >= 0:
```
seanlaw, line 52: Does this mean that we'll have values of `k` that are less than zero? Are these referring to the lower triangle?

stumpy/scrump.py
```diff
@@ -22,8 +22,11 @@ def _get_max_order_idx(m, n, orders, start, percentage):
     m : int
         Window size
 
-    n : int
-        Matrix profile length
+    n_A : int
+        Length of the first time series
```
seanlaw, line 26: "Length of time series T_A"

stumpy/scrump.py
```diff
@@ -22,8 +22,11 @@ def _get_max_order_idx(m, n, orders, start, percentage):
...
+
+    n_B : int
+        Lenght of the second time series
```
seanlaw, line 29: "Length of time series T_B"

seanlaw: Should only one of these be necessary (i.e., should one of `n_A` or `n_B` be optional) if we are doing a self-join?

stumpy/scrump.py
```diff
@@ -75,8 +85,11 @@ def _get_orders_ranges(n_split, m, n, orders, start, percentage):
     m : int
         Window size
 
-    n : int
-        Matrix profile length
+    n_A : int
+        Length of the first time series
```
seanlaw, line 89: "Length of time series T_A"

stumpy/scrump.py
```diff
@@ -75,8 +85,11 @@ def _get_orders_ranges(n_split, m, n, orders, start, percentage):
...
+
+    n_B : int
+        Lenght of the second time series
```
seanlaw, line 92: "Length of time series T_B"

seanlaw: Again, should `n_B` be optional for self-joins?

stumpy/scrump.py
```diff
@@ -114,34 +133,61 @@ def _get_orders_ranges(n_split, m, n, orders, start, percentage):
...
-    μ : ndarray
-        Sliding window mean for T
+    M_T : ndarray
+        Sliding mean of time series, `T`
 
-    σ : ndarray
-        Sliding window standard deviation for T
+    Σ_T : ndarray
+        Sliding standard deviation of time series, `T`μ_Q : ndarray
+        Mean of the query sequence, `Q`, relative to the current sliding window
```
seanlaw, line 183: Looks like "μ_Q : ndarray Mean of the query sequence, `Q`, relative to the current sliding window" can be removed here

stumpy/scrump.py
```diff
@@ -158,25 +204,43 @@ def _compute_diagonal(
     I : ndarray
         Matrix profile indices
 
+    thread_idx : int
+
```
seanlaw, line 208: Can we add the description to this argument and move `thread_idx` above `P` so that it is consistent with the function arguments list?

stumpy/scrump.py
```diff
@@ -186,31 +250,48 @@ def _compute_diagonal(
 
 
 @njit(parallel=True, fastmath=True)
-def _scrump(T, m, μ, σ, orders, orders_ranges, percentage=0.1):
+def _scrump(T_A, T_B, m, M_T, Σ_T, μ_Q, σ_Q, orders, orders_ranges, ignore_trivial):
     """
     A Numba JIT-compiled version of SCRIMP (self-join) for parallel computation
```
seanlaw, line 255: Can probably remove "(self-join)" if it works for AB-joins too

stumpy/scrump.py
```diff
@@ -186,31 +250,48 @@ def _compute_diagonal(
...
-    μ : ndarray
-        Sliding window mean for T
+    M_T : ndarray
+        Sliding mean of time series, `T`
 
-    σ : ndarray
-        Sliding window standard deviation for T
+    Σ_T : ndarray
+        Sliding standard deviation of time series, `T`μ_Q : ndarray
+        Mean of the query sequence, `Q`, relative to the current sliding window
```
seanlaw, line 275: Remove "μ_Q : ndarray Mean of the query sequence, `Q`, relative to the current sliding window" as it is redundant from next line

stumpy/scrump.py
```diff
@@ -472,46 +568,90 @@ def scrump(T, m, percentage=0.01, pre_scrump=False, s=None):
...
     """
 
-    T = np.asarray(T)
-    T = T.copy()
-    T[np.isinf(T)] = np.nan
-    core.check_dtype(T)
+    if T_B is not None and pre_scrump:  # pragma: no cover
+        raise NotImplementedError(
+            "AB joins with preprocessing are not yet implemented."
+        )
```
seanlaw, line 574: What would we need to do to implement this? Should we start a new issue to cover this?

stumpy/stump.py
```diff
@@ -370,7 +370,6 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
             f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
-    n = T_A.shape[0]
```
seanlaw, line 373: Good catch!

stumpy/scrump.py
```diff
@@ -472,46 +568,90 @@ def scrump(T, m, percentage=0.01, pre_scrump=False, s=None):
...
         for i in range(P.shape[0]):
             if out[i, 0] > P[i]:
                 out[i, 0] = P[i]
                 out[i, 1] = I[i]
 
-    orders = np.random.permutation(range(excl_zone + 1, n - m + 1))
+    if ignore_trivial:
+        orders = np.random.permutation(range(excl_zone + 1, n_B - m + 1))
+    else:
+        orders = np.random.permutation(range(-(n_A - m + 1) + 1, n_B - m + 1))
```
seanlaw, line 640: Do you mind explaining the logic of having negative values in `orders`? Each element of `orders` should be referring to a diagonal. So, I am assuming that a negative value refers to the lower triangle. However, typically, NumPy indexing with negative numbers refers to indexing from the end of an array. In this case, IIUC, it looks like `orders` is relative to the main diagonal that runs through `[0, 0]` and has nothing to do with NumPy indexing? Is that about right?

tests/test_scrump.py
```diff
@@ -194,9 +299,9 @@ def test_scrump_self_join_full(T):
 
 
 @pytest.mark.skip(reason="naive PRESCRUMP is not yet implemented")
```
seanlaw, line 301: Do we have a plan for how to complete this? What else is needed in order to get our test coverage back to 100%?

stumpy/scrump.py
mexxexx: Yes that's right

stumpy/scrump.py
mexxexx: That's correct. Take a look at the following distance matrix:

```
 0   1   2   3   4
-1   0   1   2   3
-2  -1   0   1   2
```
Then the length of the Matrix profile is 5 (number of columns = `n_B - m +1`) while the number of rows is `n_A - m + 1`.

tests/test_scrump.py
mexxexx: The coverage is not at 100% because of PRESCRIMP, and I haven't changed anything there, so I can't say...

For this test we need to implement a naive prescrimp to have a naive scrimp_plus_plus available, i.e. to be able to compute only a certain number of diagonals but with preprocessing included.

stumpy/scrump.py
mexxexx: For this we would need to adapt prescrimp to work on AB joins.

stumpy/scrump.py
mexxexx: Sure, what description do you suggest?

tests/test_scrump.py
seanlaw: Okay, we'll need to get the `naive_prescrimp` tests in so that our tests are back up to 100% before we can release v1.4.0. Let me know if you need help with that

stumpy/scrump.py
seanlaw: Cool. Thanks

tests/test_scrump.py
mexxexx: Yes it would be great if you could help, since I don't know how prescrimp works and don't have to much time at the moment...

stumpy/scrump.py
seanlaw: Do you need any help with this?

tests/test_scrump.py
seanlaw: Cool. Then I can merge this and work on pre-scrimp

seanlaw:
@mexxexx Thanks again for this work!

Issue #166, opened 2020-05-08
Handling NaN/Inf in core.mass and stamp.mass


seanlaw:
In [this comment](https://github.com/TDAmeritrade/stumpy/issues/149#issuecomment-625534271) a user is trying to use `core.mass` (or `stamp.mass`) to compute a distance profile but their data contains `np.nan/np.inf`. Currently, `nan/inf` handling is handled at a higher abstraction level (which is fine) but it would be beneficial to allow `core.mass`/`stamp.mass` to automatically handle `nan/inf` since these functions will inevitably be called upon when finding matches to a single known pattern/motif.

seanlaw:
@mexxexx Taking a look at the previous comment above, can you think of a way where the user can easily use `core.mass` or `stamp.mass` directly with data that contains `np.nan/np.inf` without having to manually perform the pre-processing step themselves? Right now, you'd have to do something like:

```
T = np.random.rand(1000)
m = 50

# Some pre-processing steps that is usually handled by `stumpy.stump` but you need to do this manually
T_copy = T.copy()  # Leave original time series unchanged
T_copy[np.isinf(T_copy)] = np.nan
M_T, Σ_T = core.compute_mean_std(T_copy, m)
T_copy[np.isnan(T_copy)] = 0
zone = excl_zone = int(np.ceil(m / 4))

# You compute the matrix profile and find a motif at `i = 10`
Q = T_copy[10: 10 + m]
P, I = stamp.mass(Q, T_copy, M_T, Σ_T, trivial_idx=10, excl_zone=zone)
```

There are several issues that one would need to solved:

1. Making a copy of `T` so as not to overwrite the original time series
2. Convert all `inf` to `nan` before computing the sliding window mean and standard deviation
3. Convert `nan` to zero before calling `core.mass` or `stamp.mass`

Do you have any thoughts?

mexxexx:
There is some nan inf handling in `core.mass`, so it might be possible to do all this work there. We would need to have a look where core.mass is called and of this would lead to any conflicts, but from the first glance it should be fine.

If the query contains nans, an infinite distance profile is returned, so there should be no problem doing the same if the query contains infs. And if there are nans in the time series, they should be handled correctly too, so just adding some more checks shouldn't be a problem.

mexxexx:
I think in general there should be only one mass function public to the user, and I would choose core.mass, because no exclusion zone is applied there and if you only want the distance profile you probably don't need the exclusion zone anyways.

[...]

PR #183, opened 2020-05-24
Added preprocess function

mexxexx:
Following some discussions in issue #166 I created a preprocessing function (and an `apply_exclusion_zone` function) that is used by all (currently only 1D) algorithms. Furthermore, `core.mass` now supports time series with NaN/Inf.

### DIFF
--- a/stumpy/gpu_stump.py
+++ b/stumpy/gpu_stump.py
@@ -437,30 +437,32 @@ def gpu_stump(T_A, m, T_B=None, ignore_trivial=True, device_id=0):
     Note that left and right matrix profiles are only available for self-joins.
     """
 
-    T_A = np.asarray(T_A)
+    if T_B is None:  # Self join!
+        T_B = T_A
+        ignore_trivial = True
+
+    # Swap T_A and T_B for GPU implementation
+    # This keeps the API identical to and compatible with `stumpy.stump`
+    tmp_T = T_A
+    T_A = T_B
+    T_B = tmp_T
+
+    T_A, M_T, Σ_T = core.preprocess(T_A, m)
+    T_B, μ_Q, σ_Q = core.preprocess(T_B, m)
+
     if T_A.ndim != 1:  # pragma: no cover
         raise ValueError(
             f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
-    T_A = T_A.copy()
-    T_A[np.isinf(T_A)] = np.nan
-    core.check_dtype(T_A)
 
-    if T_B is None:  # Self join!
-        T_B = T_A
-        ignore_trivial = True
-    T_B = np.asarray(T_B)
-
-    T_B = np.asarray(T_B)
-    T_B = T_B.copy()
     if T_B.ndim != 1:  # pragma: no cover
         raise ValueError(
             f"T_B is {T_B.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
 
-    T_B[np.isinf(T_B)] = np.nan
+    core.check_dtype(T_A)
     core.check_dtype(T_B)
 
     core.check_window_size(m)
@@ -473,22 +475,10 @@ def gpu_stump(T_A, m, T_B=None, ignore_trivial=True, device_id=0):
         logger.warning("Arrays T_A, T_B are not equal, which implies an AB-join.")
         logger.warning("Try setting `ignore_trivial = False`.")
 
-    # Swap T_A and T_B for GPU implementation
-    # This keeps the API identical to and compatible with `stumpy.stump`
-    tmp_T = T_A
-    T_A = T_B
-    T_B = tmp_T
-
     n = T_B.shape[0]
     k = T_A.shape[0] - m + 1
     l = n - m + 1
     excl_zone = int(np.ceil(m / 4))  # See Definition 3 and Figure 3
-
-    M_T, Σ_T = core.compute_mean_std(T_A, m)
-    μ_Q, σ_Q = core.compute_mean_std(T_B, m)
-
-    T_A[np.isnan(T_A)] = 0
-    T_B[np.isnan(T_B)] = 0
 
     T_A_fname = core.array_to_temp_file(T_A)
     T_B_fname = core.array_to_temp_file(T_B)
--- a/stumpy/stamp.py
+++ b/stumpy/stamp.py
@@ -51,9 +51,7 @@ def mass(Q, T, M_T, Σ_T, trivial_idx=None, excl_zone=0, left=False, right=False
     D = core.mass(Q, T, M_T, Σ_T)
 
     if trivial_idx is not None:
-        zone_start = max(0, trivial_idx - excl_zone)
-        zone_stop = min(T.shape[0] - Q.shape[0] + 1, trivial_idx + excl_zone)
-        D[zone_start : zone_stop + 1] = np.inf
+        core.apply_exclusion_zone(D, trivial_idx, excl_zone)
 
         # Get left and right matrix profiles
         IL = -1
@@ -61,7 +59,7 @@ def mass(Q, T, M_T, Σ_T, trivial_idx=None, excl_zone=0, left=False, right=False
         if D[:trivial_idx].size:
             IL = np.argmin(D[:trivial_idx])
             PL = D[IL]
-        if PL == np.inf or zone_start <= IL < zone_stop:
+        if PL == np.inf:
             IL = -1
 
         IR = -1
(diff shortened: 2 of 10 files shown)


stumpy/gpu_stump.py
```diff
@@ -437,30 +437,32 @@ def gpu_stump(T_A, m, T_B=None, ignore_trivial=True, device_id=0):
...
-    T_B = np.asarray(T_B)
-    T_B = T_B.copy()
     if T_B.ndim != 1:  # pragma: no cover
         raise ValueError(
             f"T_B is {T_B.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
 
-    T_B[np.isinf(T_B)] = np.nan
+    core.check_dtype(T_A)
```
seanlaw, line 465: Can you please move these two lines into `core.preprocess`?

stumpy/stamp.py
```diff
@@ -130,25 +128,22 @@ def stamp(T_A, T_B, m, ignore_trivial=False):
...
-    T_A = T_A.copy()
-    T_A[np.isinf(T_A)] = np.nan
-    core.check_dtype(T_A)
-
-    T_B = T_B.copy()
-    T_B[np.isinf(T_B)] = np.nan
     if T_B.ndim != 1:  # pragma: no cover
         raise ValueError(f"T_B is {T_B.ndim}-dimensional and must be 1-dimensional. ")
+
+    core.check_dtype(T_A)
```
seanlaw, line 141: Please move to `core.preprocess`

stumpy/stomp.py
```diff
@@ -107,23 +105,22 @@ def _stomp(T_A, m, T_B=None, ignore_trivial=True):
...
-
-    T_A[np.isnan(T_A)] = 0
-
     out = np.empty((l, 4), dtype=object)
 
     # Handle first subsequence, add exclusionary zone
-    if ignore_trivial:
-        P, I = stamp.mass(T_B[:m], T_A, M_T, Σ_T, 0, excl_zone)
-        PR, IR = stamp.mass(T_B[:m], T_A, M_T, Σ_T, 0, excl_zone, right=True)
+    if not np.isinf(μ_Q[0]):
```
seanlaw, line 111: Can you please swap the order of the `if/else` statement so that it is?

```
if np.isinf(μ_Q[0]):
    .
else:
    .
```

This avoids having to perform some "mental gymnastics" to think about what `not np.isinf(μ_Q[0])` means

stumpy/stump.py
```diff
@@ -364,29 +380,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True):
...
-    T_B = np.asarray(T_B)
-    T_B = T_B.copy()
     if T_B.ndim != 1:  # pragma: no cover
         raise ValueError(
             f"T_B is {T_B.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
-    T_B[np.isinf(T_B)] = np.nan
+
+    core.check_dtype(T_A)
```
seanlaw, line 402: Please move to `core.preprocess`

stumpy/stomp.py
```diff
@@ -73,24 +73,22 @@ def _stomp(T_A, m, T_B=None, ignore_trivial=True):
...
+    if T_A.ndim != 1:  # pragma: no cover
+        raise ValueError(f"T_A is {T_A.ndim}-dimensional and must be 1-dimensional. ")
+
+    n = T_A.shape[0]
+
     if T_B.ndim != 1:  # pragma: no cover
         raise ValueError(f"T_B is {T_B.ndim}-dimensional and must be 1-dimensional. ")
-    T_B[np.isinf(T_B)] = np.nan
+
+    core.check_dtype(T_A)
```
seanlaw, line 91: Please move to `core.preprocess`

stumpy/stumped.py
```diff
@@ -83,30 +83,26 @@ def stumped(dask_client, T_A, m, T_B=None, ignore_trivial=True):
...
-    T_B = np.asarray(T_B)
     if T_B.ndim != 1:  # pragma: no cover
         raise ValueError(
             f"T_B is {T_B.ndim}-dimensional and must be 1-dimensional. "
             "For multidimensional STUMP use `stumpy.mstump` or `stumpy.mstumped`"
         )
-    T_B = T_B.copy()
-    T_B[np.isinf(T_B)] = np.nan
+
+    core.check_dtype(T_A)
```
seanlaw, line 105: Please move to `core.preprocess`

tests/test_core.py
```diff
@@ -32,20 +33,18 @@ def test_check_window_size():
 def naive_compute_mean_std(T, m):
```
seanlaw, line 33: I have a concern that this function is no longer "naive" and is actually a copy of what we are testing against.

mexxexx:
I honestly just forgot about scrump, there it will definitely be possible (although the exclusion zone is not necessary). For mstump I'll have to see, but it might work.

I didn't move the `check_dtype` stuff inside, because I thought it doesn't belong there.

As for the naive mean/std don't you think it's even more naive this way? Computing it in a rolling manner for each subsequence?

seanlaw:
> I honestly just forgot about scrump, there it will definitely be possible (although the exclusion zone is not necessary). For mstump I'll have to see, but it might work.

Cool! If you can see if you can add this to `mstump`, `mstumped` then that would be great! Let's leave `scrump` out of it since I'm doing some work on it right now and we can deal with it afterward.
 
> I didn't move the check_dtype stuff inside, because I thought it doesn't belong there.

That's fair. It's more of a "check" and not really changing anything.

> As for the naive mean/std don't you think it's even more naive this way? Computing it in a rolling manner for each subsequence?

Okay, I see it now. I didn't realize when I first looked at it. I take back my "concern" 👍

mexxexx:
How would you like to rename the function?

seanlaw:
> How would you like to rename the function?

I don’t know 🤷‍♂️ 

So, I say let’s be pragmatic and we just go with `preprocess` for now and adjust it later if needed

mexxexx:
Okay :smile: I included the preprocessing in MSTUMP/MSTUMPED and had to make small changes.

seanlaw:
@mexxexx It looks like `apply_exclusion_zone` is already `njit`. Can we use this in `mstump` and `mstumped`?  I think that if we updated `_get_first_mstump_profile` and `_compute_multi_D` to use `apply_exclusion_zone` then we should be good.

mexxexx:
I think we need a different function for `mstump`, because we have one extra dimension there.
```
D[:, zone_start : zone_stop + 1]
```
in `mstump` vs 
```
D[zone_start : zone_stop + 1]
```
in the 1D case.

seanlaw:
According to [this](https://stackoverflow.com/questions/42916029/numpy-indexing-over-the-last-axis-when-you-dont-know-the-rank-in-advance), I think it would be fine to do:

```
zone_start = max(0, idx - excl_zone)
zone_stop = min(D.shape[0], idx + excl_zone)
D[..., zone_start : zone_stop + 1] = np.inf
```

mexxexx:
Ah cool, I didn't know, thanks!

I'll try it out tomorrow.

seanlaw:
@mexxexx Everything looks good. Do you think you could also rename `stamp.mass` to `stamp.mass_PI`? Then maybe we can close issue #166?

stumpy/stamp.py
```diff
@@ -130,25 +128,22 @@ def stamp(T_A, T_B, m, ignore_trivial=False):
     T_B.shape[0]-m+1
     """
 
+    T_A, M_T, Σ_T = core.preprocess(T_A, m)
+    T_B = T_B.copy()
+    T_B[np.isinf(T_B)] = np.nan
```
seanlaw, line 133: Is there a reason why we aren't called `core.preprocess` here?

tests/test_core.py
```diff
@@ -171,6 +170,136 @@ def test_mass(Q, T):
...
+    left = np.linalg.norm(
+        core.z_norm(core.rolling_window(T, m), 1) - core.z_norm(Q), axis=1
+    )
+    left[np.isnan(left)] = np.inf
+
+    right = core.mass(Q, T)
+    npt.assert_almost_equal(left, right)
+
+
+def test_apply_exclusion_zone():
```
seanlaw, line 201: I think it is probably important to test the "edge cases: where `index = 0` and `index = 9`

tests/test_core.py
```diff
@@ -171,6 +170,136 @@ def test_mass(Q, T):
...
+    left = np.array([0, 1, np.inf, np.inf, np.inf, np.inf, np.inf, 7, 8, 9])
+    right = T.copy()
+    core.apply_exclusion_zone(right, index, exclusion_zone)
+
+    utils.replace_inf(left)
+    utils.replace_inf(right)
+    npt.assert_array_equal(left, right)
+
+
+def test_apply_exclusion_zone_multidimensional():
```
seanlaw, line 233: Same as above: I think it is probably important to test the "edge cases: where index = 0 and index = 9

seanlaw:
I made some mistakes trying to resolve the merge conflict. I'll fix this and also handle the renaming of `stamp.mass` as well as the extra unit tests requested above

seanlaw:
@mexxexx Thanks again for refactoring this!

Issue #206, opened 2020-06-14
Add NaN/inf support to FLOSS


seanlaw:
Currently, this does not handle NaN/inf for incoming data points

JosephTLucas:
I'm trolling for a potential contribution.  How would you want to handle them?  Fill forward?

seanlaw:
@JosephTLucas Welcome (from Twitter) and thank you for your consideration!

> How would you want to handle them? Fill forward?

IIRC, this is a little bit more complicated but is worth expounding upon. So, unlike typical missing data, we're not simply trying to impute data. If a new data is streamed in and it is a floating point number, then the distance profile computed (using MASS) for the new subsequence that is formed (as a result adding a new data point), and we're all good. However, if that new incoming data point (the literature calls this "ingress" of a new data point) is a `np.nan`/`np.inf` then any subsequence that contains this bad data point should only be allowed to have a distance of `np.inf` (i.e., any subsequence with an `np.nan`/`np.inf` should be "illegal").  However, in practice, to prevent MASS from failing to compute a distance profile, we simply set the value of the new data point to zero after the mean and stddev for the subsequence is computed (which keeps track of the fact that it is an illegal subsequence). This is usually fine for batch jobs but this is a problem for streaming data as we now need to keep track of the long-term presence of this bad/illegal data point and to make sure that we always account for it in our MASS calculation.

To solve this, we should just add a boolean 1D array that is the same length as `self._T` and that gets updated and cross referenced and then we can update the distance profile computed by MASS and the distances of all illegal subsequences to `np.inf` and that should be it. 

I'd be happy to work with you on this!

JosephTLucas:
Thanks for the thorough rundown, Sean.  Let me take a couple days to poke through the code, but I should be able to do that!

seanlaw:
@JosephTLucas Please do not hesitate to share intermediate work through a temporary PR and feel free to ask me anything! I am here to help

seanlaw:
@JosephTLucas I found some time to take a crack at it and there were some additional parts that were needed to make it work. Thank you again for helping us push this one forward and I look forward to seeing other contributions from you!

josephlucas-arc:
Nice work, Sean!

JosephTLucas:
Nice work, Sean!

seanlaw:
It was a team effort!

PR #224, opened 2020-08-15
[WIP] Add finite tracking dictionary

JosephTLucas:
Work in Progress for Issue #206 

Really clean codebase, Sean.  Just updating as I work so you know there's progress.  Created finite mask attribute that tracks with _T.

TODO: Find where to apply finite mask
TODO: Add test to include nan/inf data

# Pull Request Checklist

- [x] Make necessary changes
- [x] Install `black` (i.e., `python -m pip install black` or `conda install black`)
- [x] Install `flake8` (i.e., `python -m pip install flake8` or `conda install flake8`)
- [x] Install `pytest-cov` (i.e., `python -m pip install pytest-cov` or `conda install pytest-cov`)
- [x] Run `./setup.py && ./test.sh` in the root stumpy directory


stumpy/floss.py
```diff
@@ -429,6 +429,7 @@ def __init__(
         self._n = self._T.shape[0]
         self._last_idx = self._n - self._m + 1  # Depends on the changing length of `T`
         self._n_appended = 0
+        self._finite_mask = np.full(self._n, True)
```
seanlaw, line 432: In other parts of the code, we have a similar mask that is called `T_subseq_isfinite` that basically marks where a subsequence is "illegal". It is computed by doing:

```
T_subseq_isfinite = np.all(np.isfinite(core.rolling_window(T, m)), axis=1)
```

Would you mind naming this attribute `self._T_isfinite`? I think it may be (ever so) slightly more obvious than `_finite_mask`. So, `self._T_isfinite` is for each element of `T` and `T_subseq_isfinite` is for each subsequence of `T`.

I'm wondering if we can leverage `self._T_isfinite` AFTER we compute `mass` by simply doing something similar to `T_subseq_isfinite` like:

```
D = core.mass(Q, self._T, M_T, Σ_T)
D[zone_start:] = np.inf
T_subseq_isfinite = np.all(core.rolling_window(T_isfinite, m), axis=1)
D[T_subseq_isfinite] = np.inf
```

I believe that `T_subseq_isfinite` should have the same length as `D` in this case but I could be wrong. Please let me know if that makes any sense.  Just thinking off the top of my head here :)

stumpy/floss.py
```diff
@@ -482,7 +483,13 @@ def update(self, t):
         Segmentation (FLOSS).
         """
         self._T[:] = np.roll(self._T, -1)
-        self._T[-1] = t
+        self._finite_mask[:] = np.roll(self._finite_mask, -1)
```
seanlaw, line 486: Looks like you're on the right track! Awesome!

stumpy/floss.py
JosephTLucas: Great points, Sean.  I'll make the changes to stay consistent with the rest of the code base.

seanlaw: Thank you! Let me know if you need to bounce anything off of me. I am here to help!

seanlaw:
> Sean, can you help me understand how we're simulating the stream in the test_floss.py?
> Specifically:
> https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L143
> 
> `add_data` is never referenced. Are we not actually 'streaming' data through the `floss` object?
> 
> It seems like
> https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L149
> 
> 
> should potentially be:
> `rolling_Ts = core.rolling_window(add_data[1:], n)`, so that the call to floss.update includes the new data, but with this change, even the existing test fails.

Certainly! Admittedly, after reviewing the test myself, I see two problems:

1. Indeed, `add_data` is not used and should be removed. I believe that that it use to be used but the test may have been updated and I failed to delete it.
2. Admittedly, the logic of this unit test seems a bit hard to follow and it may be worth refactoring for easier future consumption. However, I am 95% certain that the existing test is still correct but reserve 5% humility to be proven otherwise. 😄 

Perhaps, a concrete example would be useful here but here are a couple of important points before diving into this test:

1. The goal of `floss` is to compute the "corrected arc curve", CAC, for a streaming data set. Unlike batch processing of a static data set (i.e., `fluss`), `floss` takes the "right matrix profile index" as input (here "right" is referring to the directionality and not synonymous with "correct") and then computes the "uni-directional" CAC (rather than the "bi-directional CAC"). Hence, you'll see that the "global matrix profile index" (`mp[:, 1]`) and "left matrix profile index" (`mp[:, 2]`) are purposely set to invalid values of `-1` so as to avoid confusion and only the "right matrix profile index" (`mp[:, 3]`) is used for `floss`.
2. In all of our unit tests, we like to (lazily?) use "left" to represent the naive reference implementation and "right" to represent the STUMPY implementation and then we assert that "left is equal to right". I'm sure there are better ways to represent this especially since "left" and "right" are both overloaded (i.e., "left" and "right" matrix profile indices) and could lead to confusion.
3. Overall, the basic steps of `floss` is to i) update the time series window, ii) compute the distance profile for the new subsequence, iii) update the matrix profile and right matrix profile index, and iv) compute the corrected arc curve for the current time series window

With this in mind, let's dive in. Let's say we can look into the future and know that our full data set is `data = np.arange(100)`. Our `old_data = data[:30]` which is just `[0, 1, 2, ..., 28, 29]`. We can compute a matrix profile by choosing a subsequence length of `m = 5`:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L145-L146

This serves as the start of our matrix profile and we will update it repeatedly below as we receive/ingress "new" data. Note that `floss` will keep the time series length the same (i.e., `n = old_data.shape[0]` which is `30`), appends/ingresses the newest data point, and discards/egresses the oldest data point (from the front of the time series. So, effectively:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L149

This contains:

```
rolling_Ts =[
  [1, 2, 3, ..., 29, 30],
  [2, 3, 4, ..., 30, 31],
  [3, 4, 5, ..., 31, 32],
  ...
  [69, 70, ..., 97, 98],
  [70, 71, ..., 98, 99]
]
```

Notice that if we iterate through each row `rolling_Ts`, it mimics what would've happened if we had ingressed/egressed a data point. Now, if we were to be "naive" about it, we could simply compute the matrix profile for each row of `rolling_Ts`. This is precisely what is happening in:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L157-L181

So, in the first iteration of the for-loop, we set `left_T = [1, 2, 3, ..., 29, 30]` (remember, this has already ingressed/egressed a single data point), compute the distance profile:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L164-L165

update the existing matrix profile:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L167-L169

and, finally, compute the corrected arc curve using the new "right matrix profile index":

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L171-L177

Similarly, for `floss`, we initiated our streaming object outside of the for-loop:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L153

then, inside of the for loop, we update the time series and `floss` automagically updates the corrected arc curve (performed internally by `floss`:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L183

Notice that for the first iteration of the for-loop, `left_T = [1, 2, 3, ..., 29, 30]` and so `left_T[-1]` is nothing but simply plucking the last point off of `left_T` (which is `30`) and feeding it into our stream. So, prior to the first update, the stream contained `[0, 1, 2, ..., 28, 29]` as its time series. However, after the `stream.update(left_T[-1])` is executed, the internal time series in the stream is updated to `[1, 2, 3, ..., 29, 30]` (i.e., the first data point, `0`, is egressed and the new data point, `30`, is ingressed), the internal matrix profile is updated, and the corrected arc curve is computed. 

Once we have both the naive corrected arc curve, `left_cac_1d`, and the `right_cac_1d` from `floss`, then we can assess the assertion:

https://github.com/TDAmeritrade/stumpy/blob/de6df5881863d04aef7f0a346691c08abb9ecaf5/tests/test_floss.py#L192

Then, the next iteration of the for-loop is executed which updates `left_T` to `left_T = [2, 3, 4, ..., 30, 31]` and the same process described above repeats. This continues until `left_T = [70, 71, ..., 98, 99]` is processed via the naive approach and then compared to `floss`.

Hopefully, that makes things a little clearer but, if not, please feel free to ask for any clarification as I'm sure it is a lot to process and digest.

> Also, just as an FYI, I tend to find time for open source/hobby work on the weekends, but want to respect your time. I have no expectation of a weekend response!

Thank you! I appreciate your conscientiousness. Tech needs more of this!

JosephTLucas:
This walk through was really great, Sean.  Thank you.  I have the tests working now!  And am working through the failures.  More to come!

Your second point (about the overloading of `left` and `right`) is exactly where I got confused.  I see the same convention is used in other tests, so I'll retain that consistency here... maybe a test-wide refactor for this is something I could tackle in a separate issue/PR.

seanlaw:
> maybe a test-wide refactor for this is something I could tackle in a separate issue/PR.

Yes, that would be great and I would welcome such a PR! When we first started out, we only had "matrix profile" and "matrix profile index" and so "left" and "right" were never had an issue but, later, we started including left/right matrix profile and indices. Unfortunately, hindsight is 20/20 😄  

I'm not sure what they should be renamed to. Do you have any suggestions?

JosephTLucas:
Okay Sean, I think `floss.py` can now handle nan/inf and that functionality is appropriately tested in `test_floss.py`.  Tomorrow I'll try and figure out why it's failing the CI.  All tests and provided scripts are passing on my computer (insert timeless developer joke here).

Regarding test naming: Your description of the `left` values as `reference` made immediate sense to me, so that's a good candidate.  We could also re-purpose the `naive_*` moniker and rename the functions to `calc_naive_*`.  I think we just drop `right`.  So `npt.assert_almost_equal(left_cac_1d, right_cac_1d)` becomes `npt.assert_almost_equal(reference_cac_1d, stream.cac_1d)` or `npt.assert_almost_equal(reference_cac_1d, cac_1d)`.

seanlaw:
It looks like `black` is failing

seanlaw:
Okay, I ran into the same issue with the `black` version and `pip install`-ing the latest version resolved it.

seanlaw:
> Regarding test naming: Your description of the left values as reference made immediate sense to me, so that's a good candidate. We could also re-purpose the naive_* moniker and rename the functions to calc_naive_*. I think we just drop right. So npt.assert_almost_equal(left_cac_1d, right_cac_1d) becomes npt.assert_almost_equal(reference_cac_1d, stream.cac_1d) or npt.assert_almost_equal(reference_cac_1d, cac_1d).

I do like this idea and even simply `ref_cac_1d` would be sufficient. Let's tackle this across all tests in #235

JosephTLucas:
I was on `black 19.10b0`.  I upgraded to black `20.8b1`, but it didn't change the formatting results of `tests/test_floss.py` or `stumpy/floss.py`.  

If I run it on the whole project, it changes:
`stumpy/aamp.py
stumpy/mstump.py
stumpy/scrump.py
tests/naive.py
tests/test_gpu_stump.py`
which obviously I didn't want to touch with this PR.  I'll push a commit with these format updates and if that passes, we can figure out if we want those changes as part of this PR.

seanlaw:
> I was on black 19.10b0. I upgraded to black 20.8b1, but it didn't change the formatting results of tests/test_floss.py or stumpy/floss.py.

> If I run it on the whole project, it changes:
stumpy/aamp.py stumpy/mstump.py stumpy/scrump.py tests/naive.py tests/test_gpu_stump.py
which obviously I didn't want to touch with this PR. I'll push a commit with these format updates and if that passes, we can figure out if we want those changes as part of this PR.

I've already upgraded to v20.8b1 myself and committed some changes but, to be pragmatic, just go ahead and and commit those changes to this PR (they may be redundant to mine) and I'll recognize it as formatting differences. This can be handled on my end as they'll be short and obvious.

seanlaw:
I'm not sure what you see on your end in Github but do you see a window like this above your comment box?

fail

If so, on the right-hand-side, you can click on "Details" and it will give you more information as to what the failure is when you look for the text "View more details on Azure Pipelines".

JosephTLucas:
It looks like we're finally good to go, @seanlaw !

seanlaw:
Great! I will take some time to review it. Thank you for all of your hard work @JosephTLucas! I really appreciate you taking the time to contribute

stumpy/floss.py
```diff
@@ -429,6 +429,7 @@ def __init__(
         self._n = self._T.shape[0]
         self._last_idx = self._n - self._m + 1  # Depends on the changing length of `T`
         self._n_appended = 0
+        self._T_isfinite = np.isfinite(core.rolling_window(self._T, m))
```
seanlaw, line 432: Can you please rename this to `self._T_subseq_isfinite`? This way it's obvious that this is storing the booleans for each subsequence in T rather than the individual elements of T.

stumpy/floss.py
```diff
@@ -482,7 +483,9 @@ def update(self, t):
         Segmentation (FLOSS).
         """
         self._T[:] = np.roll(self._T, -1)
+        self._T_isfinite[:] = np.roll(self._T_isfinite, -1)
```
seanlaw, line 486: I'm a little bit confused here. I see that you are indeed using `self._T_isfinite` to track whether individual elements in T are finite. However, above, it is clear that it contains subsequences rather than elements. I'm surprised that the final output is "right"

tests/test_floss.py
```diff
@@ -164,6 +170,8 @@ def test_floss():
         D = naive_distance_profile(left_T[-m:], left_T, m)
         D[zone_start:] = np.inf
 
+        left_T_isfinite = np.all(np.isfinite(left_T))
```
seanlaw, line 173: I'm going to ask you to leave the `test_floss` alone (or revert it back to the original test) and create a separate `test_floss_nan_inf` test. So, `test_floss` tests the best case scenario where the input data does not contain `np.nan`/`np.inf` and `test_floss_nan_inf` is a variation of `test_floss` where the inputs contain `np.nan` and/or `np.inf`. For sanity, we'd want to see that the original unaltered `test_floss` actually passes

stumpy/floss.py
```diff
@@ -502,6 +505,8 @@ def update(self, t):
         D = core.mass(Q, self._T, M_T, Σ_T)
         D[zone_start:] = np.inf
 
+        T_subseq_isfinite = np.all(self._T_isfinite[-self._m :])
```
seanlaw, line 508: I think this should be:

```
T_subseq_isfinite = np.all(core.rolling_window(self._T_isfinite, self._m), axis=1)
```

tests/test_floss.py
seanlaw: I believe that this line should be:

```
left_T_isfinite = np.isfinite(left_T)
left_T_subeq_isfinite = np.all(core.rolling(left_T_isfinite, m), axis=1)
D[~left_T_subseq_isfinite] = np.inf
```

josephlucas-arc:
hmm, I must have made some version control mistake.  I definitely made those changes (that you'd already recommended).  Thanks for the patience, I'll go double tap that and break test_floss into two tests.

seanlaw:
No problem, @JosephTLucas! I'm more surprised that the tests passed as I would've expected things to fail. Take your time and happy to review

seanlaw:
I'm super excited to see the failed test!! 👍

JosephTLucas:
I haven't forgotten about this, Sean!  Just enjoyed the (US) holiday weekend and catching up at work.  I expect to make progress this weekend.

seanlaw:
> I haven't forgotten about this, Sean! Just enjoyed the (US) holiday weekend and catching up at work. I expect to make progress this weekend.

No worries and I figured that was the case! I hope you and your family had a nice time off

tests/test_floss.py
```diff
@@ -193,3 +192,73 @@ def test_floss():
...
+
+test_data_nan = np.insert(
+    np.random.uniform(-1000, 1000, [64]), [i for i in range(30, 60, 5)], np.nan
+)
+test_data_inf = np.insert(
+    np.random.uniform(-1000, 1000, [64]), [i for i in range(30, 60, 5)], np.inf
+)
+
+
+@pytest.mark.parametrize("data", [(test_data_nan), (test_data_inf)])
```
seanlaw, line 205: @JosephTLucas I apologize for the late curveball here so feel free to ignore and I can handle this after the merge if it's too much to ask. Usually, when we do `nan`/`inf` testing, we like to define the substitution of real float values at the top like:

https://github.com/TDAmeritrade/stumpy/blob/164368e97e2acae5f882a3acb5e7caefc144b7a0/tests/test_stump.py#L9-L21

and then the unit test itself looks like:

https://github.com/TDAmeritrade/stumpy/blob/164368e97e2acae5f882a3acb5e7caefc144b7a0/tests/test_stump.py#L176-L197

For consistency with other tests, would you mind changing this to:

```
# Add this at the top of `test_floss.py`
substitution_locations = [(slice(0, 0), 0, -1, slice(1, 3), [0, 3])] 
substitution_values = [np.nan, np.inf] 
```

and then in the test:

```
@pytest.mark.parametrize("substitute", substitution_values) 
@pytest.mark.parametrize("substitution_locations", substitution_locations) 
def test_floss_nan_inf(substitute, substitution_locations): 
    T = np.random.uniform(-1000, 1000, [64])
    m = 5
    data = T.copy()

    for substitution_location in substitution_locations:
        data[:] = T[:]
        data[substitution_location] = substitute
        old_data = data[:30]
        n = old_data.shape[0]
```

seanlaw: We've found that testing like this (i.e., moving the `nan`/`inf` around to multiple substitution locations but keeping the same original data within a run) has covered a lot of edge cases and is also consistent across other tests. Again, feel free to let me know if you'd rather I handle it after the merge. Happy to do so!

JosephTLucas: Of course, I'll make that change!

seanlaw: Great! Thank you and please let me know if you have any questions.

tests/test_floss.py
```diff
@@ -194,19 +194,21 @@ def test_floss():
...
-    old_data = data[:30]
-    n = old_data.shape[0]
+    data = T.copy()
+    for substitution_location in substitution_locations:
+        data[:] = T[:]
+        data[substitution_location] = substitute
+        old_data = data[:30]
+        n = old_data.shape[0]
 
     mp = naive_right_mp(old_data, m)
```
seanlaw, line 213: I think this entire section starting from this line and below need to be indented one level deeper so that sits within the `for stubstition_location in substitution_locations:` for-loop

JosephTLucas: doh!  Good catch. You're always on top of it, Sean.

seanlaw: > You're always on top of it, Sean.

I think that the only conclusion that can be drawn from this is that I've stared at this code for far too long and, therefore, I'm probably dropping many, many other balls! 😭

JosephTLucas:
Just to level, I'm currently trying to troubleshoot the assertion failure to decide if:
A) The current implementation of `floss` doesn't handle nan/inf appropriately.
or
B) The naive test doesn't handle nan/inf appropriately.

Going to step through them more carefully with a toy example, but wanted to keep us in step.

seanlaw:
> Just to level, I'm currently trying to troubleshoot the assertion failure to decide if:
> A) The current implementation of `floss` doesn't handle nan/inf appropriately.
> or
> B) The naive test doesn't handle nan/inf appropriately.
> 
> Going to step through them more carefully with a toy example, but wanted to keep us in step.

I think it's likely the latter. I think you might need to add some additional checks after the distances are computed like:

```
D = naive_distance_profile(left_T[-m:], left_T, m)
D[zone_start:] = np.inf

left_T_subseq_isfinite = np.all(np.isfinite(rolling_window(left_T, m)), axis=1)
# If the query subsequence contains nan/inf, set all distances to infinity
if np.any(~left_T_subseq[-1]):
    D[:] = np.inf
# Replace all distances to infinity where the subsequence is non-finite  
D[~left_T_subseq_isfinite] = np.inf
```

I'm sure I'm missing something but this might be the right direction

seanlaw:
> I think it's likely the latter. You might need to add some additional checks after the naive distances are computed like:

> D = naive_distance_profile(left_T[-m:], left_T, m)
> D[zone_start:] = np.inf
> left_T_subseq_isfinite = np.all(np.isfinite(rolling_window(left_T, m)), axis=1)
> # If the query subsequence contains nan/inf, set all distances to infinity
> if np.any(~left_T_subseq[-1]):
>    D[:] = np.inf
> # Replace all distances to infinity where the subsequence is non-finite  
> D[~left_T_subseq_isfinite] = np.inf

Ahhh, I think I see what the issue is. It looks like `test_floss.py` is calling it's own `naive_distance_profile` function (this is at the top of the file) rather than calling the `naive.distance_profile` (notice the underscore in the former and the dot in the latter). So, if you replace `naive_distance_profile` with `naive.distance_profile` then I think you can ignore my comments above since `naive.distance_profile` automatically handles `nan`/`inf`

JosephTLucas:
Hoping to dig back in soon, Sean.  Just overextended myself a bit!

seanlaw:
> Hoping to dig back in soon, Sean. Just overextended myself a bit!

No problem! I just re-ran the pipeline to remind myself where things were as I had forgotten. Please take your time and let me know if/how I can help.

JosephTLucas:
I've set aside some time to work on this this weekend.  Mainly looking at the test implementation as the potential cause for the failing tests.

On a side note, I don't see any any Hacktoberfest PRs (spam or otherwise).  Oh well... it was pretty dramatic this year.

Hope all is well, Sean!

seanlaw:
Sounds good. FYI, [this related issue was fixed/merged](https://github.com/TDAmeritrade/stumpy/issues/235) and so you'll want to grab this

JosephTLucas:
@seanlaw I feel like I haven't been able to make suitable progress on this.  I'm okay if you want to mark it as `stale` and welcome other contributors.

seanlaw:
> I feel like I haven't been able to make suitable progress on this. I'm okay if you want to mark it as stale and welcome other contributors.

That's fair. I think it's pretty close so I may hop on it and see what I can do :)

seanlaw:
This was resolved in commit 908ba6513bf594dea3a6d775dca2ca71cbb34411

seanlaw:
Thank you for helping us make progress @JosephTLucas!

Issue #243, opened 2020-09-11
Add Tutorial for meter-swapping detection


Attol8:
A tutorial that reproduces section 3.6 of [this paper](https://link.springer.com/article/10.1007/s10618-019-00668-6). 

The supporting page with code and data for the paper is not present. I have downloaded the original dataset from [here](https://pureportal.strath.ac.uk/en/datasets/refit-electrical-load-measurements-cleaned). @seanlaw do you think we should email them to provide us with the data and cleaning code that they have used? it would solve many time-consuming tasks and ensure 100% reproducibility.

As usual, I am going to submit a pull request with the notebook, so we can discuss the technical details in there.

seanlaw:
@Attol8 Thank you for diving into this. For efficiency's sake, I think we may want to take a two pronged approach and:

1. Reach out to Eamonn Keogh directly and explain that the [accompanying website for the paper](https://sites.google.com/site/matrixprofiletopten/home) is empty and see if he is able to help. Unfortunately, I've tried to reach out to first author, Yan Zhu, before but haven't had much luck. 
2. Simultaneously, we should try to dig into it ourselves (I can help with this) and see how far we can get. I've often uncovered other issues when doing this and so it's a good sanity check.

What do you think?

Attol8:
Hi @seanlaw. I agree that dig into it ourselves may be the best approach. But, as you will see in the pull request here, the data they use is simply different from the original dataset, which implies that they took some preprocessing steps which would be interesting to uncover. However, the method works anyway so we can post a tutorial without reproducing the exact steps taken in the paper, which also proves that the methods actually work on slightly different datasets. 

Also, the paper is not meant to be as detailed as possible (they often do not list key parameters and are vague on the code implementations), thus a slightly different implementation may still fit our use-case. 

Let me know if you agree with me

seanlaw:
> But, as you will see in the pull request here, the data they use is simply different from the original dataset, which implies that they took some preprocessing steps which would be interesting to uncover. However, the method works anyway so we can post a tutorial without reproducing the exact steps taken in the paper, which also proves that the methods actually work on slightly different datasets.

I understand and agree. The most important thing is to convey the core idea (i.e., that one can actually detect a swap) using a clear example. To be pragmatic, I'll time box the data reproducibility part to Monday so that we move forward with what your PR proposes if we can't figure it out by then. I'll try to play around with the data a little and see what is going on.

jaydurant:
[...]

PR #244, opened 2020-09-11
Meter swap tutorial

Attol8:
Hi @seanlaw, as mentioned in issue #243 I have added a notebook in `/docs` containing a tutorial on how to use the MP to detect meter-swapping electricity thefts. 

Please Note:

* Each House in the consumption data is represented by a pretty large .csv. There are 20 houses in totals and I have only used 5 (1, 2, 3, 4, and 11). I have loaded the data from Google Drive (I was using Google Colab) in a folder called "consumption".
* I could not reproduce the same time series of the papers. The original [paper](https://www.nature.com/articles/sdata2016122.pdf) describing the dataset is pretty clear on the fact that there is a lot of missing information (Figure 3), while their time-series do not display this behaviour. I have tried many steps before coming to the conclusion that it is best to use the original dataset without preprocessing.
* Due to the time series size, I have resampled the data from 8-6s measurements to 5 mins records, I hope the notebook is clear on this part. 
* Due to data size, I used STUMPY GPU implementation on Google Colaboratory
* The notebooks leverage the Python library `tqdm`
* In the last section "Plot the discovered motifs", I could not plot the motifs found by the MP as the index returned is always `-1`. I tried to solve this issue but couldn't figure it out, maybe you already encountered this bug. Also, this part is not strictly necessary so you can just take it out from the notebook.

Let me know what are your thoughts on the above points. Overall, this section proved to be much more difficult to implement compared to the Shapelet Discovery one, mainly because of the nature (size and shape) of the dataset. 

I will also proceed with the other sections from the paper 😃 

# Pull Request Checklist

- [x] Create a new branch
- [x] Install `black` (i.e., `python -m pip install black` or `conda install black`)
- [x] Install `flake8` (i.e., `python -m pip install flake8` or `conda install flake8`)
- [x] Install `pytest-cov` (i.e., `python -m pip install pytest-cov` or `conda install pytest-cov`)


seanlaw:
Thank you, @Attol8! I will take a look

seanlaw:
@Attol8 I'm going to provide some feedback below and it may come across as harsh/strong so I want to make sure you understand that my goal is to be straightforward. Please do not take it personally as I would like for this to be a discussion.

> Each House in the consumption data is represented by a pretty large .csv. There are 20 houses in totals and I have only used 5 (1, 2, 3, 4, and 11). I have loaded the data from Google Drive (I was using Google Colab) in a folder called "consumption".

> I could not reproduce the same time series of the papers. The original paper describing the dataset is pretty clear on the fact that there is a lot of missing information (Figure 3), while their time-series do not display this behaviour. I have tried many steps before coming to the conclusion that it is best to use the original dataset without preprocessing.
Due to the time series size, I have resampled the data from 8-6s measurements to 5 mins records, I hope the notebook is clear on this part.
Due to data size, I used STUMPY GPU implementation on Google Colaboratory

So, I think it is fine that we use only a few houses but I did see that the paper only takes data between "January 1st" to "December 31st". I did a quick analysis and found that only these files contain a full year of data (in 2014):

```
House_1.csv
House_2.csv
House_3.csv
House_4.csv
House_5.csv
House_6.csv
House_7.csv
House_8.csv
House_9.csv
House_10.csv
House_15.csv
```

For now, it is fine to use Colab but when the notebook gets published then we'll need to switch everything to using `stumpy.stump`. Alternatively, we could compute approximate matrix profiles (and AB-joins) using `stumpy.scrump` and apply this to a sub-sampled data set and also for data in 2014. Once we get to around 100K data points or less, it should be okay for CPUs (for STUMPY v1.5.0).


> The notebooks leverage the Python library tqdm

I get why you are using `tqdm` but, for the tutorials, I think it's unnecessary and rather distracting. These tutorials will mostly be read-only so the status bar is mostly irrelevant since the user wouldn't actually be waiting. And, sure, some might try to run the notebooks locally but I think that isn't the majority. Would you mind removing the `tqdm` components?

> In the last section "Plot the discovered motifs", I could not plot the motifs found by the MP as the index returned is always -1. I tried to solve this issue but couldn't figure it out, maybe you already encountered this bug. Also, this part is not strictly necessary so you can just take it out from the notebook.

In your notebook, which column are you referring to? I wonder if you are simply referencing the wrong column? Since you are performing an AB-join, the third and fourth columns are the left and right matrix profile indices, respectively, but they will always be `-1` for AB-joins. Note that the third column has an index of `2` and the fourth column has an index of `3` (since it is zero-based indexing in NumPy). To get the global matrix profile index (column 2), make sure you are using an index of `1`. Let me know if that helps

Attol8:
>@Attol8 I'm going to provide some feedback below and it may come across as harsh/strong so I want to make sure you understand that my goal is to be straightforward. Please do not take it personally as I would like for this to be a discussion.

@seanlaw I am here to learn, so no need to point this out! Please be as straightforward as possible in your feedback, as I find it extremely useful.

>So, I think it is fine that we use only a few houses but I did see that the paper only takes data between "January 1st" to "December 31st". 

Yes, that's true but even if we take  "January 1st" to "December 31st". the time series plots still look different from those of the paper. Also, the paper states that they used 2013 data, which cannot be possible.

Given the divergences between data and paper, I would keep the data that I have used in the notebook, we show a full example with 5 Houses with different patterns in consumption and it works with incomplete and heterogeneous data. Also, please note that when I do: `consumptions[f"House{i}"] = consumptions[f"House{i}"].resample('5T').mean()`, we reduce our time series from roughly 4M data points to 100,000. Thus, the notebook already contains a subsample of the original data.

>we could compute approximate matrix profiles (and AB-joins) using `stumpy.scrump`

Yes, what do you suggest? should we keep the nb like this for now and rewrite it when you put it into the docs?

>I get why you are using tqdm but, for the tutorials, I think it's unnecessary and rather distracting.

Totally agree here, I have removed `tqdm` from the notebook.

>In your notebook, which column are you referring to? I wonder if you are simply referencing the wrong column?

Sorry, I was really unclear on this issue. My aim is to plot the top discovered motif from the Heads and Tails join (like Fig. 9 -bottom- in the paper). To do so, I need 2 indices of the top motif: one from the 'Heads' query subsequence and one from the 'Tails' with its nearest neighbour. I thought that by accessing the third and fourth columns of the AB-join MP min value (`mp[np.where(mp[:, 0] == mp[:,0].min())].flatten()`) I would get the desired result. Maybe I am getting confused here, but I do not see how to accomplish this task

seanlaw:
> @seanlaw I am here to learn, so no need to point this out! Please be as straightforward as possible in your feedback, as I find it extremely useful.

Great! I am glad that we have established a proper understanding. It is great to work with you!

> Yes, that's true but even if we take "January 1st" to "December 31st". the time series plots still look different from those of the paper. Also, the paper states that they used 2013 data, which cannot be possible.

Yes, I noticed this as well. My goal is to avoid having to explain/justify the pre-processing too much. I think it is reasonable to say "the data is too big and so we'll reduce our dataset by subsampling and..." (anything contained nicely in a sentence). This probably isn't an issue now. I'll think about it more after the this PR is merged.

> Given the divergences between data and paper, I would keep the data that I have used in the notebook, we show a full example with 5 Houses with different patterns in consumption and it works with incomplete and heterogeneous data. 

I am cool with this! 👍 

> Also, please note that when I do: consumptions[f"House{i}"] = consumptions[f"House{i}"].resample('5T').mean(), we reduce our time series from roughly 4M data points to 100,000. Thus, the notebook already contains a subsample of the original data.

Yes, I noticed the resampling. Have you tried downsampling to 10K data points (instead of 100K)? If the result is the same with 10K data points then we might want to show that instead and then compare that with `stumpy.scrump` in a bonus section. Just a thought and by no means obligated.

Why are we doing `.mean()` here?

> we could compute approximate matrix profiles (and AB-joins) using stumpy.scrump
> Yes, what do you suggest? should we keep the nb like this for now and rewrite it when you put it into the docs?

Yes, let's just use `stumpy.stump` for now (this should be fast enough on a modest laptop/desktop and with the data downsampled to 10K).

As for `stumpy.scrump`, I would strongly recommend looking at this tutorial.

Essentially, you can replace all of your `stumpy.gpu_stump` function calls with

```
mp = stumpy.scrump(T_A, m, T_B, ignore_trivial=False, percentage=0.01, pre_scrump=True)
```

These matrix profiles should be approximately equal to computing the full matrix profile. When `percentage=1.0` then the matrix profile should be exactly the same as calling `stumpy.stump` or `stump.gpu_stump`. The difference is that we might be able to get away with only computing 1% of the distances for the matrix profile.

> I get why you are using tqdm but, for the tutorials, I think it's unnecessary and rather distracting.
> Totally agree here, in my newest commit I have removed tqdm from the notebook.

Thank you!

In your notebook, which column are you referring to? I wonder if you are simply referencing the wrong column?

Sorry, I was really unclear on this issue. My aim is to plot the top discovered motif from the Heads and Tails join (like Fig. 9 -bottom- in the paper). To do so, I need 2 indices of the top motif: one from the 'Heads' query subsequence and one from the 'Tails' with its nearest neighbour. I thought that by accessing the third and fourth columns of the AB-join MP min value (mp[np.where(mp[:, 0] == mp[:,0].min())].flatten()) I would get the desired result. Maybe I am getting confused here, but I do not see how to accomplish this task

Generally, I think what you want is something like:

```
idx = np.argmin(mp[:, 0])  # This is the index location for the smallest matrix profile distance
match_idx = mp[idx, 1]  # This is the index location for the nearest neighbor
```

I don't remember off the top of my head which index points to the "Head" time series and which index points to the "Tail" time series (i.e., you may need to swap them). Let me know if this makes sense.

seanlaw:
Now that I think about it, the reason why I like the idea of using only the houses that contain data from "January 1st-December 31st, 2014" is that then the user no longer has to think about the year and they only need to focus on the month. It is much easier to convey, "Each of our houses contains a year's worth of data from January 1st to December 31st" (and ignore 2014 completely). Then, we choose a "Tail" that starts on "November 1st" or "October 1st" instead of "November 10th". "November 10th" just seems very arbitrary. "October 1st" is very clearly the last quarter of the year and feels like a natural choice that requires very little justification. 

What do you think?

seanlaw:
I feel like the Head Tail swapping parts and pairwise matrix profile calculations need to be shortened. It looks and feels very long for something that is claimed to need only a few lines of code

seanlaw:
> Actually, instead of downsampling, I think it would be a lot cleaner take each dataframe, groupby every five minutes, and sum up the data every 5 minutes so that they'd be, say, 5 minute aggregates. This way, we're not throwing away data. In practice, one should probably be doing it this way. What are your thoughts?

Thinking about this more, if everything is grouped by the same 5-10 minute bins then we might be able to store all of the data for the 5 houses into a single Pandas dataframe where the index is the datetime and each column is the aggregated energy consumption for each house.

Attol8:
Hi @seanlaw ! First of all sorry if it took me some time to answer but I have started uni again and I have a lot of coursework piling up, so I can dedicate less time to this. Hope that is not an issue 😃 
I like your recommendations for future tutorials, I will make sure to follow it to avoid useless implementations. 

In my latest commit I have:

* Went to great lengths to make the code shorter and easier to read. Unluckily, this tutorial requires a bit of data pre-processing that we cannot exclude.
* Thanks to your suggestions, Plotted join discovered motifs. I believe they really do make everything clearer. 
* Downsampled data to hourly samples (which is roughly 10K data point for each time series). Thus, switched from `gpu_stump` to `stump` with great success, the code is fast on Google Colab cpu.

Lastly, I could not implement `stumpy.scrump` because it meant that I had to add 4-6 lines of codes in the algorithm part. Also, I incurred in some `DivisionByZero` errors in the preprocessing part. 

With regards to the usefulness of this notebook, I will let you decide on that. I do think that the current notebook is pretty accurate and to the point. I will now start another tutorial with your recommendation  👍

seanlaw:
> Hi @seanlaw ! First of all sorry if it took me some time to answer but I have started uni again and I have a lot of coursework piling up, so I can dedicate less time to this. Hope that is not an issue 😃

Congratulations on the successful start of another school year, @Attol8! Please focus on your coursework and no need to worry about delays or absences from STUMPY development. I appreciate everything that you've done and hope that you'll have an amazing semester! 

> I like your recommendations for future tutorials, I will make sure to follow it to avoid useless implementations.

Awesome!

> In my latest commit I have:
> Went to great lengths to make the code shorter and easier to read. Unluckily, this tutorial requires a bit of data pre-processing that we cannot exclude.

Thank you for digging into it. I really appreciate your persistence! I'm still working my way through the Shapelet tutorial but once that's done then I can spend some time on this tutorial. 

> Thanks to your suggestions, Plotted join discovered motifs. I believe they really do make everything clearer.
Downsampled data to hourly samples (which is roughly 10K data point for each time series). Thus, switched from gpu_stump to stump with great success, the code is fast on Google Colab cpu.

Very cool! I took a quick look and the motif looks pretty close so nice work. I can work on porting it over from Colab so that everything runs locally and maybe see if there's a difference if we aggregate by hour.

>Lastly, I could not implement stumpy.scrump because it meant that I had to add 4-6 lines of codes in the algorithm part. Also, I incurred in some DivisionByZero errors in the preprocessing part.

Okay, no problem. `stumpy.scrump` was a "nice to have" and by no means required. 

> With regards to the usefulness of this notebook, I will let you decide on that. I do think that the current notebook is pretty accurate and to the point. I will now start another tutorial with your recommendations in mind 👍

Perfect! I can't stress enough how thankful I am for your efforts and I'm sure that STUMPY users will also find the tutorials useful. Thank you!

seanlaw:
@Attol8 I think this is ready to merge as the core of the ideas are all there. Similar to the Shapelet tutorial, I'm going to make some changes and then push those once they are ready. I'll let you know once I've pushed the updates. Thank you!

Attol8:
Thanks for the kind words @seanlaw !

PR #288, opened 2020-12-02
Fix mass absolute nan

### LINKED ISSUE #286: mass_absolute returns NaN for some negative time-series
I found this while playing with the PAMAP dataset. Haven't been able to reproduce this with random numbers, so here's the relevant data and code.
```python
Q = np.array([-13.09, -14.1 , -15.08, -16.31, -17.13, -17.5 , -18.07, -18.07,
       -17.48, -16.24, -14.88, -13.56, -12.65, -11.93, -11.48, -11.06,
       -10.83, -10.67, -10.59, -10.81, -10.92, -11.15, -11.37, -11.53,
       -11.19, -11.08, -10.48, -10.14,  -9.92,  -9.99, -10.11,  -9.92,
        -9.7 ,  -9.47,  -9.06,  -9.01,  -8.79,  -8.67,  -8.33,  -8.  ,
        -8.26,  -8.  ,  -7.54,  -7.32,  -7.13,  -7.24,  -7.43,  -7.93,
        -8.8 ,  -9.71])
print(stumpy.core.mass_absolute(Q, Q))
```
I get `nan`, but of course it should be `0`. The problem is some float imprecision in `_mass_absolute` that leads to negative values that can't be properly `np.sqrt`ed.

Easy to fix, of course. A test for this edge case should be added as well.

DanBenHa:
Fixes #286 

# Pull Request Checklist

- [x] Create a new branch

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -861,7 +861,9 @@ def _mass_absolute(Q_squared, T_squared, QT):
     `See Mueen's Absolute Algorithm for Similarity Search \
     <https://www.cs.unm.edu/~mueen/MASS_absolute.m>`__
     """
-    return np.sqrt(Q_squared + T_squared - 2 * QT)
+    D = Q_squared + T_squared - 2 * QT
+    D[D < 0] = 0.0
+    return np.sqrt(D)
 
 
 def mass_absolute(Q, T):
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -430,6 +430,66 @@ def test_mass_absolute_T_inf(Q, T):
     npt.assert_almost_equal(ref, comp)
 
 
+def test_mass_absolute_sqrt_input_negative():
+    Q = np.array(
+        [
+            -13.09,
+            -14.1,
+            -15.08,
+            -16.31,
+            -17.13,
+            -17.5,
+            -18.07,
+            -18.07,
+            -17.48,
+            -16.24,
+            -14.88,
+            -13.56,
+            -12.65,
+            -11.93,
+            -11.48,
+            -11.06,
+            -10.83,
+            -10.67,
+            -10.59,
+            -10.81,
+            -10.92,
+            -11.15,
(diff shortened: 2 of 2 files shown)


seanlaw:
@DanBenHa Thank you for the PR! I couldn't find much information on when negative values may occur except for one article that simply set those values to zero before taking the square root. I am wondering if there are situations where setting the negative values to zero is wrong?

Otherwise, everything else looks good and this should be ready to merge.

DanBenHa:
> I am wondering if there are situations where setting the negative values to zero is wrong?

Good question. I don't know. But that would imply calculating a complex value for the distance of two real-valued vectors. Would that make sense?

seanlaw:
> Good question. I don't know. But that would imply calculating a complex value for the distance of two real-valued vectors. Would that make sense?

Since negative values imply some sort of deficiency in the numerical precision, I think what I'm really asking is if there might be situations where the amplitude of the negative value may be indicative of something that is a lot worst (while a small, near zero amplitude is due to numerical precision).  Just thinking out loud here though :)

DanBenHa:
> Since negative values imply some sort of deficiency in the numerical precision, I think what I'm really asking is if there might be situations where the amplitude of the negative value may be indicative of something that is a lot worst (while a small, near zero amplitude is due to numerical precision). Just thinking out loud here though :)

I see. I guess it could be indicative of problems in the computation of `Q_squared`, `T_squared` and `QT`, so primarily in the functions `rolling_is_finite`, `sliding_dot_product` and `rolling_window`.

seanlaw:
Okay, I spent most of my morning looking for an ideal solution and, while there were many discussions, there were no obvious solutions

DanBenHa:
> Okay, I spent most of my morning looking for an ideal solution and, while there were many discussions, there were no obvious solutions

Haha, reminds me of some of my mornings ;)

Thanks for merging and the interesting question raised.

Issue #211, opened 2020-07-02
Add Geometric Time Series Chains


seanlaw:
There has been recent work to help improve the robustness of time series chains. 

See this [published conference paper](https://github.com/TDAmeritrade/stumpy/files/6536717/Matrix.Profile_XXI_Geometric.Chain.Paper.pdf)

See this [supplementary site](https://sites.google.com/site/timeserieschains/)

The data for Fig 12 can be found [here](https://www.cs.ucr.edu/~eamonn/public/Lab_C_060314.mat)

See this short [YouTube video](https://www.youtube.com/watch?v=VQpzVGYrm_g)

Ninimama:
Hi Seanlaw,

Thanks for creating this wonderful package. I was wondering if I could help you with implementing this. 

I was using TSC17 for my data and got some weird result, and, in my case study, the problem was curse of dimensionality and the possible noises. Then I realized there is this new algorithm that can enhance the quality of the discovered chains. 

According to the paper, we need to implement Table 4 and 5 and a few evaluation metrics. I haven't contributed to any project before (in terms of adding a code to a project), so it would be great if you could provide me with a  link that shows the rules/styles I should follow. Your package already provides the left matrix profile, which is referred to as "backward matrix profile index" in the Table 4. So, I think we have things ready. 

And, one last question for now: The formula for the angle provided in the section 4.1 of the paper doesn't need to normalize each vector. Right?

seanlaw:
@Ninimama Thank you for your interest in contributing to STUMPY! We recommend going over this [Contributor's Guide](https://stumpy.readthedocs.io/en/latest/Contribute.html) just to familiarize yourself with the general process.

As for Geometric Time Series Chains, I haven't yet gone over the paper in detail so we certainly welcome contributions! As you can see with all of our [Tutorials](https://stumpy.readthedocs.io/en/latest/tutorials.html), we always try to reproduce one or more examples directly from the published work (and using their datasets too) by reproducing one or more of the published figures in the paper. This serves to ensure that what we are doing is consistent with the published work and it allows the user to go back to the paper for more details beyond the tutorial.

[...]

Issue #270, opened 2020-10-18
Add Tutorial Data to Zenodo


seanlaw:
All of our tutorials reproduce published work by downloading them from the original source. These sources are typically on Google sites and may eventually become stale or unsupported. It would be best to back them up on Zenodo and have one consistent place to download the data from. Then, we can update our tutorials to point to Zenodo instead.

seanlaw:
In issue #107 @MokaPot said:

> You mentioned Zenodo in another issue? That seems like a good way to do this? Could preprocess the data into clean pandas friendly csvs, upload and the just call url from read_csv? Guess it would be good to establish some structure upfront?

Moving this part of the conversation here 😄 

This Zenodo idea is still swimming around in my head and isn't well formed yet. What I'm trying to avoid is storing data in a Github repository but I am also concerned that the original data sources from the published papers may one day disappear. I tend to be a purist when it comes to "raw data" and I prefer retrieving them from the source and showing the pre-processing in code (there are tradeoffs to this but I prefer to be transparent where possible). 

I've actually tried creating rough CSVs of only the un-processed data that we need in our [Github Wiki](https://github.com/TDAmeritrade/stumpy/wiki) (see the list on the right of this page). I was hoping that we could point our Tutorials there. However, I found that Github was actually limiting the number of times of text and truncating/deleting some of the data so this turned out to be a bad choice. So, we'll need a better alternative.

As you are aware, in our available Tutorial notebooks, we are currently retrieving data from the original data source and reading it into a Pandas DataFrame. I would like to store the un-processed data as a CSV into Zenodo. What's unclear to me is the rights to maintain that data (not ownership of the data) can be managed. Essentially, if you or somebody else uploads the data, how can I maintain it (from an authorization perspective)? Perhaps, there is a way for me to create a "community" and then give other people read/write access? This part is still unclear to me.

Until we figure it out, I don't mind having the Tutorial notebooks download the data from its original source. Perhaps, you had some ideas or thoughts?

MokaPot:
Yep looks like you can be the community curator you then accept/reject data or other media. However there is currently a long standing issue on co-curating or transfering  (https://github.com/zenodo/zenodo/issues/810) which doesn't seem to have gained much traction. 

The only other alternative I know of off the top of my head is kaggle, but I'm sure there are plenty of other options. I'll keep an eye out.

[...]

PR #384, opened 2021-05-24
[WIP] Tutorial on Time Series Geometric Chain 

Ninimama:
Initial attempt on the Issue #211:   Creating Tutorial for Time Series Geometric Chain
===============================================
## Pull Request Checklist

Below is a simple checklist, but please do not hesitate to ask for assistance!

- [x] Create a new branch


Ninimama:
Please feel free to let me know anything that I should better consider in the future. I am not sure if the amount of changes corresponding to each commit is good, or I should do commit more frequently. 

I also need to work with Zenodo.ipynb and upload the data into Zonodo. For now, I just directly retrieve the data from the source provided by the authors of the journal.

seanlaw:
@Ninimama Thank you for this PR. Please give me some time to review and provide feedback

Ninimama:
@seanlaw: 
As you suggested, I am about to do small changes and make commit and do the PR again. However, the changes I am about to make is going to be on top of the file whose pull request is pending. (I just realized I need to modify a few things and also add some new functions to the notebook.). 

So, I checked stackoverflow and read different answers. I read somewhere that I can create a new branch from the existing branch (the one for which I already did PR):

> main -> my first branch -> second branch

Then, I can do PR for the second branch. But I also read in another post that it is usually better to do my changes in my second branch, but not do the second PR. And instead, I should better wait for the feedback from the owner for the first PR. Then, (after updating my local repo), do the second pull request.

I was wondering if you could help me with this matter or provide me with a link through which I can better understand the work flow. I searched different websites. However, I think I am still confused regarding this matter.


Thanks,
Nima

seanlaw:
> As you suggested, I am about to do small changes and make commit and do the PR again. However, the changes I am about to make is going to be on top of the file whose pull request is pending. (I just realized I need to modify a few things and also add some new functions to the notebook.).

@Ninimama Sounds good!

> I was wondering if you could help me with this matter or provide me with a link through which I can better understand the work flow. I searched different websites. However, I think I am still confused regarding this matter.

Great question (and likely something that isn't covered in our Contributor Guide). I remember doing this when I first started and it was somewhat confusing and magical at the same time. I tried looking for a description but I couldn't find one immediately so let me outline the process flow below:

1. Fork the main repo (TDAmeritrade/stumpy)
2. Clone your forked repo to your computer (Ninimama/stumpy)
3. Create a new branch (Geometric_Chain)
4. Make changes to the code base (i.e., add a "docs/Tutorial_Geometric_Chains.ipynb" notebook)
5. Commit changes to the "Geometric_Chain" branch and add a useful commit message
6. Push the changes back to your forked repo (Ninimama/stumpy)
7. Submit a PR to the TDAmeritrade/stumpy repo from Ninimama/stumpy

Hopefully, this process should look familiar to you as this is exactly what you've done so far! So, what happens when you have additional changes that are coming after the initial PR was submitted? Let's continue:

8. Continuing with the existing "Geometric_Chain" branch (you do **not** need a new branch), you make changes to the existing "docs/Tutorial_Geometric_Chains.ipynb" notebook
9. Commit these additional changes to the existing "Geometric_Chain" branch and add a useful commit message
10. Push the changes back to your forked repo (Ninimama/stumpy)

And here is where the Github "magic" happens. ALL changes that you make to your local "Geometric_Chain" branch that get pushed back up to your forked repo (Ninimama/stumpy) will automatically be pushed to the PR in the TDAmeritrade/stumpy repo. In other words, just like above where you see your initial commit messages, anything that you push back up to your forked repo will end up here as well without you having to do anything extra! There is no need to create a separate PR as long as all of the work (commits) are going toward the same goal. 

To convince yourself, continue in the "Geometric_Chain" branch of your local clone (on your machine) and, say, add a comment to the notebook. And follow steps 8-10. Then, you should see the new commit message above!

Go ahead. Try it out! Let me know what you learn :)

If you don't mind, it would be great if you could describe this in our Contributor Guide too! 

>  I should better wait for the feedback from the owner for the first PR

Generally speaking, if I see an initial PR waiting for me then I will try to find time to take a look. However, we will use this PR comment space to keep each other informed. Similar to how I informed you that I might need a bit of time to review, you can also inform me by communicating things like:

"This is work-in-progress and I I will message you in 2 days when I think it is in a reasonable spot for you to provide some feedback"

or

"I've started working on the tutorial but I'm stuck on X, Y, and Z and I would like some feedback"

So, it really depends on what support you need from me and this can be different for different PRs. Just know that I am here to help! To reiterate, the most important thing is to have me look at things in small, easily consumable chunks. I trust that you will make good decisions and, when in doubt, please do not hesitate to ask and err on the side of caution. Again, I am here to  help!

Ninimama:
@seanlaw 
**Thanks! That helps a lot! I will try!!!** 

So, I think I should better revise some parts and probably get rid of the implementation for now (I mean the function that I wrote in the notebook which is in my last commit)! So, it will be just the introduction part, and I will do its commit soon. Hence, it will be a smaller size to be reviewed by you. (If you think I should still downsize it, please let me know) Then, in the following commits, I will add the functions (one after anther) to provide the implementation of Geometric chain in the notebook. 

**Thanks again for  the clarification. Just one last question for now:**  
You mentioned that it is preferable to "look at things in small, easily consumable chunks." So, just to confirm, I can do several commits and push it back to my fork one after another. **What is important**  is that each commit corresponds to a small-size change. Right? If not, then I assume you meant I should better commit for a small change **and wait for your feedback**  before doing the next commit.

seanlaw:
> You mentioned that it is preferable to "look at things in small, easily consumable chunks." So, just to confirm, I can do several commits and push it back to my fork one after another. What is important is that each commit corresponds to a small-size change. Right? If not, then I assume you meant I should better commit for a small change and wait for your feedback before doing the next commit.

Right. This was referring to having each commit be "small" but also know that this isn't a hard rule and more of a recommendation for you to consider. From my experience, approximately 50 lines of code or less (excluding docstrings) is reasonable and likely corresponds to 5 or fewer Python functions.  So, I think this comment is much more important when writing code (vs tutorials as tutorials allow you to annotate and explain the surrounding code). To be clear, I may not review every commit and might allow commits to accumulate before reviewing but it is much nicer to have commits broken down into related chunks. Then again, at the end of a PR, all commits may be "squashed" into one giant commit anyways since each PR should solve ONE Github issue. 

I'm not at all recommending this but if you wanted to be really disciplined, you could break up the geometric chains issue into smaller sub-issues and then submit indidividual PRs for each sub-issue. But this is extreme and the "right" approach is likely somewhere in the middle but this choice is more art than science.

Ninimama:
@seanlaw 
> But I try to make sure that the commits are pieces or work that I am reasonably satisfied
I usually use **bold**  font to make the reader pay attention. Not sure if that's a good style for STUMPY.

Ninimama:
@seanlaw 
(As discussed in  #270)
> For now, I think what we'd want to do is to only upload data that is currently being used in our public tutorials (i.e., not any tutorials that are still a work-in-progress)

If you are still having the same opinion, I will read the data directly from the main resource.

seanlaw:
> If you are still having the same opinion, I will read the data directly from the main resource.

Yes, that sounds good! We can add it to Zenodo once we feel that the tutorial is at a good place (I can help with this). Then I will ask you to update the tutorial to pull directly from Zenodo instead of directly from the main source.

Ninimama:
@seanlaw:

I implemented the function __DirectionAngular( )__ that is used in Table 4 of the [Matrix Profile XXI](https://dl.acm.org/doi/abs/10.1145/3394486.3403164). The function calculates the __direction angle__ as defined in __Definition 8__ in the paper. 

Similar to what the authors did, I first used the stumpy.allc() function to find non-geometric chains (a.k.a bi-directional chains). The goal was to calculate the direction angle for such chain.  So, I got the same chain that was shown in __Fig 11 of the paper__. However, the __direction angles__ are NOT the same as the ones provided in the __Table 1 of the paper__. I checked out the function __DirectionAngular( )__ several times and everything seems fine to me. 

It is worthwhile to note that I reproduced the __Geometric Chains__ as the ones provided in the [supplementary materials ](https://sites.google.com/site/timeserieschains/home) for the same Data (I haven't shown it in this tutorial yet). So, I think my function should work correctly. The numbers provided in the __Table 1 of the paper__ might be wrong though!!!

So, Could you please check out the __definition 8 of the paper__ and the function __DirectionAngular( )__ to see if there is anything that I am missing or not?

seanlaw:
> @seanlaw:
> 
> I implemented the function **DirectionAngular( )** that is used in Table 4 of the [Matrix Profile XXI](https://dl.acm.org/doi/abs/10.1145/3394486.3403164). The function calculates the **direction angle** as defined in **Definition 8** in the paper.
> 
> Similar to what the authors did, I first used the stumpy.allc() function to find non-geometric chains (a.k.a bi-directional chains). The goal was to calculate the direction angle for such chain. So, I got the same chain that was shown in **Fig 11 of the paper**. However, the **direction angles** are NOT the same as the ones provided in the **Table 1 of the paper**. I checked out the function **DirectionAngular( )** several times and everything seems fine to me.
> 
> It is worthwhile to note that I reproduced the **Geometric Chains** as the ones provided in the [supplementary materials ](https://sites.google.com/site/timeserieschains/home) for the same Data (I haven't shown it in this tutorial yet). So, I think my function should work correctly. The numbers provided in the **Table 1 of the paper** might be wrong though!!!
> 
> So, Could you please check out the **definition 8 of the paper** and the function **DirectionAngular( )** to see if there is anything that I am missing or not?

Yes, sounds good! Thanks for letting me know. I will take a look.

seanlaw:
@Ninimama I looked through your code as well as the Matlab code and nothing stands out to me.

> It is worthwhile to note that I reproduced the Geometric Chains as the ones provided in the supplementary materials for the same Data (I haven't shown it in this tutorial yet). So, I think my function should work correctly. The numbers provided in the Table 1 of the paper might be wrong though!!!

Can you please elaborate on this point? What did you reproduce? So, the chains identified by `atsc` is nothing new and it is rather important that we are computing the direction angles correctly. Are you able to reproduce any of the direction angles?

Ninimama:
@seanlaw:
Thanks for taking a look at the code. Exactly! That's why I have concerns about those angles. I also tried non-normalized version of subsequences but that didn't help to get the angles provided in the Table 1 of the paper. 

> Can you please elaborate on this point? What did you reproduce?

I got the results pointed out in Section 5.2 of the paper using the same settings. For instance, I considered the user-defined THETA=40 degree as chosen by the authors (please see below for the claims provided by the authors of the paper):

> We ran both ___TCS17(atsc)__ and __TSC20(geometric chain)__ on both datasets using a length of 100 and found:
* For sine, TSC20 discovered no chain. This is exactly what we would hope to discover.
* For damped, TSC20 discovers a chain of length seven that reflects the subtle evolving flattening of the sine waves.
* For sine, TSC17 discovered a chain of length six. This is disquieting. Clearly by any reasonable definition, this dataset is not evolving, this suggests that TSC17 can find spurious chains.
* For damped, TSC17, discovers a chain of length four. Compared to TSC20, this is a much shorter chain, barely longer than we might expect by chance.
-----------------------------------------------------------------

I also got the same result provided in the [damped_sine_wave(page 4,5)](https://sites.google.com/site/timeserieschains/materials/Fig_16_1_ver_02.pdf?attredirects=0&d=1). 

In page 4: The indices of my subsequences in Geometric chain are the same as the ones provided there.
In page 5: The histogram I got for the length of chains are the same as the ones provided (with title: TSC20)

----------------------------------------------------------------
I recently checked the angles for a different chain of the same data (i.e. sine_with_noise data), mentioned in the [sine_wave(page 3)](https://docs.google.com/viewer?a=v&pid=sites&srcid=ZGVmYXVsdGRvbWFpbnx0aW1lc2VyaWVzY2hhaW5zfGd4OjdmZmNmMTNiZDU4YzdiNWQ). I pushed the commit, so you can see it in the notebook.

In page 3: The Direction Angle I got are similar to the ones provided by the author for this chain. 
__NOTE__: However, as mentioned in my previous comment, my angles are different from the ones provided in the Table 1 of the paper (The angles are for the same data (i.e. sine_wave) but a different chain)

---------------------------------------------------------------

So, there are three possibilities:

1. My code is wrong! 
2.  The values provided in the Table 1 of the paper might be calculated for the sine_wave but without noise. (However, according to the caption of Fig. 11 of the paper, this shouldn't be true)
3. The values of the Table 1 of the paper is not correct.

--------------------------------------------------------------
Please let me know what you think.



__P.S.__ 
By the away, could you check out the commit message and let me know whether I should have better made it shorter and explained the things in the commit body instead.

Ninimama:
@seanlaw:
Thanks for taking a look at the code. Exactly! That's why I have concerns about those angles. I also tried non-normalized version of subsequences but that didn't help to get the angles provided in the Table 1 of the paper. 

> Can you please elaborate on this point? What did you reproduce?

I got the results pointed out in Section 5.2 of the paper using the same settings. For instance, I considered the user-defined THETA=40 degree as chosen by the authors (please see below for the claims provided by the authors of the paper):

> We ran both ___TCS17(atsc)__ and __TSC20(geometric chain)__ on both datasets using a length of 100 and found:
* For sine, TSC20 discovered no chain. This is exactly what we would hope to discover.
* For damped, TSC20 discovers a chain of length seven that reflects the subtle evolving flattening of the sine waves.
* For sine, TSC17 discovered a chain of length six. This is disquieting. Clearly by any reasonable definition, this dataset is not evolving, this suggests that TSC17 can find spurious chains.
* For damped, TSC17, discovers a chain of length four. Compared to TSC20, this is a much shorter chain, barely longer than we might expect by chance.
-----------------------------------------------------------------

I also got the same result provided in the [damped_sine_wave(page 4,5)](https://sites.google.com/site/timeserieschains/materials/Fig_16_1_ver_02.pdf?attredirects=0&d=1). 

In page 4: The indices of my subsequences in Geometric chain are the same as the ones provided there.
In page 5: The histogram I got for the length of chains are the same as the ones provided (with title: TSC20)

----------------------------------------------------------------
I recently checked the angles for a different chain of the same data (i.e. sine_with_noise data), mentioned in the [sine_wave(page 3)](https://docs.google.com/viewer?a=v&pid=sites&srcid=ZGVmYXVsdGRvbWFpbnx0aW1lc2VyaWVzY2hhaW5zfGd4OjdmZmNmMTNiZDU4YzdiNWQ). I pushed the commit, so you can see it in the notebook.

In page 3: The Direction Angles that I got are similar to the ones provided by the author for this chain. 
__NOTE__: However, as mentioned in my previous comment, my angles are different from the ones provided in the Table 1 of the paper (The angles are for the same data (i.e. sine_wave) but a different chain)

---------------------------------------------------------------

So, there are three possibilities:

1. My code is wrong! 
2.  The values provided in the Table 1 of the paper might be calculated for the sine_wave but without noise. (However, according to the caption of Fig. 11 of the paper, this shouldn't be true. Furthermore, considering sine without noise will probably lead to 0 in the denominator of arccos() provided in the Definition 8 of the paper.)
3. The values of the Table 1 of the paper is not correct.

--------------------------------------------------------------
Please let me know what you think.

seanlaw:
> So, there are three possibilities:
> My code is wrong!
> The values provided in the Table 1 of the paper might be calculated for the sine_wave but without noise. (However, according to the caption of Fig. 11 of the paper, this shouldn't be true. Furthermore, considering sine without noise will probably lead to 0 in the denominator of arccos() provided in the Definition 8 of the paper.)
> The values of the Table 1 of the paper is not correct.

So, the short answer is that "I don't know". This is precisely why we start with a notebook and why we try to reproduce the results from the paper. I've definitely discovered problems with their code/paper before and have reached out to the authors in the past to get clarification. It might make sense to step away for a few days and come back to it with fresh eyes. For what it's worth, your code looks reasonable from a "correctness" standpoint. Don't be surprised if the work is not reproducible 😄 

In the past, I have found the Matlab code to be useful for cross referencing. The files `TSC_Chain_calculate_CosAdj_glb.m` and `TSC_cosine.m` seem to be important. Let's focus on getting the direction angle right and take some time to understand what is really going wrong.

seanlaw:
@Ninimama I am thinking whether it is possible to recreate the direction angles for Table 2?

seanlaw:
@Ninimama Okay, after reviewing Table 4, I think it that it is incorrect to call `stumpy.allc` to find the longest chain(s). Instead, you identify the all chain set by leveraging the backward matrix profile indices (i.e., `IL`) and recreating Table 4.

seanlaw:
Actually, I take it back. It sounds like Table 1 and Table 2 were created by using `atsc` (from TSC17) and then only computing the direction angles for these chains. However, `atsc` doesn't seem to produce the same indices for Table 2. So, I can't really tell where the indices are coming from for Table 2.

Ninimama:
@seanlaw:
__In Table1:__ 
They use TSC17 (`atsc`) to get indices. Although the showed one chain with length 6, my notebook shows there are four of them. The authors showed two of these: One in their paper (let's call it Chain_I) and another one (let's call it Chain_II) in their supplementary document available here: [sine_wave(page 3)](https://docs.google.com/viewer?a=v&pid=sites&srcid=ZGVmYXVsdGRvbWFpbnx0aW1lc2VyaWVzY2hhaW5zfGd4OjdmZmNmMTNiZDU4YzdiNWQ). 

* Regarding Chain_II:  the Direction Angle that I got, and the one provided in the supplementary docs are the same. 

* Regarding Chain_I: It is worthwhile to note that I get the same subsequences of the chain (as we use `astc`). But the problem is the angles provided in that Table 1. __So, I run the MATLAB code:__ `TSC_Chain_calculate_CosAdj_glb.m` with the same input and I get the following output: [0.6504,    0.6337,    0.7047,    0.6631]

* __Now, if you look at the MATLAB file, you will see that it calculates the cos(theta) rather than theta__. So, the aforementioned values are the cos() of angles. So, after I took its COS_INVERSE, I got the same Direction Angle as the one I got with my Python Code for Chain_I. (Just in case you want to check out the results yourself, please note that, as opposed to my python code, the MATLAB code considers the indices of subsequences of the chain in descending order. And, consecutively, the output of the MATLAB code shows the Direction Angle of subsequence with the largest index at the beginning of the list and the Direction Angle of subsequence with the smallest index at the end of the list. So, if you reverse it, you can see the values are exactly the ones I got with my python code.) 

Therefore, I can say the Table 1 is probably incorrect.


__In Table 2:__ 
As the paper explains, __the Table 2 indices are extracted from the Geometric Approach (NOT `atsc`)__. I used the MATLAB code `TSC_Chain_calculate_CosAdj_glb.m`  first to see if I can get the Direction Angle for TABLE 2. This time, I get the same numbers (with MATLAB code).  So, the numbers of Table 2 should be correct. 

My implementation of the Geometric Time Series Chain leads to the same set of indices of the chain illustrated in Fig. 2. However, my Direction Angle are slightly different as shown below.

[6.335101215236893,
 10.434212996793763,
 10.894232140239529,
 11.495886582668518,
 10.184783440620357,
 23.592215094943352,
 27.03457573592252]

Compared to the values of Table 2, my values of Direction Angle are not EXACTLY the same BUT they are really CLOSE to the ones provided. We also calculate the cos() values of these angles, the outcome will be close to the cos() of the angles of Table 2. 

How can we figure out if this small error is NOT because of different numerical computation in Python and Library?

Ninimama:
@seanlaw 
And, one more thing: Finding the settings to reproduce the result of this paper has been difficult. For instance, for Table 2 (which is related to Fig. 2), you might think the WindowLength is 300 (according to the figure). However, if you check out the MATLAB code, it says 360 !!! Or the other time when I was trying to reproduce the result, I had to find the corresponding .m file to get the correct setting.

seanlaw:
> They use TSC17 (atsc) to get indices. Although the showed one chain with length 6, my notebook shows there are four of them. The authors showed two of these: One in their paper (let's call it Chain_I) and another one (let's call it Chain_II) in their supplementary document available here: sine_wave(page 3).

> Regarding Chain_II: the Direction Angle that I got, and the one provided in the supplementary docs are the same.

Great!

> Regarding Chain_I: It is worthwhile to note that I get the same subsequences of the chain (as we use astc). But the problem is the angles provided in that Table 1. So, I run the MATLAB code: TSC_Chain_calculate_CosAdj_glb.m with the same input and I get the following output: [0.6504, 0.6337, 0.7047, 0.6631]

> Now, if you look at the MATLAB file, you will see that it calculates the cos(theta) rather than theta. So, the aforementioned values are the cos() of angles. So, after I took its COS_INVERSE, I got the same Direction Angle as the one I got with my Python Code for Chain_I. (Just in case you want to check out the results yourself, please note that, as opposed to my python code, the MATLAB code considers the indices of subsequences of the chain in descending order. And, consecutively, the output of the MATLAB code shows the Direction Angle of subsequence with the largest index at the beginning of the list and the Direction Angle of subsequence with the smallest index at the end of the list. So, if you reverse it, you can see the values are exactly the ones I got with my python code.)

> Therefore, I can say the Table 1 is probably incorrect.

Okay, so this is good. I actually think that you should be following the paper's convention where `C1 > C2 > C3 > ... < Ck` (i.e., a chain in descending order).

> In Table 2:
> As the paper explains, the Table 2 indices are extracted from the Geometric Approach (NOT atsc). I used the MATLAB code TSC_Chain_calculate_CosAdj_glb.m first to see if I can get the Direction Angle for TABLE 2. This time, I get the same numbers (with MATLAB code). So, the numbers of Table 2 should be correct.

Great!

> My implementation of the Geometric Time Series Chain leads to the same set of indices of the chain illustrated in Fig. 2. 

Great! Can you please add this work to the Tutorial?

> However, my Direction Angle are slightly different as shown below.

> [6.335101215236893,
> 10.434212996793763,
> 10.894232140239529,
> 11.495886582668518,
> 10.184783440620357,
> 23.592215094943352,
> 27.03457573592252]

> Compared to the values of Table 2, my values of Direction Angle are not EXACTLY the same BUT they are really CLOSE to the ones provided. If we calculate the cos() values of these angles, the result will be close to the cos() of the angles of Table 2.

> How can we figure out if this small error is NOT because of different numerical computation in Python and MATLAB?

I don't believe that this is due to numerical precision. 

> And, one more thing: Finding the settings to reproduce the result of this paper has been difficult. For instance, for Table 2 (which is related to Fig. 2), you might think the WindowLength is 300 (according to the figure). However, if you check out the MATLAB code, it says 360 !!! Or the other time when I was trying to reproduce the result, I had to find the corresponding .m file to get the correct setting.

Yes, I noticed this as well.

Consider the fact that they've made the claim that "empirically any value between 30 and 50 gives near identical results". Based on Table 1 and Table 2, I'm not sure setting theta to 40 degrees is great since we're actually seeing numbers between 40 and 50 when we compute direction angles for Table 1 on our own (using `acos`)

> If you think we should better put our effort into something else (another issue), please let me know.

Sometimes, people make small/subtle mistakes (which is acceptable) but there seem to be many, many problems here. The quality of this work seems below the standard that we need in order to confidently reproduce the first two tables. The next thing I'd try is to see if your Python Geometric Chains implementation can reproduce the exact chains shown in Fig 12, Fig 14, and Fig 21.

Ninimama:
> Okay, so this is good. I actually think that you should be following the paper's convention where `C1 > C2 > C3 > ... < Ck` (i.e., a chain in descending order).

Sure. Will do.


> Great! Can you please add this work to the Tutorial?

Will push the commit soon.  I need to check one more time the functions and then will add them to the tutorial.

 

> I don't believe that this is due to numerical precision

So, I recently calculated one of the angles in python but without using the DirectionAngle( ) function. In other words, I just got two subsequences and calculated its Direction Angle. I got the same result as the one produced by the function DirectionAngle() in my notebook. So, I cannot see what's missing here. 
I installed MATLAB. I am going to cross-ref the values and discover the thing that is missing.




> I'm not sure setting theta to 40 degrees is great

I have the same opinion. For the sake of reproducing their work, I set its default value to 40 degree as the authors stated that they use this 40 degree for all of their results. However, it is not hard coded, and it can be changed later to a more reasonable value.



> The next thing I'd try is to see if your Python Geometric Chains implementation can reproduce the exact chains shown in Fig 12, Fig 14, and Fig 21

I took a look at the paper and supplementary files. And, I found No supplementary PDFs/CODE/DATA for these three figures. 
So, I will check out the [UCR Time Series Classification Archives](https://www.cs.ucr.edu/~eamonn/time_series_data_2018/) to get the data sets. I will try to reproduce Fig12 and 14. 
__NOTE:__ Fig 21 shows no indices for the subsequences in the chain, so I will probably skip that.

Ninimama:
@seanlaw:

Found the problem!
Please take a look at this MATLAB code  (a part  of `TSC_Chain_calculate_CosAdj_glb.m`):

image

As you can see, instead of taking the z-norm of subsequences and then calculates their difference,  it takes the difference and then z-normalize it! I did the former one myself and I think it makes more sense as the matrix profile is based on the z-norm of each subsequence for this case study.  To make sure that's the only problem, I changed my DirectionAngle function and made it similar to the one used int his MATLAB code and I got exactly the same angles. 


So, in terms of reproducing the work, we should do the same (i.e. normalizing after taking the difference of two subsequences) but I think the one that I did make more sense.

seanlaw:
> Found the problem!

Amazing detective work, @Ninimama! I had gone through that function and completely overlooked it myself. Well done!

> So, in terms of reproducing the work, we should do the same (i.e. normalizing after taking the difference of two subsequences) but I think the one that I did make more sense.

I completely agree with you here. We should always z-normalize our subsequences first and not the difference of two subsequences. Now that we understand the source of the problem. I think it is okay to "fix" it (i.e., ignore their approach) and do it the logical way (i.e., z-normalizing individual subsequences).

I think this is an error and we should document this in the Notes section of the `direction_angle` docstring (we've done this before with other functions when we found errors). Now the key question is whether or not it makes sense for us to set a different default `theta` value?

seanlaw:
I just want to reiterate that this is very exciting! Thank you for sticking with this!

Ninimama:
@seanlaw 
Thanks for your support and input throughout the process. I truly appreciate your patience 😃  

I will continue working on the tutorial. The narrower the theta the more restriction we are putting on the chains and therefore fewer chains (and maybe shorter ones) may be achieved. So, we may use significance score or maybe a new metric (?!) to find an optimal one. 

For now, I can work on the Tutorial and complete the functions. I will try and see how the results would like on the data used for Fig 12 and 14 only if I can find the correct data! Although the results might be different compared to the one presented in the paper,  it would be nice to add that to the notebook. I will probably complete it in a couple of days.

seanlaw:
> I will continue working on the tutorial. The narrower the theta the more restriction we are putting on the chains and therefore fewer chains (and maybe shorter ones) may be achieved. So, we may use significance score or maybe a new metric (?!) to find an optimal one.

So, this may be a good place to reinforce that the goal of this PR should be strictly limited to reproducing the paper. With the exception that we z-norm the subsequences, everything else should be the same try to be same as the paper where possible. In parallel, it would be good to create a separate Github issue to track all of the "additional" work that may need to be explored in addition to the original issue. In other words, we want the scope of each PR to always be reasonably well defined and to NOT allow for "scope creep" into the issue (i.e., let's try to keep issues in bite sized pieces and accomplish them with a corresponding PR). But always feel empowered to create a new issue that defines/scopes the NEXT logical stage (i.e., maybe a new metric or not?) or maybe even put it into the discussion section and we can have an open debate before coding. Again, the goal is to make sure that we don't waste time and continue to work collaboratively. 

To reiterate, this PR should really be about recreating Table 4 in code. This will include creating a, say, `geom_chains` function that calls a `_direction_angle` function and a `_significance_score` function. Anything else should probably be in a separate issue (beyond switching to z-normalizing subsequences).

> For now, I can work on the Tutorial and complete the functions. I will try and see how the results would like on the data used for Fig 12 and 14 only if I can find the correct data! Although the results might be different compared to the one presented in the paper, it would be nice to add that to the notebook. I will probably complete it in a couple of days.

Sounds good. The sine wave examples are good for validation but, personally, I find them to be less informative and would likely not want them to be in the tutorial if other data is available. Let me know if you aren't able to locate the data. I may be able to help!

Ninimama:
@seanlaw 
Could you please guide me on how to locate the data? I downloaded the UCR archives, but I think it is not cleat what data I should use. 

For instance, the gun-point data and insect EPG data (like any other data sets) has two train/test parts for the purpose of classification. However, there is one single time series data in the paper.  If you already contact the authors and have the data, please share it with me if you can. If not, I can contact the authors myself.

seanlaw:
Okay, so it looks like Fig. 14 and Fig. 21 are the same data. I will reach out to Eammon and see if he can provide that dataset. 

In the meantime, can you use Fig.1/Fig. 2 (it is the same data as Fig. 17) and compute the indices and geometric chains?

Ninimama:
> I got the results pointed out in Section 5.2 of the paper using the same settings. For instance, I considered the user-defined THETA=40 degree as chosen by the authors (please see below for the claims provided by the authors of the paper):

Apparently, at that time, I myself forgot to do z-normalization in calculating the Direction Angle. So, when I added that part later I forgot to analyze the damped_sine data again. I recently calculated the geometric chains for the damped_sine data and realized that geometric chain discovers the chain with max length=4 (as opposed to 7 claimed in the paper). You can see it in the new version.

==================================

Regarding Fig.1 and 2, I got the results but haven't added it to the notebook yet. I will add and push another commit soon.

==================================



> Okay, so it looks like Fig. 14 and Fig. 21 are the same data.
I think you meant Fig. 12 and 21. Right?

seanlaw:
> Apparently, at that time, I myself forgot to do z-normalization in calculating the Direction Angle. So, when I added that part later I forgot to analyze the damped_sine data again. I recently calculated the geometric chains for the damped_sine data and realized that geometric chain discovers the chain with max length=4 (as opposed to 7 claimed in the paper). You can see it in the new version of the notebook.

Okay, I will take a look at the notebook. Does it matter if we increase `theta`? Does the max length increase?

> Fig. 12 and 21 (so, we want the data for Fig. 12 and Fig. 14.).  I think Fig. 21 is not useful as it shows no indices.

Yes, sorry, I meant Fig. 12 and Fig.21 are using the same data and not Fig 14 (I've edited my response above). Perhaps, maybe we don't need the indices? We can get the indices using your method, plot their locations, and compare them visually with Fig. 21? 

I guess the main question right now is whether the chains mean anything using our "new" approach (z-normalizing subsequences)? If, for some reason, the resulting chains aren't meaningful then we should really think about "why" and if the original published method is possibly "correct"?

Ninimama:
> Okay, I will take a look at the notebook. Does it matter if we increase `theta`? Does the max length increase?

Yes. So, for instance, if you put `theta=180`, it means there is no restriction, and it will find ALL backward chains.

For that damped_sine time series case, I found a chain with length 4 (instead of 7) using the default 40 degree for `theta`. I showed it in the notebook.

Now, if I change the `theta` from 40 to 50, I will get the chain with length 7 achieved by the authors with their implementation where they normalize the difference and consider `theta=40`. I didn't show it in the notebook.




> and compare them visually with Fig. 21?

Didn't think about that. Sure.



> I guess the main question right now is whether the chains mean anything using our "new" approach (z-normalizing subsequences)

 I checked the MATLAB code for  the __sig_score (Definition 10)__, `TSC_subsequence_distance.m`. (please see image below)

image

Its output was used in `TSC_BChain_calculate_LDist.m` (which calculates the numerator of sig_score in Definition 10) and it was also used in `TSC_BChain_calculate_AmaxDist.m` (which calculates the denominator of sig_score in Definition 10).

Now, if you take a look at the paper, and see both Definition 8 and 10, you will notice that they both involve the calculation of the norm-2 of the difference of two subsequences. So, it doesn't make sense to normalize the difference for one but not the other. 



> If, for some reason, the resulting chains aren't meaningful

Well, this is interesting. Having assumed that my implementation and the authors' are the same __FOR sig_score function__ (as we both normalize the subsequences before calculating the difference), I found something that might help us make the right decision:

The authors claim that their method found a longer chain (with 7 subsequences) for damped_sine and that's better. However, they didn't provide the sig_score for it. On the other hand, the same paper stated that it is better to sort the discovered chains according to their sig_score and select the ones with top-k high scores. (top-5 or top-10). 

__With their implementation:__ there are many 4-subsequence chains and there exist some 4-subsequence chains with sig_score that is higher than the sig_score of 7-subsuence chain! But it has not been  mentioned in the paper. 

Then, I tried __my implementation__, and here is a summary of the result (all the following result are for the longest possible chain discovered by the alogirthm):
* `atsc` finds (bi-directional) 4-subsequence chain with sig_score=1.17
* with geometric approach and `theta=40`, ONLY one longest chain exist, a four-subsequence chain with sig_score=1.11
* with geometric approach and `theta=50`, There are three longest chains (length seven) and highest sig_score among them: 1.19

__NOTE:__
with geometric approach and `theta=50`, There are many 4-subsequence chains and  there are some with sig_score>1.2 


Therefore, although my implementation with `theta=50` get the longer chain, but its sig_score is lower than the ones with length four (And this is similar to the case of authors implementation with `theta=40`  which was not provided in their paper as I mentioned earlier above)



So, I think this brings us to this point: Should we care about longer chain or higher significance score (or any other new metric)?

Ninimama:
@seanlaw:
After I wrote the results for the chains in the respiration data, I got an opinion:
I think the geometric chain algorithm (TSC20) is not necessarily better than the regular chain algorithm (TSC17). It can, however, be considered as an alternative approach in analyzing chains and get some "new" cases that might mean something. No matter what approach one might take to get the chains (TSC17 or TSC20), a metric is probably needed to measure the quality of chains. 

I can take a look at the chain paper and see what the results would be if the geometric approach will be applied to those data set.

seanlaw:
@Ninimama I've emailed Eammon regarding the data and I've also asked him about the z-normalization regarding the direction angle and so let's hope to hear back. In the meantime, I'll try to find some time to go over your notebook but I think now might be a good time to pause and revisit this once we gain some clarity. 

So, while `atsc` may not be the best, the `atsc` approach is reproducible and explainable. However, so far, geometric chains seems inconsistent and still requires the user to guess at some parameters. From a software development standpoint, since there will be many, many users who will try to compute chains with STUMPY, we really need something that stands on solid ground as the users will have a lot of questions when things "don't work" (or work differently than they had expected). Unfortunately, it is very hard (nearly impossible) to support software that we don't fully understand and it will more than likely decrease/harm the credibility of the package. To be absolutely clear, this is not a reflection on you and the hard work that you've put into it thus far! In fact, I am really grateful that you have been meticulous and thorough in your investigations as this is the standard that our users expect. So, thank you! Regardless of the outcome of this PR, I hope that you've come away having felt welcomed to our community and that you've learned as much from me as I have learned from you. We'd love for you to continue contributing to our project!

Ninimama:
I totally agree with you. In fact, according to what we've seen so far, there is no valid evidence that shows TSC17 outperforms (?!) the TSC20 algorithm. It is really reasonable to stop rather than FORCING something to be happened if that thing may not 
be correct or consistence in providing reasonable outcomes. 

It has been a WONDERFUL  experience for me in terms of getting involved in the process. Thanks for your support and constructive feedbacks.

For the time being, I can through the issues and see what topic I can work on.

seanlaw:
@Ninimama I received the following response from the authors:

```
(1) Psyllid data is here.

https://www.cs.ucr.edu/~eamonn/public/Lab_C_060314.mat

  Mat = load(Lab_C_060314.mat);
  Data = Mat.volts;
  Data = Data(from:downsample:to);

 I probably use the following values in Fig. 12.
  from = 6770000; to = from + 60000; downsample = 3; subsequence_length = 200;

 Sorry I missed the parameters for Fig. 21.

(2) Code

As far as I saw that part, the following seems to be correct,
Now I don't have enough time to check. I will try this revision around the late June.

---------------------------------------
if Znorm == 1
 sub_1 = TSC_znorm_trans(sub_1);
 sub_i1 = TSC_znorm_trans(sub_i1);
 sub_i2 = TSC_znorm_trans(sub_i2);
end

diff_1 = sub_i1 - sub_1; 
diff_2 = sub_i2 - sub_1;
CosAdj_i = TSC_cosine(diff_1, diff_2);
---------------------------------------
```

There is a bit of a language barrier here but I believe that they are trying to say is that they should have z-normalized at the subsequence level and not at the subsequence difference level.

Ninimama:
@seanlaw 
Thanks for the update. You are right.  According to what they said in their code, apparently they should have done the z-normalization first at subsequent level.

seanlaw:
@Ninimama I received a response back from the authors:

> I have revised and uploaded the code related to z-normalization. https://sites.google.com/site/timeserieschains/
> The performance has been improved slightly. A little larger direction angle seems to be better, so I select 45 instead of 40 used in the previous version.
> With regard to Fig 12, I could not reproduce it due to this revision or the difference in data. However, Fig. 12 is an example for showing a week point of Geometrical TSC, therefore it is not an issue. With regard to Fig. 21, I got a similar result as a top-3 chain by a new command 'Demo_TSC_Psyllid'. This result is included in the revised release version.

I haven't looked at it yet

Ninimama:
Thanks for the follow up. It has been a while. I can go through it if I get some time in the nearby future. I still think it might slightly improve the result and one should also be careful of the theta parameter as its value can change the result. 

I will go through it and will update you after resolving the git conflict issue in the snippet project. I would like to finish that first. It is hard for me to go back and forth on different branches.

Ninimama:
@seanlaw 
What do you think? Should we check it out again to see how the results sounds now?

seanlaw:
> I will go through it and will update you after resolving the git conflict issue in the snippet project. I would like to finish that first. It is hard for me to go back and forth on different branches.

I would say that the priority order should be:

1. Snippets Tutorial
2. Understanding MERLIN
3. Geometric Chains

The bottom two are longer term goals in my mind. Does that help?

Ninimama:
@seanlaw 

> I would say that the priority order should be:
> 
> 1. Snippets Tutorial
> 2. Understanding MERLIN
> 3. Geometric Chains
> 
> The bottom two are longer term goals in my mind. Does that help?

Definitely. That makes sense as we are about to change snippets in the newer version as you mentioned before.

seanlaw:
@Ninimama I'm just doing some house cleaning here. Even though geometric chains may not be working, I am thinking about merging this notebook and then moving it into a "WIP" sub-directory. This way, we can at least capture some of this work for future reference. What do you think?

Ninimama:
@seanlaw 
I think that would be a good idea. 

By "merging", I assume users can go through the notebook. Right?  If yes, then it might be a good idea to add a short note in the beginning of the notebook that briefly explains the difference between the paper's MATLAB implementation and the current Python implementation. Also, add a note that says we didn't observe any (significant) improvement after correcting the implementation.  We should let users know that they should better use the `stumpy.atsc` first before digging into the notebook.

Feel free to close this PR and do any changes that is appropriate.

PR #407, opened 2021-06-14
Fix issue 358

mexxexx:
I updated the fast pattern matching tutorial and welcome any suggestions.

A question from my side: I wrote this some while ago, when there was still an `excl_zone` parameter. I wanted to demonstrate that if setting `excl_zone` to zero in `match`, you could get the same behavior as when doing the pattern search by hand as in the first part of the tutorial. Now this is no longer possible. Do you have any suggestions on what to do? It feels natural to be able to disable the exclusion zone if necessary, but this would require an `ignore_trivial` parameter or so that is not there.


mexxexx:
Also, please ignore the ton of commits. They are just some residue. I guess you will squash the history in the end.

seanlaw:
> A question from my side: I wrote this some while ago, when there was still an excl_zone parameter. I wanted to demonstrate that if setting excl_zone to zero in match, you could get the same behavior as when doing the pattern search by hand as in the first part of the tutorial. Now this is no longer possible. Do you have any suggestions on what to do? It feels natural to be able to disable the exclusion zone if necessary, but this would require an ignore_trivial parameter or so that is not there.

I'm not entirely sure but let me take a look and see if I can understand what you mean.

seanlaw:
@mexxexx I went through the tutorial really, really quickly and I love what you've done and how you've laid everything out! Great work!  Regarding the `excl_zone`, if I understand correctly, you can still set the exclusion zone to zero by doing:

```
from stumpy.config import STUMPY_EXCL_ZONE_DENOM
STUMPY_EXCL_ZONE_DENOM = np.inf
```

Then, internally inside of `match()`,  `excl_zone = int(np.ceil(m / STUMPY_EXCL_ZONE_DENOM))` should resolve to zero. So, the exclusion zone is now set globally rather than locally. This ensures that the user doesn't "forget" to set it across multiple functions. Does that make sense? Hopefully, I'm not misunderstanding your meaning.

mexxexx:
Sure, take your time and then we can see if anything needs adjustment.

> Regarding the excl_zone, if I understand correctly, you can still set the exclusion zone to zero by doing:

That is what I needed :) I did not think about setting the denominator to `np.inf`! It would still make sense to reset the denominator afterwards, right? Otherwise it might screw up the following computations. 

What about adding a small context manager to `config.py` looking something like this:

``` 
@contextlib.contextmanager
def set_environment_variable(name, value):
    thismodule = sys.modules[__name__]

    environment_vars = [v for v in dir(thismodule) if v.startswith("STUMPY")]
    if name not in environment_vars:
        raise AttributeError("No STUMPY environment variable with the name {} exists.".format(name))

    value_old = getattr(thismodule, name)
    setattr(thismodule, name, value)
    yield
    setattr(thismodule, name, value_old)
```

Then one could to the following:
```
with stumpy.config.set_environment_variable("STUMPY_EXCL_ZONE_DENOM", np.inf):
    dosomething()
```
And then it would automatically reset to the original value:
```
with stumpy.config.set_environment_variable("STUMPY_EXCL_ZONE_DENOM", np.inf):
    print(stumpy.config.STUMPY_EXCL_ZONE_DENOM) # prints np.inf

print(stumpy.config.STUMPY_EXCL_ZONE_DENOM) # prints 4
```

However, I realize that this might be a bit over the top and only for a very specific use case.

seanlaw:
> What about adding a small context manager to config.py

Initially, I thought that using a context manager was an interesting idea. However, as I thought through it:

1. It does feel somewhat over-the-top and it feels like a lot of unnecessary coding 
2. I don't love that the reset behavior feels implicit (i.e., it isn't clear to first time users that variables will reset or turn back to the previously set value and not necessarily to the ORIGINAL value?) 

My vote is to not use a context manager.

seanlaw:
Also, if you wouldn't mind handling this, I've added a Matplotlib style file so the lines:

```
plt.rcParams["figure.figsize"] = [20, 6]  # width, height
plt.rcParams['xtick.direction'] = 'out'
```

can now be replace with:

```
plt.style.use('stumpy.mplstyle')
```

seanlaw:
@mexxexx I really enjoyed your addition! Great work! I provided some minor edits for you to consider

seanlaw:
We are adding a Binder link to the top of each tutorial. Would you mind adding the following below the main `H1` header:

```
[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/TDAmeritrade/stumpy/main?filepath=notebooks/Tutorial_Pattern_Searching.ipynb)
```

Right now it links to `Tutorial_Pattern_Searching` but you may change that accordingly

mexxexx:
@seanlaw I finally got around implementing your feedback into the tutorial. What do you think?

seanlaw:
@mexxexx I think everything looks great! I find some minor things but I can go over it after this PR gets merged. Thank you for this contribution!

Meng6:
> Also, if you wouldn't mind handling this, I've added a Matplotlib style file so the lines:
> 
> ```
> plt.rcParams["figure.figsize"] = [20, 6]  # width, height
> plt.rcParams['xtick.direction'] = 'out'
> ```
> 
> can now be replace with:
> 
> ```
> plt.style.use('stumpy.mplstyle')
> ```

Hi @seanlaw ,

I installed the latest version of `stumpy` package (v1.9.2) via Conda, but I got an error while running the following line of code:
```
plt.style.use('stumpy.mplstyle')
```
Here is the error message:
```
File "/Users/xxx/opt/anaconda3/lib/python3.7/site-packages/matplotlib/style/core.py", line 127, in use
    "available styles".format(style)) from err
OSError: 'stumpy.mplstyle' not found in the style library and input is not a valid URL or path; see `style.available` for list of available styles
```
I think that's because the stumpy's matplotlib file cannot be found. Do you know how to solve this problem? 

Thanks so much for your help!
Meng

seanlaw:
@Meng6 I think what you are experiencing is a separate issue. Would you mind posting your question in our Github Discussions:

https://github.com/TDAmeritrade/stumpy/discussions

And please describe what it is that you are trying to do? The style file is not needed for STUMPY (it is only there to keep our plots looking somewhat consistent) and you can safely comment/ignore that line out if you are trying to copy the steps in the tutorial.

seanlaw:
@Meng6 Alternatively, you can replace:

```
plt.style.use('stumpy.mplstyle')
```

with:

```
plt.style.use('https://raw.githubusercontent.com/TDAmeritrade/stumpy/main/docs/stumpy.mplstyle')
```

And it should work. I've updated all of the tutorials with that now

Meng6:
Hi @seanlaw, it works! Thanks so much for your help & quick response!

PR #418, opened 2021-06-20
[WIP] Add Tutorial for Matrix Profile XXI: MERLIN algorithm #Issue 417

Ninimama:
# Pull Request Checklist

- [x] Create a new branch


seanlaw:
@Ninimama Will you let me know if/when you think it would make sense for me to take a look?

Ninimama:
@seanlaw 
The commit itself is a good size for review. However, I need to check the notebook out again and plot the discords. I will let you know in a day or two after I push the next commit. Discords are found by DRAG algorithm (Table 1 and 2 of the paper)

Do you want me to implement the MERLIN algorithm (Table 3 of the paper) and then let you know? Because, for now,  it is just Table 1 and 2 (DRAG algorithm) that is used in MERLIN algorithm.

seanlaw:
> Do you want me to implement the MERLIN algorithm (Table 3 of the paper) and then let you know? Because, for now, it is just Table 1 and 2 (DRAG algorithm) that is used in MERLIN algorithm.

Yes, it would be great if you could implement the MERLIN algorithm. In the meantime, I'll try to find some time to review the DRAG parts and I may have some questions for you.

seanlaw:
In case I forget to mention it, these initial versions of the code are excellent candidates for the unit tests. If you look at `tests/naive.py` you'll see basic/slow but accurate implementation of all of the algorithms and then we compare the results from our (future) fast implementation with the naive implementation.

docs/Tutorial_MERLIN.ipynb
```diff
@@ -0,0 +1,407 @@
+{
```
seanlaw, line 1: <p>Here are a couple of suggestions:</p><p> </p><ol><li>Use <code>T</code> instead of <code>T_A</code></li><li>The z-norm for <code>T_i</code> can be moved outside of the <code>for j</code> loop</li><li>Instead of returning <code>failure</code> , I think it is fine to simply <code>return C</code>  with length <code>0</code>  <span class="ql-cursor">﻿</span></li></ol><p> </p><p>It also looks like we may be able to parallelize the inner <code>for j</code> loop</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <p>Is there a reason why we are storing the subsequence length <code>m</code> inside of <code>D</code> ? It looks like it doesn't change and would be a constant. In fact, I would considering using two separate 1D arrays (<code>idx</code> and <code>distance</code>), do the computation, and combine the results into <code>D</code> at the end before returning. This would make the code a lot easier to read as the variable names are obvious. Eventually, I can help rename things to be more consistent with the variable names in the code base (i.e., we typically use <code>D</code> for a distance profile and so we'd likely just use <code>discords</code> to represent the set of discords, etc).</p><p> </p><p>Why are we converting <code>C</code> into a <code>list</code> ? Is this necessary?</p><p> </p><p>Instead of <code>D[j_ind,-1]&nbsp;</code> please be explicit and use <code>D[j_ind, 3]</code> .</p><p> </p><p>I noticed that <code>is_discord</code> isn't actually being used anywhere. You are setting it to <code>True</code>/<code>False</code> and that's it.</p><p> </p><p>The use of <code>C_indices_tot = np.arange(len(C_lst))</code> and <code>C_indices = C_indices_tot[:]</code> seem complicated to me. I wonder if there is a way to simplify all of this referencing. The Table 2 in the paper does not seem to look as complicated</p><p> </p><p>Why return a set of indices to remove? Why not just remove them before and only return the final set of <code>discords</code>  ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

Ninimama: > It also looks like we may be able to parallelize the inner `for j` loop

I need to read a few articles to see how this can be done.

Ninimama: > Is there a reason why we are storing the subsequence length `m` inside of `D` ?

I didn't understand this part. By subsequence, you mean the its (starting) index stored in the first column? If so, then the index itself is not going to change as I will keep them all and remove the unwanted ones later. However, its NN and its distance will be changed. So, if you are talking about `[j_ind,-1] = min(D[j_ind,-1], dist)`, then I think it is necessary as we try to update the NN distance of the discord located at index `j_ind`. 



> Why are we converting `C` into a `list` ? Is this necessary?
> 
> Instead of `D[j_ind,-1] ` please be explicit and use `D[j_ind, 3]` .
> 
> I noticed that `is_discord` isn't actually being used anywhere. You are setting it to `True`/`False` and that's it.

Thanks. I  will take care of it. In fact, that was when I tried to understand the difference between the proposed algorithm in the paper and the one provided in MATLAB and what `dj` is. So, I wrote as I tried to figure things out and didn't pay much attention to small details. So, I called it [WIP]. I will refine the code and will consider your comments in the next commit for sure.







> The use of `C_indices_tot = np.arange(len(C_lst))` and `C_indices = C_indices_tot[:]` seem complicated to me. I wonder if there is a way to simplify all of this referencing. The Table 2 in the paper does not seem to look as complicated
> 
> Why return a set of indices to remove? Why not just remove them before and only return the final set of `discords` ?

I totally agree. I need to go through it. I remember I got an error about the change in the size of `set` object  in the for loop. So, I tried to resolve it. I need to go back to that `set` approach.

seanlaw: Don't worry about it for now. Let's focus on getting it to work first before making it fast

seanlaw: <p>I have experience with this and can help assist. We'll be using the Numba package for parallelization</p>

seanlaw: <p>Sounds good and no rush! I just wanted to provide feedback while it was still fresh in my head. I can tell you that I don't do well jumping back-and-forth from PR to PR. The context switching hurts my brain :)</p>

Ninimama: > The Table 2 in the paper does not seem to look as complicated

image

I tried to go through the algorithm one more time. I believe the Table 2 algorithm is not complete. Let's take a look at `dj`: It is initially set to `inf`. And, it is being updated through the outer loop . It is not shown in the TABLE 2; but, that is what authors did in their MATLAB code. They store it and updated it. (please see below)
```
% Algorithm 1: Discord Refinement Phase
cand_nndists2 = containers.Map(C.keys, inf(length(C), 1), 'UniformValues', true);
cand_nn_pos = containers.Map(C.keys, NaN(length(C), 1), 'UniformValues', true);
r2 = r^2;

% I diverge from the pseudocode in the paper here. First, I have to use 
% squared distances for correctness reasons in order to avoid taking a 
% square root at every step where we check against best so far. Second,
% I had to change the way elements are removed from the candidate set, or
% rather I had to change the loop tracking variable. If elements are 
% removed from an array while iterating over that same array, it will 
% result in skipped elements in almost any environment. Instead I'm 
% iterating over a copy of its hash keys, which remains constant over 
% the inner 2 loops. Elements are still progressively removed from the 
% candidate set.

for i = 1 : subseqcount
    if isempty(C)
        break;
    end
    keys = C.keys;
    Si = (TS(i : i + L - 1) - mu(i)) ./ sig(i);
    for j = 1 : length(keys)
        if abs(i - keys{j}) < min_separation
            continue;
        end
        dist2 = 0;
        candidx = keys{j};
        Candj = C(candidx);
        nndist2 = cand_nndists2(candidx); 
        for k = 1 : L
            dist2 = dist2 + (Si(k) - Candj(k))^2;
            if dist2 >= nndist2
                break;
            end
        end
        if dist2 < r2
            C.remove(candidx);
            cand_nndists2.remove(candidx);
            cand_nn_pos.remove(candidx);
            continue;
        elseif  dist2 < cand_nndists2(candidx)  
            cand_nndists2(candidx) = dist2;
            cand_nn_pos(candidx) = i;
        end
    end    
end
```

I can use dictionary to store `dj` (and also delete the ones (i.e. keys) that should  be removed throughout the for loop). I used numpy array instead for now. If you thin dictionary or any other structure is better to store/update the discords, pleases let me know.

========================================================================

Now, there is one thing left that should be noticed for now! I took a look at the documents on their webpage and realized: 

image

There are a few differences:

(1) It says the NN is at index 470. But I didn't get that in the result of my implementation

(2)  [PROBABLY NOT IMPORTANT] The length of time series in the image above is 3000, similar to what provided in the paper. However, the length of the time series provided as a data set on the webpage of the paper is 2000. 

(3) It says the actual discord value is 10.27 (above which the algorithm returns no discord). However, my implementation shows that the actual discord value is 10.43.

===================================================================

I already revised the first two functions according to your suggestion. I will implement MERLIN and then push the commit.

seanlaw: What value(s) do you get when you run the Matlab code?

Ninimama: > What value(s) do you get when you run the Matlab code?

Will let you know as I haven't tried it yet.

seanlaw: @udf2457 I was wondering if you had any thoughts on this?

Ninimama: @seanlaw @udf2457 @alvii147 
So, I checked out the code again. As @alvii147 mentioned before, the Introduction Link contains a .zip file in which we can find the latest version of DRAG implemented by the author: `discord_discovery_gemm3.m`

I run the MATLAB code and compare it with mine using the same input:

The discord location and its distance to NN are the same. HOWEVER, the index of the NN subsequence are not the same.

seanlaw: > The discord location and its distance to NN are the same. HOWEVER, the index of the NN subsequence are not the same.

Does that imply that there are two subsequences that have the same distance to the discord? Or do you think that they reported the wrong nearest neighbor index? Can you compute the "other" nearest neighbor distance by hand and compare?

Ninimama: > > The discord location and its distance to NN are the same. HOWEVER, the index of the NN subsequence are not the same.
> 
> Does that imply that there are two subsequences that have the same distance to the discord? Or do you think that they reported the wrong nearest neighbor index? Can you compute the "other" nearest neighbor distance by hand and compare?

I just deleted my previous code. I am going to test the output of my implementation as I think it doesn't return the corrrect NN distance. (I checked the distance between subseq and its NN (provided in my output) and their distance is 0! I am now trying to debug it.

seanlaw: Also, don't forget that you can find the top discord using the more expensive `stump` (i.e., compute the matrix profile and take the max). So you can at least use it to verify your results

Ninimama: UPDATE:

EVERYTHING looks good! Thanks for your support. Now, I am going to do the MERLIN main algorithm,

udf2457: @seanlaw Just to say sorry for the delay in reply, yesterday was a bit nuts at $work.  @Ninimama good to hear you found a solution.

Ninimama:
@seanlaw  @udf2457 

I implemented the MERLIN algorithm using Table 3. And, I ALMOST successfully reproduce one of the figures of the paper (Fig. 11). 

I also compared it to the results of MATLAB code (MERLIN3.m) and the results are ALMOST the same.

**I used the term ALMOST because there are only two cases (out of 92 cases) where the discord indices were different**

========================================

**Some notes:**

* The python implementation is naive and needs to be improved to enhance the efficiency in time and memory.
* It seems the algorithm provided in the MATLAB code (`MERLIN3.m`) is different from the one provided in the Table3 of the paper. I think the main difference is "updating the parameter `r` " . Please take a look at how the author updated the parameter r in their MATLAB code, and compared it with Table 3.
* Although different data sets are provided on the webpage of the paper, it is not really clear which samples of those data sets are used in the paper. For instance, CVP data (Section IV-B, Fig. 9) is provided as .mat file that contains 9 different sets of time series. 
* After checking out four data sets, It seems TAXI cab is a good option for reproducing the result. TAXI cab has 30min resolution. According to Fig. 11, it seems we should only consider the time series values at timestamps 00:00, 01:00, 02:00, ... (and skip the values recorded at --:30)

seanlaw:
@Ninimama  When you say 92 cases is it only for the taxi cab dataset? Or is this for other datasets too? I think it is worthwhile applying MERLIN to a few different datasets in order to validate the approach.

Ninimama:
> @Ninimama When you say 92 cases is it only for the taxi cab dataset? Or is this for other datasets too? I think it is worthwhile applying MERLIN to a few different datasets in order to validate the approach. If it is for other datasets, can you show those in the tutorial so that I can take a look? We probably won't include all of them but I'd like to see them for my own reference.

That is only for TAXI cab. I will plot those two discords that are different in Python compared to MATLAB.

=================================================================

I will try to see if I can apply the MERLIN algorithm on any other data. As I said, the problem is that each data set provided on their webpage contains some sets of time series data, and it is hard to know which one is used in the paper. I will go through the paper again and the data sets to see if I can find at least two more data sets

udf2457:
@Ninimama  `MERLIN3.m` seems to be (mostly) an attempt at optimising the algorithm with a view to introducing additional features under Merlin 3.0 in the future (see hints in the ppt file in the same zip)

Ninimama:
@seanlaw @udf2457 

So, I applied MERLIN on Mars Science Laboratory data and visually reproduce Fig. 17 (top) and Fig. 17 (bottom). These two figures corresponds to two different sets of time series. I also got the results of MATLAB (marlin3.m) for Fig.17 (top) and I realized that out of 100 discords (with length m=`100` to `m=200`), there are only three discords that have different index. (however, their starting position are close to each other.)

======================================================================

**Some additional notes:**

* As @udf2457 mentioned, the merlin3.m is supposed to be a more efficient algorithm. (And, it is faster to some extent! I will explain it below in the next paragraph) However, I think DRAG algorithm plays a more important role as the implantation of this algorithm is changed to speed up the process of discovering the discords candidates. If you take a look at the comments in `discord_discovery_gemm3.m` file, you can see the authors do some eliminations to speed up the discovery process, and they avoid comparing a subsequence will ALL other subsequences. So, for instance, my implementation (which is based on Table 1 and 2 and 3) takes more than 1 hour to provide the results for the Fig 17 (bottom).

* There is also another reason that their implementation is faster, and that is how they update `r`. In the paper, they reduce r by 1% if they don't find a discord for a subsequence of length `m`. And, as soon as they find a discord (and that is the top discord with length m), the go to discover discord of length `m+1` and reset r to the dist of previous discord to its NN subsequence.
HOWEVER, in the paper, they reduce `r` by a factor of `(m-1)/m` which is smaller than 0.99 for m<100. Therefore, r will decrease faster and, subsequently, a discord with dist>r will be discovered sooner. However, if m goes beyond 100, then 0.99 reduces the `r` faster. However, since probably the major time of the algorithm is usually spent on lower m, their approach in marlin3.m is more efficient than the one proposed in Table3.

=====================================================================

Final word (my suggestion):
I think we should understand their approach in MATLAB code and implement that in Python, as it is very faster than the ones they proposed in their paper.

udf2457:
Thank you @Ninimama for all your ongoing work and your nice succinct summary of the enhanced algorithm. 

I would agree with your general sentiment , especially given the headline description of MERLIN:

> MERLIN, an algorithm that can efficiently and exactly find discords of all lengths in massive time series archives. 

The downside of course, as @Ninimama also alludes to is the increase in implementation complexity.

seanlaw:
Thank you @Ninimama! 

> So, I applied MERLIN on Mars Science Laboratory data and visually reproduced Fig. 17 (top) and Fig. 17 (bottom). These two figures corresponds to two different sets of time series. I also got the results of MATLAB (marlin3.m) for Fig.17 (top) and I realized that out of 100 discords (with length m=100 to m=200), there are only three discords that have different index. (however, their starting position are close to each other.)

Can you explain why the three discords have different indices? Is it the difference between MERLIN and MERLIN3? 

> I think we should understand their approach in MATLAB code and implement that in Python, as it is faster compared to the one proposed in their paper.

I agree with both you and @udf2457. If MERLIN3 is essentially producing the same results but it is drastically faster than the pseudocode in the paper then we should try to implement MERLIN3. @Ninimama do you see any negative reasons not to use MERLIN3?

Ninimama:
@seanlaw 
**I hope the following explanation finds you well.**  
> Can you explain why the three discords have different indices? Is it the difference between MERLIN and MERLIN3?

I believe the reason is related to the implementation of MERLIN3. This is what I understand:

Let's say `disc_m` is the discord of length `m` that is different in MERLIN and MERLIN3. To see how much the two different discords, found by MERLIN and MERLIN3, are different, we can compare their distance to their NN subsequence. The ideal scenario is to see if the two different discords have the same distance to their NN subsequence. Then, we can say both discords are equally good. And therefore, MERLIN and MERLIN3 have the same quality.

=================================================================

**I compared the results for both NYC TAXI data (Fig. 11) and also MSL-A-4 (Fig. 17-top).**  
Comparing the results shows that the distance of `disc_m` to its NN subsequence discovered by MERLIN is bigger compared to the one discovered by MERLIN3. This shows that MERLIN discovered better discords in those few cases for both data sets. 

MSL-A-4 (Fig. 17-top), the difference is negligible and in TAXI case study the difference is small.

===========================================================================

**MERLIN vs MERLIN3:**  
MERLIN uses three different ways to update 'r' (please see line 5, 12, 16-18 of Table 3). However, MERLIN3 only uses 'r=r * (m-1)/m'. I think the value of 'r' (as threshold of distance to NN subsequence) causes such difference. Also, one might be interested to investigate the impact of MERLIN3-DRAG compared to the DRAG provided in Table 1&2 of the paper.



==========================================================================
==========================================================================

> @Ninimama Aside from the added code complexity, do you see any negative reasons not to use MERLIN3?

All in all, I believe MERLIN3 is a better choice as it provides almost the same discords. Some of them might be false positive which is totally okay as the authors mentioned the same challenge for MERLIN in their paper. 



**The question that I have in mind and might be a good idea to confirm it by asking the authors is this:**  
Is there any particular reason why we should update 'r' more slowly as we increase the length of subsequence, `m`?

udf2457:
@Ninimama in relation to the false-positives I agree.  Infact if you look at the penultimate slide of the PPT in the MERLIN3 zip, the authors mention three scenarios that may lead to false-positives (constant regions, twin freaks and large magnitude changes).

In relation to your suggestion of a question to ask the authors, would you like me to try to reach out to them again or will/are you do/doing that already ?  If me, then are there any other questions at the back of your mind we could present at the same time ?  I will also see if I can entice one of them to review this PR and comment.... ;-)

Ninimama:
@udf2457 

Thanks for the input. You are right. They discussed some of the scenarios that can lead to false positive.

 > In relation to your suggestion of a question to ask the authors, would you like me to try to reach out to them again or will/are you do/doing that already ? If me, then are there any other questions at the back of your mind we could present at the same time ? 

I haven't asked them yet. 
@seanlaw: Do you have any questions in mind?


> I will could also see if I can entice one of them to review this PR and comment.... ;-)
This is MERLIN implementation. Do you think they are interested in reviewing MERLIN as they are currently working on MERLIN3?

udf2457:
> This is MERLIN implementation. Do you think they are interested in reviewing MERLIN as they are currently working on MERLIN3?

Fair point, and given one of them did make a passing comment to me I think you're correct in your thinking:

> In the meantime, you may be interested in our expanded "introduction to MERLIN"

(i.e. "introduction to MERLIN" being the name of the zip with the MERLIN3 code in it).

Ninimama:
@seanlaw @udf2457 

Just a quick note on my recent comment:

I thought again and realized updating r doesn't affect the discord and can only impact the computation time. 

Those few cases where the discovered discord index was different is because of the enhanced version of DRAG algorithm used by the authors. they try to eliminate some of the subsequences to speed up the computation. And that results in those few errors but it is worth it as it can reduce the computation time.

udf2457:
@Ninimama 

Just landed in my Inbox in reply to your earlier question:

> I believe the single update of  'r=r * (m-1)/m' our first implementation. We later found that 'r = M - (2xS)' performed better in our tests, but it also required the two initializing steps with different update equations. 
> 
> Now to answer why we thought  'r=r * (m-1)/m' seemed reasonable. As we consider larger subsequence lengths, the change in maximum possible euclidean distance changes from sqrt(4*L)  to sqrt(4*L+1). The percent difference in these sequential values decreases with longer lengths.

Ninimama:
Now that I finished the `Snippet_Tutorial`, I will try to see if I can the current python implementation of MERLIN as provided in the paper. Then, I can implement the version the authors provided on the paper's support webpage. I will resume this work around this weekend.

seanlaw:
> Now that I finished the Snippet_Tutorial, I will try to see if I can improve the current python implementation of MERLIN as provided in the paper. Then, I can implement the version that the authors provided on their paper's support webpage.

> I will resume this PR around this weekend.

Sounds great!

udf2457:
> Now that I finished the `Snippet_Tutorial`, I will try to see if I can improve the current python implementation of MERLIN as provided in the paper. Then, I can implement the version that the authors provided on their paper's support webpage.
> 
> I will resume this PR around this weekend.

Super duper trooper @Ninimama  :)

Ninimama:
In the following, fastMERLIN refers to MERLIN3, the updated algorithm of MERLIN, provided in the supporting page of the original paper)

**NOTE:** the author of MERLIN3 is a not the author of the original paper


1. As you can observe in the notebook, I converted the fastMERLIN MATLAB code to Python. The new algorithm, which has three main phases, is a little bit hard to read due to the a lot of IF/For statements considered by the author for considering different scenarios.  Furthermore, it was hard to break it down to smaller parts. One approach might be to separate each phase and write it in a separate function. However,  those functions may not be used in other cases as these three phases are designed for fastMERLIN. The current implementation is not clean.

2. Current Python / MATLAB Implementation of fastMERLIN and MERLIN (original paper version) give the same result. 

4. I also tried MP (with Stumpy) to discover discords. I realized Stumpy is faster! (you can see the computation time at the end of the notebook) I compared fastMERLIN with MP regarding the discovered discords (their indices and also their dist to their NN). There are a few discords that don't have the same index in both fastMERLIN and MP. However, their difference is negligible. It is noteworthy to mention the following reasons:
* exclusion zone: fastMERLIN (by default) consider 100% exclusion. 
* z-normalizing subsequences:  z-normalizing a sequence means mu=0 and std=1, by using z=(x-mu)/std. [Right?]  However, in the MATLAB implementation, the subsequences were divided by their norm.

5. Next step: Try fastMERLIN on a big time series (?) and see if it can beat MP.

udf2457:
You are a star in the github galaxy @Ninimama.  Thank you for all your work on this.

seanlaw:
> the author of MERLIN3 is a not the author of the original paper

Can you please explain what you mean by this? What changed?

Ninimama:
> > the author of MERLIN3 is a not the author of the original paper
> 
> Can you please explain what you mean by this? What changed?


Well, the author is from the Eamon Keogh's team but he is not the author of the paper. And, he mentioned it in the comment of the MATLAB file. I will try to explain the changes in the next few days.



> I think it may still make sense to break things down into logical components. Considering that you've identified that there are three main phases then I think that's a good place to start.

Agreed. I will think about breaking it down to different parts. Then, I might get some help from you on how to create unit test. MERLIN can be used as unit test for the whole algorithm. but to test each phase of fastMERLIN, I think we should write a unit test.


> Great! Is fastMERLIN actually faster? If so, how much faster? Given that they produce the same results, it might make sense to use MERLIN as the `naive.py` implementation for testing.

Since the python implementation of MERLIN might not be efficient, for now I can compare the computation time of MERLIN with fastMERLIN using their MATLAB implementation.

> What does "100% exclusion" mean?

It means if a subsequence of length L starts at index i, it will compare with the subsequence that starts at i+L. (was that correct that I said 100% exclusion? If not, what is the correct term?)

> Is it possible to use z-normalization in (fast)MERLIN and get the same discords as STUMPY?
> 
> This inconsistency in (fast)MERLIN bothers me as it implies that it isn't as closely related to matrix profiles as I would have hoped and, perhaps, may be beyond the scope of STUMPY. This isn't to say that MERLIN isn't valuable but that it may be worth splitting out into a separate package.

I checked the code again. They scaled the vector by their norm AFTER making their mean equal to 0. Therefore, I think they should be the same. I will check it out. 

> You'll probably want to perform your timing comparisons on time series that are at least 1 million (2^20) data points long.

Good idea. I can repeat the time series to make its length bigger. Then, I can apply MP and fastMERLIN and see how they perform.

=================
If you have any further suggestion, please let me know. Thanks.

seanlaw:
> Well, the author is from the Eamon Keogh's team but he is not the author of the paper. And, he mentioned it in the comment of the MATLAB file. I will try to explain the changes in the next few days.

Got it. Thanks

> Agreed. I will think about breaking it down to different parts. Then, I might get some help from you on how to create unit test. MERLIN can be used as unit test for the whole algorithm. but to test each phase of fastMERLIN, I think we should write a unit test.

Let's worry about the unit tests later. In fact, what matters the most in this case is that the discord output (both distance to NN and NN index location) are identical to STUMPY's discords. Based on the last two plots in the tutorial ("comparing indices of discords of length 5-96" and "comparing dist (to NN) of discords of length 5-96"), it is clear that (fast)MERLIN does not reproduce exactly the NN indices and distances. We should fix this in (fast)MERLIN and make sure that the output matches STUMPY's discords.

> It means if a subsequence of length L starts at index i, it will compare with the subsequence that starts at i+L. (was that correct that I said 100% exclusion? If not, what is the correct term?)

Ahhh, I see. I would just say that the exclusion zone is set to `[i-L, i+L+1]` (in NumPy indexing syntax) in the current (fast)MERLIN implementation and STUMPY uses `[i - L/4, i+L/4]`. For consistency, I think it is best to set the same exclusion zone as STUMPY. 

> I checked the code again. They scaled the vector by their norm AFTER making their mean equal to 0. Therefore, I think both scaling approaches should have the same impact. I will check it out.

Based on your final plot of "comparing dist (to NN) of discords of length 5-96", I think it is clear that their distance calculation is different from STUMPY's z-normalization. Otherwise, those distances would be identical.

felipekosta:
Dear Sean and Nima,

I am a doctoral student and intend to use algorithms (Matrix Profile and/or Merlin) in my thesis.

I have been following your work regarding the conversion of matlab code from version 3 of the Merlin algorithm.

I would like to thank you since the information was very helpful to me.

When I accessed the page https://github.com/TDAmeritrade/stumpy/pull/418 you were analyzing the results of the first python version of the Merlin algorithm and talking about the possibility of converting version 3 as well.

Today I accessed this page again and saw that the version 3 conversion work is well under way.

As the expectation was that version 3 would be faster than the previous one, I also started work on converting this version to python.

Last week I managed to run the first version of the code, using the dataset referring to the single bad heartbeat (qtdbSel100MLII) and the results were interesting (only 3 differences - 72, 77 and 190 - for the sequence sizes between 30 and 300).

Today I will run tests with other datasets available at https://sites.google.com/view/merlin-find-anomalies/datasets.

If you can contribute to the process, in something that is within my possibilities, let me know.

seanlaw:
> I am a doctoral student and intend to use algorithms (Matrix Profile and/or Merlin) in my thesis. 
> I have been following your work regarding the conversion of matlab code from version 3 of the Merlin algorithm.
> I would like to thank you since the information was very helpful to me.

@felipekosta Welcome to the STUMPY community! 
 
> If you can contribute to the process, in something that is within my possibilities, let me know.

I welcome you to jump right in. This is an open community and we'll accept all of the help that we can get. Feel free to review and comment on (or even respectfully debate) the code that @Ninimama has put together. Right now, we are trying to figure out if the published (fast)MERLIN work actually functions properly and, if so, is it worth adding to STUMPY. The answer may be "no" and that's okay but we want to make sure we understand why things do/don't work.

What are you comfortable with helping with?

Ninimama:
> Ahhh, I see. I would just say that the exclusion zone is set to `[i-L, i+L+1]`

Just to confirm, is this correct? so, for instance, if `i=100`, and `L (as exclusion) = 5`, I thought we are excluding: [96....104] as I assume the `i` itself is included in the exclusion process... so [i-L+1, i+L-1].

However, after fixing bugs, I realized there are some inconsistencies. So, I assume STUMPY does exclude  `[i-L, i+L+1]`. Right?

seanlaw:
> Just to confirm, is this correct? so, for instance, if i=100, and L (as exclusion) = 5, I thought we are excluding: [96....104] as I assume the i itself is included in the exclusion process... so [i-L+1, i+L-1].

To be clear, I am **not** saying that what you have applied for (fast)MERLIN wrong. Instead, I think that in order to be able to compare (fast)MERLIN's discords with STUMPY's discords, you will likely need to consistent and apply the same exclusion zone. 

> However, after fixing bugs, I realized there are some inconsistencies. So, I assume STUMPY does exclude [i-L, i+L+1]. Right? (and not the way I provided above)

Correct, STUMPY applies `[i-L, i+L+1]` and where `L = int(np.ceil(m / 4))` and `m` is the typical subsequence window size.

> Because, Considering the same example, I think if we consider exclusion based on what you mentioned, we are ignoring index 95 (however, the sequence {95-96-97-98-99}does not have overlap with the one that starts at 100.

Again, I don't know what the protocol for (fast)MERLIN is but let's assume that you are computing a bunch of distances and then you ignore the distances for any index locations within the exclusion zone. Whatever this exclusion zone is, I think it needs to be the same as the exclusion zone in STUMPY. Does that help?

> And one last question: The L is math.ceil(m/4). Right?

Please see above

felipekosta:
> > I am a doctoral student and intend to use algorithms (Matrix Profile and/or Merlin) in my thesis.
> > I have been following your work regarding the conversion of matlab code from version 3 of the Merlin algorithm.
> > I would like to thank you since the information was very helpful to me.
> 
> @felipekosta Welcome to the STUMPY community!
> 
> > If you can contribute to the process, in something that is within my possibilities, let me know.
> 
> I welcome you to jump right in. This is an open community and we'll accept all of the help that we can get. Feel free to review and comment on (or even respectfully debate) the code that @Ninimama has put together. Right now, we are trying to figure out if the published (fast)MERLIN work actually functions properly and, if so, is it worth adding to STUMPY. The answer may be "no" and that's okay but we want to make sure we understand why things do/don't work.
> 
> What are you comfortable with helping with? We are here to support you.

Sean, of course!

I took a quick look at the code written by Nima and it's much more elegant than mine...:-)

Anyway, I ran the comparisons (removing the old/slow version of MERLIN) and generated the attached images.
I called the version I wrote (converted) as MERLIN3_1. Maybe they can help in some way.

image

image

Processing times are for runs on my computer.

>>> computation time with each approach: 
Stumpy (Python):  18.38765001296997
fastMERLIN (MATLAB):  21.569621 
fastMERLIN (Python):  95.08298659324646
MERLIN3_1 (Python):  [42.703125]

Ninimama:
**Good NEWS!**

I fixed the bugs and also use the setting of stumpy for fastMERLIN (same z-normalization and same exclusion zone! Here are some of the very good news I want so share:

* I also changed the default of MERLIN to consider m/4 exclusion (same as STUMPY). I confirmed that the results of MERLIN, fastMERLIN and MP are the same in discord indices and their distances to NN. 
* Regarding NN indices: It is possible that we have more than 1 NN. MERLIN cannot preserve the order of subsequences as they appear in the time series. Therefore, the NN indices are different in MERLIN. HOWEVER, as the author of MERLIN3 (i.e. fastMERLIN) stated in MATLAB file (and confirmed by results provided n Jupyter notebook), fastMERLN can preserve the order. Therefore, the result of fastMERLIN is the same as STUMPY for NN indices. 
* just to let you know: since I changed the setting MERLIN (to have m/4 exclusion zone), it may not reproduce the results of paper (which considers [i-m+1, i+m] as the exclusion zone)
* Since fastMERLIN gives EXACTLY the same result as STUMPY for discords in terms of discord indices, their NN, and their distances, I am going to try them on long time series. 

======================
The results for the following notes are not yet provided in the notebook.

**Additional NOTE (1) ** 

To compare fastMERLN and MERLIN, I used their MATLAB implementation and I realized fastMERLIN is 14 times faster on a time series with length around 2000. (also, it provides NN indices that are the same as STUMPY. Although having the same dist is more important, but generating the same NN indices is a bonus for fastMERLIN!)
 
**Additional NOTE (2) ** 

I already checked the computation time of fastMERLIN and MP on long time series. fastMERLIN is faster (because  it doesn't need to calculate distance profile. It reject the ones that have small distances and then search among candidates. Furthermore, it gets help from  the distance discovered for discord of length L to find discord of length L+1 !) In some occasions, it might take some time to find the discord distance in the first iteration. However, overall, it is a lot faster in finding discords. In fact, in one test case, the STUMPY was still running to find the discord of the length minL while the fastMERLIN already found the discords for length minL to maxL.

Ninimama:
@felipekosta :

Thanks for the feedback! Your input is much appreciated. Just be careful when you want to compare the results. During the debugging process, I realized that the discord indices might be the same while their distance-to-NN and/or NN indices might be different. 

Although my implementation of fastMERLIN is now correct (at least according to the tests I performed), its speed is a lot faster. Could you please check it  out and see how it performs in terms of computation time? I tried to make it a little bit more efficient. Also, please remove the print() statements in the functions as they might slow down the performance. I used the print statement to get update as the algorithm is being run. 

It seems your implementation gives result in a lower computation time. You can see mine and modify yours to make sure the results are correct and then push it to repo!

seanlaw:
> Good NEWS!

> I fixed the bugs and also use the setting of stumpy for fastMERLIN (same z-normalization and same exclusion zone! Here are some of the very good news I want so share:

> I also changed the default of MERLIN to consider m/4 exclusion (same as STUMPY). I confirmed that the results of MERLIN, fastMERLIN and MP are the same in discord indices and their distances to NN.
Regarding NN indices: It is possible that we have more than 1 NN. MERLIN cannot preserve the order of subsequences as they appear in the time series. Therefore, the NN indices are different in MERLIN. HOWEVER, as the author of MERLIN3 (i.e. fastMERLIN) stated in MATLAB file (and confirmed by results provided n Jupyter notebook), fastMERLN can preserve the order. Therefore, the result of fastMERLIN is the same as STUMPY for NN indices.
just to let you know: since I changed the setting MERLIN (to have m/4 exclusion zone), it may not reproduce the results of paper (which considers [i-m+1, i+m] as the exclusion zone)
Since fastMERLIN gives EXACTLY the same result as STUMPY for discords in terms of discord indices, their NN, and their distances, I am going to try them on long time series.

Wow! This is really exciting news @Ninimama! It's great to hear that not only do the NN distances match but that the NN indices match as well. In general, I am less concerned about index as long as the NN distances are all correct and match those of STUMPY. I think it is perfectly okay if the results don't match the published work when we use `m/4` as long as the results match STUMPY. At the end of the day, the goal is to show that fastMERLIN is way faster for finding discords than `stump`. I'll try to take a look at the fastMERLIN code as that seems to be the one that we want to use.

seanlaw:
@Ninimama Can you tell me where I can find the "fastMERLIN" Matlab code? I see the `MERLIN_topK.m` code from the [supporting website](https://sites.google.com/view/merlin-find-anomalies/MERLIN) but I can't tell if this is simply the DRAG-based MERLIN code. Can you please clarify for me?

Ninimama:
@seanlaw 
> Can you tell me where I can find the "fastMERLIN" Matlab code?
Sure. If you click on the Documentation Tab on the website, you can see `Introduction Link: [*.zip]`. If you download the zip file, there is a folder "introducingMERLIN". You can find the m file there. Please let me know if you cannot find it.

Ninimama:
> Wow! This is really exciting news @Ninimama! It's great to hear that not only do the NN distances match but that the NN indices match as well. In general, I am less concerned about index as long as the NN distances are all correct and match those of STUMPY

In one of the case studies (for comparing fastMERLIN and MP in finding discords), I found a difference only in one of NN indices. However, their dist-to-NN are the same. Also, the difference of their dist-to-NN is in the order of 1e-12). I might go through it later to make sure it is not a bug. (just in case)

* I modified the fastMERLIN using the initialization idea provided in the older version MERLIN. It can find `min_dist` faster compared to fasterMERLIN (when min_dist is not initialized by the user) in the first iteration (for discord with length = minL)

* I also compared fastMERLIN with STUMPY for three different data sets with short (regular) length and long (expanded) length. Throughout the process, I also checked the indices and dist-to-NN to make sure they give the same results for each data set (and for its short and long version). The results are comparable when the time series is short (removing verbose print statement might help to make the results close to each other in this case). And, when time series is long, the fastMERLIN is the winner for finding discords.


> So, I would recommend that the next steps be:
> 
> 1. Remove the MERLIN code
> 2. Simplify the fastMERLIN code by breaking it up into smaller functions
> 3. Remove all of the verbose print statements

Thanks for the feedback. I will remove MERLIN and think about breaking the fastMERLIN into smaller parts.

seanlaw:
Btw, based on the PPT, I asked Ryan Mercer (one of the authors) if there was a newer version of MERLIN3 and he pointed me to MERLIN3.1 [here](https://drive.google.com/file/d/14Gu3jJ7ali3DMvoY3b0-KQQQtXpZmAaA/view)

Ninimama:
@seanlaw 
Thanks for letting me know!!! I can check it out for sure. I understand the algorithm of Merlin3. So, if there are just small changes, I can go through it and change the fastMERLIN accordingly.

Ninimama:
I took a quick look at MERLIN3.1. Interestingly, they added the same initialization process provided in MERLIN (as I stated earlier in my previous comment above, I already applied the same change to fastMERLIN). 

They put two m files together and that is why there are many lines. In my opinion, the version 3 was cleaner. It seems they added some `if` statements to consider some pre-processing steps. They also changed a few for loops. I will take a look at them. 

I think it would be better if I can clean my own version and then change them according to version 3.1 as I realized most (and main) parts have not been changed. What do you think @seanlaw ?

seanlaw:
Sounds good @Ninimama

felipekosta:
> @felipekosta :
> 
> Thanks for the feedback! Your input is much appreciated. Just be careful when you want to compare the results. During the debugging process, I realized that the discord indices might be the same while their distance-to-NN and/or NN indices might be different.
> 
@Ninimama 
Thanks for the guidelines. I will pay attention to the indices and also the values.

> My implementation of fastMERLIN is now correct (at least according to the tests I performed) and it is similar to the output of STUMPY Matrix Profile. I tried to make it a little bit more efficient.
> 
> It seems your implementation gives result in a lower computation time (at least compared to my older version of fastMERLIN). Please feel free to see the new version and modify yours to make sure the results are correct (i.e. it should be the same as what STUMPY produces) and then push it to repo if you have more efficient code! :)

I did the tests with the last version you made available (removing the prints) and the times are great. My version, in the case of long datasets, took a long time to finish, besides having problems in the execution of the taxi dataset.
I think I'll abandon my version and use yours... :-) If you agree, of course!

part-1: compare fastMERLIN and MP on short time series
===================================
                     ts_sine       ts_taxi       ts_lab
stumpy        12.587932  0.097942  0.249189
fastMERLIN   1.479352  0.685565  1.495272
MERLIN3_1   11.421875  1.875000  4.625000

part-2: compare fastMERLIN and MP on long time series
===================================
                    ts_sine        ts_taxi         ts_lab
stumpy        90.207195  68.653600  16.097634
fastMERLIN  52.509698  23.510859  25.661848
MERLIN3_1   302.25000  -                80.453125

Two other comments I would like to make, follow below.

About fastMERLIN (MATLAB) author's comment:
"MERLIN may report false positive anomalies in subsequences with near constant values. A sign that the problem exists is a spike near zero when plotting the histogram of movstd. Adding a large linear trend will solve this issue."

Did you ever worry about this issue? I saw that this happens in some datasets and that when applying the linear trend the anomalies show considerable differences.

The other point is that I realized that comparing stumpy's execution with fastMERLIN's, the CPU usage is pushed to the limit by stumpy, which doesn't happen in fastMERLIN, even the latter showing a better performance (in the image below the first half reflects the execution of stumpy and the second that of fastMERLIN).

stumpy_vs_fastMerlin_CPU_usage

seanlaw:
I wouldn't say that STUMPY was "pushed to the limit". Out of the box, STUMPY will always use all of the CPU cores by default and, in addition to discords, it's also computing all of the motifs as well. Where possible, I plan to parallelize MERLIN as well to try and leverage all of the CPU cores (and reduce total computational time).

Ninimama:
@felipekosta 
Thanks for running it on your end. As @seanlaw stated, STUMPY provides more info that is helpful in doing several tasks by getting help from matrix profile.  However, (fast)MERLIN works ONLY for discords (and not doing anything else such as snippet or regime discovery, etc.).

Also: if you set `MERLIN_initializer=True`, it sometimes lower the computation time. (so if you want to use it in your application, this might help you better. You can just try it out and see how it works)

Ninimama:
I am going to break down the code. In the first step of cleaning process, I might do very small changes as I want the code to work. Then, after being implemented and compared to MatrixProfile, we can make it cleaner.

I also provided a lot of comments to better understand the algorithm. They will be removed later.

seanlaw:
> I am going to break down the code. In the first step of cleaning process, I might do very small changes as I want the code to work. Then, after being implemented and compared to matrix profile, we can make it cleaner.
> 
> I tried to consider STUMPY mindset for naming the variables. I also provided a lot of comments to better understand the algorithm. MOST of these comments can be removed later.

Sounds good. Let me know when it is ready for review

Ninimama:
> Let me know when it is ready for review

I think you can take a look at the code now. As a reminder, (fast)MERLIN algorithm tries to find discords of arbitrary length. The minimum length and maximum length of discords are provided by the user.

For now, I just focused on the main part of the algorithm where the goal is to find discord of a specific length. It has three phases. I wrote each of these phases in its own function, and unite them in one single function `MERLIN_helper`. 

There are some overlaps between the three phases. For instance, the `min_dist` is converted to `r` in all three phases, which can be refactored later. Since I wanted to check out if each phase works properly or not, I leave it as is. We can later refactor it by converting the `min_dist` to `r` in the `MERLIN_helper` function.

seanlaw:
@Ninimama Sounds good. It will take me some time to review so I hope that is okay

seanlaw:
@Ninimama Before I look at the other phases, I'd like to fully understand the first phase. I've added some questions/comments but I have more questions still.

I understand that the code is meant to be optimized for speed but it would be great if you could provide a general guide/overview of what the goal/insight is and walk us through the rationale of why/what they optimized things this way. Let's start with phase I first and fully understand it before we move on. Once I know what the goal is, then I can better appreciate/anticipate what the code is trying to do. Right now, I still feel lost

Ninimama:
> @Ninimama Before I look at the other phases, I'd like to fully understand the first phase. I've added some questions/comments but I have more questions still.

That totally makes sense. I will try to go through them by end of this week. 

By the way, I was wondering if you prefer Top-Down design or Bottom-up? Because, as you noticed, I am doing bottom-up and  min_dist value is considered in the main function (like how to initialize and update it). So, just to get things going and see how  the code works, I chose the arbitrary value provided by the author of MERLIN paper for this toy time series.   
(I can explain this in the notebook as well)

Is that a reasonable approach?

seanlaw:
> By the way, I was wondering if you prefer Top-Down design or Bottom-up? Because, as you noticed, I am doing bottom-up and min_dist value is considered in the main function (like how to initialize and update it). So, just to get things going and see how the code works, I chose the arbitrary value provided by the author of MERLIN paper for this toy time series.
(I can explain this in the notebook as well)

What you've done is perfectly fine. I just like to write down comments as I see them so that I don't forget. I like how you've structured things so far so keep up the good work!

udf2457:
I second the comment from @seanlaw , your work on this has been nothing short of exemplary @Ninimama. Keep up the very good work indeed.

Ninimama:
<p>By "compared", I just meant calculating their (z-normalized) Euclidean distance. In Tutorial_STUMPY_Basics.ipynb, there is a similar sentence "... all subsequences within a time series can be compared by computing the pairwise z-normalized Euclidean distances." I think I should better do the same thing here and explain "compare" by adding some context to it.</p><p> </p><p>I will explain the "why" and how this first phase is compared to MP and the first phase of original MERLIN</p><p> </p><p> </p><p> </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>Good to know!!! Thanks!</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
> Before I look at the other phases, I'd like to fully understand the first phase. I've added some questions/comments but I have more questions still.

I am changing the notebook. Should I reply to every comment you provided to let you know where I address it in the notebook? Or, are you good to go through the changes yourself? I just want to see if I need to anything to make the back-and-forth process easier.

Ninimama:
<p>I removed the -1 and change &lt;= to &lt;. I think it is cleaner. I am providing the explanation in the notebook just below the if condition.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>Would you please explain this comment further? because, the way I see it, is that I have if/else statement and each of those assigns different value to last_no_conflict. And, then, this variable is going to be used in the for-loop that is outside of the if-else block. </p><p> </p><p>I can say last_no_conflict+1 is equivalent to first_with_conflict. Then, I can change the if/else and the for-loop that comes after accordingly. </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

seanlaw:
> @seanlaw
> 
> > Before I look at the other phases, I'd like to fully understand the first phase. I've added some questions/comments but I have more questions still.
> 
> I am changing the notebook. Should I reply to every comment you provided to let you know where I address it in the notebook? (because as I change the code, the line number you mentioned in your comments are going to change)
> 
> Or, are you good to go through the changes yourself? I just want to see if I need to do anything to make the back-and-forth process easier.

It's usually best to reply to each comment within the notebook so at least the context (rough location of the original question) is preserved. Unfortunately, this review process is an imperfect system.

I'm starting to feel like this PR is getting way too long. We might want to consider starting a new PR in the near future

seanlaw:
<p>I see your point in that it is being assigned later. Unfortunately, I still don't understand what the goal is from a conceptual level. However, from experience, when I see this kind of <code>range(x + 1)</code> and then <code>x = -1</code>, it suggests that we might be able to do something else to simplify the logic of the code as this usually means that there is a lot of bookkeeping going on and the it is rather hard for others to follow. </p><p> </p><p>Now that you have working code, the next part is refactoring to make sure that anybody who reads it can immediately see the logic. I can help but only if I understand what the goal is and I'll need your help with that.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
> I'm starting to feel like this PR is getting way too long. We might want to consider starting a new PR in the near future. I recommend starting a new PR after we've completed the review of Phase 1. What do you think?

Agreed. Since we now find the direction, we can focus better in  a new PR.



> It's usually best to reply to each comment within the notebook so at least the context (rough location of the original question) is preserved. Unfortunately, this review process is an imperfect system but it is clear to me that notebook comments are annotated with `View entire conversation on ReviewNB` at the bottom of the comment box while PR comments do not. However, instead of continually adding more and more comments to the code, keep the code unchanged (and therefore the line numbers remain mostly unchanged) and add other text before/after the code that refers to the specific line numbers as I've done so that it can be easily cross referenced.

Just to confirm, you mean I should, for instance, write a comment in another cell of notebook and asks the reader to go there to read about the comment that should be inside the code? I can do this next time for sure.

> Now that you have working code, the next part is refactoring to make sure that anybody who reads it can immediately see the logic.

I see. I am trying to see if I can change the variables (and their names accordingly) to make it more legible.

Ninimama:
@seanlaw 
I think you can check out the implementation of phase 1 and the markdown in which I tried to explain the difference between MP, DRAG, MERLIN, and fastMERLIN. 

Also:
> It's usually best to reply to each comment within the notebook so at least the context (rough location of the original question) is preserved.

I am going to answer each of your comments in its corresponding reply section.

Ninimama:
<p>This variable is automatically initialized by (fast)MERLIN and will be passed to this function when it is called inside the main (fast)MERLIN function. I added a note to the docstring. </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

seanlaw:
> Just to confirm, you mean I should, for instance, write a comment in another cell of notebook and asks the reader to go there to read about the comment that should be inside the code? I can do this next time for sure.

Maybe it's a personal preference but I like how the matrix profile papers lay things out by first defining the phase and what it is trying to accomplish. Then, it shows the pseudocode and then goes through each line of the code. So, in a similar fashion, you can have one (Markdown) cell that describes what the overall goal is and what the general thought process is for achieving that goal. Then, start a new (Code) cell where you write the corresponding code. Finally, in a third cell, walk through the relevant sections/lines of code to explain what those lines are doing. 

Usually, when you do this, you'll realize that there just might be too much code and that there is an opportunity to refactor and simplify things OR even split things out into a separate function.

Ninimama:
<p>Resolved! I change the code accordingly.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>Resolved! I also avoid pre-generating the array by creating an empty array as follows:</p><p> </p><p><code>cand_index = np.empty(shape=(0, ), dtype=np.int64)</code></p><p> </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>I changed the code accordingly. </p><p> </p><p><code>cand = np.empty(shape=(0, m))</code></p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>esolved. Outside of for-loop:</p><p><code>cand_in_block_flag = np.full(max_block_size, True)</code></p><p> </p><p>And, inside the for-loop, I re-use it as follows:</p><p><code>is_cand_in_block = np.copy(cand_in_block_flag[:block_size])</code></p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>#explanation:</p><p> </p><pre class="ql-syntax" spellcheck="false">#NOTE: 
            #we are trying to see if we are "allowed" to compare the so-far-discorverd candidates
            #(i.e. the ones found among indices [0,i]) with "at least" one subsequence from the current block!
            #please note that cand_index[0] is ALWAYS less than i; because, so far, we just searched indices [0, i]. 
            #so, the condition might seem to be always true. However, what if excl_zone&gt;block_size? see example below:
          
            #EXAMPLE: 
            #let's say we are now at a block that starts at i=100, and block_size=100, and excl_zone=120! 
            #Then, the right side of inequation becomes 79.
            #therefore, EVEN in a case where the very-first-discovered candidate (i.e. the one at cand_index[0]) 
            #is located at 79, there is still one subsequence in the block that the candidate should be compared with.
            #And, that is the last subsequence of the block locarted at index 199.
            #(because 199-79=120 which is the exclusion zone!)
</pre><p> </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>the variable is changed from <code>last_no_conflict</code>  to <code>first_with_conflict</code>  to avoid the confusion. (Additional note: it is possible that all so-far-discovered candidates have no conflict with the subsequences of  the current block. In that case, the value of <code>first_with_conflict</code>  is the same of <code>last_cmp + 1</code> )</p><p> </p><p>The code is changed accordingly.</p><p> </p><p>Also, in line #70, <code>np.where(cand_index[0] ...</code> is corrected and changed to <code>np.where(cand_index ...</code> <span class="ql-cursor">﻿</span></p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p><code>if cand_index[0] &gt; i - excl_zone:</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;first_with_conflict = 0</p><p>else:</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# Previously, in if condition above, we checked to see if the very-first-discovered candidate is&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#located at an index after i - excl_zone. If yes, that means ALL candidates are going to have conflict&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#with the subsequences of the current block. So, the first_with_conflict candidate...</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#is the very-first-discovered candidate!!!</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# ELSE:&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#that means there exist some candidates that do not have conflict with this current block. So,</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# to speed up computation, we can perform the matrix product between those candidates and</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# ALL the subsequences of the current block!!!&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# the following line finds the very first candidate that has conflict with the current block!&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;first_with_conflict = np.where(cand_index &lt;= i - excl_zone)[0][-1] + 1</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#so, all the discovered candidates in cand[:first_with_conflict, :] has no conflict with the block.</p><p></code></p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p><code>if block_size &gt; excl_zone: </code></p><p>		</p><p>		#i.e. we are allowed to do within-block comparison</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#if not, that means any two subsequences of the block (i.e. their starting index is in the block) are&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#going to have conflict. Therefore, this block will be always skipped! The only downside now is that</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#more candidates are going to be returned at the end of this phase!&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#Therefore, we can set block_size higher than excl_zone to make sure it considers the within-block comparison&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#to prune some candidates.&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#( NOTE:&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# let's consider an extreme case where block-size is the len of time series.&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# In this case, we are not getting that much advantage from the block-by-block approach.&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# The idea behind the block-by-block approach is to narrow down the pair-wise comparison to&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# the so-far-discovered candidates and the sequences (of each block) as we move forward in the time series.</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;# So, what is the optimal block_size? one can think about it later.</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;#)</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>the very first comparison is between the subsequence at index i (the starting index of block)&nbsp; and the one at index i+excl_zone. Alternatively, here, we first start at subsequence "i+excl_zone" in the block&nbsp; and compare it with the preceding subsequences&nbsp;</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>Already resolved and answered in a similar comment.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
> `if block_size > excl_zone: `
> 
> #i.e. we are allowed to do within-block comparison

To be honest, this is getting very confusing for me. If you click on `View entire conversation on ReviewNB` you should my answers (replies) have better format. I don't know why it became messy like this. Also, I replied to each of your comments separately using its corresponding reply section. However, the whole conversation is not being appeared here!!! So, should I do `Quote Reply` then?

seanlaw:
> Also, I replied to each of your comments separately using its corresponding reply section. However, the whole conversation is not being appeared here!!! 

Did you click "Resolved"? I don't see the conversations either :(

Ninimama:
<p>we are trying to see if we are "allowed" to compare the so-far-discorverd candidates (i.e. the ones found among indices [0,i]) with "at least" one subsequence from the current block! Please note that cand_index[0] is ALWAYS less than i; because, so far, we just searched indices [0, i].&nbsp;So, the condition might seem to be always true. However, what if excl_zone&gt;block_size? see example below:</p><p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</p><p>EXAMPLE:</p><p>let's say we are now at a block that starts at i=100, and block_size=100, and excl_zone=120!&nbsp; Then, the right side of inequation becomes 79. Therefore, EVEN in a case where the very-first-discovered candidate (i.e. the one at cand_index[0])&nbsp;is located at 79, there is still one subsequence in the block that the candidate should be compared with. And, that subsequence is the last subsequence of the block located at index 199.</p><p> </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/418/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

Ninimama:
<p>I just checked out all the comments again and clicked on "Resolved" button. </p>

seanlaw:
@Ninimama Actually, I think what is happening is that the comments/replies for the notebook follows the notebook version. Each time you push a new version, reviewNB will display the newest version of the notebook (which will have no comments since it is "new"). So, rather than continually pushing little changes one at a time, you should commit your changes to your local branch (so I do not see them) and do not push those changes to Github until ALL of the comments are addressed. I think that this will help reduce the amount of noise and confusion.

This is starting to be unmanageable. Would you mind starting a new PR (for phase I) and close this one?  And, when you are ready for review, stop making changes until we've agreed that you have addressed all of my questions. This way, each version of the notebook is in a "resolved state". How does that sound?

udf2457:
@seanlaw Yes, my inbox has been swamped with notifications. I decided against commenting about at the time, but it would be nice if the volume could be quietened down a little. ;-)

Ninimama:
@seanlaw 
I was about to say exactly the same thing. I think we should better start a new PR. I am going to close it now. And, I will not push any changes until we resolve all the issues/comments at each step of the reviewing process.

Issue #417, opened 2021-06-19
Add Matrix Profile XXI MERLIN Algorithm for Anomaly Detection


seanlaw:
### Discussed in https://github.com/TDAmeritrade/stumpy/discussions/415

<div type='discussions-op-text'>

<sup>Originally posted by **udf2457** June 19, 2021</sup>
I was reading about MERLIN which was presented at  ICDM 2020:

- Matrix Profile XXI: MERLIN: Parameter-Free Discovery of Arbitrary Length Anomalies in Massive Time Series Archives. Takaaki Nakamura, Makoto Imamura, Ryan Mercer and  Eamonn Keogh. ICDM 2020 [[pdf](http://www.cs.ucr.edu/~eamonn/MERLIN_Long_version_for_website.pdf),[author's Matlab reference implementation]( https://sites.google.com/view/merlin-find-anomalies)]

It would be really neat to see MERLIN implemented in `stumpy`

Ninimama:
@seanlaw  @udf2457

I tried to implement the `Table I` and `Table II` of the paper, referred to as `DRAG` algorithm by the authors. The `DRAG` algorithm will be used later in the `MERLIN` algorithm presented in Table III. 


There are some confusing parts about the algorithm. I tried to check out both the paper and the MATLAB code. I am going to share my understanding with you:

In contrast to what presented in the Table I of the paper, the MATLAB code has a break in the for loop (please see below):

![image](https://user-images.githubusercontent.com/38519522/122677163-61284880-d19e-11eb-8d7a-dbb88335ea03.png)

It changes the output of the Table I algorithm, known as the first phase of DRAG. However, the output of the Table II algorithm, the second phase of DRAG, will remain unchanged (at least for the top-10 discords )

============================================================

In Table II, there is `dj` variable, but no definition is provided for this variable. So, I took a look at the MATLAB code and I realized that `dj` is the distance of a discord to its Nearest Neighbor. And, it should be stored and updated as we move from one subsequence to another through the `outer loop`. 

To keep the same structure for the loops as  presented in Table II, I did a similar thing. The code is not efficient as I tried to get the result first and then refine the code. I return both the ndarray that contains discord info and also the indices of the discords that should be removed. 

===========================================================

I submitted PR #418. 

@udf2457: If you are planning to work on this issue, I can stop and continue on my other projects.

udf2457:
[...]

PR #497, opened 2021-12-19
[WIP] Add Tutorial for Matrix Profile XXI: MERLIN algorithm [Issue #417]

Ninimama:
This PR  addresses the issue #417. As previously discussed in #418, we now know that (fast)MERLIN can discover discords quickly and its output (i.e. index of discords, NN index of discords, the distance between a discord and its NN) is the same as what STUMPY MatrixProfile provides (based on a few tests I conducted before on different data sets)

I started to write a new notebook. For now, I only included the introduction part and explained the first phase of (fast)MERLIN as I want to slow down and work well at each step rather than rushing into things. 

To make the code more understandable and legible, I revised some of the comments. 

@seanlaw 
I will answer all of your previous comments (provided in #418) in the next  comment.


Ninimama:
@seanlaw 
Responses to the comments provided on the first phase of (fast)MERLIN:

> What do you mean by "compared"? And why are we comparing the subsequences? What is the goal?

> It would be great if you can explain what one could do naively and then how this first phase is "optimized" compared to the naive approach.

By "compared", I just meant calculating their z-normalized Euclidean distance. In `Tutorial_STUMPY_Basics.ipynb`, there is a similar sentence: "... all subsequences within a time series can be compared by computing the pairwise z-normalized Euclidean distances." I revised the markdown similarly to clarify the term "compare"

I also explained the disadvantage of MP when it comes to finding discords and explained how the first phase of (fast)MERLIN is faster than the first phase of original MERLIN.

---

> I don't think you need this function. You can do all of this in one line of STUMPY code:

`core.z_norm(core.rolling_window(T, m), axis=1)`

Thanks for the cool shortcut! I removed the function and use the suggested snippet!

---

> Line 2.    `def find_candidates(subsequences, min_dist, excl_zone=None, max_block_size=256):`
How does one decide what value min_dist should be?

The variable `min_dist` is automatically initialized by (fast)MERLIN and will be passed to this function when it is called inside the main (fast)MERLIN function. I added a note to the docstring.

---

> Line 35.        `if excl_zone is None:`
Instead of providing an exclusion zone, we have moved away from that and are using:

> `excl_zone = int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))` 
So, users define the exclusion zone by setting the config.STUMPY_EXCL_DENOM value (this is rare and should be avoided)

Resolved.

---

> Line 40.        `cand_index = np.zeros(n, dtype='int') #contains the starting index of the candidates in ascending order`
Please use dtype=np.int64

Resolved! I also avoid pre-generating this array and revise the line of code as follows:
`cand_index = np.empty(shape=(0, ), dtype=np.int64)`

---

> Line 42.       `#NOTE:`
I understand the sentiment behind the note. However, in Python, if we don't know the size ahead of time then it is often better to create list and append to it since it may be wasteful to pre-generate a 2D array of size n x m. Just something to consider and then you won't have to keep track of which parts of the cand and cand_index are valid

I now do `cand = np.empty(shape=(0,m))` 

---

> Line 51.           `is_cand_in_block = np.full(block_size, True) #at first, we assume all subsequences in the block can be a candidate.`
Move this outside of the for-loop and reuse the same array for each iteration of the for-loop

Resolved! It is changed as follows:

Outside of for-loop:
`cand_in_block_flag = np.full(max_block_size, True)`

And, inside the for-loop, I re-use it as follows:
`is_cand_in_block = np.copy(cand_in_block_flag[:block_size]`

---

> Line 61.                `if cand_index[0] <= i + block_size - excl_zone - 1: `
I don't understand `i + block_size - excl_zone - 1` Can you please explain why we are subtracting the `excl_zone` and also subtracting `1` ?

I removed `-1` and change the inequality sign from <= to <. 
Explanation of the `if condition`:
 i=100, and block_size=100, and excl_zone=120. The right side of inequation becomes 80.  So, if the very first selected candidate is at index 80, it means there are no selected candidates that can be compared with (at least) one subsequence of the current block as their separation will be less than excl_zone..

---


> Line 68.                       ` last_no_conflict = -1`
Why not set `last_no_conflict = 0` instead of `-1` ? Later, you still add +1 back to it and it feels like you really just want it to be zero?

The variable is changed from `last_no_conflict` to `first_with_conflict` to avoid the confusion.  I modified the code accordingly.


---

>Line 70.                       `last_no_conflict = np.where(cand_index[0] <= i - excl_zone)[0][-1]`
I don't understand this

The `[0]` in `cand_index[0]` is a mistake. Also, the code is now changed to:
`first_no_conflict = np.where(cand_index <= i - excl_zone)[0][-1]` + 1

Explanation:
The idea is to determine at what point the conflict begins between the selected candidates and the subsequences of the current block

---

> Line 100.            `if block_size > excl_zone: #i.e. we are allowed to do within-block comparison`
Why are we comparing `block_size` and `excl_zone` ? What happens if `excl_zone` is greater than `block_size`? This can certainly happen if your time series is very long

if `block_size > excl_zone` becomes `True,` then, it means we are allowed to do within-block comparison
if not, that means any two subsequences of the block (i.e. their starting index is in the block) are going to have conflict. Therefore, there will be no within-block comparison for this particular block.
**NOTE:** <br>
The only downside is that more candidates might be returned at the end of this phase! Therefore, we might set the value of block_size greater than excl_zone to make sure we can use the power of vectorization for within-block comparison. Also, doing within-block comparison might help us to narrow down the selected candidates more. 
Choosing bigger block is not always better. For instance, let us consider an extreme case where block-size is the length of time series. In this case, we are NOT getting that much advantage from the vectorization as we are losing the within-block comparisons. 

---

Line 103.                    `corr = np.matmul(subsequences[i + pos,:], subsequences[i:i + pos - excl_zone + 1 ,:].T) `
Can you explain why we choose `subsequences[i + pos, :]` and `subsequences[i:i + pos - excl_zone + 1, :]` ? It isn't clear to me why this selection of subsequences are being compared

Please note that this is for within-block comparisons. Since there are conflicts between a subsequence and some other subsequences of the same block, we should iterate through all subsequences one by one. 

the first comparison is between the subsequence at `index i` (the starting index of block) and the one at index `i+excl_zone.` Alternatively, we can start the comparing process by calculating the z-normalized Euclidean distance between the subsequence at `i+excl_zone` (located in the block) with its preceding subsequences 

---

> Line 4.    `min_dist = 10.27`
10.27 seems really arbitrary.

previously answered.

seanlaw:
@Ninimama After going through the code so far, I think it is too complicated to maintain/understand. It's probably best if we try and implement the basic MERLIN in a separate PR. What do you think?

Ninimama:
@seanlaw 
I totally agree. I am now working on MERLIN. I will update you by creating a new PR if I get to a point where I see the code is working AND it is clean. 

Thanks.

Issue #509, opened 2022-01-03
stumpy snippets: mistake in plotting the regimes in the tutorial notebook


Ninimama:
I think there might be a small issue in plotting the regimes in the [Snippets Tutorial Notebook](https://github.com/TDAmeritrade/stumpy/blob/main/docs/Tutorial_Time_Series_Snippets.ipynb). 

---

I took a look at the paper [Matrix Profile XIII: Snippets](https://www.cs.ucr.edu/~eamonn/Time_Series_Snippets_10pages.pdf), and I checked the implementation of snippet in STUMPY. It is all good! However, The problem is when the regimes are being plotted (section: `plotting the regimes of snippets`)

```
for i in range(k):
    slices_of_indices = regimes[np.where(regimes[:,0]==i)][:,1:]
    print(f'indices slice for the regime of the snippets #{i} is: {slices_of_indices}')
    
    for per_slice in slices_of_indices:
        start_idx = per_slice[0]
        stop_idx = per_slice[1]+m-1
        plt.plot(range(start_idx,stop_idx),T[start_idx:stop_idx], lw=5, label='snippt #{}'.format(i))

plt.legend(loc='upper right',fontsize=15)
plt.show()
``` 

As you can see, it is assumed that the indices included in the regimes are just the beginning of a subsequence. 
(please notice:  **stop_idx = per_slice[1]+m-1**)

I think we should do **stop_idx = per_slice[1]+m** (that `-1` was also wrong) ONLY for the most-right regime as the last index returned by the regime is `len(T) - m` in that particular case.

seanlaw:
@Ninimama Yes, that sounds right. Would you mind submitting a PR to correct this?

Ninimama:
@seanlaw 
Thanks for the confirmation. I will submit a PR regarding this issue.

Ninimama:
I submitted a PR. But, before you accept the changes, you may want to take a look at this:

I tried to use `stumpy,snippets` on electricity demand data (with length 8760). For `m=7*24`, one of the slices in a regime was `[3832 3833]` which doesn't make sense unless we add `m-1` to each index included in that range. Then, I think more about it. Please correct me if I am wrong. Since snippets is based on the matrix profile and the matrix profile values correspond to the starting index of each subsequence, we should add `m-1` to the indices returned in the regimes. (Maybe, overlap is inevitable in snippets)

seanlaw:
[...]

PR #510, opened 2022-01-04
Tutorial snippets (#509)

Ninimama:
This PR resolved the issue discussed in #509 regarding the plots of snippets in the [Tutorial_Time_Series_Snippets notebook](https://github.com/TDAmeritrade/stumpy/blob/main/docs/Tutorial_Time_Series_Snippets.ipynb).

We  just update the `stop_idx` of a regime only if that is the most-right index. This is to make sure the snippets are not going to overlap, and to cover the end part of time series as well. (Note that the length of matrix profile is `len(T) - m + 1` and not `len(T)`.)


Ninimama:
@seanlaw 
Do you know why my commits from the other branch (for `Discord`) appeared here? (Because, I created another branch and just pushed that here)

seanlaw:
> Do you know why my commits from the other branch (for Discord) appeared here? (Because, I created another branch and just pushed that here)

I'm not sure but all signs point to `Tutorial_Snippets` not branching directly from `main`. We should not accept this PR and figure this out as we do not want to mix the branches

Issue #552, opened 2022-02-26
Add Unit Tests for `mmotifs` and `aamp_mmotifs`


seanlaw:
The `mmotifs` and `aamp_mmotifs` features were recently added but they need unit tests to accompany them:

- [ ] <s>Add naive implementation for `mmotifs` and `aamp_motifs` in `tests/naive.py`</s>
- [ ] Add `pytest -x -W ignore::RuntimeWarning -W ignore::DeprecationWarning tests/test_motifs.py` to the `test_unit` function in `test.sh`
- [ ] Add `pytest -x -W ignore::RuntimeWarning -W ignore::DeprecationWarning tests/test_aamp_motifs.py` to the `test_unit` function in `test.sh`
- [ ] Add unit tests for `mmotifs` in `tests/test_mmotifs.py`
- [ ] Add unit tests for `aamp_motifs` in `tests/test_aamp_mmotifs.py` 
- [ ] Add `mmotifs` and `aamp_motifs` to `stumpy/__init__.py` 
- [ ] Add `mmotifs` to `docs/api.rst`
- [ ] Add Example section to `mmotifs()` docstring:

```
    Examples
    --------
    >>> mp = stumpy.stump(np.array([584., -11., 23., 79., 1001., 0., -19.]), m=3)
    >>> stumpy.motifs(
    ...     np.array([584., -11., 23., 79., 1001., 0., -19.]),
    ...     mp[:, 0],
    ...     max_distance=2.0)
    (array([[0.        , 0.11633857]]), array([[0, 4]]))
```

seanlaw:
@SaVoAMP [had asked](https://github.com/TDAmeritrade/stumpy/pull/544#issuecomment-1052243504):

> Unfortunately, I have never worked with unit tests, which means that I am a complete beginner. Are there any resources available since I have absolutely no clue how to start?

SaVoAMP:
All right, so how to implement functions naivly? 🤔 
Can I then also start with the tutorial in parallel?

seanlaw:
> I will need to investigate why mmotifs and aamp_motifs aren't triggering missing coverage.

@SaVoAMP I fixed this. Can you please pull and merge the latest commit to your branch?

SaVoAMP:
> Btw, did you have to add any additional steps that may have been omitted in my description above?

No, I reforked the repo, added a branch, cloned it inside my linux vm, activated a virtual environment for python and ran `pip.sh` as you described.

[...]

PR #553, opened 2022-02-27
[WIP] Add `mmotifs` and `aamp_mmoitfs` Unit Tests

SaVoAMP:
# Pull Request Checklist

- [x] Create a new branch
#552


seanlaw:
@SaVoAMP Before pushing your commits, please remember to execute `black . && flake8 .`

SaVoAMP:
> ./test_mstump.py:93:1: D103 Missing docstring in public function
./test_mstump.py:111:1: D103 Missing docstring in public function
./test_mstump.py:130:1: D103 Missing docstring in public function
./test_mstump.py:141:1: D103 Missing docstring in public function
./test_mstump.py:155:1: D103 Missing docstring in public function
./test_mstump.py:166:1: D103 Missing docstring in public function
./test_mstump.py:182:1: D103 Missing docstring in public function
./test_mstump.py:194:1: D103 Missing docstring in public function
./test_mstump.py:210:1: D103 Missing docstring in public function
./test_mstump.py:216:1: D103 Missing docstring in public function
./test_mstump.py:227:1: D103 Missing docstring in public function
./test_mstump.py:241:1: D103 Missing docstring in public function
./test_mstump.py:252:1: D103 Missing docstring in public function
./test_mstump.py:267:1: D103 Missing docstring in public function
./test_mstump.py:284:1: D103 Missing docstring in public function
./test_mstump.py:304:1: D103 Missing docstring in public function
./test_mstump.py:317:1: D103 Missing docstring in public function
./test_mstump.py:338:1: D103 Missing docstring in public function
./test_mstump.py:359:1: D103 Missing docstring in public function
./test_scrump.py:27:1: D103 Missing docstring in public function
./test_scrump.py:44:1: D103 Missing docstring in public function
./test_scrump.py:61:1: D103 Missing docstring in public function
./test_scrump.py:79:1: D103 Missing docstring in public function
./test_scrump.py:95:1: D103 Missing docstring in public function
./test_scrump.py:102:1: D103 Missing docstring in public function
./test_scrump.py:136:1: D103 Missing docstring in public function
./test_scrump.py:170:1: D103 Missing docstring in public function
./test_scrump.py:205:1: D103 Missing docstring in public function
./test_scrump.py:239:1: D103 Missing docstring in public function
./test_scrump.py:277:1: D103 Missing docstring in public function
./test_scrump.py:315:1: D103 Missing docstring in public function
./test_scrump.py:343:1: D103 Missing docstring in public function
./test_scrump.py:371:1: D103 Missing docstring in public function
./test_scrump.py:412:1: D103 Missing docstring in public function
./test_scrump.py:457:1: D103 Missing docstring in public function
./test_scrump.py:486:1: D103 Missing docstring in public function
./test_scrump.py:515:1: D103 Missing docstring in public function
./test_scrump.py:544:1: D103 Missing docstring in public function
./test_scrump.py:580:1: D103 Missing docstring in public function
./test_scrump.py:621:1: D103 Missing docstring in public function
./test_scrump.py:662:1: D103 Missing docstring in public function

I am not sure what to do because I get many problems when running flake8 .

seanlaw:
Can you try running those commands in the root STUMPY directory instead on inside the `tests/` directory?

SaVoAMP:
Ohh, sry, that was pretty stupid! Probably better not to make pull requests with a headache 😆

SaVoAMP:
I tried to write the first testing function. I'm not sure whether I did it as expected 🤔

SaVoAMP:
I think we have got a problem with `mmotifs`:
```
 query_matches = match(
            Q=T[subspace_k, motif_idx : motif_idx + m],
            T=T[subspace_k],
            M_T=M_T[subspace_k],
            Σ_T=Σ_T[subspace_k],
            max_matches=max_matches,
            max_distance=max_distance,
            atol=atol,
            normalize=normalize,
            p=p,
        )

        query_length = len(query_matches)

        if len(query_matches) > min_neighbors:
            motif_distances.append(query_matches[:, 0])
            motif_indices.append(query_matches[:, 1])
            motif_subspaces.append(subspace_k)
            motif_mdls.append(mdls)
```

I think we need something like `query_matches.size` instead of `len(query_matches)` since `query_matches` is an array.
I just noticed that I didn't get the expected results while trying to construct the test example:

I have got `array([[0.0, 3]], dtype=object)` for `query_matches` and when writing `query_length = len(query_matches)` I get `1` as result but I should get `2`. Thus the `if`-statement gets skipped and I get wrong results.

seanlaw:
> I have got array([[0.0, 3]], dtype=object) for query_matches and when writing query_length = len(query_matches) I get 1 as result but I should get 2. 

Based on `array([[0.0, 3]], dtype=object)`, I would agree that the length is `1` as expected. Why would you expect to get 2? Please let me know if there's anything I can do to help.

SaVoAMP:
I'm sorry for the confusion!

The problem is that I have `min_neighbors=1`. Shouldn't this mean that I should get one motif pair without the other matches at least? Or is the `max_distance` parameter superior?

SaVoAMP:
I get exit code 5 on pytest which means, there are no tests to run.

I don't see any assertions (using the assert keyword) as seen in the documentation of pytest in your code.

I tried to use your code but I still get the same error message. What am I missing?

seanlaw:
> I get exit code 5 on pytest which means, there are no tests to run.
I don't see any assertions (using the assert keyword) as seen in the documentation of pytest in your code.
I tried to use your code but I still get the same error message. What am I missing?

Let me take a look.

seanlaw:
> I get exit code 5 on pytest which means, there are no tests to run.
> I don't see any assertions (using the assert keyword) as seen in the documentation of pytest in your code.
> I tried to use your code but I still get the same error message. What am I missing?

I think I understand what the issue is after looking at the unit tests that are being executed [here](https://github.com/TDAmeritrade/stumpy/runs/5359804415?check_suite_focus=true):

```
============================= test session starts ==============================
platform linux -- Python 3.7.12, pytest-7.0.1, pluggy-1.0.0
rootdir: /home/runner/work/stumpy/stumpy, configfile: pytest.ini
collected 7 items

tests/test_aamp_motifs.py .......                                        [100%]

============================== 7 passed in 1.09s ===============================
============================= test session starts ==============================
platform linux -- Python 3.7.12, pytest-7.0.1, pluggy-1.0.0
rootdir: /home/runner/work/stumpy/stumpy, configfile: pytest.ini
collected 0 items

============================ no tests ran in 0.00s =============================
Error: pytest encountered exit code 5
Error: Process completed with exit code 5.
```

This is consistent with what you are observing. However, it wasn't clear which test file was experiencing an `exit code 5` though the test file that last succeeded was for `tests/test_aamp_motifs.py`. Since this is a unit test (and not coverage testing), I took a look at the `test_unit` section to see what the next test was after `tests/test_aamp_motifs.py`:

```
tests/test_aamp_motifs.py
check_errs $?
pytest -x -W ignore::RuntimeWarning -W ignore::DeprecationWarning tests/test_aamp_mmotifs.py
check_errs $?
```

Okay, so the next test is `tests/test_aamp_mmotifs.py` and when I look inside that file I see that it is empty, which is then consistent with your observation of:

>  I get exit code 5 on pytest which means, there are no tests to run.

Because there are, indeed, no tests to run inside of `tests/test_aamp_mmotifs.py`. I'm guessing that if you add a simple dummy test inside of `tests/test_aamp_mmotifs.py` like:

```
def increment(x):
    return x + 1


def test_answer():
    assert increment(3) == 4
```

Then you should be able to overcome this `pytest` exit 5 error. Now, to address your comment:

> I don't see any assertions (using the assert keyword) as seen in the documentation of pytest in your code.

Yes, it is true that I don't explicitly use the `assert` keyword in my tests. This is because `numpy` arrays are especially tricky (i.e., not only do you need to compare each element between two arrays but you need to also consider things like the `shape` and the `dtype`, etc) to compare with each other using plain vanilla `assert` and we perform a ton of comparisons between `numpy` arrays in STUMPY. So, instead, [`numpy` provides a very nice set of testing tools](https://numpy.org/doc/stable/reference/routines.testing.html) that will handle a lot of this for us. Thus, in place of Python's `assert` which is typically used for comparing single (scalar-like) values, we use `numpy.testing` assertions. Here is a simple example (I have only written this down from the top of my head so I hope it has minimal errors):

```
import numpy as np
import numpy.testing as npt

def naive_increment(x, incr):
    x = x.copy()
    for i in range(x):
        x[i] += incr
    return x


def test_answer():
    ref = np.array([3, 4, 5, 6, 7, 8, 9, 10, 11, 12], dtype=np.int64)
    x = np.arange(10)
    cmp = naive_increment(x, 3)
    npt.assert_almost_equal(cmp, ref)
```

Notice that we have a reference `ref` that is basically the "expected outcome" and then the thing that we are comparing to, `cmp`. To compare the two `numpy` arrays, we use `numpy.testing` (aliased to `npt` here to save us from typing) and then we assert that the two arrays are "almost equal" via `npt.assert_almost_equal(cmp, ref)`. We typically use `assert_almost_equal` (which is overkill here for integers) since floating point values will have poorer precision and so we allow for some slight differences but not a lot. 

In fact, you can combine both sets of `numpy` assertions and Python `assert` into one test file if you like:

```
import numpy as np
import numpy.testing as npt

def increment(x, incr):
    x = x.copy()
    for i in range(x):
        x[i] += incr
    return x


def test_answer():
    ref = np.array([3, 4, 5, 6, 7, 8, 9, 10, 11, 12], dtype=np.int64)
    x = np.arange(10)
    cmp = increment(x, 3)
    npt.assert_almost_equal(cmp, ref)


def increment(x):
    return x + 1


def test_answer():
    assert increment(3) == 4
```

Please let me know if I can help fill any knowledge/understanding gaps.

seanlaw:
Awesome! It looks like all [tests are passing now](https://github.com/TDAmeritrade/stumpy/actions/runs/1911115218) (even the dummy one). Now, you can focus on writing the tests for coverage:

```
Name                     Stmts   Miss  Cover   Missing
------------------------------------------------------
stumpy/aamp_mmotifs.py      51     44    14%   108-195
stumpy/core.py             407      1    99%   4[78](https://github.com/TDAmeritrade/stumpy/runs/5361908563?check_suite_focus=true#step:8:78)
stumpy/mmotifs.py           51      6    88%   144, 157, 170-171, 182, 207
------------------------------------------------------
TOTAL                    10684     51    99%
```

Please ignore the missing coverage for `stumpy/core.py`. I just pushed some new code that fixes that (so please merge the newest addition at your earliest convenience)

SaVoAMP:
Alright, perfect!

Would you mind taking a look at my first test function in `test_mmotifs.py` before I start with the other tests? I would appreciate any feedback as I'm pretty unconfident with all of this.

seanlaw:
> Would you mind taking a look at my first test function in test_mmotifs.py before I start with the other tests? I would appreciate any feedback as I'm pretty unconfident with all of this.

Yes, please give me some time to review

tests/test_mmotifs.py
```diff
@@ -0,0 +1,137 @@
...
+        ]
+    )
+    m = 3
+
+    P, I = mstump(T, m)
+    motif_distances, motif_indices, motif_subspaces, motif_mdls = mmotifs(
+        T, P, I, max_distance=np.inf, max_matches=2, k=1
+    )
+
+    # Assert
```
seanlaw, line 107: Yep, if the tests pass then you are on the right track!  The goal here is to compare your manual calculations (without calling the `mmotifs()` function) with what the `mmotifs()` function would produce.

Then, you try to test all possible combinations of inputs by hand (note your expected manual outputs) and then see if the function you are testing also produces the same expected outputs. If not then there is something wrong with your function. 

After that, the goal is to think about nasty edge cases and to write tests to ensure that we are handling those edge cases properly. We don't have to catch everything (it's impossible to) but at least think about how a user might use/abuse the function and hopefully we test and make sure that we consider all reasonable (i.e., rationally obvious) use cases and have good coverage on those. 

Also, as you may have noticed, the goal of having a "naive" implementation means that you don't have to do things manually and, instead, it allows you to try random inputs of different small arrays (in addition to a pre-determined array with known outputs). The only difference between the naive implementation vs your implementation should be speed (because the naive implementation would use a bunch of slow for-loops but is easier to read and follow versus faster algorithms are usually VERY HARD to read/follow since they often employ hyper specialized tricks).

seanlaw:
> Alright, perfect!
> 
> Would you mind taking a look at my first test function in `test_mmotifs.py` before I start with the other tests? I would appreciate any feedback as I'm pretty unconfident with all of this.

@SaVoAMP I've left a few comments but, seriously, (no surprise) you're doing a great job! You'll be teaching me about unit testing soon :)

tests/test_mmotifs.py
```diff
@@ -0,0 +1,137 @@
...
+        np.array([0.1, 1.0, 2.0, 3.0, -1.0, 0.1, 1.0, 2.0, -0.5]),
+    ),
+    (np.random.uniform(-1000, 1000, [8]), np.random.uniform(-1000, 1000, [64])),
+]
+
+
+def test_motifs_multidimensional_one_motif_all_dimensions():
+    # Find the two dimensional motif pair
+
+    # Arrange
```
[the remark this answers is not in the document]
SaVoAMP: I was using already covered functions to obtain them. Is this ok?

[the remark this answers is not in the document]
seanlaw: Yes, that should be fine. However, if for some reason those functions are not covered then we should also cover them first.

SaVoAMP:
I think we might have got a problem with `mmotifs` (another problem as the one above). 
In the first function I was testing whether I can find the two dimensional motif pair correctly. The pair is shown in red:
grafik

I was searching for the top two motif pairs. With MDL I get a one dimensional motif both times. The first moitf pair is located in the third time series (T_4 in my plot) at the positions 2 (3) and 12 (13) (don't be confused since I am counting from 1 in the plot).
This would mean that I obtain the following (please ignore the two dimensional motif pair):
grafik
Afterwards we are setting the exclusion zones 
grafik
and try to find the next candidate motif (and here I think that the problem arises) via:
```
for idx in query_matches[:, 1]:
     core.apply_exclusion_zone(P, idx, excl_zone, np.inf)
candidate_idx = np.argmin(P, axis=1)
nn_idx = I[np.arange(len(candidate_idx)), candidate_idx]
```
The result for the new `candidate_idx` is `[8, ..., ..., ...]` and  `[11, ..., ..., ...]` for the `nn_idx`. In the next iteration mdl tells me that `k=0` and that the motif is present in T_2. After plotting the motif pair on positions 8 (9) and 11 (12) I see the following problem, since we are only setting the matrix profile values to infinity, but aren't doing anything with the MP indices:
grafik
Should this be like this? I haven't thought about such a case so far. Or can we say that it is ok, since it really may be another motif and not exactly the same in another dimension?

seanlaw:
> puhh, that's strange. I was debugging the new function before pushing and checking if the results are the same. And they were the same, as you can see 🤔 I have absolutely no clue what might be wrong. I can't explain what causes the test to produce a `0` instead of a `2` in the array as the debugger doesn't yield any clues.

This is strange indeed! I tried running your tests locally and the tests passed as expected. I just re-ran the tests so let's see if anything changes.

seanlaw:
@SaVoAMP I found one thing that may be missing in `mmotifs()` and `aamp_mmotifs()` and that is we need to make a copy of the original input `P`:

```
T, M_T, Σ_T = core.preprocess(T, m)
P = P.copy()
```

This way, as we apply the exclusion zone, we don't directly affect/change the `P` that the user passed into us. Can you please fix this?

seanlaw:
> Should this be like this? I haven't thought about such a case so far. Or can we say that it is ok, since it actually may be another motif and not exactly the same in another dimension?

I think this is okay. I can't recall if I mentioned this to you previously but my personal definition of a "motif" is for a single subsequence (or multi-dimensional subsequence all starting at the same, single index location) and **not** for a pair of subsequences. Hopefully that is clear.

So, if we have a single candidate motif then I think that it's single nearest neighbor is allowed to be anywhere along the time series (except within the candidate motif's exclusion zone). For the sake of clarity, let's say call the index location of this first candidate motif `idx_1` and the index location of its nearest neighbor as `nn_idx_1. After we record this motif (by appending it to our growing list), this candidate is no longer allowed to be considered a candidate motif and we enforce this by applying an exclusion zone around `idx_1`. However, there is absolutely nothing stopping us from assigning the location at `idx_1` as a nearest neighbor location for a different candidate motif. In fact, it is perfectly fine for a single subsequence to be a nearest neighbor to multiple motifs! In other words, maybe we get down to the 8th candidate motif (located at `idx_8`) and we find that its nearest neighbor is actually located at `nn_idx_8 = idx_1` and again, later on, for the 52nd candidate motif (located at `idx_52`) we find that its nearest neighbor is at `nn_idx_52 = idx_1`. I don't see anything wrong with that. Just because a subsequence was discovered to be a motif, we should **never** eliminate that subsequence from being identified as the nearest neighbor for other subsequences. At least, I don't have any reason to do so. Do you?

SaVoAMP:
Ok! I was wondering if we might need the value of the nearest neighbor for something? Because this information would be lost since the matrix profile is infinity, if the nearest neighbor is located inside of an exclusion zone.

seanlaw:
> I was wondering if we might need the value of the nearest neighbor for something? Because this information would be lost since the matrix profile is infinity, if the nearest neighbor is located inside of an exclusion zone.

I'm not sure what you mean by this. What "value of the nearest neighbor" are you referring to? Is this the distance or the index location? What information would be lost?

SaVoAMP:
yes, I mean the distance.

seanlaw:
> yes, I mean the distance.

But the distance to the nearest neighbor is stored in `motif_indices` and `motif_distances`, no?

seanlaw:
> puhh, that's strange. I was debugging the new function before pushing and checking if the results are the same. And they were the same. I have absolutely no clue what might be wrong. 

@SaVoAMP  Did you manage to figure out what the problem was?

SaVoAMP:
> Did you manage to figure out what the problem was?

Yes, sorry! Looks like if I was too stupid for copy and paste. It doesn't even make sense to only return one subspace, if two motifs are found. Don't know what I was doing there on that day. I have now switched from Visual Studio Code to Pycharm and found that you can easily do pytests there. Now such stupid errors should hopefully no longer occur!

seanlaw:
> Yes, sorry! Looks like if I was too stupid for copy and paste. It doesn't even make sense to only return one subspace, if two motifs are found. Don't know what I was doing there on that day. I have now switched from Visual Studio Code to Pycharm and found that you can easily do pytests there. Now such stupid errors should hopefully no longer occur!

I had that happen to me so no worries at all! I think what's more important is that you understood clearly why the test was failing and that it wasn't some random unexpected behavior. This is precisely why we write simple unit tests as they help us make sure that our expectations are "correct".

SaVoAMP:
```
E       Mismatched elements: 1 / 2 (50%)
E       Max absolute difference: 2
E       Max relative difference: 0.
E        x: array([[ 2, 12]])
E        y: array([[ 0, 12]])

tests/test_mmotifs.py:132: AssertionError
=========================== short test summary info ============================
FAILED tests/test_mmotifs.py::test_motifs_multidimensional_with_deafault_parameters
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
================== 1 failed, 618 passed in 959.41s (0:15:59) ===================
```

Seems as if we have the same strange problem now as the last time. Could you figure out what's wrong? To me it doesn't make sense at all 🤔

seanlaw:
> Seems as if we have the same strange problem now as the last time. Could you figure out what's wrong? To me it doesn't make sense at all 🤔

I will need to find some time to take a look. One observation (or possible red herring) is that it is happening during coverage testing when `NUMBA_DISABLE_JIT=1` (i.e., the numba JIT compiler is turned off). This doesn't affect `mmotifs()` but it may affect `mstump`, though, the distances should still be the same. 

I also noticed that if we change one element in `T[3, 4] = 6.01` then `mmotifs()` ends up getting stuck in the `while` loop and does not satisfy the condition to break out of the loop. Would you mind taking a look?

seanlaw:
Okay, I can confirm that I am able to reproduce the same behavior on my laptop when I set:

```
import numba
numba.config.DISABLE_JIT = True
```

Please give me some time to dig into this but I have a feeling it has to do with `mstump`

seanlaw:
@SaVoAMP I am inclined to believe that there may be a bug in `mstump` that rears its head when `njit` (numba JIT compilation) is turned on and your simple test case has revealed it! If you look at `T`:

```
T = np.array(
    [
        [5.0, 0.0, 3.0, 3.0, 7.0, 9.0, 3.0, 5.0,  2.0, 4.0, 7.0, 6.0, 8.0, 8.0, 1.0],
        [7.0, 3.0, 5.0, 9.0, 8.0, 7.0, 4.0, 5.0, 10.0, 8.0, 4.0, 3.0, 2.0, 0.0, 1.0],
        [6.0, 7.0, 7.0, 8.0, 1.0, 5.0, 9.0, 8.0,  9.0, 4.0, 3.0, 0.0, 3.0, 5.0, 0.0],
        [0.0, 1.0, 3.0, 2.0, 6.0, 1.0, 9.0, 10.0, 1.0, 2.0, 2.0, 5.0, 1.0, 0.0, 4.0],
    ]
)
m = 3
```
There is actually a perfect subsequence match in the second row specifically at `T[1, 0:0+m]` and `T[1, 12:12+m]` and is revealed by comparing their z-normalized distances:
```
y = np.array(T[1, 0:0+m])
x = np.array(T[1, 12:12+m])
(x - x.mean())/x.std(), (y - y.mean())/y.std()

(array([ 1.22474487, -1.22474487,  0.        ]),
 array([ 1.22474487, -1.22474487,  0.        ]))
```
Thus, a result of `[[0, 12]]` is (also?) correct! For some reason, if I turn off `njit` for `mstump` (and all functions that is called by `mstump` then we get `[[0, 12]]`. However, if we turn on `njit` then this `[[0, 12]]` in row `1` gets overlooked and `[[2, 12]]` from row `3` is selected. This needs to be fixed first and I will try to get to the bottom of it!

seanlaw:
In fact, in the 1-D case, it looks like there are a possible 8 different subsequences that produce a perfect (zero-distance) match after z-normalization and it can come from different rows of `T`.

SaVoAMP:
oh ok, that is strange. I had created time series T_2 and T_4 by hand. Time series T_1 and T_3 are random. Originally I wanted to find the two-dimensional motif in the time series T_2 and T_4.

> I also noticed that if we change one element in T[3, 4] = 6.01 then mmotifs() ends up getting stuck in the while loop and does not satisfy the condition to break out of the loop. Would you mind taking a look?

Ok, that is strange too. I probably won't get to it today, but I'll check it out tomorrow at the latest!

seanlaw:
This is good though! This is precisely why we do unit testing as it helps us uncover edge cases that we had not considered.

SaVoAMP:
> I also noticed that if we change one element in T[3, 4] = 6.01 then mmotifs() ends up getting stuck in the while loop and does not satisfy the condition to break out of the loop. Would you mind taking a look?

I'm not sure why 🤔 When I'm setting `T[3, 4] = 6.01` (zero count I guess) and `max_distance=np.inf` and `max_matches=2` to get exactly one motif pair, I get the following:
`motif_distances=[[0.0, 0.0]]`
`motif_indices_arr=[[8, 11]]`
`motif_subspaces=[np.array([1])]`
`motif_mdls=[np.array([176.0, 177.509775, 191.26466251, 235.01955001])]`
So I'm not getting stuck in the while loop 🤔 
What parameters did you pass to `mmotifs`?

SaVoAMP:
Do you have any suggestions what I should also test in `test_mmotifs` and `test_aamp_mmotifs`? I think maybe testing the exclusion zone as in `test_motifs` and what else am I missing?

seanlaw:
> oh ok, that is strange. I had created time series T_2 and T_4 by hand. Time series T_1 and T_3 are random. Originally I wanted to find the two-dimensional motif in the time series T_2 and T_4.

So, I've come to the conclusion that there isn't anything wrong with `mstump` beyond minor errors in numerical precision. First, instead of calling `stumpy.mstump`, please use `naive.mstump` instead. Second, your time series is contains 8 subsequences that have a nearly identical nearest neighbor. When I do:
```
T = np.array(
    [
        [5.0, 0.0, 3.0, 3.0, 7.0, 9.0, 3.0, 5.0,  2.0, 4.0, 7.0, 6.0, 8.0, 8.0, 1.0],
        [7.0, 3.0, 5.0, 9.0, 8.0, 7.0, 4.0, 5.0, 10.0, 8.0, 4.0, 3.0, 2.0, 0.0, 1.0],
        [6.0, 7.0, 7.0, 8.0, 1.0, 5.0, 9.0, 8.0,  9.0, 4.0, 3.0, 0.0, 3.0, 5.0, 0.0],
        [0.0, 1.0, 3.0, 2.0, 6.0, 1.0, 9.0, 10.0, 1.0, 2.0, 2.0, 5.0, 1.0, 0.0, 4.0],
    ]
)
m = 3
m = 3
naive.mstump(T, m, excl_zone)
```
I get:
```
(array([[0.00000000e+00, 5.20740757e-16, 3.33066907e-16, 0.00000000e+00,
         1.57405307e-01, 1.75788205e-01, 2.18688574e-15, 9.98752684e-02,
         3.55444798e-16, 4.19103734e-01, 0.00000000e+00, 3.55444798e-16,
         0.00000000e+00],
        [3.73957587e-01, 5.43661105e-01, 1.93596696e-01, 4.54076717e-02,
         1.78206916e-01, 3.73957587e-01, 5.96436567e-01, 1.93596696e-01,
         4.27594911e-01, 5.32427565e-01, 4.54076717e-02, 1.78206916e-01,
         2.38422270e-01],
        [4.24764519e-01, 1.37723753e+00, 8.64372909e-01, 2.82997680e-01,
         9.54173183e-01, 4.24764519e-01, 1.41488677e+00, 8.64372909e-01,
         7.14473894e-01, 9.12442982e-01, 2.82997680e-01, 9.54173183e-01,
         9.12442982e-01],
        [5.79539036e-01, 1.64167328e+00, 1.31307407e+00, 7.14495646e-01,
         1.38336362e+00, 5.79539036e-01, 1.64167328e+00, 1.31307407e+00,
         1.03810281e+00, 1.35206597e+00, 7.14495646e-01, 1.38336362e+00,
         1.35206597e+00]]),
 array([[12, 11, 12, 10, 11,  8, 10,  2, 11,  2,  3,  8,  0],
        [ 5, 11,  7, 10, 11,  0, 10,  2,  3,  2,  3,  4,  2],
        [ 5,  3,  7, 10, 11,  0,  1,  2,  3, 12,  3,  4,  9],
        [ 5,  6,  7, 10, 11,  0,  1,  2,  3, 12,  3,  4,  9]]))
```
And, as you can see, the first dimension in the multi-dimensional matrix profile is filled with values that are exceptionally small and, for all intents and purposes, are "zero". I think it would be best to alter your time series a little (even simply after the decimal place) to ensure a tiny bit of variability.

seanlaw:
> So I'm not getting stuck in the while loop 🤔 What parameters did you pass to mmotifs?

```
T = np.array(
    [
        [5.0, 0.0, 3.0, 3.0, 7.0, 9.0, 3.0, 5.0,  2.0, 4.0, 7.0, 6.0, 8.0, 8.0, 1.0],
        [7.0, 3.0, 5.0, 9.0, 8.0, 7.0, 4.0, 5.0, 10.0, 8.0, 4.0, 3.0, 2.0, 0.0, 1.0],
        [6.0, 7.0, 7.0, 8.0, 1.0, 5.0, 9.0, 8.0,  9.0, 4.0, 3.0, 0.0, 3.0, 5.0, 0.0],
        [0.0, 1.0, 3.0, 2.0, 6.1, 1.0, 9.0, 10.0, 1.0, 2.0, 2.0, 5.0, 1.0, 0.0, 4.0],
    ]
)
m = 3
P, I = mstump(T, m)
motif_distances, motif_indices, motif_subspaces, motif_mdls = mmotifs.mmotifs(T, P, I)
```

This loops forever. Can you please confirm?

seanlaw:
> Do you have any suggestions what I should also test in test_mmotifs and test_aamp_mmotifs? Maybe the include parameter? But it wasn't tested in test_motifs.

In general, it is good to test all input parameters. And then, test all branches of the function (i.e., every time there is an `if/else` statement, you want to ensure that you traverse both paths). At the end of the day, I think it is reasonable once you see that there are no "Missing" lines mentioned in code coverage.

SaVoAMP:
> This loops forever. Can you please confirm? Considering that this is the basic function call, we need to ensure that things don't unnecessarily run forever. Unfortunately, this may be a poor choice of using a while loop and something that we may want/need to change before releasing this feature to the public.

I really don't understand what is going on. If I pass in exactly the same that you did, the while loop is ending after the first iteration. And I get the same outputs as before.

SaVoAMP:
>I think it would be best to alter your time series a little (even simply after the decimal place) to ensure a tiny bit of variability.

Sure, I'll try.

SaVoAMP:
Now I know what you mean. The problems only arise when using the naive implementation of stumpy. When using the original function, everything is working.

seanlaw:
> Now I know what you mean. The problems only arise when using the naive implementation of stumpy. When using the original function, everything is working. I don't understand why.

To reiterate, both `naive.mstump` and `stumpy.mstump` are "correct" up to roughly six significant digits but `naive.mstump` is more precise as it computes the distances using a much slower but more precise approach. However, since your input time series contains multiple "best" motifs (i.e., with zero distance), it is hard to choose which subsequence is the "best" candidate since they are all technically the same. Additionally, the likelihood of this occurring in real world applications is exceptionally rare as the size of your window size `m` also increases. For toy cases or unit tests we use `m=3` so that the tests are fast but, in reality, nobody should ever choose such a small window size. 

So, please stick to calling `naive.mstump` for the unit tests since it gives slightly better numerical precision than `stumpy.mstump`. Then again, considering that your original input time series only provided precision up to one decimal place, we should only expect `mstump`, `mmoitfs`, and `match` to be accurate to one decimal place.

SaVoAMP:
I think I've got the problem why the while loop is not ending!

After debugging my way through `mmotifs` when calling the naive version of `mstump`, I get the following values for the parameters:
`k=0`
`subspace_k=1`
`motif_idx=10` and
`nn_idx[0]=3`

That means that we hand over the second (`subspace_k=1`, zero based) time series and the query that can be found in this time series beginning at the index location `motif_idx=10` with length `m`. And here the problem arises, I guess.
When calling `match` we expect to find the self match in the first array. That means, that we expect to find `query_matches=[[0.0, 10], ...]`. In the second array we would then expect to find the distance to the nearest neighbor and its position. That means, that we want to have something like this after calling the `match`-function: `query_matches=[[0.0, 10], [distance_to_nearest_neighbor, 3], ...]`. 
But when taking a look at the corresponding distance profile `D` in the `match`-function, we see the following:

image

As you can see, there is a `0.00000` distance on two positions and not only on the position `10` of the self match. That means that the nearest neighbor is as far away as the motif is to itself. Now, when debugging through line 431 in `motifs.py` (in the `match`-function) we try to find the minimum distance of the motif on position `10`. And this distance should be the self-match and therefore also be at position 10. But since the distance to the nearest neighbor at position '3' is exactly the same, position `3` is chosen instead of the self-match (since it comes earlier in the distance array). We get `candidate_idx = np.argmin(D)=3` and therefore the function returns the following to `mmotifs.py`: `matches=[[0.0, 3]]` and not `[[0.0, 10]]` as expected. Then, back in `mmotifs.py` we do not fulfill the if-condition, since `len(query_matches) = min_neighbors` and not bigger than. Thus the exclusion zone is set to the index position `3` in lines 203 and 204 of `mmotifs.py` and not to the position of the motif (self-match). If the next motif index is now selected, the motif at position '10' is selected again, since it was not eliminated. And the whole process starts again... Therefore, the while loop does not end.

seanlaw:
@SaVoAMP I think we've uncovered a fundamental flaw in `stumpy.match`. Firstly, given that `stumpy.match` takes `Q` and `T` as inputs, it has no knowledge that `Q` may actually be a subsequence that is extracted from `T` and nor should it ever make that assumption. In other words, `stumpy.match`, which can be used independent of `stumpy.motifs`, basically assumes that `Q` and `T` come from different time series. Based on the inputs to `stumpy.match`, I don't think any of the parameters can be used to infer that `Q` comes from `T`. Thus, it might make sense to add a single parameter called `query_idx=None` to `stumpy.match`, where we can enforce that the subsequences at `query_idx` must be used as the first "self-match/trivial-match". Thus, most people who use `stumpy.match` directly likely wouldn't care about `query_idx` but for `stumpy.motifs` and `stumpy.mmotifs`, we'd use it. And maybe we'd update `stumpy.match` to:

```
    if query_idx is not None:
        candidate_idx = query_idx
    else:
        candidate_idx = np.argmin(D)

    while (
        D[candidate_idx] <= atol + max_distance
        and np.isfinite(D[candidate_idx])
        and len(matches) < max_matches
    ):
        matches.append([D[candidate_idx], candidate_idx])
        core.apply_exclusion_zone(D, candidate_idx, excl_zone, np.inf)
        candidate_idx = np.argmin(D)
```

I haven't thought this through fully and I'm sure that I'm missing something but this is how I'm leaning. This certainly seems like something that we need to address first and we should not rush this decision because it is an important one!

SaVoAMP:
I've tested it and it seems to work but another problem arises. However, for `candidate_idx=10` no match is returned at all, not even the self-match. Therefore an exception is thrown in line 204 in `mmotifs.py` when trying to set an exclusion zone. I think this is because of the `atol` parameter of `mach`. What is this parameter needed for anyway? I think we need to make sure that either way the self-match is found, right?

I used the following lines to make sure that the candidate index is the same as the query index:
   ```
 if query_idx is None:
        candidate_idx = np.argmin(D)
    else:
        candidate_idx = query_idx
        query_idx = None
```

It's the same that you suggested, but I think we should set the `query_idx` to `None` afterwards.

**Comment by seanlaw** [comment]
If you look at the `while` loop, you'll see that `atol` is added to `max_distance` before comparing this to the candidate distance. Normally, you would simply compare `if D[candidate_idx] <= max_distance` but due to the limited numerical precision, we have allowed some very small "tolerance" (this is borrowed from `numpy` https://numpy.org/doc/stable/reference/generated/numpy.isclose.html) in the `max_distance` to allow for this. So, instead, we compare the candidate distance to the `max_distance` plus some small tolerance, `atol`, which means `if D[candidate_idx] <= max_distance + atol`. This is pretty common in numerical computing due to the fact that we get rounding errors, etc. 

> I think we need to make sure that either way the self-match is found, right?

Yes, for matrix profiles derived from a self-join, we should always return the self-match first. `query_idx=10` really serves two purposes:
1. When `query_idx is not None` then this automatically implies a self-join and not an AB-join
2. When `query_idx is not None` then we should set the first `candidate_idx` to `query_idx`

 However, for AB-joins, `query_idx` should always be set to `None` and this (`query_idx=None`) should be the default behavior.

**Comment by seanlaw** [comment]
> I used the following lines to make sure that the candidate index is the same as the query index:
```
if query_idx is None:
     candidate_idx = np.argmin(D)
 else:
     candidate_idx = query_idx
     query_idx = None
```

I don't understand why this would be any different or better. Why would you need to set `query_idx=None` after `candidate_idx=query_idx`? In my original code (to be added to `stumpy.match`:

```
    if query_idx is not None:
        candidate_idx = query_idx
    else:
        candidate_idx = np.argmin(D)

    while (
        D[candidate_idx] <= atol + max_distance
        and np.isfinite(D[candidate_idx])
        and len(matches) < max_matches
    ):
        matches.append([D[candidate_idx], candidate_idx])
        core.apply_exclusion_zone(D, candidate_idx, excl_zone, np.inf)
        candidate_idx = np.argmin(D)
```

`query_idx` is only used once by `candidate_idx` if and only if `query_idx is not None` and, even when it is used, it is only used once. `query_idx` is not used inside of the `while` loop.

> However, for candidate_idx=10 no match is returned at all, not even the self-match.

I don't understand this. If you add a new `query_idx` parameter to `stumpy.match` (along with the snippet of code above):

```
def match(
    Q,
    T,
    M_T=None,
    Σ_T=None,
    max_distance=None,
    max_matches=None,
    atol=1e-8,
    normalize=True,
    p=2.0,
    query_idx=None  # ADDED THIS NEW PARAMETER
):
```

And then, in `mmotifs()` you call `match` with:

```
query_matches = match(
            Q=T[subspace_k, motif_idx : motif_idx + m],
            T=T[subspace_k],
            M_T=M_T[subspace_k],
            Σ_T=Σ_T[subspace_k],
            max_matches=max_matches,
            max_distance=max_distance,
            atol=atol,
            normalize=normalize,
            p=p,
            query_idx=motif_idx  #ADDED THIS LINE
        )
```

Then the first match MUST be the self-match, right? I'm assuming all default parameters for everything else.

> Therefore an exception is thrown in line 204 in mmotifs.py when trying to set an exclusion zone

I'm guessing that the exception is due to the fact that there is NOTHING in `query_matches` but it should not be due to `atol`. `atol` only makes it EASIER to find a match. Can you please provide the exact code and inputs for recreating your problem?

**Comment by SaVoAMP** [comment]
> I don't understand why this would be any different or better. Why would you need to set query_idx=None after candidate_idx=query_idx? In my original code (to be added to stumpy.match:

You're right, sorry, I had a thinking error. But that shouldn't be the problem why there is no match returned.

>  def match(
>     Q,
>     T,
>     M_T=None,
>     Σ_T=None,
>     max_distance=None,
>     max_matches=None,
>     atol=1e-8,
>     normalize=True,
>     p=2.0,
>     query_idx=None  # ADDED THIS NEW PARAMETER
> ):
> 
> And then, in mmotifs() you call match with:
> 
> query_matches = match(
>             Q=T[subspace_k, motif_idx : motif_idx + m],
>             T=T[subspace_k],
>             M_T=M_T[subspace_k],
>             Σ_T=Σ_T[subspace_k],
>             max_matches=max_matches,
>             max_distance=max_distance,
>             atol=atol,
>             normalize=normalize,
>             p=p,
>             query_idx=motif_idx  #ADDED THIS LINE
>         )
> 
> Then the first match MUST be the self-match, right? I'm assuming all default parameters for everything else.

Yes, that is exactly the same that I did. But strangely there is nothing returned for `query_matches`.

**Comment by SaVoAMP** [comment]
The problem arises in the while-loop of `match()` in `motifs.py` since `D[candidate_idx] <= atol + max_distance` doesn't seem to be fulfilled for `candidate_idx=10` and therefore no match is returned to the `mmotifs`-function.
The distance profile `D` has the following rounded values:
image
But if we take a look at the more accurate values we get the following:
image
And if we now pick out the value at the position `10`, we see that `1.5*10⁻7 = 15*10⁻8` which is larger than `atol=1*10⁻8`. So the problem actually seems to be in this line of code. I just don't understand why the number here is so relatively high since this is the self-match in the distance profile.

> I'm guessing that the exception is due to the fact that there is NOTHING in query_matches but it should not be due to atol. atol only makes it EASIER to find a match. Can you please provide the exact code and inputs for recreating your problem?

I used the code that is inside the `test_mmotifs.py` script to understand why the loop doesn't end. Specifically I was debugging my way through the `test_motifs_multidimensional_with_default_parameters()`-function using every parameters and the time series exactly as they can be found in the script. I best push everything, including `query_idx`. Then you can see everything.

**Comment by seanlaw** [comment]
> I used the code that is inside the test_mmotifs.py script to understand why the loop doesn't end. Specifically I was debugging my way through the test_motifs_multidimensional_with_default_parameters()-function using every parameters and the time series exactly as they can be found in the script. I best push everything, including query_idx. Then you can see everything.

Thank you. Let me take a look

**Comment by seanlaw** [comment]
@SaVoAMP While the math behind matrix profiles are sound, I think that the fundamental issue that we continue to deal with is numerical precision. In `stumpy.match`, we are calling `core.mass`, which computes our distance profiles and, instead of computing the z-norm distances in a naive way, it uses some fancy dot products, means, and stddevs values to save on some of the computation. Unfortunately, for exceptionally small window sizes (i.e., like the `m=3` that you are using), the loss in precision in computing the distance profile is dramatically amplified. I know that we use a small window size for unit testing as it helps keep our tests small/fast. However, in this case, I think that using `m=3` is causing us a lot more headache than it is worth and it may not be at all representative of real-world datasets. Unfortunately, I just don't see any good way to improve the precision for small `m`, save for checking the size of `m` and then switching to the naive method for computing the distance. For now, may I recommend keeping the same `T` and using `m=4` or `m=5` instead?

**Comment by seanlaw** [comment]
> All right. I should change `T` so that the values don't end with `.0`. Then I'll also try to make my desired two dimensional motif of length 3 to a motif of length four. Hopefully everything is working afterwards.

Thank you. Hopefully, this is a lesson-learned for me as well to always remember that small window sizes are a source of amplified imprecision 😄

**Comment by SaVoAMP** [comment]
Should I add the `query_idx` parameter to the docstring or are you doing this?

**Comment by seanlaw** [comment]
> Should I add the `query_idx` parameter to the docstring or are you doing this?

If you don't mind, would you mind attempting it and then I can provide feedback? It should likely reference something about "being used when the matrix profile is a self-join"?

**Comment by SaVoAMP** [comment]
`Flake8 . ` throws an error in `__init__.py` that says `./__init__.py:1:1: D104 Missing docstring in public package` but I didn't change anything there.  Do I have to add `from .mmotifs import mmotifs` and `from .aamp_mmotifs import aamp_mmotifs`?

**Comment by seanlaw** [comment]
> `Flake8 . ` throws an error in `__init__.py` that says `./__init__.py:1:1: D104 Missing docstring in public package` but I didn't change anything there. Do I have to add `from .mmotifs import mmotifs` and `from .aamp_mmotifs import aamp_mmotifs`?

No. I think the problem is that you are running `flake8 .` inside of the `stumpy` subdirectory (i.e., inside the same directory where `__init__.py` is. Instead, you should run `flake8 .` in the parent directory (i.e., where the `setup.py` file is). Inside of that directory, there is a `.flake8` file that will help you ignore that `flake8` error

> By the way, I think we also have to add `query_idx` to `aamp_match`.

Yes, please.

File: stumpy/motifs.py [hunk 3]
```diff
@@ -365,6 +366,11 @@ def match(
         The absolute tolerance parameter. This value will be added to `max_distance`
         when comparing distances between subsequences.
 
+    query_idx : int, default None
+        This parameter is used when the matrix profile is a self-join.
```
seanlaw, line 370: This is great! I might consider elaborating or providing a little bit more context like:

```
This is the index position along the time series, `T`, where the query subsequence, `Q`, is located. `query_idx` should only be used when the matrix profile is a self-join and should be set to `None` for matrix profiles computed from AB-joins. If `query_idx` is set to a specific integer value, then this will help ensure that the self-match will be returned first.
```

seanlaw:
> By the way, I think we also have to add query_idx to aamp_match.

Are you also adding `query_idx` to `aamp_mmotifs`?

SaVoAMP:
>Are you also plannging to add query_idx to aamp_mmotifs? I am thinking if we need to add it to aamp_motifs and motifs as well for consistency?

I neither added it to `mmotifs` nor `aamp_mmotifs`. I only used the parameter inside the `mmotifs` function. But you are right, I forgot to pass the parameter in `aamp_mmotifs` to the `aamp_match` call.

I added the `query_idx` parameter to the docstring and the definition of the match-functions in `aamp_motifs` and `motifs`. But I think that it's not necessary in `mmotifs` and `aamp_mmotifs` since we only pass the parameter to the match-function and the user shouldn't have to set this parameter, I think.

SaVoAMP:
>Right now, when we call the multi-dimensional versions, the first match is "guaranteed" to be the self-match but this is not guaranteed in the 1-D versions.

Ohh, yes, you're right! Totally forgot.. 
Of course, the parameter must be passed to the `match` functions in `motifs` and `aamp_motifs` too! Should I do this?
But if I understood you correctly, the parameter is also supposed to be an input parameter in the `motifs` and `aamp_motifs` function, so that in the one-dimensional case the user himself specifies whether it is a self-join or an AB-join?

seanlaw:
> Of course, the parameter must be passed to the match functions in motifs and aamp_motifs too! Should I do this?

Yes, please! Thanks in advance

> But if I understood you correctly, the parameter is also supposed to be an input parameter in the motifs and aamp_motifs function, so that in the one-dimensional case the user himself specifies whether it is a self-join or an AB-join?

I made the realization that, currently, both `motifs` and `aamp_motifs` (1-D version) technically assume a self-join. This is evidenced by the fact that the input only accepts a single time series `T`. The only way to support an AB-join is to allow the user to supply an additional time series, `T_B`, which is currently not possible (we may consider this in the future but not now). So, I think we can safely assume that both `motifs` and `aamp_motifs` only support matrix profiles derived from self-joins and so `query_idx` is not needed as an input parameter to these functions and, instead, the `query_idx` is assumed to be the same as the `candidate_idx` inside of `_motifs`/`_aamp_motifs` (i.e., `query_idx=candidate_idx` needs to get added to these private functions, not as input parameters but when they call `match`/`aamp_match`).

Does that make sense?

SaVoAMP:
Yes, that makes sense!

Just added the parameter to the `match` calls in the `_motifs` and `_aamp_motifs` functions.

SaVoAMP:
Shouldn't we allow `max_matches` to be an array or a list of the size of `max_motifs` like we have done with `cutoffs`?
I'm asking since I'm trying to find out how well the motif discovery using matrix profiles is suited to automatically annotate multidimensional time series data. I wrote a script that calculates the matrix profile for each new subsequence length and tries to find the motifs with `mmotifs`. I have got different sublabels that tell me that I'm searching for 5 sublabels of one type and 7 of another type for example.  Then I get 2 different subsequence lengths (for example by computing the average of each sublabel type). If the two subsequences are different, I only have to compute the matrix profile and find one motif with `mmotifs` both times. But if they have the same length, the matrix profile should only be computed once but I want to find two motifs with `mmotifs`. Here I've got the problem that I can't tell `mmotifs` that I'm searching for 5 matches for the first motif and for 7 matches for the second motif.

Wouldn't it make more sense to also allow `max_matches` to be an array but set a default array that only contains the number 10, so that nothing changes to the user if he doesn't want to have a different number of matches for each motif? We could do exactly the same as we have done with `cutoffs` and therefore, if only one value is set, this value could be applied to every dimension. 🤔

seanlaw:
> Shouldn't we allow `max_matches` to be an array or a list of the size of `max_motifs` like we have done with `cutoffs`? I'm asking since I'm trying to find out how well the motif discovery using matrix profiles is suited to automatically annotate multidimensional time series data. I wrote a script that calculates the matrix profile for each new subsequence length and tries to find the motifs with `mmotifs`. I have got different sublabels that tell me that I'm searching for 5 sublabels of one type and 7 of another type for example. Then I get 2 different subsequence lengths (for example by computing the average of each sublabel type). If the two subsequences are different, I only have to compute the matrix profile and find one motif with `mmotifs` both times. But if they have the same length, the matrix profile should only be computed once but I want to find two motifs with `mmotifs`. Here I've got the problem that I can't tell `mmotifs` that I'm searching for 5 matches for the first motif and for 7 matches for the second motif.
> 
> Wouldn't it make more sense to also allow `max_matches` to be an array but set a default array that only contains the number 10, so that nothing changes to the user if he doesn't want to have a different number of matches for each motif? We could do exactly the same as we have done with `cutoffs` and therefore, if only one value is set, this value could be applied to every dimension. 🤔

@SaVoAMP I hope that I am understanding your intent correctly but what you are describing (in terms of computing multiple multi-dimensional matrix profiles for different window sizes) sounds more complex than the average use case. Instead of changing `mmotifs`, I find myself wondering if you are able to get what you need in a post-processing step after `mmotifs` is called? 

> Here I've got the problem that I can't tell `mmotifs` that I'm searching for 5 matches for the first motif and for 7 matches for the second motif.

In this case, I would call `mmotifs` with `max_matches=7` (or whatever the largest number needs to be) and then post-process the results afterward by removing the parts that are not needed or are too much. Even if you changed/modified `max_matches` to be variable for each motif, you'd still need to post-process the results anyways because the width of, say, `motif_distances` would still (possibly) be as wide as `7` matches for all motifs (i.e., the ones where you request `max_matches=5` will have `np.inf` padded to it). I do not wish to return a jagged/ragged `motif_distance` as this would cause inconsistencies with `motifs` (1-D).

I understand that it may seem like "less things to compute" by providing an array/list of `max_matches` but that gain is likely only very, very small. I'm not sure it is something that we'd want to optimize. Again, I'm not opposed to your idea but I am trying to reduce/minimize the code complexity especially when the computational gain isn't substantial.

SaVoAMP:
>I do not wish to return a jagged/ragged motif_distance as this would cause inconsistencies with motifs (1-D).

I'm not sure if I understand what you mean. Won't I get results of different lengths anyway if I don't set the `max_distance` parameter to `np.inf`? For every motif I get as many matches as are lying in `max_distance`. But then there may be different numbers of matches from motif to motif 🤔  or am I missing something?

seanlaw:
> I'm not sure if I understand what you mean. Won't I get results of different lengths anyway if I don't set the `max_distance` parameter to `np.inf`? For every motif I get as many matches as are lying in `max_distance`. But then there may be different numbers of matches from motif to motif 🤔 or am I missing something?

Yes, this is true but at the end of `mmotifs`, we convert our jagged list into a perfectly rectangular array:

```
motif_distances = core._jagged_list_to_array(
        motif_distances, fill_value=np.nan, dtype=np.float64
    )
```

So, you would need to post-process this array. This is the same thing that happens at the end of `motifs` (1-D version)

SaVoAMP:
ahh and that's what you meant with "Even if you changed/modified max_matches to be variable for each motif, you'd still need to post-process the results anyways because the width of, say, motif_distances would still (possibly) be as wide as 7 matches for all motifs". I also would have to exclude those values that are `nan`.

seanlaw:
Exactly! We had made a decision earlier on with `motifs` to return a rectangular `numpy` array so that it would allows the user to leverage `numpy` functions. We try to limit our use of Python lists as return types.

SaVoAMP:
I added the test functions with the `cutoffs` parameter with the aim to get into the `if`-condition and return `break` since this is the only thing that is not covered yet:
image
image
However, if `motif_value > cutoffs[k]` I'm not stepping into the `break` but the program is stepping out of the `while`-loop and so is doing the same that is expected. I don't understand why 🤔

seanlaw:
> However, if motif_value > cutoffs[k] I'm not stepping into the break but the program is stepping out of the while-loop and therefore is doing what is expected

I'm assuming that you've confirmed the values of `motif_value` and `cutoffs[k]` (i.e., print them out) and ensured that `motif_value` is indeed greater than `cutoffs[k]`?

SaVoAMP:
Yes, I tried to set different values for 'cutoffs' but wasn't able to get into the 'brake' while debugging. The way it is now, we should step into the 'if'-condition since 'motiv_value=np.inf'. 
But I don't think that it is a problem with 'mmotifs' sine it I'd behaving as it should. So when proving the conditions in 'if' and one condition is true, the debugger is jumping out of the 'while' loop as expected, but I am not able to get into the 'brake' line.

seanlaw:
@SaVoAMP Can you please provide me with the set of input parameters for `mmotifs` that you believe should allow you to enter the `break` condition?

SaVoAMP:
image
I'm able to get to `break` if I place a `print`-function before it (without doing anything else).
I only call the function `test_motifs_multidimensional_more_motifs_when_cutoffs_is_set()
` in `test_mmotifs.py`. Once `motif_value=np.inf` the `if`-condition is fulfilled, thus I traverse into the condition and reach the `break`. Without the `print` it's just jumping over the `break` keyword, but with it I'm able to get to the `break`. Nonetheless the `break` is working even though I can't reach that exact line directly. The only problem is that this line doesn't pass the coverage tests, neither in `test_mmotifs.py` nor in `test_aamp_mmotifs.py`.

seanlaw:
@SaVoAMP In this rare case, please add `  # pragma: no cover`:

```
if (
            motif_value > cutoffs[k]
            or not np.isfinite(motif_value)
            or (isinstance(max_distance, float) and motif_value > max_distance)
        ):  # pragma: no cover
            break
```

I don't usually recommend this but this is an exceptional case that requires it.

SaVoAMP:
```
====================== 1271 passed in 2050.37s (0:34:10) =======================
Name    Stmts   Miss  Cover   Missing
-------------------------------------
-------------------------------------
TOTAL   10871      0   100%

76 files skipped due to complete coverage.

```

It looks as if everything is covered now.

seanlaw:
> It looks as if everything is covered now.

Awesome! Maybe you can focus on the tutorial now and I will find some time to thoroughly review the tests and code changes. Thank you for all of your efforts!

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
+import numpy as np
+import numpy.testing as npt
+import naive
+
+from stumpy.aamp_mmotifs import aamp_mmotifs
+from stumpy import config
+
+
+# These are tests for non-normalized multidimensional motif discovery
```
seanlaw, line 9: Please remove this comment

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
...
+
+
+def test_motifs_multidimensional_one_motif_all_dimensions():
+    # Find the two-dimensional motif pair
+
+    # Arrange
```
seanlaw, line 15: Please remove comment

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
...
+    motif_distances_expected = np.array([[0.0, 2.87778559]])
+    motif_indices_expected = np.array([[0, 5]])
+    motif_subspaces_expected = [np.array([2, 1])]
+    motif_mdls_expected = [np.array([244.0, 260.67970001, 279.86313714, 281.35940001])]
+
+    # Act
```
seanlaw, line 21: Please remove this comment

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
...
+    T = np.array(
```
seanlaw, line 22: If this array is the same in all tests, can you please follow other tests and declare the array once at the beginning like (see `tests/test_stumpy.py`: 

```
test_data = [np.array([9, 8100, -60, 7], dtype=np.float64)]
```
and then you can pass it to your test with:

```
@pytest.mark.parametrize("T", test_data)
def test_aamp_mmotifs_2D_motif_pair(T):
    .
    .
    .
    P, I = naive.maamp(T, m, excl_zone)
```

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
...
+import numpy.testing as npt
+import naive
+
+from stumpy.aamp_mmotifs import aamp_mmotifs
+from stumpy import config
+
+
+# These are tests for non-normalized multidimensional motif discovery
+
+
```
seanlaw, line 11: For the first unit test, can you add a test that simply executes with all default values?

```
def test_aamp_mmotifs():
    .
    .
    .
    motif_distances, motif_indices, 
    motif_subspaces, motif_mdls = 
    aamp_mmotifs(T, P, I)
```

seanlaw: The reasoning is that experienced developers will often look to the unit tests for the simplest reproducible example of how to use the function. So, we'll want to provide a simple example with default input parameters.

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
...
+    npt.assert_array_almost_equal(motif_distances_expected, motif_distances)
+    npt.assert_array_almost_equal(motif_indices_expected, motif_indices)
+    npt.assert_array_almost_equal(motif_subspaces_expected, motif_subspaces)
+    npt.assert_array_almost_equal(motif_mdls_expected, motif_mdls)
+
+
+def test_motifs_multidimensional_with_one_motif_max_matches_none():
+    # Find the motif pair while setting 'max_matches=None'
+
+    # Arrange
```
seanlaw, line 148: Please remove comment

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
...
+                0.5,
+                4.3,
+            ],
+        ]
+    )
+    m = 4
+
+    excl_zone = int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
+    P, I = naive.maamp(T, m, excl_zone)
+    motif_distances, motif_indices, motif_subspaces, motif_mdls = aamp_mmotifs(
```
seanlaw, line 201: Please append `_cmp` (see above)

seanlaw (changes requested):
@SaVoAMP I think things look pretty good. I provided some comments for you to consider. All comments for `test_aamp_mmoitfs.py` should also apply to `test_mmotifs.py`. Please, let me know if you have any questions, comments, or concerns.

tests/test_aamp_mmotifs.py
SaVoAMP: I tried to do this, but then I didn't get results since we get empty lists/arrays.

tests/test_aamp_mmotifs.py
SaVoAMP: >If this array is the same in all tests, can you please follow other tests and declare the array once at the beginning like (see tests/test_stumpy.py:

The array has to be the same every time we are searching for the first motif. But once we discover the top two motifs we get the following: 
```
motif_mdls_ref = [
        np.array([244.0, 260.67970001, 279.86313714, 281.35940001]),
        np.array([254.33985, 260.67970001, 279.86313714, 291.20703549]),
    ]
```

seanlaw: I'm not sure I understand what you mean. I'm just asking whether `T` is the same in all of the unit tests? If so, then please define `T` once at the top of the test file as `test_data` and then pass `test_data` into each of your tests via:
```
@pytest.mark.parametrize("T", test_data)
def test_aamp_mmotifs(T):
```

SaVoAMP: Ohh, I'm sorry! 
Sometimes it's a bit difficult to tell which line you mean 😄

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,351 @@
...
+
+    # Assert
+    npt.assert_array_almost_equal(motif_distances_expected, motif_distances)
+    npt.assert_array_almost_equal(motif_indices_expected, motif_indices)
+    npt.assert_array_almost_equal(motif_subspaces_expected, motif_subspaces)
+    npt.assert_array_almost_equal(motif_mdls_expected, motif_mdls)
+
+
+def test_motifs_multidimensional_with_one_motif_max_matches_none():
+    # Find the motif pair while setting 'max_matches=None'
```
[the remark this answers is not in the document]
SaVoAMP: Do you mean that I should delete all  comments or only the `assert` and so on?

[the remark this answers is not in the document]
seanlaw: Sorry, I mean that the comment should be combined into one with the name of the test. The convention is that the name of the test function should clearly represent which stumpy function is being tested and then what specifically about that function is being tested. In most of your cases, it is either `mmotifs` or `aamp_mmotifs` that is being tested and, sometimes, along with different parameters. And so that should be reflected clearly in the test function name where possible. The fact that you have added a comment directly below the function suggests that the function name isn't sufficiently named.

tests/test_aamp_mmotifs.py
seanlaw: Maybe I'm missing something but I simply did:

```
import numpy as np
from stumpy.aamp_mmotifs import aamp_mmotifs
import stumpy
T = np.random.rand(500).reshape(5, 100)
m = 5
mp, indices = stumpy.mstump(T, m, normalize=False)
aamp_mmotifs(T, mp, indices)
```

I'm not saying that these are the most interesting results but it seems to be sufficient.

tests/test_aamp_mmotifs.py
seanlaw: @SaVoAMP In case it matters, in Github, I usually assign the comment to the precise line of code (or start of the code chunk) that I am referring to.  Github usually displays the comment directly under the "line of interest" and, in this case, I chose the starting line of:

```
T = np.array(
```

Although, based on what is displayed, I can see why you thought I was referring to the `motif_mdls_expected =` line now.

tests/test_aamp_mmotifs.py
seanlaw: Great. Thank you!

seanlaw:
@SaVoAMP Thank you for putting in all of this work. I hope that I'm not coming across as too "nit-picky" (I probably am 😄). However, since you had expressed that you are relatively new to (some?) of this software development process, I wanted to at least provide the appropriate feedback and "thinking" regarding best practices. While I could have accepted your PR "as is", I didn't want to take away your opportunity to learn, try, and gain more "real world" exposure on the topic. As you can see, this part is often more art than science but I'm sure you can understand why it takes me a long time (working by myself) to release new features as I care deeply about maintaining the high quality of that last Kilometer. Things like unit testing and documentation are often the hardest part of the development process that gets heavily neglected because it isn't glamorous work (like adding a new feature) and you've really excelled!  However, unit testing and documentation are an integral (if not the most important) part of open source software so please know that you are really helping STUMPY uphold its standards as well as its commitment to its community! Thank you

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,248 @@
...
+                1.3,
+                0.5,
+                4.3,
+            ],
+        ]
+    )
+]
+
+
+def test_aamp_mmotifs_default_parameters():
```
seanlaw, line 54: Perfect!

seanlaw: I don't know if you notice this but after we take out the chunk of code for `T`, the test function(s) are all super simple and it is easy to digest what is happening. This little bit of refactoring, cleans up the code and everything can be read without scrolling. This is exactly how a "good" unit test should look and this is about as close to "perfection" as I could expect, which makes my job as a reviewer really, really easy. The rest of the unit tests look equally as great too! Excellent work!

"Future Sean" thanks you as well. 😄

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,248 @@
...
+    ) = aamp_mmotifs(T, P, I, max_distance=np.inf, max_matches=None)
+
+    npt.assert_array_almost_equal(motif_distances_ref, motif_distances_cmp)
+    npt.assert_array_almost_equal(motif_indices_ref, motif_indices_cmp)
+    npt.assert_array_almost_equal(motif_subspaces_ref, motif_subspaces_cmp)
+    npt.assert_array_almost_equal(motif_mdls_ref, motif_mdls_cmp)
+
+
+@pytest.mark.parametrize("T", test_data)
+def test_aamp_mmotifs_exactly_one_motif_pair_k_chosen(T):
```
seanlaw, line 137: Maybe a small suggestion that we could even shorten and make the function name:

```
def test_aamp_mmotifs_max_motifs_1_max_matches_2_k_1(T):
```
This way, just from reading the function name, I can anticipate what parameters matter in this test? What do you think? I'm good either way.

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,248 @@
...
+    ) = aamp_mmotifs(T, P, I, max_distance=np.inf)
+
+    npt.assert_array_almost_equal(motif_distances_ref, motif_distances_cmp)
+    npt.assert_array_almost_equal(motif_indices_ref, motif_indices_cmp)
+    npt.assert_array_almost_equal(motif_subspaces_ref, motif_subspaces_cmp)
+    npt.assert_array_almost_equal(motif_mdls_ref, motif_mdls_cmp)
+
+
+@pytest.mark.parametrize("T", test_data)
+def test_aamp_motifs_max_distance_max_matches_none(T):
```
seanlaw, line 111: I think it should say `_mmotifs` (missing an "m")

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,248 @@
...
+    ) = aamp_mmotifs(T, P, I, max_distance=np.inf, max_matches=2, k=1)
+
+    npt.assert_array_almost_equal(motif_distances_ref, motif_distances_cmp)
+    npt.assert_array_almost_equal(motif_indices_ref, motif_indices_cmp)
+    npt.assert_array_almost_equal(motif_subspaces_ref, motif_subspaces_cmp)
+    npt.assert_array_almost_equal(motif_mdls_ref, motif_mdls_cmp)
+
+
+@pytest.mark.parametrize("T", test_data)
+def test_aamp_mmotifs_more_motif_pairs_when_cutoffs_is_set(T):
```
seanlaw, line 161: Minimally, I would say `cutoffs_3` in the function name

tests/test_aamp_mmotifs.py
```diff
@@ -0,0 +1,248 @@
...
+    )
+
+    npt.assert_array_almost_equal(motif_distances_ref, motif_distances_cmp)
+    npt.assert_array_almost_equal(motif_indices_ref, motif_indices_cmp)
+    npt.assert_array_almost_equal(motif_subspaces_ref, motif_subspaces_cmp)
+    npt.assert_array_almost_equal(motif_mdls_ref, motif_mdls_cmp)
+
+
+@pytest.mark.parametrize("T", test_data)
+def test_aamp_mmotifs_exactly_two_motif_pairs(T):
```
seanlaw, line 217: Maybe `_max_motifs_2`

tests/test_mmotifs.py
```diff
@@ -0,0 +1,203 @@
...
+                0.5,
+                4.3,
+            ],
+        ]
+    )
+]
+
+
+@pytest.mark.parametrize("T", test_data)
+def test_mmotifs_with_deafault_parameters(T):
```
seanlaw, line 55: "default" instead of "deafault" (extra "a")

seanlaw (changes requested):
@SaVoAMP We are very, very close to having your PR merged! Just a few minor comments and suggestions for you to consider.

SaVoAMP:
Thank you for your support!

>However, since you had expressed that you are relatively new to (some?) of this software development process, I wanted to at least provide the appropriate feedback and "thinking" regarding best practices.

Yes, as I've already mentioned these are my first pull requests ever. I started my major without any prior experience with computers (I didn't even had one and bought my first computer ever in the third semester 😆) as my specialization in "high school" was social work. And although my current major included some computer science modules (C# Programming, data structures and algorithms, very basic software engineering without git) the focus is still on Math and Physics, we are only required to do those modules to be able to do numerical simulations with tools like COMSOL Multiphysics and the like.

Since the start of my mandatory apprenticeship and successive bachelors thesis at the Fraunhofer Institute I'm working on data science projects and had to learn Python, Linux and git from scratch over the last months.

Therefore I especially value your professional and detailed feedback.

seanlaw:
> Yes, as I've already mentioned these are my first pull requests ever. I started my major without any prior experience with computers (I didn't even had one and bought my first computer ever in the third semester 😆) as my specialization in "high school" was social work. And although my current major included some computer science modules (C# Programming, data structures and algorithms, very basic software engineering without git) the focus is still on Math and Physics, we are only required to do those modules to be able to do numerical simulations with tools like COMSOL Multiphysics and the like.

😮 🤯 

I hope you realize that you are well on your way in your professional career! You've picked things up really, really quickly and, if this is your first time, I can only imagine how successful you will be when all of this becomes second nature to you. 😄 I've given you only a bit of remote guidance and you've been resourceful and proactive. It's been a pleasure working with you!

> I also brush up my english skills along the way

I was actually thinking about this the other day. Your English is excellent and it is not lost on me that you are needing to push through the language barrier in addition to learning brand new software development skills as well as balancing your responsibilities as a student. Really amazing! I sincerely appreciate your dedication

seanlaw:
> I'm jokingly referring to this since I was surprised about you answering my questions on weekends. 

STUMPY is a labor of love and 100% volunteer. The unsolicited advice is "don't be like me" 😅 . As a former academic (scientific researcher), I have accrued some really bad habits and you should **not** use me as a "good example". I am lucky that I love STUMPY and so I don't feel that this is "work". But you must find your balance and figure out what is important to you and what you are passionate about.

seanlaw:
@SaVoAMP Not trying to rush you but will you let me know when it is appropriate for me to review this PR again?

SaVoAMP:
Ohh, I'm sorry! I was just about to ask you the same question 😆 
I've already tried to change what you suggested. It's in the last commit.

seanlaw:
Great! Glad I asked. I will find some time to review it

seanlaw:
@SaVoAMP Congratulations on another wonderful PR being merged. Thank you for this wonderful contribution!

Issue #518, opened 2022-01-13
Add Top-K Motif Discovery Tutorial


seanlaw:
Demonstrate the use of `stumpy.motifs` and `stumpy.match`

seanlaw:
In the paper, it says:

> we tested our method on an electrical load measurement dataset [21]. The dataset consists of electrical load measurements for individual appliances (and an aggregated load) from households in United Kingdom. Five appliances are considered: fridge-freezer, freezer, tumble dryer, dishwasher,
and washing machine. The data was collected from April 19, 2014 to May 15, 2014, where the length is 17,000. The subsequence length was set to 4 hours. 

Cross referencing the cleaned version of the [REFIT dataset](https://pureportal.strath.ac.uk/en/datasets/refit-electrical-load-measurements), it was found that only "CLEAN_House3.csv" [in the "Access Dataset > REFIT Power Data - 111215" ](https://pureportal.strath.ac.uk/files/45646018/REFITPowerData111215.7z) contained the aforementioned appliances. It was not possible to reproduce the paper's data preprocessing steps but this should be sufficient:

```
%matplotlib inline

import matplotlib.pyplot as plt
import pandas as pd

plt.style.use('https://raw.githubusercontent.com/TDAmeritrade/stumpy/main/docs/stumpy.mplstyle')

start_date, end_date = '2014-04-19', '2014-05-15'
df = pd.read_csv("CLEAN_House3.csv")
df["Time"] = pd.to_datetime(df["Time"])
colnames = {
    "Appliance1": "Toaster",
    "Appliance2": "Fridge-Freezer",
    "Appliance3": "Freezer",
    "Appliance4": "Tumble Dryer",
    "Appliance5": "Dishwasher",
    "Appliance6": "Washing Machine",
    "Appliance7": "Television",
    "Appliance8": "Microwave", 
    "Appliance9": "Kettle",
}
df = (df.rename(colnames, axis="columns")
        .loc[:, ["Time",
                 "Fridge-Freezer",
                 "Freezer", 
                 "Tumble Dryer", 
                 "Dishwasher",
                 "Washing Machine"]]
        .query('Time >= @start_date and Time <= @end_date')
        .groupby(pd.Grouper(key='Time', freq="T")).sum()
     )

for colname in df.columns:
    plt.plot(df[colname])
```

SaVoAMP:
Hey!

[...]

PR #557, opened 2022-03-02
[ADD] Added first draft notebook for the multidimensional motif and m…

SaVoAMP:
…atches discovery tutorial

# Pull Request Checklist

- [x] Create a new branch
#518


SaVoAMP:
I was not sure if I am supposed to run black and flake8 on jupyter notebooks too, so I skipped these steps.

seanlaw:
> I was not sure if I am supposed to run black and flake8 on jupyter notebooks too, so I skipped these steps.

Nope. black/flake8 is not needed for notebooks. Thanks for checking

seanlaw:
@SaVoAMP [Said](https://github.com/TDAmeritrade/stumpy/issues/518#issuecomment-1059751004):

>Alright!
My problem is that setting values with zero standard deviation to np.nan doesn't work and I can't explain why. I wrote two if-conditions and when I debug, I also get into the conditions. But when setting the values to np.nan it doesn't work, although the constant_idx vector is occupied. I probably need to take a closer look at the data frame operations here.
> Could this be because the timestamps are now here in the left column and no longer the data points 0, 1, 2, ..., n?
But I think constant_idx is giving me the data point positions since constant_idx=[239, 240, 241, ..., 4650, 4651, 4652]. Is there a simple (data science) trick? I would insert another column that counts the data_points and work with them instead of the timestamps. But that can certainly be done more elegantly.

I will take a look at the notebook (click on the big dark purple button at the top of the page that says "ReviewNB"). Your best bet is to make your changes and push the notebook so that I can directly see what you are saying and we can comment (roughly) inline.

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
```diff
@@ -0,0 +1,241 @@
+{
```
seanlaw, line 1: <p>So, it looks like you are manipulating the values within a column and then you are trying to stuff the results back into the same column of your dataframe. Unfortunately, you cannot do that via <code>df.values[:, i] = X</code> as <code>df[:, i].values</code> only allows you to access a <strong>copy</strong> of the values as a <code>numpy</code> array but, since it's a copy, it does not allow you to write back into the original Pandas dataframe. Would something like this work for what you are trying to achieve?:</p><p> </p><pre class="ql-syntax" spellcheck="false"># m was set to 4 hours
m = 4 * 60


padded_stddev = np.zeros(len(df))&nbsp; # reusuable array filled with zeros
for colname in df.columns:
&nbsp; &nbsp; stddev = core.rolling_nanstd(df.loc[:, colname].values, m)&nbsp; # Compute rolling stddev. Constant regions will have a 0.0 stddev
&nbsp; &nbsp; padded_stddev[-len(constant_mask):] = stddev&nbsp; # Copy stddev into array that is pre-padded with (m-1) zeros
&nbsp; &nbsp; modified_values = df.loc[:, colname].where(padded_stddev != 0.0, np.nan)&nbsp; # Replace original value with np.nan where condition is False
&nbsp; &nbsp; df.assign(colname=modified_values)
</pre><p>Stepping through each line of the for-loop:</p><ol><li>This should look familiar as we are computing the stddev for a particular column. In Pandas, the preferred way to access the column (or Pandas series) by name is via the <code>loc</code> function where the first element corresponds to a list of named indices (in this case, our index is a timestamp) and then followed by a list of column/series names. In our case, we want to grab all rows (i.e., all timestamps) but only a single column at a time, so, <code>df.loc[:, colname]</code> specifies that all rows with  <code>:</code>  and only the column with <code>colname</code> should be analyzed. The <code>.values</code> makes sure to convert that column/series into a copy of a <code>numpy</code> array for that column/series. Then, we pass that copy of values into <code>core.rolling_nanstd</code> (a STUMPY function) and compute the rolling standard deviation. Of course, it is shortened by <code>m-1</code> values and so we'll need to eventually pad it with some additional values</li><li>Here, we use our pre-generated array that has been padded with <code>m-1</code> zeros at the start and then we copy over our standard deviation values to fill in the rest of the array. So, the first <code>m-1</code> values are all zeros and then followed by all of the <code>stddev</code>  values</li><li>Then, we create a new array of <code>modified_values</code> of the <code>df.loc[:, colname]</code> series based on whether the each row of <code>padded_stddev</code>  is equal/not equal to <code>0.0</code>. Note that the Pandas series <code>where</code> function assesses whether the condition provided as the first argument is <code>True</code>  or <code>False</code>. When <code>True</code>, it keeps the value from the Pandas series. However, if the condition is found to be <code>False</code> then it will assign the value in the second argument, which is <code>np.nan</code> in this case. So, for each row of <code>padded_stddev</code> , if that row has a value that is not equal to <code>0.0</code> (i.e., <code>True</code>), then  keep the value from row in the the Pandas series. However, If the row in <code>padded_stddev</code> is equal to <code>0.0</code>  (i.e., <code>False</code>), then replace the value in row of the Pandas series with <code>np.nan</code>. Note that all of this is written out to a new Pandas series called <code>modified_values</code></li><li>Finally, we assign the values of the series in <code>df</code> with column name <code>colname</code>  with the our new Pandas series <code>modified_values</code> </li></ol><p> </p><p>We could have combined steps 3 and 4 into one:</p><pre class="ql-syntax" spellcheck="false">df.assign(colname=df.loc[:, colname].where(padded_stddev != 0.0, np.nan))
</pre><p>but this is a bit harder to digest/consume.</p><p> </p><p>Please ask for any clarification and I would be happy to discuss this further.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

seanlaw:
@SaVoAMP I came across [this paper](https://arxiv.org/pdf/1911.09218.pdf) by Jessica Lin (who has worked with Eamonn Keogh in the past). If you look at Section D, I think that the PAMAP2 dataset (Overview found [here](http://archive.ics.uci.edu/ml/datasets/PAMAP2+Physical+Activity+Monitoring) - Download zip file [here](http://archive.ics.uci.edu/ml/machine-learning-databases/00231/)) may be good for our tutorial. In the paper, they focused on the "ironing" activity and used a 12.5 second (12500 data point) window size, which they claimed allowed them to identify relevant hand motions (from the Y and Z hand coordinates) that are consistent with ironing while, presumably, other coordinates (from the chest and ankles) may have been irrelevant for ironing. 

The dataset is suppose to contain 1.3 million datapoints in length which may be too much for a tutorial but I wonder if we could downsample the data by 10x (i.e., only look at every 10th data point and thereby analyzing only 130K data points for each dimension) and still get be able to convey our point. I wanted to bring it to your attention in case it may be useful.

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
SaVoAMP: Thank you very much! This was pretty useful for my understanding.

SaVoAMP:
> The dataset is suppose to contain 1.3 million datapoints in length which may be too much for a tutorial but I wonder if we could downsample the data by 10x (i.e., only look at every 10th data point and thereby analyzing only 130K data points for each dimension) and still get be able to convey our point. I wanted to bring it to your attention in case it may be useful.

I'm not sure what data set would be better 🤔 Just added your suggestions to eliminate constant regions. Now we have much better results. But I'm not sure why there is not eliminated more of the 'zero'-regions in the Tumble Dryer and Washing Machine data. Maybe we get good results if I also eliminate regions where the power demand is zero, since this regions are not important? Or to you think that would do to much?
Do you think we can still get something out of the dataset, or would you rather switch to the other dataset?

seanlaw:
> > The dataset is suppose to contain 1.3 million datapoints in length which may be too much for a tutorial but I wonder if we could downsample the data by 10x (i.e., only look at every 10th data point and thereby analyzing only 130K data points for each dimension) and still get be able to convey our point. I wanted to bring it to your attention in case it may be useful.
> 
> I'm not sure what data set would be better 🤔 Just added your suggestions to eliminate constant regions. Now we have much better results. But I'm not sure why there is not eliminated more of the 'zero'-regions in the Tumble Dryer and Washing Machine data. Maybe we get good results if I also eliminate regions where the power demand is zero, since this regions are not important? Or to you think that would do to much? Do you think we can still get something out of the dataset, or would you rather switch to the other dataset?

I think we should exhaust our options fully before moving on. The key thing is that there is no way for anybody (including your or I) to know what the best option is until we try. The good thing is that there is no rush here. We should assess as we go. If you are willing, let's "try something and see what happens". Let me know if you are getting frustrated by anything or if there are any other ways that I can help. However, I see that you are a good problem solver too. Know that we are in this together!

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
seanlaw: Great!

SaVoAMP:
I also tried to eliminate those regions that are zero since they aren't interesting. Unfortunately this is eliminating too much and therefore the program doesn't work anymore. I think it might be generally difficult to find more than one motif with the dataset. Preprocessing alone is a major problem here. Either you eliminate too much or too little 😞

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
```diff
@@ -0,0 +1,250 @@
...
```
seanlaw, line 1: <p>What if instead of setting things to <code>np.nan</code>, set things to random noise?</p><p> </p><pre class="ql-syntax" spellcheck="false">m = 4 * 60

padded_stddev = np.zeros(len(df))
for colname in df.columns:
&nbsp; &nbsp; noise = np.random.uniform(0.0, 0.1, size=len(df))
&nbsp; &nbsp; stddev = core.rolling_nanstd(df.loc[:, colname].values, m)
&nbsp; &nbsp; padded_stddev[-len(stddev):] = stddev
&nbsp; &nbsp; modified_values = df.loc[:, colname].where(padded_stddev != 0.0, noise)
&nbsp; &nbsp; df = df.assign(**{colname: modified_values})
&nbsp; &nbsp; # # Also eliminate 0 regions
&nbsp; &nbsp; zero_idx = df.loc[:, colname].where(df.loc[:, colname] != 0, noise)
&nbsp; &nbsp; df = df.assign(**{colname: zero_idx})  
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

seanlaw: After this, I did:

```
mps, indices = stumpy.mstump(df.astype(float), m)
motifs_distances, motifs_indices, motifs_subspaces, motifs_mdls = mmotifs(
    df.astype(float), mps, indices, max_distance=np.inf, max_matches=2
)
``` 

I think we need to allow `max_motifs=5` if we wanted to find more motifs as the default is `max_motifs=1`

SaVoAMP: >What if instead of setting things to np.nan, set things to random noise

That's a pretty good idea! Thank you!

>I think we need to allow max_motifs=5 if we wanted to find more motifs as the default is max_motifs=1 

Yes, I agree. But since we want to show as much of the `mmotifs` function as we can, I would do everything step by step. Or what would you say? 

My suggestion would be to proceed as follows:
I would try to set different parameter combinations in `mmotifs` and try to find different things. For example we could try to find the first motif unconstrained (as it is now), explain how to choose the parameters in `mmotifs` to do so, plot the found motif and explain what we have done, in the first step. Afterwards we could do exactly the same steps but try to find a 2-dimensional motif (instead of the unconstrained one) for example. If we also have the 2-dimensional motif, we could search for its nearest neighbors (matches) and so on. And after that I would try to find some kind of story that explains what we are doing in each step and what we are finding / trying to find.  

That means, that I would try to call `mmotifs` in different ways in order to explain our input parameters. Then I would plot what we have found and explain our findings. We also could show the MDL results after the unconstrained search, to show the user what `mmotifs` is doing if no `k` is set and explain that. And so on ...
I think it might be easiest to do the programming steps first, look what parameter combinations lead to meaningful results and write the context/explanation for the jupyter notebook at the end since it depends on the parameter settings. 

What do you think?

```
num_motifs = motifs_distances.shape[0]

for motif_num in range(num_motifs):
    motif_indices = motifs_indices[motif_num]
    mdl = motifs_mdls[motif_num]
    S = motifs_subspaces[motif_num]
    subspace = [df.columns.values[s] for s in S]

    k = len(S)
    f, axs = plt.subplots(k, 1, figsize=(20, 20), sharex=True)
    plt.suptitle(f"The {k}-dimensional motif pair")

    for j, s in enumerate(S):
        time_series = df[df.columns[s]]
        if k > 1:
            ax = axs[j]
        else:
            ax = axs
        ax.plot(time_series, c="0.75")
        ax.set_title(df.columns.values[s])

        motif_idx = motif_indices[0]
        nn_idx = motif_indices[1]
        ax.plot(time_series[motif_idx : motif_idx + m], c="r", linewidth=4)
        ax.plot(time_series[nn_idx : nn_idx + m], c="r", linewidth=4)
        # ax.axvline(x=motif_idx, linestyle='dashed', c='k', linewidth=2)
        # ax.axvline(x=nn_idx, linestyle='dashed', c='k', linewidth=2)

        # Only relevant if you want to find further matches
        # Set motif_indices[2:] to avoid double counting the motif pair
        [ax.plot(time_series[match : match + m]) for match in motif_indices[2:]]
```

I think it might be useful to wrap that code in a separate function for this tutorial since it will be repeated many times. I'm asking because I haven't seen any function definitions in your tutorials yet. I would define the function at the beginning of the tutorial and explain that it will be used many times to visualize our results/findings.

seanlaw: I think the first demonstration should always be calling the function with default parameters as it is the easiest and most straightforward:

```
mps, indices = stumpy.mstump(df.astype(float), m)
motifs_distances, motifs_indices, motifs_subspaces, motifs_mdls = mmotifs(df.astype(float), mps, indices)
```

So, you get something reasonable back with very little effort. Then, I would motivate the next step with "but this only gave us one motif and, at most, 10 matches", so how would we find more motifs beyond this first "top" one? And, by asking logical questions, we can motivate the introduction/need for the other parameters.

>  I'm asking because I haven't seen any function definitions in your tutorials yet. I would define the function at the beginning of the tutorial and explain that it will be used many times to visualize our results/findings.

You are very observant and thank you for asking! In general, for STUMPY code, I completely agree that we should write reusable functions rather than repeating ourselves everywhere. However, for tutorials, I personally hate needing to scroll all the way up to the top of the notebook in order to remember what the function does. I just can't keep it all in my head and I'd rather having all of the relevant code all in one place for a tutorial. Let me be absolutely clear that I'm not claiming to be "right" here but I think that by not defining functions in tutorials makes the barrier to entry a little lower for those who are new to Python. Considering that you've been a consumer of the tutorials, I'm open to your feedback and criticism as well! How do you feel about it? What are your thoughts?

SaVoAMP: >You are very observant and thank you for asking! In general, for STUMPY code, I completely agree that we should write reusable functions rather than repeating ourselves everywhere. However, for tutorials, I personally hate needing to scroll all the way up to the top of the notebook in order to remember what the function does. I just can't keep it all in my head and I'd rather having all of the relevant code all in one place for a tutorial. Let me be absolutely clear that I'm not claiming to be "right" here but I think that by not defining functions in tutorials makes the barrier to entry a little lower for those who are new to Python. So the repeated code in our tutorials was not a lazy choice but one that was aiming to provide a more "inclusive"/"locally self-contained" programming approach. Considering that you've been a consumer of the tutorials, I'm open to your feedback and criticism as well! How do you feel about it? What are your thoughts?

Since different visualizations were always chosen in the previous tutorials, I thought this made sense. However, I think that in a tutorial where you want to explain the functionality of a function by setting different parameters it may be clearer to define one visualization function since we are using exactly the same code many times (which was not the case in the previous tutorials) in order to explain how everything works and what the differences between the function calls are. 

>However, for tutorials, I personally hate needing to scroll all the way up to the top of the notebook in order to remember what the function does. I just can't keep it all in my head and I'd rather having all of the relevant code all in one place for a tutorial.

Generally I agree with you here. But in this specific case I think it might be easier/clearer to define a `visualize()` or `show_matches()` function. Such a function is so simple that I think the name says enough about it that everyone knows what is happening without having to scroll up for it 🤔 

>I think the first demonstration should always be calling the function with default parameters as it is the easiest and most straightforward:

Yes, that sounds reasonable! I wanted to make sure at the beginning that I actually get exactly one motif pair. I'll take a look if we get a solution with the default parameters (in my testing data I had to at least set the `max_distance` parameter in order to find a motif).

seanlaw: > However, I think that in a tutorial where you want to explain the functionality of a function by setting different parameters it may be clearer to define one visualization function since we are using exactly the same code many times (which was not the case in the previous tutorials) in order to explain how everything works and what the differences between the function calls are.
> Such a function is so simple that I think the name says enough about it that everyone knows what is happening without having to scroll up for it

@SaVoAMP I trust your judgement. Go for it and let's see how it turns out!

> Yes, that sounds reasonable! I wanted to make sure at the beginning that I actually get exactly one motif pair. I'll take a look if we get a solution with the default parameters (in my testing data I had to at least set the max_distance parameter in order to find a motif).

Sounds good!

SaVoAMP: I'm not sure why but every time I rerun the notebook, I get different outputs, for example:
image
image

seanlaw: <p>This does sound strange. It implies that you may be overwriting your original data structures. Are your <code>otifs_distances, motifs_indices, motifs_subspaces, motifs_mdls</code> still the same when you re-run the notebook or are they changing after each re-run? </p>

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
```diff
@@ -0,0 +1,242 @@
...
```
seanlaw, line 1: <p>I don't think that it is wise to have a function with no inputs as this can cause unintended behavior. Instead, you should always pass in the latest <code>motif_distances</code>, <code>motif_distances</code> , <code>df</code>, etc.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

seanlaw:
> [MOD] Rap visualization in a fuction

I can't lie, @SaVoAMP, I was really looking forward to seeing how you combine "Rap" and "visualization" 😸

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
SaVoAMP: Ok, thank you, I'll change it!

SaVoAMP:
>I can't lie, @SaVoAMP, I was really looking forward to seeing how you combine "Rap" and "visualization" smile_cat

Oh god, that is very embarrassing 😆 It's supposed to mean 'wrap' 🤣

seanlaw:
Hey, it's been attempted once before in [this tutorial](https://stumpy.readthedocs.io/en/latest/Tutorial_AB_Joins.html#Finding-Similarities-in-Music-Using-STUMPY) so I'm keeping my fingers crossed for a repeat 🤣

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
SaVoAMP: Unfortunately this doesn't change anything and I still get different motifs. I have to take a closer look at this the next days. I don't like Jupyter notebooks since I am often debugging my way through all the code which is inconvenient with them in my opinion. So I resort to copy-pasting it to Pycharm and back as needed 😆

SaVoAMP:
>but this only gave us one motif and, at most, 10 matches

> max_matches : int, default 10
        The maximum amount of similar matches of a motif representative to be returned.
        
I notice while trying to write the tutorial that we use matches and motifs completely synonymously. 
When saying that `10 is the maximum amount of similar matches of a motif representative to be returned` I would expect to get `11` subsequences (the motif and its matches). The same also applies to `gave us one motif and, at most, 10 matches`.  Do you think we should reword the description a bit or explain in the tutorial how this is meant?

seanlaw:
@SaVoAMP No rush but please let me know when this is ready for review.

SaVoAMP:
Yes, I will! 
It's just quite a bit stressful for me with the bachelor thesis, because I'm trying to write an evaluation function that takes all possible special cases into account. My goal is to have it ready in the middle of next week and then I have a little more time for the tutorial!

At most, if you have a little more time, you could take a quick look to see if I'm going in the right direction with the tutorial 😄 Otherwise, I would first write down the complete tutorial once and then let you know if I have considered everything that is important or interesting for the tutorial in my eyes, so that you can then give me overall feedback.

seanlaw:
> At most, if you have a little more time, you could take a quick look to see if I'm going in the right direction with the tutorial 😄 Otherwise, I would first write down the complete tutorial once and then let you know if I have considered everything that is important or interesting for the tutorial in my eyes, so that you can then give me overall feedback.

Everything looks pretty good so far. Maybe the only major suggestion is that for Figure 3 (Tumble Dryer), it would be good to make it 2 subplots (stacked one on top of the other) and where the top plot is what you currently have and the bottom plot is an overlay of all of the motifs+matches (don't forget to z-normalize them). Like [this overlay](https://stumpy.readthedocs.io/en/latest/Tutorial_Pattern_Matching.html#Levaraging-STUMPY-To-Do-Some-Work-For-You)

SaVoAMP:
All right, good suggestion!
I was actually planning to do that, however, only after the second plot where only the motif pair was drawn in. But of course I  also can do that at the first plot with default parameters!

SaVoAMP:
I think I'm ready for the first review now 😄

seanlaw:
Thank you for letting me know. I will find some time to review it

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
```diff
@@ -0,0 +1,827 @@
...
```
seanlaw, line 1: <p>Maybe highlight the zoomed in section with a grey background (above)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

seanlaw, line 1: <p>Since the first example below focuses on "Tumble Dryer", it may be less confusing to focus on an area where there was significant dishwasher activity (i.e., the motif region)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

seanlaw, line 1: <p>The colors above refer to different appliances. I think it is confusing for the same range colors to be used here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

seanlaw, line 1: <p>What should we take away from this image?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

seanlaw, line 1: <p>"Tumble", not "Tuble"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

seanlaw, line 1: <p>This is actually pretty cool! However, shouldn't we be able to capture the spikes in the "Washing Machine" usage? It's hard to tell without zooming in but it feels like we're only capturing the noise?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

seanlaw, line 1: <p>I feel like we should split up the images for the first motif pair and the second motif pair. Additionally, I think we'd want to explain that not all matrix profile distances are created equal and that, likely, the matrix profile value for the second motif pair is much higher than the matrix profile value for the first motif pair. In other words, there may be no significant relationship between "Dishwasher" and "Freezer" (right??)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

seanlaw, line 1: <p>I don't think it's entirely clear why we get three appliances. At least, it's not clear to me :)</p><p> </p><p>Also, what is the difference between "Fridge-Freezer" and "Freezer"? Perhaps the latter is what American's might refer to as a "Chest Freezer"? We might want to clarify it here if we can.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='43'/>

seanlaw:
@SaVoAMP Overall, I really like the contents! You've covered a lot of important points and explained a lot of the concepts well.

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
SaVoAMP: <p> </p><p>Yes, I was also thinking about this. But I'm not sure what to do about the noise problem. Do you think that increasing the amount of noise would change anything? Do you have any ideas?</p>

SaVoAMP: > By any chance, have you ever tried:
> 
>     Not adding noise
>     Replace zero regions with np.nan
>     Use a variety of smaller window sizes

Yes, I was trying to do so, but it seems like as if it was eliminating too much :(

> Additionally, I think you had mentioned this before but it does sound like there may be an annotation vector play here. What's challenging is that, from the multi-dimensional matrix profile (aka "multi-MP"), we don't immediately know which dimensions contributed to each distance value (in the multi-MP). So, we would need to do something like:
> 
>     Start with the multi-MP
>     For each value, determine the precise subspace that generated the distance value (we know what the precise subsequence pairs are)
>     For a given matrix profile distance and subspace, determine if the (multi-dimensional) subsequences within that subspace contain zero regions
>     if so, set the multi-MP distance value for that dimension and subsequence index to np.inf
> 
> Essentially, in steps 1-3, we're generating a multi-dimensional annotation vector. I think this is a bit of work to do but it may actually be a process flow that I think users will inevitably encounter with their real-world data. Would you care to play this?

Yeah I think it may be worth investigating some time into it and see if it is working as expected. However, I am currently quite stressed with the writing of the bachelor thesis, because my program does not yet want the way that I would like it to be and I also have to evaluate quite a lot. It may therefore take some time until I can look at the problem. But I'll definitely check it out as soon as I have some more free time again!

seanlaw: > It may therefore take some time until I can look at the problem. But I'll definitely check it out as soon as I have some more free time again!

Please take your time! I'll try to offer some code suggestions related to the three points above.

SaVoAMP: All right, sounds good :)

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #8.    from stumpy.mmotifs import mmotifs
</pre><p>We don't need this line anymore. Instead, with the latest version of <code>stumpy</code>, you should be able to call <code>stumpy.mmotifs</code> directly   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #30.        .sum()
</pre><p>I think we can add <code>.astype(float)</code> after this so we don't need to say <code>df.astype(float)</code> everywhere else below  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

seanlaw, line 1: <p>I think this should be:</p><pre class="ql-syntax" spellcheck="false">for colname in df.columns:
&nbsp; &nbsp; plt.plot(df[colname], label=colname)
plt.legend()
plt.show()
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/557/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

seanlaw:
@SaVoAMP What joke am I missing/overlooking? :)

SaVoAMP:
> Now our dataset contains the time series of the five appliances with timestamps now sampled in minutes. Let's visualize it!
fridge freezerfreezerdishwasherWashing Machine and Tumble Dryer
>Ok, now that we've marveled at those beautifully visualized household appliances we're ready to visualize those timeseries as well.

The humble joke was visualizing the appliances with pictures (instead of visualizing the dataset by plotting the data).
Don't worry, not all jokes in Germany are that serious, I just didn't want to go overboard with that one 😄

SaVoAMP:
My initial intention was to clarify the difference between a fridge-freezer and a freezer and I guess I've kind of embedded that clarification as a humble joke 🥲

seanlaw:
Gotcha! 😂

SaVoAMP:
> 
>     Shouldn't we allow max_matches to be an array or a list of the size of max_motifs like we have done with cutoffs? I'm asking since I'm trying to find out how well the motif discovery using matrix profiles is suited to automatically annotate multidimensional time series data. I wrote a script that calculates the matrix profile for each new subsequence length and tries to find the motifs with mmotifs. I have got different sublabels that tell me that I'm searching for 5 sublabels of one type and 7 of another type for example. Then I get 2 different subsequence lengths (for example by computing the average of each sublabel type). If the two subsequences are different, I only have to compute the matrix profile and find one motif with mmotifs both times. But if they have the same length, the matrix profile should only be computed once but I want to find two motifs with mmotifs. Here I've got the problem that I can't tell mmotifs that I'm searching for 5 matches for the first motif and for 7 matches for the second motif.
> 
>     Wouldn't it make more sense to also allow max_matches to be an array but set a default array that only contains the number 10, so that nothing changes to the user if he doesn't want to have a different number of matches for each motif? We could do exactly the same as we have done with cutoffs and therefore, if only one value is set, this value could be applied to every dimension. thinking
> 
> @SaVoAMP I hope that I am understanding your intent correctly but what you are describing (in terms of computing multiple multi-dimensional matrix profiles for different window sizes) sounds more complex than the average use case. Instead of changing mmotifs, I find myself wondering if you are able to get what you need in a post-processing step after mmotifs is called?
> 
>     Here I've got the problem that I can't tell mmotifs that I'm searching for 5 matches for the first motif and for 7 matches for the second motif.
> 
> In this case, I would call mmotifs with max_matches=7 (or whatever the largest number needs to be) and then post-process the results afterward by removing the parts that are not needed or are too much. Even if you changed/modified max_matches to be variable for each motif, you'd still need to post-process the results anyways because the width of, say, motif_distances would still (possibly) be as wide as 7 matches for all motifs (i.e., the ones where you request max_matches=5 will have np.inf padded to it). I do not wish to return a jagged/ragged motif_distance as this would cause inconsistencies with motifs (1-D).
> 
> I understand that it may seem like "less things to compute" by providing an array/list of max_matches but that gain is likely only very, very small. I'm not sure it is something that we'd want to optimize. Again, I'm not opposed to your idea but I am trying to reduce/minimize the code complexity especially when the computational gain may not be substantial. Maybe I'm missing your point but it's starting to feel like we are on the verge of "doing too much".

By the way, today I was thinking about this again. I don't think that it would be enough to post process the results of the mmotifs function and search for the largest number of matches for every motif. Let's imagine again that we want to find one motif with `max_matches=5` and another motif with `max_matches=7`. If the whole time series only contains 7+5=12 regions where the motifs/matches are present with only little space between them, we would eliminate too much when searching for 7 matches for both motifs in the fist step. Therefore if we want to find 5 matches for the first motif and 7 for the second one, we would first find 7 matches of the first one (that should be post processed later to only include 5 matches instead of 7) with mmotifs. Afterwards we are setting an exclusion zone and that is the reason why we are only able to find 5 more matches for the second motif and the last two matches can't be found since we have also applied the exclusion zone to the two matches that were found unsuitable for the fist motif. That means that we have found 7 matches for the first motif (which should only contain 5 matches) and could therefore only find 5 matches for the second motif that should contain 7 matches. Post processing wouldn't help here. Although we could post process the first motif to only contain the 5 matches we wouldn't be able to find the two matches that should belong to the second motif.

seanlaw:
> By the way, today I was thinking about this again. I don't think that it would be enough to post process the results of the mmotifs function and search for the largest number of matches for every motif. Let's imagine again that we want to find one motif with max_matches=5 and another motif with max_matches=7. If the whole time series only contains 7+5=12 regions where the motifs/matches are present with only little space between them, we would eliminate too much when searching for 7 matches for both motifs in the fist step. Therefore if we want to find 5 matches for the first motif and 7 for the second one, we would first find 7 matches of the first one (that should be post processed later to only include 5 matches instead of 7) with mmotifs. Afterwards we are setting an exclusion zone and that is the reason why we are only able to find 5 more matches for the second motif and the last two matches can't be found since we have also applied the exclusion zone to the two matches that were found unsuitable for the fist motif. That means that we have found 7 matches for the first motif (which should only contain 5 matches) and could therefore only find 5 matches for the second motif that should contain 7 matches. Post processing wouldn't help here. Although we could post process the first motif to only contain the 5 matches we wouldn't be able to find the two matches that should belong to the second motif.

Hmm. In this contrived example, I'd argue that the time series data is simply too short and somewhat of a "worst-case-scenario" or extreme edge case. I'm not convinced that it's as common as you are presenting it to be. Even if that were true, `mmotifs` should only be used to find multi-dimensional motifs in the "easiest" cases and should not be used to solve ALL cases (especially not edge cases). Instead, if a user's data/results were this limited, they should really stop using `mmotifs` and pivot toward writing their own motif discovery function or establishing their own custom process. For STUMPY, I want us to keep in mind that "we don't want (`mmotifs`) to be everything for everyone" and should avoid being worried about these extreme edge cases. They are certainly important but they are well beyond the scope of what STUMPY can/should support. Otherwise, we'd spend all of our time chasing "perfection" when really what we should be aiming for is "good enough for obvious use cases". Does that make sense?

SaVoAMP:
I came to the conclusion that it wouldn't work at all since we don't know which motif will be found first. Therefor we aren't able to tell the function that the first motif should be found with 5 repetitions and the other with 7, for example. So it wouldn't make sense at all.

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
SaVoAMP: Ahh, cool, thank you very much! Unfortunately I'm very stressed with my thesis and final presentation at the moment. There is only one month left and afterwards I have got it! Then I'll have much time to think about this and try different things out. If you want to, I could then also help you with the one-dimensional motif and matches tutorial that you wanted to create or help with something different :)

seanlaw: <p>That would be great! Best of luck</p>

SaVoAMP:
Today I came across this paper: https://core.ac.uk/download/pdf/287941767.pdf

In the paper the authors talk about the implications of z-normalization in the matrix profile. They claim that, when comparing subsequences that are relatively flat and noisy, the resulting distance is high despite the visual similarity of these subsequences. Therefor they derived a method to eliminate worse performance of the z-normalized euclidean distance, when series contain flat and noisy subsequences. I thought that this could also be helpful for our problem in the tutorial 🤔 If you haven't looked at it yet, maybe I could take a closer look at the paper after I finish my bachelor thesis and find out if it may help.

seanlaw:
> Today I came across this paper: https://core.ac.uk/download/pdf/287941767.pdf

@SaVoAMP I did come across that paper a few years ago when it was published but I have not looked at it recently. While the research may be valid, I am concerned that the tutorial will be too distracted from being focused around the `mmotifs` parameters. I don't want to throw too many different concepts at the reader and have it feel too overwhelming. What do you think?

SaVoAMP:
Yes, that makes sense! 
So we could first finish the tutorial with an easier concept and I could dive into the other problem afterwards.

notebooks/Tutorial_Top-K_Multidimensional_Motif_and_Matches_Discovery.ipynb
SaVoAMP: Hey :)

I finally finished my bachelor thesis. I'll try to integrate your suggested Annotation Vector approach the days!

seanlaw: <p>Welcome back and congratulations on wrapping up your bachelors thesis! </p>

SaVoAMP:
Hey, sorry, I was ill for a while 😞 

I have already integrated your suggested Annotation Vector code and updated all of the text according to the motifs that are found with this new approach. I think that the new motifs make more sense than the old ones. However, the three-dimensional motif that is found in the "Constrained Search" section doesn't look really meaningful since many no-action regions are found for the matches.

At this time I didn't find out why the "Comparing the motif with its nearest neighbors" plot only works sometimes. That's pretty strange 🤔  I think I have to check this with another IDE since I don't like to debug directly on jupyter notebooks 😆

By the way, do you have a programming task I could contribute to? Maybe something like developing the `mmotifs` function? I would love to contribute some code and improve my programming skills, but jupyter notebooks drive me crazy 😆

seanlaw:
Hi @SaVoAMP! How are you? I just wanted to check-in here and see how things were going?

SaVoAMP:
Hey 😃 

> At some point, if the process and results aren't clear then maybe we merge the tutorial and simply accept it in a "work-in-progress" state and allow people to discover it (or we point people to it that ask in the discussions) but, due to the lack of clarity, we don't add it to the Tutorial section. Maybe you'd agree that while the tools are reasonable, the example use case will likely cause more confusion. I think you've done a tonne of work and I also want to give you an opportunity to break from it as well. We've done this many, many times with other tutorials where we just didn't publish it (not because of lack of effort) but because it wasn't "ready" yet and, sometimes, it may never be. What do you think?

Sorry, I interpreted the meaning of this message so that further work on this tutorial section may not be worth the effort. 

The only idea that comes to my mind regarding this tutorial would be to change the data set and thus redo the text of the tutorial section as well. 
I might have a suggestion for an appropriate data set (I think this one should be suitable for the multidimensional pattern matching tutorial) [here](https://github.com/andrebert/body-weight-exercises). It is a data set about human activity recognition and consists of 30-dimensional time series data. Different participants perform various exercises (such as Squats, Mountain Climbers, Push-Ups and so on) in three sets. I already used this data set for my Bachelors Thesis. Unfortunately the data is stored in .zip-archives and sorted as separate files for different sensors (where one sensor file includes the time series data of the three spatial dimensions) and would require some preprocessing to bring it into a suitable .csv format. What do you suggest?

seanlaw:
@SaVoAMP I think the pragmatic thing to do is:

1. Declare this notebook as a "work-in-progress"
2. Clean up this notebook so that we emphasize the parts that worked well (i.e., the tools/functions made it easy for the user to find what they need in the data) and what parts didn't work well and needs better data/examples
3. In other words, we clearly state what is "done" and what "isn't done" (also reminding the reader that we might need ot change the data and try again)
4. Merge the notebook as-is

This way, your work can still be captured and anybody (including you) can take another stab at improving it in a future PR to this file (whether it's using the same dataset or something else). If you want to try it with another dataset then I would keep the PR separate. How does that sound?

SaVoAMP:
All right, that sounds good!

Unfortunately I'll probably only get to do it on the weekend since I'm busy with my masters degree and work on weekdays.

seanlaw:
> Hey, I'm sorry but I was very busy on the weekend with studying. But I promise that I'll get to do it as soon as I have some free time. 😃

No worries! Thank you for letting me know @SaVoAMP

seanlaw:
@SaVoAMP [This](https://github.com/TDAmeritrade/stumpy/issues/694) might be relevant

SaVoAMP:
Uhh, this looks very interesting, especially since I particularly like to work with sports data.

I'm very sorry that I have not found the time to go through the tutorial again yet. In addition to the currently quite stressful everyday life, I was also sick last week. I really hope to get to it this coming weekend!

Probably it would make the most sense to work though your mentioned steps (17 days ago) first and experiment with the new data set afterwards as you already said?

> 
>     Declare this notebook as a "work-in-progress"
>     Clean up this notebook so that we emphasize the parts that worked well (i.e., the tools/functions made it easy for the user to find what they need in the data) and what parts didn't work well and needs better data/examples
>     In other words, we clearly state what is "done" and what "isn't done" (also reminding the reader that we might need ot change the data and try again)
>     Merge the notebook as-is

These steps I think (and hope) I should be able to do soon. 

> If you want to try it with another data set then I would keep the PR separate. How does that sound?

This will probably take a while. I find the data set really exciting, but unfortunately I can't guarantee how fast I will make progress besides studying and working. If it continues to be this stressful, I may not be able to work on it decently until my exam period in late January / early February is over. Unless you just happen to have an idea how to combine some machine learning stuff with matrix profiles so that it has some use for stumpy (I'm currently looking for a project idea for a machine learning subject I have chosen) 😆.

seanlaw:
> I'm very sorry that I have not found the time to go through the tutorial again yet. In addition to the currently quite stressful everyday life, I was also sick last week. I really hope to get to it this coming weekend!

No worries! This is all volunteer and you are allowed to say, "I don't have the time" too.

> Probably it would make the most sense to work though your mentioned steps (17 days ago) first and experiment with the new data set afterwards as you already said?

Yes, I think that is the right plan. Finish this up and, separately (whether it is you or somebody else), we can explore the other dataset.

> This will probably take a while. I find the data set really exciting, but unfortunately I can't guarantee how fast I will make progress besides studying and working. If it continues to be this stressful, I may not be able to work on it decently until my exam period in late January / early February is over. 

This is why I created a new issue. Sometimes, issues sit there and nobody has time/interest to work on them and that is okay. It's just a place to track thoughts/ideas and anybody can give it a try if they have the availability. No pressure.

> Unless you just happen to have an idea how to combine some machine learning stuff with matrix profiles so that it has some use for stumpy (I'm currently looking for a project idea for a machine learning subject I have chosen) 

I recommend that you post something in our Discussion section as others may have some thoughts/suggestions!

SaVoAMP:
Today I took a closer look at the "comparing the motif with its nearest neighbors" plot. It seems to be correct after all. I enabled `%matplotlib notebook` and zoomed into the plot. It turned out that there are 10 matches plotted but they look so similar that it's not possible to see them without zooming very close into the plot. I think everything should fit now if one disregards the fact that the data set is not suitable in order to find meaningful motifs.

seanlaw:
Great! If you are satisfied with it then I will merge this notebook after the tests complete? I will await your final confirmation @SaVoAMP

PR #565, opened 2022-03-06
Fixed #563 Corrected description of parameter p in docstrings

### LINKED ISSUE #563: misleading description of parameter `p` in `mass_absolute`
I found this before, and not sure if it is a mistake or not.

In `core.mass`, the description of parameter `p` is as follows:
```
    p : float, default 2.0
        The p-norm to apply for computing the Minkowski distance. This parameter is
        ignored when `normalize == False`.
```

So, I assume `p` is only for normalized version. However, we see the following description of the same parameter in `core.mass_absolute`:

```
p : float, default 2.0
```

Maybe we should remove parameter `p` here in non-normalized version, or add a note to let user know it has not been implemented yet.

NimaSarajpoor:


seanlaw:
Thank you @NimaSarajpoor.  I will merge this once the tests pass

Issue #585, opened 2022-04-05
Variable Length Motifs and Discords Discov ery


seanlaw:
There is the [2018 VALMOD paper](https://arxiv.org/pdf/2008.13432.pdf) and, later the [2020 MAD paper](https://arxiv.org/pdf/2008.13447.pdf)

NimaSarajpoor:
I am trying to go through the extended version of paper. I already finished the introduction part of Jupyter Notebook, which provides the gist of concept behind the paper. However, before pushing it, I wanted to get equation (2) myself to make sure it is correct. And, then I noticed there is a typo:

![image](https://user-images.githubusercontent.com/38519522/161904985-fcdfb211-df61-43fe-bb4a-e6e86745f133.png)

So, I applied the same typo in my calculation, and I got eq (2).  Now, I am trying to solve it without typo.... I will push the notebook so you can see the solution...

seanlaw:
Great. I'll need to go over the paper myself

PR #586, opened 2022-04-06
[WIP] Add Tutorial Notebook for VALMOD #585

NimaSarajpoor:
This notebook addresses issue #585. In this notebook, we would like to implement the VALMOD method proposed in [VALMOD_2018](https://helios2.mi.parisdescartes.fr/~themisp/publications/sigmod18-valmod.pdf) and [VALMOD_2020](https://arxiv.org/pdf/2008.13447.pdf).

**What I have done so far:**
* Provide introduction that provides gist of concept proposed in the paper
* Calculate Lower-Bound distance profile after correcting the typo in eq(1) of paper....and verify the calculation with a `np.random.uniform` time series data.

For now, I calculated LB given `q>0` (see eq(2) in paper.) However, we still need to find LB when `q <= 0`. Since we already noticed the difference between the paper's LB and the one provided in the notebook (after correcting the typo), we need to calculate LB for `q<=0` as our answer might be different than what provided in the paper for this case as well.


NimaSarajpoor:
@seanlaw 
Please allow me some time to see if I can calculate `LB` for q<=0 (see eq(2) of paper). I will let you know when I am done...

NimaSarajpoor:
I had a miscalculation. Although there is a typo in the paper, it seems the eq(2) of paper is correct. I fixed the typo of paper when I was doing calculation. However, I had a miscalculation somewhere else...so I corrected it and I got the eq(2)... for `q>0`. I will fix the notebook...

NimaSarajpoor:
@seanlaw 

Notebook is ready. The notebook covers the first 12 pages of VALMOD_2020 paper. I FIX the miscalculation and things are good! I also implemented the Low-Bound distance profile function for now (we may use it later in VALMOD algorithm).

NimaSarajpoor:
Just wanted to let you know that you can ignore the function '_calc_LB_dist_profile' function at the end of notebook (it is working..but I think it is not clean. I may probably remove it as VALMOD algorithm does not use such function. I just created it to get Lower-Bound of distance profile)

seanlaw:
I will first need to go over the initial 12 pages myself and then I will review the notebook :)

seanlaw:
@NimaSarajpoor I've gone over your notebook quickly but haven't verified the derivation. Usually, with derivations, I like to write things out fully without skipping any steps (see https://github.com/TDAmeritrade/stumpy/blob/main/docs/Matrix_Profile_Derivation.ipynb). Some of your equations don't seem to be rendering for me and it's a bit hard for me to follow. I can try to find some time to work through the derivation to verify your work if that's helpful?

NimaSarajpoor:
I see. Please let me re-write it. I will try to follow the same approach/style you used in the link you provided. I will check ReviewNB and if it is rendered well, I will let you know. Sounds good? 
(Btw, is it necessary to provide the derivation? I did that because of the typo in the paper, but then I realized the result is the same as what provided in eq(2) for q>0)

seanlaw:
Yes, that would be great!

Personally, I think writing out the derivation clearly will help (me) and others reduce any doubt in understanding. Also, I find that it provides an opportunity to help maintain consistency in the code regarding variable names.

NimaSarajpoor:
Weird...still not rendering well.... please let me do some investigation on my end to see what's going on...

NimaSarajpoor:
@seanlaw 
So, I tried the Welford_Review/ Matrix_Profile_Derivation/ Pearson notebooks to see if ReviewerNB on Github can render them. Unfortunately, it seems that it cannot render them properly for the first two notebooks. Peason notebook is rendered properly though!  

I guess you wrote the notebooks on your end and pushed them to the repo...and things were good when rendered locally in .ipynb. Did you, by any chance, try to check your notebooks via ReviewerNB of Github?

---
It seems the problem is related to the ReviewerNB of Github. I enclosed the math equations with `$$`  and it seems the problem is resolved. Well, almost resolved! I checked out  the ReviewNB here and it seems there is still one error. I pushed the same notebook to a test_repo I created and it seems that single error does not appear when I checked it out with ReviewerNB on my end.

[my_test_repo](https://app.reviewnb.com/NimaSarajpoor/test_repo/)

NimaSarajpoor:
@seanlaw 
Just for the records:
if I click on the ReviewNB purple button (on top of this PR page), it seems there is still one error related to rendering. However, when I clicked on ReviewNB blue hyperlink (on top of this PR page just below the purple button), and navigated to the notebook from my fork of STUMPY, everything seems to be fine and there is no error in rendering....

[STUMPY_my_fork](https://app.reviewnb.com/NimaSarajpoor/stumpy/blob/Tutorial_VALMOD/docs%2FTutorial_VALMOD.ipynb)

seanlaw:
Sounds good

docs/Tutorial_VALMOD.ipynb
```diff
@@ -0,0 +1,688 @@
+{
```
seanlaw, line 1: <p>IIUC, Keogh often refers to this as the "twin freak" problem where the top-1 neighbor is similar but the neighbors after that are quite far away.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <p>Do you think that this will be an issue for non-normalized Euclidean distance or other p-norm distances?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

seanlaw, line 1: <p>In the future, we would want to be consistent and common with STUMPY as we have tried very hard to make all of our naming conventions as consistent as possible. Whereas each different author of a matrix profile has often chosen to use different naming conventions.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

seanlaw, line 1: <p>"distance" rather than "distane" (missing a "c")</p><p> </p><p>What is "inequation"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

seanlaw, line 1: <p>Thank you. This is much clearer than the paper!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

seanlaw, line 1: <p>Why are we using lowercase <code>t</code> here? Lowercase <code>t</code> typically refers to the element inside of a subsequence or inside of the time series <code>T</code>   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

seanlaw:
Apologies, these comments are for an older commit. I forgot to hit "Finish Review" along with my last comment.

docs/Tutorial_VALMOD.ipynb
```diff
@@ -0,0 +1,606 @@
...
```
seanlaw, line 1: <p>For clarity, I would add one additional step in between these two equations that takes the <code>m + k</code> summation and splits them into two separate summations (both staying under the square root) that go from <code>t = 1</code>  to <code>t = m</code>  and then <code>t = m + 1</code> to <code>t = m + k</code> . Then, it becomes obvious why the "greater than or equal to" must be <code>True</code> (i.e., if we remove the second summation which must be greater or equal to zero, then the bottom equation becomes a lower bound, so to speak)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

seanlaw, line 1: <p>It might be obvious to you, but we are skipping steps here. Starting with the upper equation, I would recommend adding the following explicit steps</p><p> </p><ol><li>In the inner parentheses, you factor out <code>1/sigma_j,m+k</code> </li><li>In the inner parantheses, you multiply by a factor of <code>sigma_j,m/sigma_j,m</code> , which is basically multiplying by <code>1</code>   </li><li>Resolve/simplify the denominators inside of the parentheses</li><li>Finally, move the factors from inside the parentheses to outside of the square root</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

seanlaw, line 1: <p>Should we really do this substitution now? Or can we do it at the end to match the paper?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

seanlaw, line 1: <p>Your math is better than mine. Can you educate me on what the <code>triangle over equal sign</code> means?</p><p> </p><p>Also, why are we calling it <code>X_t</code> ? Why <code>X</code>?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

seanlaw, line 1: <p>I see, the substitutions above makes the partial derivatives easier to do! I'd actually REALLY like to see this step-by-step derivation if you don't mind. The current notation is a bit clunky. I think if you performed the partial derivative calculation and then, near the end, set the left side to zero, then it would be pretty clear. I know it's a lot of work though</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

NimaSarajpoor: Yeah... I did that in the previous version as it was not easy for me to understand it before. But then, I skipped that step. I do not know  why? Maybe, because I saw it the second time and I was comfortable with it.  

Your are correct! I should have kept that....I will apply the steps in next set of commits.

NimaSarajpoor: I tried to follow the paper as closely as possible. In fact, we do not even need to factor out `sigma_j,m/sigma_j,m+k`. We can start taking derivatives with respect to variables `mu_i,m+k` and `sigma_i,m+k`. The manipulation was performed by the authors and I basically tried to follow their steps.

Also, note that the factor `sigma_j,m/sigma_j,m+k` extracted by the authors will remain in eq(2). So, I thought it would be easier to follow the authors' step as they created such factor right from the beginning.... and focused on minimizing the (manipulated) term under the root sign.

If you want to avoid such substitution, then I suggest to avoid the factoring-out step as well, and simply treat  `mu_i,m+k` and `sigma_i,m+k` as variables right from the beginning.

NimaSarajpoor: > what the `triangle over equal sign` means?

if I am not mistaken, it means "define/call" something. I used the same symbol when I was writing eq(6) of the notebook. So, if we want to define a term ourselves, we use this symbol. I think it is somewhat similar to assignment operator `:=`. I did this to distinguish between the name of variables in the problem, and the names I created/defined myself. 

> Also, why are we calling it `X_t` ? Why `X`?

Did you mean "why NOT `X`?" ?, or Did you mean why the letter `X`? I used the subscript `t` to show that this variable changes with `t`. Why `X`? No reason. Now that you asked, I think I should have used a better name such as `d_t` as the distance between two corresponding elements  at time stamp `t`.

NimaSarajpoor: You are definitely correct. I felt the same way but wasn't sure what was the best way to do that. Thanks for the suggestion. So, I will keep (1) and (2). Then, for (3), I first take the derivatives and then, I replace the `partial-derivation notation` with 0.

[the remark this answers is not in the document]
NimaSarajpoor: (NOTE: I forgot to provide the limits on summation. I will fix it)

The factor `-2/(sigma_square)` is constant, i.e. not depending on `t`.  So, we can take it out and  simply remove it. Now, if you expand the term by multiplying `X_t`, you can see we have: `SUM(TX - \mu X)`, which can be written as `SUM(TX) - SUM(\mu X)`. The latter term, `SUM(\mu X)` is zero because: `SUM(\mu X) = \mu SUM(X) = using eq(3) = 0` 

Please note that all `\mu` and all `\sigma` are constant. They might be known or unknown, but they are not changing with time stamp `t`. 

---

Well, now I see I should have provided the steps. When explanation goes beyond half a sentence, it should be a red flag for me that something is not written in a clear way. I can provide the steps that I skipped before. If you think that is not necessary, please let me know.

NimaSarajpoor:
@seanlaw 
Thanks for the comments! 

> Apologies, these comments are for an older commit. I forgot to hit "Finish Review" along with my last comment.

I found two of those comments in the ReviewNB. Maybe they got mixed together (?). I will address two of those comments and then I resolved the rest. Please let me know if I miss anything.

docs/Tutorial_VALMOD.ipynb
NimaSarajpoor: Thanks!!! I knew I read something about such case before but I couldn't remember the name. I will use the same terminology. I will try to provide the reference if I can find that link/file. I will revise it in the next set of commits.

docs/Tutorial_VALMOD.ipynb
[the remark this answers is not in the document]
NimaSarajpoor: I will revise it accordingly in the next set of commits.

NimaSarajpoor: So, is it okay to keep the name `X_t` (for now)?

docs/Tutorial_VALMOD.ipynb
NimaSarajpoor: I assume the comment is for the previous version of notebook. So, I ignore this. If I am not correct, please let me know.

NimaSarajpoor: I assume the comment is for the previous version of notebook. So, I ignore this. If I am not correct, please let me know.

seanlaw: Yes, it would be great if you could add the derivation for this

docs/Tutorial_VALMOD.ipynb
seanlaw: Agreed. Thinks look nicer as a result of the substitution

seanlaw: I propose a more neutral variable like the `\alpha_t` or `\beta_t`

docs/Tutorial_VALMOD.ipynb
NimaSarajpoor: Sure. I will add it to the notebook in next set of commits.

docs/Tutorial_VALMOD.ipynb
NimaSarajpoor: Sounds good! I will revise notebook accordingly in the next set of commits.

NimaSarajpoor:
I think we are all set. I can push commits after revising the notebook.

docs/Tutorial_VALMOD.ipynb
```diff
@@ -0,0 +1,1137 @@
...
```
seanlaw, line 1: <p>Looks great! Easy to follow</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

seanlaw, line 1: <p>I think it makes sense to substitute <code>alpha</code> in the right side of <code>f</code> since, down below, you use the <code>alpha</code> representation of <code>f</code> .</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

seanlaw, line 1: <p>Excellent. Very clear</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

seanlaw, line 1: <p>Interesting. So the second term in <code>alpha</code> has no effect on the partial derivative (i.e., only the first term matters).</p><p> </p><p>Can you please explain why the sum of <code>alpha</code> is zero? I see that it is zero in the other partial derivative but what does it have to do with this partial derivative? Btw, <code>mu^prime(0)</code> looks as if you are evaluating <code>mu^prime</code> as a function. Instead, you may want to use a <code>\cdot</code> instead to show that it is a multiplication </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='37'/>

seanlaw, line 1: <p>Performing the expansion here feels very disconnected from equation (3)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

seanlaw, line 1: <p>Is this Pearson relationship and LB discussed in the paper as well?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='47'/>

seanlaw, line 1: <p>This is getting really complex. I have not reviewed this Pearson part yet.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='55'/>

NimaSarajpoor: > Interesting. So the second term in `alpha` has no effect on the partial derivative (i.e., only the first term matters).

Yes, that's right. In `\alpha`, the variables '\mu_prime` and `\sigma_prime` appear only in the first term. So, when we take derivatives w.r.t to these variables, the second term of `\alpha` becomes 0.

> Can you please explain why the sum of `alpha` is zero? I see that it is zero in the other partial derivative but what does it have to do with this partial derivative?

Please note that  (1) and (2) are a single system of equations. That is, we want to find `\mu_prime` and `sigma_prime` such that they satisfy both (1) and (2). So, if we get to an equation derived from (1) or (2), the equation should hold true in the other one. So, after working with (1), we realized `\sum\alpha_prime = 0` (at the critical point we are looking for). That means the values of `\mu_prime` and `sigma_prime` should satisfy equation (3). In other words, whenever we see `\sum\mu_prime` during our solution, we can replace it with 0. Please let me know  if my answer is not clear.



> Btw, `mu^prime(0)` looks as if you are evaluating `mu^prime` as a function. Instead, you may want to use a `\cdot` instead to show that it is a multiplication

Correct! I will change it in the next set of commits. Thanks!

NimaSarajpoor: I can feel it now. So, I will change the structure. I will start expanding (3) right after eq(3). THEN, I can show the partial derivative  of  function w.r.t. to the other variable (i.e. `\sigma_prime`) followed by its expansion. I will revise  it in the next set of commits.

seanlaw: <blockquote>Please note that (1) and (2) are a single system of equations. Please let me know if my answer is not clear.</blockquote><p> </p><p>Understood. You'll have to forgive me as it's been a long time since I've had to think about this :) It would be helpful to add this reminder to the notebook</p>

seanlaw: <p>Yes. Thanks</p>

seanlaw: <p>Cool. I will have to find some time to read that part</p>

NimaSarajpoor: I couldn't find an easier way to solve it. Right after the first line, I even replaced one of parenthesis with its equivalent term `alpha` so I can use (3) and (4) to simplify it further before plugging the calculated values `\mu_prime` (8) and `\sigma_prime` (9) in `f(\mu_prime , \sigma_prime)`. 

I could have simply plugged in the values (8) and (9) to directly find `f_min` right from beginning of this cell; however, I think, that approach will make it more complex and harder to follow. 

---

I will address the previous comments. Do you want me to wait till you review this part as well?

NimaSarajpoor: Sure...I will add a note regarding this matter.

seanlaw: Yeah, why don't you take some time to see if it can be simplified before I go over it. No rush

seanlaw: <p>I was just thinking about this a little bit more and your math/engineering skills are really shining through here and I get to learn from you too! So, thank you. You've reminded me that, conceptually, we are looking for the global minima in 2-dimensional space and hence why the partial derivatives form a system of equations. This is fun :)</p>

NimaSarajpoor: It IS fun indeed! 

> conceptually, we are looking for the global minima in 2-dimensional space and hence why the partial derivatives form a system of equations.

Exactly!!!

NimaSarajpoor:
@seanlaw 
I was wondering if you could confirm my understanding of Welford's notebook [Welford_Review](https://github.com/TDAmeritrade/stumpy/blob/main/docs/Welford_Review.ipynb):

In line (23), the `S(...)` is actually CSS as its title said "The Corrected Sum of Squares". Correct? And, it is different than the standard deviation `S` in line (123).

docs/Tutorial_VALMOD.ipynb
NimaSarajpoor: I think breaking it down to smaller parts and reminder reader of some previously-calculated equations can improve readability. I also think I skipped a few steps but I believe it can help reader to follow the process easier. So, if you do not mind, please ignore this part and check it out in next set of commits.

(I can also try to see if there is a different way to get to calculate `f_min`)

seanlaw:
> @seanlaw I was wondering if you could confirm my understanding of Welford's notebook [Welford_Review](https://github.com/TDAmeritrade/stumpy/blob/main/docs/Welford_Review.ipynb):
> 
> In line (23), the `S(...)` is actually CSS as its title said "The Corrected Sum of Squares". Correct? And, it is different than the standard deviation `S` in line (123).

Yes, this seems to be correct.  Welford referred to his "Corrected Sum of Squares" as `S` and so the initial set of derivations was for Welford's paper. However, once this was completed, I tried to switch over to STUMPY notation in the "Deriving Rolling Window Variance (or Standard Deviation) with Welford's Method" so that everything was "easier" to follow. Thus, the first half was confirming Welford's equations and then the second half was applying Welford's equations in the context of STUMPY notation.

> Also...do you know the relationship between `\sigma` of length `m+1`, and `\sigma` of length `m`, both subsequences starting from the same index? I am planning to derive it but I want to know if such relationship is already discovered (so that I do not spend time on it)

Are you referring to Welford still here? If so, I believe that it is still referred to as [Welford's Online Algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) or [computing a running variance](https://www.johndcook.com/blog/standard_deviation/) (note that you can compute stddev by taking the square root of the variance). It's been a while since I've touched it but I realized that the basic concept for computing a "rolling variance" (i.e., the size of the window always stays the same) versus computing a "running variance" (i.e., the size of the window always increments by one) is pretty similar. In STUMPY, we implemented the "rolling variance" (see `welford_nanvar` in `core.py`).  Let me know if this answers your question

Please note that these rolling/running variance calculations are VERY tricky and is very susceptible to "catastrophic cancellation" and so one needs to take care to check the numerical stability!!!

seanlaw:
> Thanks for the info! That helps a lot. What I am looking for is "running variance". So, I will try to be extra careful in my calculation and provide the steps in the notebook so you can check them out as well. (I think finding that relationship should help me with finding LB for non-positive Pearson Correlation \rho.)

@NimaSarajpoor Please re-read my response one more time as I was editing it as your were reading. Especially the first paragraph!

seanlaw:
> So, I have running variance! Cool! Thanks!

Exactly! The running covariance is the superset of running variance. I'm going to add a note to the Welford Review notebook to make it clear that one is "running" and the latter is "rolling" and explain where each one is used in STUMPY

NimaSarajpoor:
I revised the notebook according to latest comment. I could not find an easier way for plugging in the values of critical point in `f(...)`. I added a few more steps and some explanation. Please let me know what you think and if it is easier to follow now.

docs/Tutorial_VALMOD.ipynb
```diff
@@ -0,0 +1,1344 @@
...
```
seanlaw, line 1: <p>Why not perform/declare this substitution above when you mention the substitution for <code>\sigma^prime</code> and <code>\mu^prime</code> ? I feel like this can be consolidated a bit nicer and more consistently?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

seanlaw, line 1: <p>Use a colon in "equations below:" rather than a period</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='32'/>

seanlaw, line 1: <p>I think this should be "Solving" and not "Deriving"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

seanlaw, line 1: <p>Solving?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='40'/>

seanlaw, line 1: <p>Please add a comment to remind the reader that the sum of <code>\alpha_t</code> is zero from equation (7) </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

seanlaw, line 1: <p>Also, the first line is too hard to follow since you're performing a factorization AND a multiplication at the same time. Can you please split up those two steps?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='46'/>

seanlaw, line 1: <p>Ummm, I think you should demonstrate how to get to this as this is not the standard representation of Pearson correlation! Typically, it is the covariance divided by the standard deviations of each subsequence. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='48'/>

seanlaw, line 1: <p>I don't think you mean "rewrite" but, instead, you mean "rearrange"?</p><p> </p><p>I had a hard time understanding when <code>i = j</code> becomes relevant until below. I think you should just substitute the rearrange Pearson correlation into (*)  first and then remind the reader that <code>\rho = 1 </code>  after the substitution</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='49'/>

seanlaw, line 1: <p>It took me quite some effort to see that you rearranged the terms and then multiplied the first term through. Please split up the steps so it is more obvious. Otherwise, it is hard to follow</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='52'/>

seanlaw, line 1: <p>I think it would be good to show the two equations (8) and (10) here and remind the reader as to what the knowns/unknowns are in this system of equations. And then show the derivation for solving them.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='53'/>

seanlaw, line 1: <p>We should make a statement here to tie back to the original goal, which is that we wanted to express these terms relative to known values in order to get a proper lower bound</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='54'/>

seanlaw, line 1: <p>Very suspenseful!  :)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='55'/>

seanlaw, line 1: <p>What does the subscript <code>c</code> mean here??  Ohh, "critical point"??</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='58'/>

seanlaw, line 1: <p>Wow, this derivation was a lot of work! I'm surprised/impressed that the original authors went through all of this trouble!</p><p> </p><p>Thank you for working through the math</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='66'/>

seanlaw, line 1: <p>Cool. Thanks for the note. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='69'/>

seanlaw:
@NimaSarajpoor Given the simplicity of the lower bound equations, I think that we should separate/split the notebook derivation for the lower bound equations from the VALMOD notebook (i.e., where we apply VALMOD). So, in the VALMOD notebook, we assume that the lower bound equations in the paper is correct and this is backed up by our LB derivation notebook.  Does that make sense? At the end of the day, the VALMOD user should not need to care about the LB derivation.

docs/Tutorial_VALMOD.ipynb
NimaSarajpoor: <p>I see. So, if I understand correctly, you mean.... put eq(3) right after the "note" above. Correct? And then, I can provide equations in this order: (1), (4), (2). </p>

NimaSarajpoor: <p>I will correct it.</p>

NimaSarajpoor: <p>It is actually both... we derive partial derivative, and then at the end, we try to solve it when we replace left-hand side with 0.</p><p> </p><p>But, as you suggested, I think "Solving" is better as it is what we are doing at the end of the day. I will change it in the next set of commits. </p>

NimaSarajpoor: <p>will change it.</p>

NimaSarajpoor: <p>Thanks for the suggestion. so, I will add the reminder just before the last two equations.</p>

NimaSarajpoor: <p>Yes. So, I can first keep the fraction format and just distribute the <code>\sum</code> into the numerators of fractions, and THEN, provide the rest of equations. I will add this step before the first line.</p>

NimaSarajpoor: I used line(14) and line(22) of the notebook [Matrix_Profile_Derivation](https://github.com/TDAmeritrade/stumpy/blob/main/docs/Matrix_Profile_Derivation.ipynb) to notice that the `q` in the paper is actually Pearson Correlation \rho.  I will add this note to the notebook. Does that suffice?

NimaSarajpoor: > I don't think you mean "rewrite" but, instead, you mean "rearrange"?

Yes. I think rearrange is the correct wording here! You have eyes of an eagle when you review something as you can see all small details.. Very cool!


> I had a hard time understanding when `i = j` becomes relevant until below. I think you should just substitute the rearrange Pearson correlation into (*) first and then remind the reader that `\rho = 1 ` after the substitution

Yes...it feels things are not connected here. I can use `\rho_i,j` and `\rho_i,i`. And, then replace the latter with 1 followed by a note to the reader.

NimaSarajpoor: <p>I will elaborate it.</p>

NimaSarajpoor: Sure. I will add them to the notebook.

NimaSarajpoor: <p>I got your point. I will add a note regarding this matter.</p>

NimaSarajpoor: Yes! But, actually, I should have removed subscript `c` as you can see I removed them in the rest of equations. I used it first and then I thought it is not necessary as the reader should know that the values `\mu_prime` and `\sigma_prime` are NOT just ANY values anymore, but in fact, they are the calculated values for the critical point.

What do you think?

NimaSarajpoor: I have the same thought... that the authors were not afraid of the equations, and continued solving this problem.

NimaSarajpoor: I haven't performed the second-derivative test yet. 

I tried to derive LB for `q<0` but I have not been successful so far. I tried different inequalities but those did not help.

NimaSarajpoor:
> Given the simplicity of the lower bound equations, I think that we should separate/split the notebook derivation for the lower bound equations from the VALMOD notebook (i.e., where we apply VALMOD).

I totally agree with this idea as every thing will be in its own place and it can result in two cleaner notebooks. So, should I change the title of this notebook / or submit a new PR? Or, do you want to take care of it yourself?

docs/Tutorial_VALMOD.ipynb
seanlaw: <p>Yes, something like that</p>

seanlaw: <p>I think one should start with a standard definition of <code>\rho</code> (see wikipedia) and then manipulate it into the form below </p>

seanlaw: <p>Sure. That is probably fine</p>

seanlaw:
> So, should I change the title of this notebook / or submit a new PR? Or, do you want to take care of it yourself?

I've updated the title of this PR and so please feel free to change the title of this notebook and add another notebook for the application of VALMOD

docs/Tutorial_VALMOD.ipynb
NimaSarajpoor: Got it. Will do it.

NimaSarajpoor: The formula I used in the notebook is actually eq(4) of this paper: [paper](https://www.cs.unm.edu/~mueen/Projects/JOCOR/joinICDM.pdf). I was wondering if I should better provide the reference rather than deriving it (which is 2 line at the most...but I think providing ref reduces the noise in the notebook and distracts readers less.)

What do you think?

NimaSarajpoor: Sure. I think it is similar to software where we want to have minimum dependencies. I will add the calculation.

NimaSarajpoor:
I think we have discussed all notes. I will push commits (I will push two notebooks).

seanlaw:
Sounds good. Thank you

NimaSarajpoor:
@seanlaw 
What I have done:
* I addressed the comments to make the calculations more readable.
* I successfully derived LB for `\rho < 0`. 
* I also showed the local minimum is global minimum. 
* I created another notebook (But, I do not know why I cannot see it here after `git push`.) I git `add/commit` that new notebook.)

Please allow me to check out my work via ReviewNB and also tried to see if I can git push the other notebook. I will let you know shortly when I am done.

NimaSarajpoor:
@seanlaw 
**UPDATE**
I think it is all good:
* I checked out the notebook via ReviewNB. (It seems I somehow skipped numbers (13)/(14) when I was numbering equations. I will fix that in the next set of reviews if that is okay with you. For now, it does not create confusion as I never refer to those numbers later in the notebook.)
* I resolved the issue with the other notebook. However, I can `git push` it next time if you do not mind.

Please feel free to review.

NimaSarajpoor:
@seanlaw 
Ahhh...Now I see your comments are shown as "Outdated"...I think it is because I changed the name of notebook and did `git rm` for the old one, and `git add` for the new one...we lost the previous comments. If you have any suggestion regarding this matter, please let me know.

seanlaw:
I have no better suggestion but I think we are fine for now.

docs/LowerBound_Dist_Profile_Derivation.ipynb
```diff
@@ -0,0 +1,2319 @@
+{
```
seanlaw, line 1: <p>Equation (4) seems redundant here. Also, I think it would be better to transform equation (1) by first substitution <code>\alpha_t</code> and then to <code>f</code>   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

seanlaw, line 1: <p>Fix spelling of "length", "subsequences"</p><p> </p><p>Also, I recommend "between two subsequences starting at locations <code>i</code> and <code>j</code> , respectively, and both of length <code>m</code> is defined as follows"  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

seanlaw, line 1: <p>This is great!</p><p> </p><p>Eventually, we will need to convert all of these equations to base-zero indexing</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

seanlaw, line 1: <p>Excellent!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

seanlaw, line 1: <p>Fix spelling of "subsequence"</p><p> </p><p>I would move this comment up by two cells so that it is closer to the <code>\rho_ii</code>  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

seanlaw, line 1: <p>Excellent!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='57'/>

seanlaw:
@NimaSarajpoor I haven't gone over the second part of the derivation (where the pearson correlation is less than or equal to zero) but I don't have any reason to believe that it is wrong. At the end of the day, for VALMOD, we're planning to use the final LB equations. Thank you for putting in the time and work to go through the derivations

docs/LowerBound_Dist_Profile_Derivation.ipynb
NimaSarajpoor: I will change it,

NimaSarajpoor: If I may ask, why do you think these equations are not base-zero indexing? You meant something like numpy indexing, right? 

I think it can be base-zero indexing. `i` is the start index of `T_i,m`. So, if someone has base-zero indexing in mind and set `i=0`, then the elements we are including in the `\sum \limits_{t=1}^{m} T[i+t-1]` will be from `0` to `m-1`.

NimaSarajpoor: I just checked out the Welford Notebook...Got your point! Cool!

NimaSarajpoor: I will fix those and revise the sentence.

NimaSarajpoor: Got it. I will provide the comment right after when `\rho_ii` appears in the calculation.

NimaSarajpoor:
> but I don't have any reason to believe that it is wrong.

Yeah. So, do you think we should better remove the second part (LB for `\rho <= 0`) and the third part (LB for `rho>0` is global minimum) of the notebook to make it cleaner?


> Thank you for putting in the time and work to go through the derivations

Thanks for your support and feedback!

seanlaw:
> Yeah. So, do you think we should better remove the second part (LB for `\rho <= 0`) and the third part (LB for `rho>0` is global minimum) of the notebook to make it cleaner?

No, I would leave it. It's good to have

docs/LowerBound_Dist_Profile_Derivation.ipynb
seanlaw: <p>Great! It just makes it easier for people to convert it into code if they need it. In our case, it's probably not an issue because we are only going to use the final LB equations, which are pretty straightforward unlike the matrix profile or Welford equations (for MPX).</p>

seanlaw: <p>Thanks</p>

NimaSarajpoor:
@seanlaw 
The notebook is ready.

seanlaw:
@NimaSarajpoor I think things look good.

NimaSarajpoor:
@seanlaw 
Great! Thanks for checking it out in your busy schedule. So, I will continue working on the implementation of VALMOD.

NimaSarajpoor:
@seanlaw 
please free to review. 
* I improved `section 2: Lower Bound of Distance Profile`
* I added `section 3: Core idea` to briefly explain VALMOD
* I added `section 4: VALMOD algorithm`. I implemented algorithm3 of paper.

docs/Tutorial_VALMOD.ipynb
```diff
@@ -0,0 +1,659 @@
...
```
seanlaw, line 1: <p>Spelling mistake in "VALMOP"</p><p> </p><p>Do you mean "bottom-up" approach?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/586/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

NimaSarajpoor: <p>Yes. I will fix both.</p>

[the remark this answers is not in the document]
NimaSarajpoor: <p>I will change <code>n</code> to <code>k</code> in next set of commits.</p>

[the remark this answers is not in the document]
NimaSarajpoor: <p>Good to know! I will replace <code>heapq</code> with <code>np.searchsorted</code>.</p>

NimaSarajpoor:
@seanlaw 
> We basically need to spend some time figuring out how to allow stumpy.stump , stumpy.stumped , and stumpy.gpu_stump return top-k nearest neighbors. 

So, should I now go and study `stump`/`stumped`/`gpu_stump`? And, then try to change all of them to return `top-k` nearest neighbors?

seanlaw:
@NimaSarajpoor  Yes, I also added a new issue #592 where we can discuss it in more detail

NimaSarajpoor:
A couple of notes that are confirmed by the main author of [VALMOD](https://arxiv.org/pdf/2008.13447.pdf):
* On Page 13,  in `Algorithm 2: updateV ALMP`, line 3 should have been: `if VALMP.normdistances[i] > lNormDist or ...`
* On page 16, the `Algorithm 4: ComputeSubMP` can be further optimized by updating `minDistABS` in the for-loop (line 28)

(This is to make sure that we do not lose this information later)

seanlaw:
> So, should I now go and study stump/stumped/gpu_stump? And, then try to change all of them to return top-k nearest neighbors?

We've come along way @NimaSarajpoor! I wonder how easy/hard it would be to implement VALMOD now that we have top-k nearest neighbors?

NimaSarajpoor:
@seanlaw 

> We've come along way

We have indeed!

> I wonder how easy/hard it would be to implement VALMOD now that we have top-k nearest neighbors?

I took a quick look at the paper. I don't remember the details but I think the first four algorithms are the core ones. The first two algorithms are easy. The third one is already done (the top-k feature added to STUMPY). In my opinion, the main remaining task is algorithm 3. I think its implementation should be straightforward.

NimaSarajpoor:
I think the algorithm presented in a paper has  a flaw. (I sent an email to the main author and am waiting for his response.)

In the page14 of the paper, the following can be read:

```
Algorithm 2 shows the routine to update the VALMP structure. The final VALMP consists of four parts. The `i` the entry of the normDistances vector stores the smallest length-normalized Euclidean distance values between the ith subsequence and its nearest neighbor, while the ith place of vector distances stores their straight Euclidean distance. The location of each subsequence’s nearest neighbor is stored in the vector indices. The structure lengths contains the length of the i th subsequences pair.
```

Let's assume `P` denotes the variable-length matrix profile obtained by VALMOD. According to my investigation `P` cannot be exact. Athough `min(P)` is exact,   `P` itself is not exact. (see algorithm 4, line 29)

PR #590, opened 2022-04-14
FIX #589

### LINKED ISSUE #589: inconsistent data type in Tutorial_DiscordMERLIN.ipynb
I was reviewing the [MERLIN NOTEBOOK](https://github.com/TDAmeritrade/stumpy/blob/main/docs/Tutorial_DiscordMERLIN.ipynb) and I noticed that the output of function `_discords` doesn't show consistent data type when it is appended by `[-1, np.inf, -1]` with `dtype=object` .

Apparently I should have defined `dtype` of each column separately and then concatenate them with `dtype=object`. 

If you think we can ignore this minor(?) issue at this stage, please let me know. If you think we should take care of it, I can try to fix it; however I am not sure if there is any clean way to  do this.

NimaSarajpoor:
This fixes issue #589. As suggested, I used pre-filled arrays with dummy values.

- In function `_refine_candidates`: return lists instead of one ndarray
- In function `_discords`: use arrays pre-filled with dummy values
- In function `_discords`: use indexer

---

**Note:** we now have if-for-if/break structure in the while-loop. You may not like it. So, if you do not like it, we may use lists instead of arrays, and as you said before, just blindly append to it and then clean it up at the end.

### DIFF
--- a/docs/Tutorial_DiscordMERLIN.ipynb
+++ b/docs/Tutorial_DiscordMERLIN.ipynb
@@ -183,7 +183,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "id": "a40d0bd2",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


seanlaw:
@NimaSarajpoor Looks great. Thank you

PR #597, opened 2022-04-30
Fixed #596 Rename naive.mass to naive.mass_PI

### LINKED ISSUE #596: Rename functions `naive.mass` and `naive.distance_profile`
It seems that the function `naive.mass` is actually the naive version of function `stumpy.stamp._mass_PI` (NOT the `stumpy.core.mass`) Therefore, it might be a good idea to do the following things:

(1) Rename function `naive.mass` to `naive.mass_PI`
(2) Rename function `naive.distance_profile` to `naive.mass`

So... `naive.stamp` calls function `naive.mass_PI`, and the function `naive.mass_PI` calls function `naive.mass`.

NimaSarajpoor:
Not sure if it is what you had in mind or not, but I only renamed `naive.mass` to `naive.mass_PI`

---

**I did NOT rename `naive.distance_profile` to `naive.mass` because:**
* I think `naive.distance_profile` handles `nan/inf` differently.
* I think the name `distance_profile` shows the purpose of function better.

### DIFF
--- a/tests/naive.py
+++ b/tests/naive.py
@@ -94,7 +94,7 @@ def aamp_distance_matrix(T_A, T_B, m, p):
     return distance_matrix
 
 
-def mass(Q, T, m, trivial_idx=None, excl_zone=0, ignore_trivial=False):
+def mass_PI(Q, T, m, trivial_idx=None, excl_zone=0, ignore_trivial=False):
     Q = np.asarray(Q)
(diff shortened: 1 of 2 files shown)


seanlaw:
Sounds good.

seanlaw:
@NimaSarajpoor  Looks good. Thank you

PR #599, opened 2022-05-06
Fixed #598 Add parameter row_wise to naive.stump

### LINKED ISSUE #598: Replace `naive.stamp` with `naive.stump`, with new parameter `row_traversal`
As discussed in PR #595, we would like to replace `naive.stamp` with `naive.stump(...., row_traversal=True)`. It seems that the function `naive.stamp` is used in several unit test files. To avoid pushing all these files, I created this issue so that we can address it first before moving forward in the PR, and make sure it passes all unit tests.

NimaSarajpoor:
This PR addresses issue #598. I added parameter `row_wise` to `naive.stump` and, then, instead of calling `mass_PI`, I simply traverse the matrix and update matrix profile and indices.

* Btw, at first, I used the name `row_traversal` for the new parameter. But  then, I thought your proposed name (i.e. `row_wise`) is good!

### DIFF
--- a/tests/naive.py
+++ b/tests/naive.py
@@ -198,7 +198,7 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
                 idx = -1
             I[i, 0] = idx
 
-            #self-join: left matrix profile
+            # self-join: left matrix profile
             IL = -1
             IR = -1
             if ignore_trivial and i > 0:
@@ -207,9 +207,9 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
                     IL = -1
                 I[i, 1] = IL
 
-            #self-join: right matrix profile
+            # self-join: right matrix profile
             if ignore_trivial and i < D.shape[0]:
-                IR = i + np.argmin(D[i:]) #shift argmin by `i` to get true index
+                IR = i + np.argmin(D[i:])  # shift argmin by `i` to get true index
                 if D[IR] == np.inf:
                     IR = -1
                 I[i, 2] = IR
--- a/tests/naive.py
+++ b/tests/naive.py
@@ -139,7 +139,6 @@ def mass_PI(Q, T, m, trivial_idx=None, excl_zone=0, ignore_trivial=False):
     return P, I, IL, IR
 
 
-# just for reference
 def stamp(T_A, m, T_B=None, exclusion_zone=None):
     if T_B is None:  # self-join
         result = np.array(
--- a/tests/naive.py
+++ b/tests/naive.py
@@ -160,9 +160,9 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
     """
     Traverse distance matrix and update the matrix profile and
     matrix profile indices
-    
-    If `row_wise=False`, then it is a diagonal traversal.
-    If `row_wise=True`, then it is a row-wise traversal.
+
+    If the parameter `row_wise` is set to `False`, then it is a diagonal traversal.
+    If the parameter `row_wise` is set `True`, then it is a row-wise traversal.
     """
     if T_B is None:  # self-join:
         ignore_trivial = True
--- a/tests/naive.py
+++ b/tests/naive.py
@@ -184,6 +184,7 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
     P = np.full((l, 3), np.inf)
     I = np.full((l, 3), -1, dtype=np.int64)
 
+    result = np.empty((l, 4), dtype=object)
     if row_wise:  # row-wise traversal in distance matrix
         if ignore_trivial:  # self-join
             for i in range(l):
@@ -213,11 +214,8 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
                     IR = -1
                 I[i, 2] = IR
 
-        result = np.empty((l, 4), dtype=object)
         result[:, 0] = P[:, 0]
         result[:, 1:4] = I[:, :]
-
-        return result
 
     else:  # diagonal traversal
         if ignore_trivial:
@@ -253,11 +251,10 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
                             P[i, 2] = D
                             I[i, 2] = i + k
 
-        result = np.empty((l, 4), dtype=object)
         result[:, 0] = P[:, 0]
         result[:, 1:4] = I[:, :]
 
-        return result
+    return result
 
 
 def aamp(T_A, m, T_B=None, exclusion_zone=None, p=2.0):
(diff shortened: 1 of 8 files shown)


seanlaw:
@NimaSarajpoor Did all of the tests pass locally for you?

tests/naive.py
```diff
@@ -139,6 +139,7 @@ def mass_PI(Q, T, m, trivial_idx=None, excl_zone=0, ignore_trivial=False):
     return P, I, IL, IR
 
 
+# just for reference
```
seanlaw, line 142: Please remove this comment

tests/naive.py
```diff
@@ -181,47 +182,83 @@ def stump(T_A, m, T_B=None, exclusion_zone=None):
...
+                IR = i + np.argmin(D[i:])  # shift argmin by `i` to get true index
+                if D[IR] == np.inf:
+                    IR = -1
+                I[i, 2] = IR
+
+        result = np.empty((l, 4), dtype=object)
+        result[:, 0] = P[:, 0]
+        result[:, 1:4] = I[:, :]
+
+        return result
```
seanlaw, line 221: Don't return in the middle of the if statement. Please initialize `result` before the first (outermost) `if`, then fill it, and return `result` at the end of the function.

tests/naive.py
```diff
@@ -181,47 +182,83 @@ def stump(T_A, m, T_B=None, exclusion_zone=None):
...
+                        if D < P[i + k, 1]:
+                            P[i + k, 1] = D
+                            I[i + k, 1] = i
+
+                        if D < P[i, 2]:
+                            # right matrix profile and right matrix profile index
+                            P[i, 2] = D
+                            I[i, 2] = i + k
+
+        result = np.empty((l, 4), dtype=object)
```
seanlaw, line 257: See comment above.

tests/test_gpu_stump.py
```diff
@@ -44,7 +44,7 @@ def test_gpu_stump_int_input():
 def test_gpu_stump_self_join(T_A, T_B):
     m = 3
     zone = int(np.ceil(m / 4))
-    ref_mp = naive.stamp(T_B, m, exclusion_zone=zone)
+    ref_mp = naive.stump(T_B, m, exclusion_zone=zone, row_wise=True)
```
seanlaw, line 47: I think this looks great! Just by looking at the `naive.stump` input parameters it is obvious what is happening.

tests/naive.py
```diff
@@ -156,7 +157,7 @@ def stamp(T_A, m, T_B=None, exclusion_zone=None):
     return result
 
 
-def stump(T_A, m, T_B=None, exclusion_zone=None):
+def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
     """
     Traverse distance matrix along the diagonals and update the matrix profile and
     matrix profile indices
```
seanlaw, line 163: Can you please add an additional note that when `row_wise=True` then what happens?

seanlaw (changes requested):

tests/naive.py
NimaSarajpoor: Will do

tests/naive.py
NimaSarajpoor: Sure. I will add note in this doc string.

tests/test_gpu_stump.py
NimaSarajpoor: Your are right! It is easy to understand what is going on.

tests/naive.py
NimaSarajpoor: Sure. I can do that. I will revise it.

NimaSarajpoor:
> Did all of the tests pass locally for you?

I think so. The message said some files are skipped. Did I misunderstand that? Should I expect 100% all the time? Since it did not show "Failure", I think things are ok. I ran the test again, and here is the message:

```
====================== 1298 passed in 2987.10s (0:49:47) ======================
Name             Stmts   Miss  Cover   Missing
tests\naive.py    1032      8    99%   120-125, 137, 144-157
TOTAL            11314      8    99%

79 files skipped due to complete coverage.
Cleaning Up

```

seanlaw:
> tests\naive.py    1032      8    99%   120-125, 137, 144-157

So, it appears that by replacing `naive.stamp` with `naive.stump`, we are not executing the lines `120-125` and line `137` in the `mass_PI` function of `naive.py` as well as in lines `144-157` of `stamp`. However, I'm not too concerned about this. Can you please add a:

```
# pragma: no cover
```

at the end of line 119, line 136, and line 143 in `naive.py`?

NimaSarajpoor:
I will add the `# pragma: no cover` in those lines. I think we have discussed all comments. I will apply changes and push them shortly.

tests/naive.py
NimaSarajpoor: @seanlaw 

> Please initialize result before the first (outermost) if, then fill it, and return result at the end of the function.

Just out of curiosity, is there any particular reason for not doing all of that at the end of function? 

```
# in tests/naive.py
def stump(...):
    # at the end
    result = np.empty((l, 4), dtype=object)
    result[:, 0] = P[:, 0]
    result[:, 1:4] = I[:, :]
    
    return result
```

seanlaw: Yes, at the end is fine too and is probably "better" in this case

seanlaw:
Great thank you!

NimaSarajpoor:
@seanlaw 
Please feel free to review.

tests/naive.py
```diff
@@ -156,10 +156,13 @@ def stamp(T_A, m, T_B=None, exclusion_zone=None):
...
 
 
-def stump(T_A, m, T_B=None, exclusion_zone=None):
+def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False):
     """
-    Traverse distance matrix along the diagonals and update the matrix profile and
+    Traverse distance matrix and update the matrix profile and
     matrix profile indices
+
+    If the parameter `row_wise` is set to `False`, then it is a diagonal traversal.
```
seanlaw, line 164: The first letter should be lowercase in this `If` since it continues the sentence.

seanlaw (changes requested):

tests/naive.py
NimaSarajpoor: I actually forgot to end the previous sentence with period (`.`)...So:

```
Traverse distance matrix and update the matrix profile and
matrix profile indices. # the sentence is  ended here

# new sentence
If ...
```

---

I think what you are suggesting is this:
```
Traverse distance matrix diagonally and update the matrix profile and 
matrix profile indices if the parameter `row_wise` is set to `False`. 
If the parameter `row_wise` is set `True`, then it is a row-wise traversal.
```

Should I go with the latter one?

seanlaw: Sure

PR #603, opened 2022-05-12
Fix #591 Handling Constant Regions

### LINKED ISSUE #591: Handle Constant Regions Consistently
Currently, `stumpy.stump` and `stumpy.stumped` can account for constant regions. So, when two subsequences are being compared and one subsequence is constant (and the second is not constant) then the pearson correlation is set to `0.5`, which means that the z-normalized Euclidean distance is `np.sqrt(np.abs(2 * m * (1 - 0.5)))` or `np.sqrt(m)` (<s>not sure if this is the best/correct choice</s> see proof below). However, when both subsequences are constant then the pearson correlation is set to `1.0`, which means that the distance is `0.0`

However, this is unaccounted for in `stumpy.mstump` and `stumpy.mstumped` and `stumpy.gpu_stump` and `stumpy.mass` and some `tests/naive.py` implementatios. This may be something we should consider doing consistently everywhere.

<details>
<summary>changes to `_mass`</summary>

Replace the return statement with:
```
distance_profile = calculate_distance_profile(m, QT, μ_Q, σ_Q, M_T, Σ_T)
if σ_Q == 0.0:
    # Both subsequences are constant
    distance_profile[Σ_T == 0.0] = 0.0
else:
    # Only one subsequence is constant
    distance_profile[Σ_T == 0.0] = math.sqrt(m)

return distance_profile
```
</details>

[This paper](https://core.ac.uk/download/pdf/287941767.pdf) may offer some direction/insight as to how best to handle this.

NimaSarajpoor:
This addresses Issue #591. I tried to open the notebook in incognito mode to see if you can open it later. But, I couldn't open it in that mode. So, I am sending this PR so you can see it. Feel free to remove it after reading it.

### DIFF
--- a/docs/Constant_Subseq.ipynb
+++ b/docs/Constant_Subseq.ipynb
@@ -1,0 +1,285 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "3e98146e",
+   "metadata": {},
+   "source": [
+    "In matrix profile obtained with z-normalized Euclidean Distance, one challenge is to handle a subsequence that contain constant values. "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6937c2bd",
+   "metadata": {},
+   "source": [
+    "Let us consider two subsequences T and T'. If both are constant subsequences, then their distance is 0 (because they are similar). But, what if only one of them, e.g. T, is constant, and the other one, T', is not constant? "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "68fe0283",
+   "metadata": {},
+   "source": [
+    "$T = \\{a, a, ..., a\\}$, with length `m`. <br>\n",
+    "$T' = \\{T'_{0}, T'_{2}, ..., T'_{m-1}\\}$"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "06ed366c",
+   "metadata": {},
+   "source": [
+    "Therefore, their z-normalized versions are:"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a619978f",
+   "metadata": {},
+   "source": [
+    "\n",
+    "$$\n",
+    "\\begin{align}\n",
+    "T_{z} ={}&\n",
+    "    \\frac{\n",
+    "    T - \\mu_{T}\n",
+    "    }{\n",
+    "    \\sigma_{T}\n",
+    "    } \\quad (1)\n",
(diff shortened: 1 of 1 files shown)


NimaSarajpoor:
> I'm not sure how I feel about this part as, realistically, everything hinges upon this

I think it is what you did in your simulation. So. for instance, when `a = np.full(m, 10)` (constant subseq with value of 10). the average is 10. right? so, `(a - a.mean())/a_std)` becomes 0 since its numerator is 0.

(You name the subsequence `a` that contains values of 10. In the notebook, I used the  name `T` for the constant subsequence and it contains the values of `a`)

seanlaw:
Yes, you're right. So, it's one of those things where, definitely, that fraction should evaluate as "undefined" when the stddev is zero (i.e., the point is not differentiable at stddev == 0.0) and, essentially, we are approximating the fraction by taking the limit as stddev approaches 0.0. Technically, `sqrt(m)` is "wrong" when `stddev == 0.0` but we're sort of accepting that it is a close approximation even though it is undefined

NimaSarajpoor:
@seanlaw 
That is a good reasoning, and taking limit with `std-->0` makes sense!!! 

Also, I would like to reiterate this fact that assuming the vector `0` for the z-normalized version of a constant subsequence can make sense in my head because: 

For a non-constant vector, we cannot have **`0`**(i.e. vector `0` with length `m`) as its z-normalized version. Any non-constant subsequence, after z-normalized transformation, can be mapped to a point in `R^{m}` space but not **`0`** ! 

Now, if we consider **`0`** as the z-normalized version of a constant subseq, then we fill the origin point in the `R^{m}`.

seanlaw:
> Now, if we consider 0 as the z-normalized version of a constant subseq, then we fill the origin point in the space R^{m} that was being missed before for non-constant subsequences!

I see, I see. So, the infinite set of non-constant subsequences would fill the entire space except for the origin, where there would be a hole and we are filling that hole with our constant subsequence?

NimaSarajpoor:
> So, the infinite set of non-constant subsequences would fill the entire space except for the origin, where there would be a hole and we are filling that hole with our constant subsequence

Exactly! That is how I see the z-normalized space in my head! (Just trying to find different justification for vector **`0`** as the z-normalized version of constant subsequences)

seanlaw:
> Exactly! That is how I see the z-normalized space in my head! (Just trying to find different justification for vector 0 as the z-normalized version of constant subsequences)

Fun! Your math/conceptual understanding of high-dimensional space is wayyy better than mine 😄

PR #614, opened 2022-06-02
Revised docstrings (as a follow up to PR #613)

NimaSarajpoor:
Revised docstrings


seanlaw:
Thank you @NimaSarajpoor

PR #595, opened 2022-04-28
[WIP] Add Top-K Nearest Neighbors  for Normalized Matrix Profile #592

NimaSarajpoor:
This PR addresses issue #592. In this PR, we want to extend the function `stump` and the related ones so that it returns Top-K Nearest Neighbors Matrix Profile (i.e. the k smallest values for each distance profile and their corresponding indices).

What I have  done so far:
* Wrote a naïve implementation of `stump_topk` 
* Provided a new unit test for function `stump` 

---
**Notes:**
(1) `np.searchsort()` is used in the `naive.stump_topk`. Another naive approach can be traversing distance matrix row-by-row, and using `np.argpartition()` followed by `np.argsort()'.

(2) I think considering the first k columns in the output of `stump_topk` is cleaner, and also easier for the user to get access to those top-k values. 

I can/will think about a clean way to change the structure of output just before returning it so that the  first four columns becomes the same for all k. However, I think that makes it hard to use the output later in other modules.

---

**Add To-Do List**

- [x] Add parameter `k` to `stump` 
- [x] Add parameter `k` to `stumped`
- [x] Add parameter `k` to `gpu_stump`
- [x] Add parameter `k` to `scrump`
- [x] Add parameter `k` to `stumpi `
- [ ] Run published tutorials to see if they still work with no problem 

----

Side notes:
* Add parameter `k` to non-normalized versions.
* Check docstrings for functions that require matrix profile `P` as input. For instance, in `stumpy.motifs`, we may say something like... "`P` is (top-1) 1D array matrix profile"

PR #616, opened 2022-06-06
Topk_matrixprofile Investigator

NimaSarajpoor:
This PR is to investigate the time-out failure in PR #595. 

This is how I created this branch:
```
# in branch main
git branch -c [old branch] [new branch]
git checkout [new branch]

# so, while  I am in this new branch
git reset --hard HEAD~[number]
```

Note that I did not `git merge main` after this. It seems the changes in `main` are preserved already. So, later, we can skip the commits regarding the `git merge main`.


NimaSarajpoor:
@seanlaw 
Apparently, the `git branch -c` does not create a completely separated branch, and it seems the new branch is created off the old branch. Please see screenshot below.

image

So, I think it would be similar:
```
git checkout -b [new_branch]  76f97cb # which is the highlighted commit in image above
```

I think this should work for us. I mean...I can now do `git cherry-pick [commit] ` (or any other approach?!) to add commits one by one from this point. I hope this works.

seanlaw:
@NimaSarajpoor In its current state, can you help me understand how this PR is different/same as PR #595? It looks to me like the files are all the same?

NimaSarajpoor:
> In its current state, can you help me understand how this PR is different/same as PR #595? It looks to me like the files are all the same?

So, I did not start from scratch. I knew that a few `push`es ago, things were good. So, now my NEW branch is at that point were things were good. Even if the files look like the same, some last changes to those files are retrieved. For instance, if you take a look at `tests/test_stump.py`, you can see the KNN test function is just for k=3 and not `range(1,4)`. Or, in `stumpy/gpu_stump.py`, the `gpu_searchsorted_left/right` are decorated with `@jit` (and not `@jit(device=True)`).

seanlaw:
@NimaSarajpoor I ask because all of the tests have now passed (perhaps this was expected). Please go ahead ahead and add one commit at a time, allow the tests to run, and let's see where it breaks.

NimaSarajpoor:
Sure. If I see a commit results in failure. I will reset it, and then try to see if I can break down that specific commit to smaller commits so we can isolate the "bad" change.

NimaSarajpoor:
I got into a `detached head` problem where things got weird and the changes got disappeared. So, I had to retrieve it! This is basically just to make sure everything is normal again. 

(I am now using `git show` to find changes in each commit and apply those changes to the files in this current branch)

seanlaw:
The test passed. Please keep adding incrementally until tests fail

seanlaw:
@NimaSarajpoor I just made some changes to `test.sh` so that:

1. The coverage test results have been split up so that we have a little bit better visibility into how long each test is taking (similar to unit tests)
2. I've updated `./test.sh custom` to be a lot more user friendly from the command line so you shouldn't need to change the `test.sh` file and so that you can quickly do:

```
./test.sh custom tests/test_stump.py

# Run multiple test files
./test.sh custom tests/test_stump.py tests/test_stumped.py

# Run 1000 rounds (default is 10 rounds)
./test.sh custom 1000 tests/test_stump.py

# Run a specific test within a test file
./test.sh custom tests/test_stump.py::test_stump_self_join
```

NimaSarajpoor:
@seanlaw 
I know that I should do `k in  range(2, 4)` as `k=1` is already tested; however, I am trying to replicate the commits so we do not miss anything. 

(Also, for now, I am skipping the commits in which I just fixed a typo or rename a function).

seanlaw:
> @seanlaw I know that I should do `k in range(2, 4)` as `k=1` is already tested; however, I am trying to replicate the commits so we do not miss anything.
> 
> (Also, for now, I am skipping the commits in which I just fixed a typo or rename a function).

Sounds good. Thanks for the heads up

seanlaw:
Great! Tests are still passing

NimaSarajpoor:
I will push the next commit in a couple of hours.

NimaSarajpoor:
@seanlaw 
I think we are going to see failure in this last commit as it is related to a function that is used in `stumped`, and I believe the remaining commits are just minor changes. 

So, if we do not see any failure this time, I will push the few remaining commits together.

seanlaw:
> I do not know why tests are passing.

@NimaSarajpoor I'd bet that it is because of the two things that I did:

1. Switch to `2022.5.0` for both `dask` and `distributed`
2. Split the coverage tests into individual tests

Let me revert the first one (i.e., allow `2022.5.2` to be installed) and then have you merge and push to see what happens.

NimaSarajpoor:
@seanlaw 
I also compared the files side by side and it seems they are the same (i.e. I did not miss any commit and all of them are replicated here in this PR).

NimaSarajpoor:
I am planning to delete this branch. Just to be on the safe side, I will back up the files in `stumpy` and `tests`. 

If there is something that I should be aware of regarding this matter, please let me know.

seanlaw:
No, I think it should be okay to delete this branch and close out this PR

Issue #615, opened 2022-06-04
Dask/Distributed Performance Regression


seanlaw:
Somewhere between Dask/Distributed version `2022.5.0` and `2022.5.1` (and subsequently `2022.5.2`), STUMPY's dask-related unit tests (e.g., https://github.com/TDAmeritrade/stumpy/blob/main/tests/test_stumped.py) are around 10x slower. We'll need to track this down and find out what may have changed in either `dask` and/or `distributed`

seanlaw:
@jrbourbeau @mrocklin @crusaderky It'll take me a few days to formulate a better reproducible example beyond switching from `2022.5.0` to `2022.5.1` but would any of you happen to have any thoughts or ideas of what might be causing this? I've tried looking at the `distributed` changelogs (https://github.com/dask/distributed/blob/main/docs/source/changelog.rst) but didn't notice anything obvious.

mrocklin:
Nothing is immediately obvious to me personally.  @crusaderky might be the best bet.

Alternatively, is `git bisect` a feasible option here?  (how easy is it to reproduce the bad behavior?)

seanlaw:
Thanks @mrocklin. I'm going to find some time and see if I can isolate a minimum reproducible example

mrocklin:
Sounds good.  `git bisect` may genuinely be faster is the stumpy test suite is easy to run.  Happy to walk you through using it if you're unfamiliar.  It's a useful tool.

seanlaw:
I had heard about `git bisect` before but this is my first time using it and it felt like magic! Thanks for pushing me @mrocklin 😄 

I was able to isolate the problem in `distributed`. This was the first "bad" commit:

```
commit 36e99465bdd80be283440ad823cb8231ccb471f5
Author: Florian Jetter <fjetter@users.noreply.github.com>
Date:   Wed May 18 11:45:38 2022 +0200

    Ensure multiple clients can cancel their key without interference (#6016)

 distributed/client.py                  |  2 ++
 distributed/scheduler.py               | 39 +++++++++++++++++------------
 distributed/tests/test_as_completed.py |  8 +++---
 distributed/tests/test_client.py       | 45 ++++++++++++++++++++++------------
 distributed/tests/test_worker.py       |  4 +--
 5 files changed, 61 insertions(+), 37 deletions(-)
```

So, to reproduce the problem, I started with a fresh `miniconda` environment, I did:

```
conda install -y -c conda-forge stumpy
cd dask
[...]

PR #617, opened 2022-06-07
Fix dask/distributed regression

fjetter:
Closes https://github.com/TDAmeritrade/stumpy/issues/615

The comment suggests that these cancellations are required to delete data from the dask cluster but this is not necessary for a couple of reasons

- Cancellation itself is not necessary unless you want to abort an actively running graph, e.g. if you wanted to stop all submitted futures before they finished. We reach this code path only if all completed successfully 
- What you are looking for is likely rather `del future` or `future.release()` which tells the scheduler that the future is no longer required and it deletes it
- The explicit release of the data is not even required. We're already at the end of the function and as soon as we exit the local context, the futures are garbage collected and dask dereferences the futures automatically and releases the data.

### DIFF
--- a/stumpy/aamped.py
+++ b/stumpy/aamped.py
@@ -150,16 +150,6 @@ def aamped(dask_client, T_A, m, T_B=None, ignore_trivial=True, p=2.0):
     out[:, 0] = profile[:, 0]
     out[:, 1:4] = indices
 
-    # Delete data from Dask cluster
-    dask_client.cancel(T_A_future)
-    dask_client.cancel(T_B_future)
-    dask_client.cancel(T_A_subseq_isfinite_future)
-    dask_client.cancel(T_B_subseq_isfinite_future)
-    for diags_future in diags_futures:
-        dask_client.cancel(diags_future)
-    for future in futures:
-        dask_client.cancel(future)
-
     threshold = 10e-6
     if core.are_distances_too_small(out[:, 0], threshold=threshold):  # pragma: no cover
         logger.warning(f"A large number of values are smaller than {threshold}.")
--- a/stumpy/maamped.py
+++ b/stumpy/maamped.py
@@ -167,15 +167,4 @@ def maamped(dask_client, T, m, include=None, discords=False, p=2.0):
         stop = min(k, start + step)
         P[:, start + 1 : stop], I[:, start + 1 : stop] = results[i]
 
-    # Delete data from Dask cluster
-    dask_client.cancel(T_A_future)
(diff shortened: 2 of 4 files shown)


seanlaw:
@fjetter Thank you for looking into this. Do you know what would happen with the data if, say, `stumped` is called multiple times with the same inputs? Will this cause a problem (i.e., the garbage collector hasn't deleted the data yet and `dask` can somehow still reference the same inputs)? I only ask because, in the past, when I did **not** `scatter` the data with `hash=False`:

```
dask_client.scatter(
    some_array, broadcast=True
)
``` 

Then, `dask` got into a state where `some_array` could still be referenced by multiple calls to the same STUMPY function but the data was corrupted (likely due to some race condition). Adding `hash=False` resolved the issue but that is also why I chose to `cancel` the data to be extra safe. I only did this to be explicit but, perhaps, this is not needed?

fjetter:
With `hash=False` you will generate a random key, see

- https://github.com/dask/distributed/blob/febd41d94321adfb8fc6254ffe9bd350a99f0189/distributed/client.py#L2339-L2341
- https://github.com/dask/distributed/blob/febd41d94321adfb8fc6254ffe9bd350a99f0189/distributed/client.py#L2230-L2233

i.e. even if you were to call scatter with the identical data twice, we would store the data twice and dask would not know that it's duplicated but would treat every instance as unique. Using this you'd never reuse any data and would re-scatter the data all the time.

It is possible that there is, for a while, still data from the previous run before the next starts but this will be cleaned up eventually. 100% sure that the runs won't mix even without cancellation.

---

If you are using `hash=True` (default) it will hash the data using [`tokenize`](https://github.com/dask/dask/blob/8c9076a5b5e7be0bd658ba7ee5fd00fe9642a9d8/dask/base.py#L922-L934) (an md5 hash) and dask would recognize that it's the same data. IIUC, we haven't implemented a shortcut where we'd skip a re-upload, i.e. you would not gain a lot by using this.

The race condition I see is not necessarily due to hash collisions but rather in how we track tasks. I could see some potential for problems with out internal state. However, we have [tests](https://github.com/dask/distributed/blob/febd41d94321adfb8fc6254ffe9bd350a99f0189/distributed/tests/test_stress.py#L58-L88) that cover this edge case but there may be something else going on, idk

seanlaw:
@fjetter Thank you for the explanation. Also, it looks like the tests have all passed and, indeed, I re-timed the test and it seems to have resolved the regression. I appreciate your help and attention on this!

PR #627, opened 2022-06-26
Fixed #625 Optimize stumpi

### LINKED ISSUE #625: stumpy/stumpi.py high computing time
```
# Input
seed = 0
np.random.seed(seed)
T = np.random.rand(100000)
m = 50
```

The computing time of `stumpy/stumpi.py` without calling `.update( )` is about 75 sec. However, it should be close to the timing of `stumpy/stump.py` (20 sec).

NimaSarajpoor:
This will address issue #625. The `for-loop` block for retrieving the left-matrix profile is optimized. 
```
# Input
seed = 0
np.random.seed(seed)
n = 100000

T = np.random.rand(n)
m = 50
```

**NOTE:**
* before optimization: the for-loop took 50 sec
* now: it is less than 1 sec

---

We might optimize the same part in `tests/naive.py: class stumpi_egress` which uses `core.mass` for evaluating distance between two single subsequences.

### DIFF
--- a/stumpy/stumpi.py
+++ b/stumpy/stumpi.py
@@ -131,20 +131,18 @@ class stumpi:
         self._left_P[mask] = self._P[mask]
 
         # the remaining ones
-        for idx in np.flatnonzero(~mask & self._left_I >= 0):
-            nn_idx = self._left_I[idx]
-            QT = np.dot(
-                self._T[idx : idx + self._m], self._T[nn_idx : nn_idx + self._m]
-            )
+        for i in np.flatnonzero(self._left_I >= 0 & ~mask):
+            j = self._left_I[i]
+            QT = np.dot(self._T[i : i + self._m], self._T[j : j + self._m])
             D_square = core._calculate_squared_distance(
                 self._m,
                 QT,
-                self._M_T[idx],
-                self._Σ_T[idx],
-                self._M_T[nn_idx],
-                self._Σ_T[nn_idx],
+                self._M_T[i],
+                self._Σ_T[i],
+                self._M_T[j],
+                self._Σ_T[j],
             )
-            self._left_P[idx] = np.sqrt(D_square)
+            self._left_P[i] = np.sqrt(D_square)
 
         Q = self._T[-m:]
         self._QT = core.sliding_dot_product(Q, self._T)
--- a/stumpy/stumpi.py
+++ b/stumpy/stumpi.py
@@ -124,10 +124,27 @@ class stumpi:
 
         self._T, self._M_T, self._Σ_T = core.preprocess(self._T, self._m)
         # Retrieve the left matrix profile values
-        for i, j in enumerate(self._left_I):
-            if j >= 0:
-                D = core.mass(self._T[i : i + self._m], self._T[j : j + self._m])
-                self._left_P[i] = D[0]
+
+        # if left neighbor of `S[i] = T[i, i+m]` is actually its matrix profile index
+        # neighbor, then their distance is matrix profile[i]!!!
+        mask = self._left_I == self._I
+        self._left_P[mask] = self._P[mask]
+
+        # the remaining ones
+        for idx in np.flatnonzero(~mask & self._left_I >= 0):
+            nn_idx = self._left_I[idx]
+            QT = np.dot(
+                self._T[idx : idx + self._m], self._T[nn_idx : nn_idx + self._m]
+            )
+            D_square = core._calculate_squared_distance(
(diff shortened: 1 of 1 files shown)


stumpy/stumpi.py
```diff
@@ -124,10 +124,25 @@ def __init__(self, T, m, egress=True, normalize=True, p=2.0):
 
         self._T, self._M_T, self._Σ_T = core.preprocess(self._T, self._m)
         # Retrieve the left matrix profile values
-        for i, j in enumerate(self._left_I):
-            if j >= 0:
-                D = core.mass(self._T[i : i + self._m], self._T[j : j + self._m])
-                self._left_P[i] = D[0]
+
+        # if left neighbor of `S[i] = T[i, i+m]` is actually its matrix profile index
```
seanlaw, line 128: Maybe change to:
"Since each matrix profile value is the minimum between the left and right matrix profile values, then we can save time by only recomputing the left matrix profile value only when the matrix profile value is equal to the right matrix profile value."

stumpy/stumpi.py
```diff
@@ -124,10 +124,25 @@ def __init__(self, T, m, egress=True, normalize=True, p=2.0):
...
+        # neighbor, then their distance is matrix profile[i]!!!
+        mask = self._left_I == self._I
+        self._left_P[mask] = self._P[mask]
+
+        # the remaining ones
```
seanlaw, line 133: This is not a good comment. Maybe:

"# Only recompute the `i`th left matrix profile value, `self._left_P[i]`,  when `self._P[i] != self._left_P[i]`"

seanlaw (changes requested):
This is great! Can you also see if `aampi` suffers from this issue (i.e., is the timing slow for the same input?) and correct it as necessary?

NimaSarajpoor:
> Can you also see if `aampi` suffers from this issue (i.e., is the timing slow for the same input?) and correct it as necessary?

will do.

stumpy/stumpi.py
NimaSarajpoor: Sure. will revise it.

stumpy/stumpi.py
NimaSarajpoor: @seanlaw 
Sorry about that. I should have put time into writing informative comments. I will revise it.

stumpy/stumpi.py
NimaSarajpoor: Are you okay with this?
"Since each matrix profile value is the minimum between the left and right matrix profile values, we can save time by re-computing the left matrix profile value only when the matrix profile index is equal to right matrix profile index."

(Since we do not have the left/right matrix profile value but just their indices, I thought it would be better to avoid it in the comment)

stumpy/stumpi.py
NimaSarajpoor: How about this?
"# Only recompute the ith left matrix profile value, self._left_P[i], when self._I[i] != self._left_I[i]"

(for the same reason I explained above)

seanlaw: Go for it!

stumpy/aampi.py
```diff
@@ -103,7 +103,18 @@ def __init__(self, T, m, egress=True, p=2.0):
         )
 
         # Retrieve the left matrix profile values
-        for i, j in enumerate(self._left_I):
+
+        # Since each matrix profile value is the minimum between the left and right
+        # matrix profile values, we can save time by re-computing the left matrix
```
seanlaw, line 108: It's subtle but I think it's better to say "re-computing ONLY the"

seanlaw (changes requested):

stumpy/aampi.py
NimaSarajpoor: sure.

seanlaw:
@NimaSarajpoor I didn't mean to emphasize "only". I was using uppercase to show that I moved "only" to a position that is earlier in the sentence. Now, there are two "only" in the sentence. I am looking for:

```
        # Since each matrix profile value is the minimum between the left and right
        # matrix profile values, we can save time by re-computing only the left matrix
        # profile value when the matrix profile index is equal to the right
        # matrix profile index.
```

NimaSarajpoor:
Ahh 😃 ... I thought that was intentional... I wanted to ask but then changed mind. I will revise the comments.

seanlaw:
I take responsibility for not being clear

seanlaw:
Thank you @NimaSarajpoor!

PR #628, opened 2022-06-28
Fixed #545 Replace while Loops

### LINKED ISSUE #545: Replace `while` Loops
Currently, `aamp_motifs.py` and `motifs.py` both utilize a `while` loop but it would be safer to change this to a `for` loop

NimaSarajpoor:
This PR solves issue #545 without increasing the computing time.

```
# tested with this input
seed = 0
np.random.seed(seed)
T = np.random.rand(100000)
stumpy.motifs(T, P, max_matches=20, max_motifs=20)
```

### DIFF
--- a/stumpy/aamp_motifs.py
+++ b/stumpy/aamp_motifs.py
@@ -97,7 +97,10 @@ def _aamp_motifs(
     motif_distances = []
 
     candidate_idx = np.argmin(P[-1])
-    while len(motif_indices) < max_motifs:
+    for i in range(l):
+        if len(motif_indices) == max_motifs:
+            break
+
         profile_value = P[-1, candidate_idx]
         if profile_value > cutoff or not np.isfinite(profile_value):  # pragma: no cover
             break
@@ -398,11 +401,14 @@ def aamp_match(
     else:
         candidate_idx = np.argmin(D)
 
-    while (
-        D[candidate_idx] <= atol + max_distance
-        and np.isfinite(D[candidate_idx])
-        and len(matches) < max_matches
-    ):
+    for i in range(len(D)):
+        if (
+            D[candidate_idx] > atol + max_distance
+            or D[candidate_idx] == np.inf
+            or len(matches) == max_matches
+        ):
+            break
+
         matches.append([D[candidate_idx], candidate_idx])
         core.apply_exclusion_zone(D, candidate_idx, excl_zone, np.inf)
         candidate_idx = np.argmin(D)
--- a/stumpy/motifs.py
+++ b/stumpy/motifs.py
@@ -96,7 +96,10 @@ def _motifs(
     motif_distances = []
 
     candidate_idx = np.argmin(P[-1])
-    while len(motif_indices) < max_motifs:
+    for i in range(l):
+        if len(motif_indices) == max_motifs:
+            break
+
         profile_value = P[-1, candidate_idx]
         if profile_value > cutoff or not np.isfinite(profile_value):  # pragma: no cover
             break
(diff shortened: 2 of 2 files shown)


seanlaw:
> The main idea is that at the end of the day, we cannot exceed the number of subsequences or available elements in our search space. So, the number of "iterations" in for-loop can be that value, and then everything else, i.e. the body of loop block, remains almost the same.

Ahhhh, I like that! That's very smart and immediately obvious/intuitive once you point out the key observation! Thank you - because I got to learn something from you 😍

stumpy/aamp_motifs.py
```diff
@@ -398,11 +401,14 @@ def max_distance(D):
...
-    while (
-        D[candidate_idx] <= atol + max_distance
-        and np.isfinite(D[candidate_idx])
-        and len(matches) < max_matches
-    ):
+    for i in range(len(D)):
+        if (
+            D[candidate_idx] > atol + max_distance
+            or D[candidate_idx] == np.inf
+            or len(matches) == max_matches
```
seanlaw, line 408: For safety, I recommend `len(matches) >= max_matches` (i.e., maybe somebody changes this function in the future and accidentally appends to `matches` twice each round and causes `len(matches)` to never be equal to `max_matches`.

stumpy/motifs.py
```diff
@@ -465,11 +468,14 @@ def max_distance(D):
...
-    while (
-        D[candidate_idx] <= atol + max_distance
-        and np.isfinite(D[candidate_idx])
-        and len(matches) < max_matches
-    ):
+    for i in range(len(D)):
+        if (
+            D[candidate_idx] > atol + max_distance
+            or D[candidate_idx] == np.inf
+            or len(matches) == max_matches
```
seanlaw, line 475: For safety, I recommend `len(matches) >= max_matches`

stumpy/aamp_motifs.py
seanlaw: I don't know if `D[candidate_idx] == np.inf` is better or if `~np.isfinite(D[candidate_idx])` is better?

stumpy/aamp_motifs.py
```diff
@@ -97,7 +97,10 @@ def _aamp_motifs(
     motif_distances = []
 
     candidate_idx = np.argmin(P[-1])
-    while len(motif_indices) < max_motifs:
+    for i in range(l):
+        if len(motif_indices) == max_motifs:
```
seanlaw, line 101: Replace with `if len(motif_indices) >= max_motifs`

stumpy/motifs.py
```diff
@@ -96,7 +96,10 @@ def _motifs(
     motif_distances = []
 
     candidate_idx = np.argmin(P[-1])
-    while len(motif_indices) < max_motifs:
+    for i in range(l):
+        if len(motif_indices) == max_motifs:
```
seanlaw, line 100: Replace with `if len(motif_indices) >= max_motifs`

NimaSarajpoor:
> Hopefully, you can see that this for-loop now feels much "safer" and is much less wasteful.

Yes! I can see there is an end to this loop, and I do not need to be worried about it! Hence, safer.

stumpy/aamp_motifs.py
NimaSarajpoor: > For safety, I recommend `len(matches) >= max_matches` (i.e., maybe somebody changes this function in the future and accidentally appends to `matches` twice each round and causes `len(matches)` to never be equal to `max_matches`.

Great perspective! I can see your point! I will revise it.

NimaSarajpoor: > I don't know if `D[candidate_idx] == np.inf` is better or if `~np.isfinite(D[candidate_idx])` is better?

Maybe this is just me...but the latter forces me to keep in mind the operator `~` and also think about the return value of `np.isfinite()`, and  then put them together.

stumpy/motifs.py
NimaSarajpoor: Sure. I will revise it.

stumpy/aamp_motifs.py
seanlaw: I agree. That is why the original code used `np.isfinite` (because it also detects `np.nan`) and, therefore, does more and is safer. Given this, I think we should go with `np.isfinite` approach

seanlaw:
@NimaSarajpoor Thanks again for fixing this!

PR #646, opened 2022-07-18
Fix algo to avoid miscalculation in AB-join

### LINKED ISSUE #644: miscalculation in `prescrump` for AB join?
So, I think there is a slight issue in the `prescrump` for AB join. 

please read the comments provided in the code snippet below:
```
# in tests/naive.py: 
# in function `prescrump`

I[i] = np.argmin(distance_profile)
P[i] = distance_profile[I[i]]
if P[i] == np.inf:
    I[i] = -1
else:
    j = I[i]
    for k in range(1, min(s, l - max(i, j))):
        d = dist_matrix[i + k, j + k]
        if d < P[i + k]:
            P[i + k] = d
            I[i + k] = j + k
        if d < P[j + k]:  # ISSUE: we can use `d` to update `P[j + k]` ONLY in self-join
            P[j + k] = d
            I[j + k] = i + k

    for k in range(1, min(s, i + 1, j + 1)):
        d = dist_matrix[i - k, j - k]
        if d < P[i - k]:
            P[i - k] = d
            I[i - k] = j - k
        if d < P[j - k]:  # ISSUE we can use `d` to update `P[j - k]` ONLY in self-join
            P[j - k] = d
            I[j - k] = i - k

return P, I
```

NimaSarajpoor:
This PR fixes #644 for non-normalized method (this is a follow up to PR #645)

### DIFF
--- a/stumpy/scraamp.py
+++ b/stumpy/scraamp.py
@@ -134,7 +134,7 @@ def _compute_PI(
                 if p_norm < P_NORM[thread_idx, i + k]:
                     P_NORM[thread_idx, i + k] = p_norm
                     I[thread_idx, i + k] = j + k
-                if p_norm < P_NORM[thread_idx, j + k]:
+                if excl_zone is not None and p_norm < P_NORM[thread_idx, j + k]:
                     P_NORM[thread_idx, j + k] = p_norm
                     I[thread_idx, j + k] = i + k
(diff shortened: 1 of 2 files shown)


seanlaw:
Thank you @NimaSarajpoor!

Issue #610, opened 2022-05-25
Fix Unit Test


seanlaw:
This unit test causes a failure:

```
def test_stump_identical_subsequence_self_join():
    seed = 27343
    np.random.seed(seed)

    identical = np.random.rand(8)
    T_A = np.random.rand(20)
    T_A[1 : 1 + identical.shape[0]] = identical
    T_A[11 : 11 + identical.shape[0]] = identical
    m = 3
    zone = int(np.ceil(m / 4))
    ref_mp = naive.stump(T_A, m, exclusion_zone=zone, row_wise=True)
    comp_mp = stump(T_A, m, ignore_trivial=True)
    naive.replace_inf(ref_mp)
    naive.replace_inf(comp_mp)
    npt.assert_almost_equal(
        ref_mp[:, 0], comp_mp[:, 0], decimal=config.STUMPY_TEST_PRECISION
    )  # ignore indices

    comp_mp = stump(pd.Series(T_A), m, ignore_trivial=True)
    naive.replace_inf(comp_mp)
    npt.assert_almost_equal(
        ref_mp[:, 0], comp_mp[:, 0], decimal=config.STUMPY_TEST_PRECISION
    )  # ignore indices
```

NimaSarajpoor:
@seanlaw 
I think I found the source of error. According to my investigation, the problem is in the precision of $\rho$. The discrepancy is in index 2 and 12. In fact `S1 = T[2:2+m]` and `S2=T[12:12+m]` are exactly the same and the $\rho$ between them must be 1. 

**However,** according to the calculation of  `stumpy.stump`, we can see that the highest value in $\rho[:, 2, 0]$ (print it in [this line](https://github.com/TDAmeritrade/stumpy/blob/576de379994896063be71d264e48a08ae17476cf/stumpy/stump.py#L405)) is `0.9999999999483034`. 

```
idx = 2 # the start index of subsequence of interest
arg = np.argmax(ρ[:, idx, 0]) 
>>> I[arg, 2, 0]
12 # this is NN to the subsequence at index idx=2

>>> ρ[arg, idx, 0]
0.9999999999483034
```

**A potential solution:**
We may prevent this by creating `config.STUMPY_RHO_PRECISION` and use it in the line of code I mentioned above before comparing the values of threads with each other (and then set it to 1 if the calculated value is too close to 1 (?!))

NimaSarajpoor:
@seanlaw 
[...]

PR #655, opened 2022-08-01
Fixed #652 Improve Docstrings

### LINKED ISSUE #652: Incorrect descriptions in Docstrings
I have been reading the module `mpdist` and I noticed:

* The description of the output of `_mpdist_vect` is missing from its docstring
* The description of `μ_Q` and `σ_Q` (in `_mpdist_vect`) are not correct. I think they should be `numpy.ndarrays` not `float`.
* The description of `μ_Q` and `σ_Q` (in `core. _mass_distance_matrix`) are not correct. I think they should be `numpy.ndarrays` not `float`.

NimaSarajpoor:
This PR addresses issue #652

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -1456,11 +1456,11 @@ def _mass_distance_matrix(Q, T, m, distance_matrix, μ_Q, σ_Q, M_T, Σ_T):
     distance_matrix : numpy.ndarray
         The full output distance matrix. This is mandatory since it may be reused.
 
-    μ_Q : float
-        Mean of `Q`
+    μ_Q : numpy.ndarray
+        Sliding mean of `Q`
 
-    σ_Q : float
-        Standard deviation of `Q`
+    σ_Q : numpy.ndarray
+        Sliding standard deviation of `Q`
 
     M_T : numpy.ndarray
         Sliding mean of `T`
@@ -2533,7 +2533,7 @@ def _select_P_ABBA_value(P_ABBA, k, custom_func=None):
 
     k : int
         Specify the `k`th value in the concatenated matrix profiles to return. This
-        parameter is ignored when `k_func` is not None.
+        parameter is ignored when `custom_func` is not None.
 
     custom_func : object, default None
         A custom user defined function for selecting the desired value from the
--- a/stumpy/mpdist.py
+++ b/stumpy/mpdist.py
@@ -192,13 +192,14 @@ def _mpdist_vect(
         Time series or sequence
 
     m : int
-        Window size
+        Window size that will be used for calculating the mpdist between Q and
+        any subsequence in T (of size `len(Q)`)
 
-    μ_Q : float
-        Mean of `Q`
+    μ_Q : numpy.ndarray
+        Sliding mean of `Q`
 
-    σ_Q : float
-        Standard deviation of `Q`
+    σ_Q : numpy.ndarray
+        Sliding standard deviation of `Q`
 
     M_T : numpy.ndarray
         Sliding mean of `T`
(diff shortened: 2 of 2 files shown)


seanlaw:
Thanks @NimaSarajpoor. Looks like there is a failing test 😢

NimaSarajpoor:
> Looks like there is a failing test

Yeah I noticed that. Although the changes I made were just on docstrings. it is good that the issue reveals itself. It is not related to this PR. But, if I can find it, I resolve it here in this PR. Is that okay?

NimaSarajpoor:
@seanlaw 
So, I took a look at the error:
```
ref_mp = naive.stumpi_egress(T, m)
ref_P = ref_mp.P_.copy()

stream = stumpi(T, m, egress=True)
comp_P = stream.P_.copy()

npt.assert_almost_equal(ref_P, comp_P, decimal=config.STUMPY_TEST_PRECISION) # Error!
```
However, both `stumpi`  and its naive version `naive.stumpi_egress` use the performant version `stump` to calculate the initial matrix profile for time series `T` and length `m`. So, we should have got the same result. Is it possible that we have two different imprecisions for the same process with same input? 

I took a look at the naive version again, and the only difference is that the naive version of stumpi does `stump(T, m)` instead of `stump(self._T, self._m)`.  Although that is most likely not the cause of error, I changed it to `self._T` and `self._m`. I ran 50 tests (i.e. 50 iterations) for that particular `test_functions` and I got no error. What do you think?

seanlaw:
@NimaSarajpoor I ran the test multiple times with different random seeds and was able to make it error out with

```
def test_stumpi_identical_subsequence_self_join_egress():
    m = 3

    seed = np.random.randint(100000)
    seed = 84451
    np.random.seed(seed)
```

It's not the exact same array but I think it's sufficiently similar. Does this help?

NimaSarajpoor:
Yes! Thanks for the seed :) I will investigate it.

NimaSarajpoor:
@seanlaw 
I think the error is not related to `stumpi`, and it is in fact related to `stumpy.stump`. In below, I showed part of the test function where the error happens.

```
# in tests/test_stumpi.py
# in test_stumpi_identical_subsequence_self_join_egress

ref_mp = naive.stumpi_egress(T, m)
ref_P = ref_mp.P_.copy()

stream = stumpi(T, m, egress=True)
comp_P = stream.P_.copy()
 
naive.replace_inf(ref_P)
naive.replace_inf(comp_P)    
>       npt.assert_almost_equal(ref_P, comp_P, decimal=config.STUMPY_TEST_PRECISION)

```
As you can see, the error is between `ref_P` and `comp_P`. These are in fact the matrix profile of time series `T` and parameter ` m`  (see the class `stumpi`/`__init__` method). The error happens because `stumpy.stump` and `naive.stump` have different outputs in this case. 

So, I started investigating `stumpy.stump` and I realized the calculated `ρ` between identical subsequences is not exactly 1 (due to imprecision in calculation). And, in this case, it is slightly more than 1 (e.g. `1.0000000000280238`) Apparently, this error is similar to issue #610. In that issue, the problem was that the value of `ρ` between two identical was not 1, and in that case, it was below 1. It was  `0.9999999999483034`.

If we go with the solution you proposed in issue #610, i.e. adding the computationally cheap `if-block`:
```
if pearson > config.STUMPY_PERFECT_CORRELATION:
            pearson = 1.0
```
we can solve both issues. I tested it with `config.STUMPY_PERFECT_CORRELATION = 0.99999999`

seanlaw:
Awesome find! Would you mind adding BOTH unit tests (with the fixed random seed), adding the `config` variable, and submitting a PR?

NimaSarajpoor:
> Would you mind adding BOTH unit tests (with the fixed random seed), adding the `config` variable, add the if-else to `naive.stump`, and submitting a PR?

Cool. Will do so shortly!

seanlaw:
@NimaSarajpoor  If it's okay with you, since this is being tracked in the other PR, I will merge this commit.

NimaSarajpoor:
> @NimaSarajpoor If it's okay with you, since this is being tracked in the other PR, I will merge this commit.

Sure. Thanks for letting me know.

seanlaw:
Thanks @NimaSarajpoor!

PR #665, opened 2022-09-01
Fix #664 fixing param s in AB-join mode

### LINKED ISSUE #664: param `s` in prescrump and scrump raises error when set to default and T_B is not None
So, I have been working on `scrump` and `prescrump`, and I think we can get error if `s=None` (default) and `T_B` is not `None`. 

Because we do:

```
# in prescrump

# lines 330-334
if T_B is None:
        T_B = T_A
    else:
        excl_zone = None

# lines 342-343
if s is None:  # pragma: no cover
        s = excl_zone
```

So, when `T_B` is not `None`,  `excl_zone` is set to None, which makes `s=None` if `s` is not provided by the user.

----

I also tried an example to make sure of it:
```
import numpy as np
from stumpy.scrump import prescrump

T_A = np.random.rand(10)
T_B = np.random.rand(10)
m = 3
prescrump(T_A, m, T_B)
``` 

And the error:
```
83 if s is None:  # pragma: no cover
     84     s = excl_zone
---> 86 indices = np.random.permutation(range(0, l, s)).astype(np.int64)
     88 return (T_A, T_B, μ_Q, σ_Q, M_T, Σ_T, indices, s, excl_zone)

TypeError: 'NoneType' object cannot be interpreted as an integer
```

---

Solution:
Maybe this suffices:

```
if s is None:  # pragma: no cover:
         s =  int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
```

NimaSarajpoor:
This PR resolves the issue raised in #664. When `s=None` and `T_B` is not `None`, the value of `s` remains `None` in scrump and prescrump and their non-normalized versions.  I used the suggestion provided by @seanlaw  (in #664) to fix the issue.

### DIFF
--- a/stumpy/scraamp.py
+++ b/stumpy/scraamp.py
@@ -324,8 +324,11 @@ def prescraamp(T_A, m, T_B=None, s=None, p=2.0):
     n_A = T_A.shape[0]
     l = n_A - m + 1
 
-    if s is None:  # pragma: no cover
-        s = excl_zone
+    if s is None:  # pragma: no cover:
+        if excl_zone is not None:  # self-join
+            s = excl_zone
+        else:  # AB-join
+            s = int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
 
     indices = np.random.permutation(range(0, l, s)).astype(np.int64)
     P, I = _prescraamp(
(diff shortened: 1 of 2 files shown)


NimaSarajpoor:
I think we got the imprecision issue again, caused by identical sequence.

seanlaw:
Thanks for addressing this @NimaSarajpoor!

Issue #85, opened 2019-12-02
Add Tutorial(s) that Reproduces “Matrix Profile Top Ten” Paper


seanlaw:
It would be nice to add a tutorial(s) that reproduces the [Matrix Profile Top Ten](https://www.cs.ucr.edu/~eamonn/Top_Ten_Things_Matrix_Profile.pdf) paper. The accompanying data at their Google sites page can be [found here](https://sites.google.com/site/matrixprofiletopten/).

It might be best to make the individual top ten sections as separate items (i.e., sub-list) that rolls under one tutorial line item in RTD and that can expand to ten sub items

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

seanlaw:
Here is the [Shapelet Discovery data ](http://alumni.cs.ucr.edu/~lexiangy/shapelet.html)

seanlaw:
[zhu2020.pdf](https://github.com/TDAmeritrade/stumpy/files/4446948/zhu2020.pdf)

Attol8:
Hi @seanlaw, I have recently bumped into STUMPY and I am a big fan of your work! Do you still need help with this issue? If so, I will start drafting a Jupyter notebook with each of the 10 use-cases described in the paper. Also, How would you title the tutorial line item in RTD?

seanlaw:
@Attol8 Thank you for your kind words and, absolutely yes, we are looking for and welcome help on this issue!

If you don't mind, the approach that I'd like to take with this is to split up each of the 10 use cases into its own individual notebook and place them into the `docs/` directory. This way, the scope for each notebook is small, achievable, and anybody can tackle a different use case. As you work on one use case, it would be best to create a new issue for that one use case so its progress can be tracked. Once we get a few (3?) use cases completed, then on my end I can handle the best way to display it in RTD (or maybe you already have some ideas and I'd be happy to discuss). This way, the writing of the notebooks will be separated from how/when it is presented to the user in our documentation (which may take a bit more thinking/planning).

I am flexible and open to your thoughts/ideas too. How does that sound?

Attol8:
Thanks for the detailed answer @seanlaw ! As you suggested, I agree that the best way to proceed is: notebooks first, reviews seconds, implementation of the docs third. Following your checklist, I will now get to work and post updates as soon as I have something ready

seanlaw:
[...]

Issue #663, opened 2022-08-28
Add (Quick) Examples


seanlaw:
Recently, I received some feedback:

1. There is too much friction to "get started". With limited time, the tutorials are too verbose for users who are looking for the one line of code to use the package (i.e., stump(data, m)). 
2. It isn't always clear what the shape of the input should be in
3. `P_Point_Point = stumpy.stump(point_df["Centroid Location"], m)[:, 0].astype(float)` is hard to read
4. `stumpy.stump` is somewhat long. Consider `st.stump` when importing
5. how quickly does the documentation allow me to implement everything on my data as someone that has spent 2 hours learning about the methodology
6. The tutorial titles are named by someone that has expert knowledge and may be off putting for beginners. Instead, the title (or description) should quickly tell you what the tutorial does

> as a data scientist i have data and a problem, and am looking for a solution tell me what solutions you can offer me is my frame of reference

Ultimately, from the user's perspective, STUMPY is package that they need to figure out how to use in as little time as possible. Even if one package is "better" under the hood, they'd probably accidentally end up using it, if I can get it to work on my data faster.

> this is probably the document that got me to spend more time with the matrix profile
https://www.cs.ucr.edu/~eamonn/100_Time_Series_Data_Mining_Questions__with_Answers.pdf
selling me solutions, not a new shiny methodology

> I guess to summarize:
> 1. get user motivated/excited (what can the thing solve for user)
> 2. get it in users hands and working
> 3. now they're more invested and more willing to get into the details

namankaushik9876:
Hi,
I want to work on this issue.
Give me chance

seanlaw:
@namankaushik9876 Thank you for your interest. Perhaps, the first thing you can do is help us identify other docs (from other packages) that could serve as a reasonable example to follow and mimic the style of. 

As notes above, I think these will be short (maybe 2 paragraphs?) and the code should focus on the STUMPY API and have east-to-understand titles so that experienced data scientists (who know what they want to accomplish) can quickly use the minimum-reproducible code snippet. This should not include code for plotting figures and, instead, we simply embed the necessary figure. But first, we need to find short examples to compare between.

How does that sound?

namankaushik9876:
sure sir

seanlaw:
Here are some crowdsourced references that may be useful:

1. [Dividing up documentation](https://documentation.divio.com/): Tutorials and How-to Guides
[...]

PR #667, opened 2022-09-05
Tutorial Discovering Motifs Under Uniform Scaling

ken-maeda:
# Discovering motifs under uniform scaling
From #85, I add Discovering motifs under uniform scaling tutorial from 3.1 of [The Matrix Profile Top Ten paper](https://github.com/TDAmeritrade/stumpy/files/4446948/zhu2020.pdf).
Add `docs/Tutorial_Matrix_Profile_Top_Ten.ipynb`

It still needs modification to plot, markdown, coding...., but please have a look through and let me know if you find problem.

# Pull Request Checklist

- [x] Create a new branch


seanlaw:
Thank you @ken-maeda. Please allow me some time to review

NimaSarajpoor:
Cool. This swiss army knife would be a great addition.

I was wondering if it would be okay to provide my feedbacks on this PR. I think this can help me improve my skills on reviewing code. I will try to avoid giving feedbacks that I am not sure of.

seanlaw:
@NimaSarajpoor Please go ahead and provide your feedback and I can follow!

seanlaw:
@NimaSarajpoor Thank you for your thorough review. I think you've provided enough valuable suggestions for a first pass. I'll chime in after there's been a few iterations/refinements

ken-maeda:
<p>This HC is from original paper. For me, it looks A is similar to B than C in even A, B ploting. But it might be too roundabout to explain these similiarities.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ken-maeda:
@NimaSarajpoor I appriciate your huge and kind feedback, I got clear path to complete this notebook.

@seanlaw and @NimaSarajpoor  

> I was really hoping that we could provide super short/quick pieces of code to perform uniform scaling quickly.

I should have shared my notebook in earlier phase to fix what it should be covered.
I completely respect your guidelines. I don't mind to shape up my notebook at all.
I just intended to reproduce paper content.

I'll make 3 different length version notebook with @NimaSarajpoor feedback.
1. original
2. bit shorter 
3. much shorter

Then Could you choose best length notebook from those?

seanlaw:
> I should have shared my notebook in earlier phase to fix what it should be covered.
I completely respect your guidelines. I don't mind to shape up my notebook at all.
I just intended to reproduce paper content.

It's my fault and I take full responsibility for this! 

> I'll make 3 different length version notebook with @NimaSarajpoor feedback. Then Could you choose best length notebook from those?

Actually, now that I've thought about it more. Let's finish this long-form first and keep it as a tutorial. Then, afterward, I think it would make sense to take each long tutorial and make an equivalent/complimentary short form with much less context and explanation and that points back to the long form if the reader needs more information/explanation. That will be independent of this PR. Let's continue that conversation in issue #663

How does that sound? Btw @ken-maeda you definitely get a voice/opinion in this too. Please treat this as a discussion and not me dictating

NimaSarajpoor:
<p>Exactly. It looks A is similar to B than C. And, I think the purpose of HC is to show that this is not true and in fact, A is similar to C than B. Right? What I am saying here is that we can show this by just calculating the distance. So, if we show <code>d(A, C) &lt; d(A, B) </code>, then we should be fine. Right? (Maybe you keep HC here for this long tutorial, and remove  the HC for the shorter version). I will leave it to you :)</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ken-maeda:
> Actually, now that I've thought about it more. Let's finish this long-form first and keep it as a tutorial. 

I fix notebook as long-from. Please take a look.
At least, maybe we could remove "chapter 3,Which kind of dataset is effective?" from notebook.
After confirmation of contents, dataset have to be stored somewhare.

ken-maeda:
<p>I think this paper is focuing on flexibility of  "time axis" by streching data.</p><p>And DTW also should bring some flexiblity on time axis. In my understanding, z-normalized ED is generally for magnitude.</p><p>I don't know maybe I just know a part of it.....</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ken-maeda:
<p>As reproduction of paper, i think it should be normalize=True.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>Got it. Thanks.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ken-maeda:
<p>In this dataset, we know that we can take different amounts of time from domain knowledge(below conclusion part ). But value range itself is not changed. So I make <code>normalize=False. </code> Should I have to explain about setting <code>nomalize=False</code>?</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
@ken-maeda 
Okay... previously I thought you did both `normalize=True` and `normalize=False` on one data. I think I was wrong. I checked the notebook again. I noticed you considered `normalize=True` for one data, and did `normalize=False` for another data. That is okay since the two cases are different.

---

image

This formula seems to be okay for non-normalize distance. You may want to explain how it works in normalized case. 

Maybe something like this? 
"The distance $d_{u}$ is the minimum of z-norm distances between $Q^{s}$ and $C$". 

Please feel free to correct it or make it better.

ken-maeda:
@NimaSarajpoor 
About `normalize `parameter, first example(chapter 2) should be `normalize=True`  according to paper matrix profile result.
(I made sure both pattern result.)
Second example(chapter 4) is uncertainm, because no matrix profile result is listed and it's not mentioned about.
So maybe we can set  `normalize=True` to second exmaple for consistency.

About formula, I thought z-normalized(magnitude axis) and uniform scaling(time axis) is independent process. If it is so, z-normlized should affect final result, but no interaction for both process.
I really don't know what should be covered. So these explaination is required, I'll add it to below non-normalize formula.

NimaSarajpoor:
> So maybe we can set `normalize=True` to second exmaple for consistency

I think second example is for electricity demand. Right? I think it is ok to do `normalize=False` as I think the magnitude matters. Also, this way we are showing that the uniform scaling can be used in both normalized and non-normalized case.

---


> About formula, I thought z-normalized(magnitude axis) and uniform scaling(time axis) is independent process. If it is so, z-normlized should affect final result, but no interaction for both process.

If I understand correctly, you are saying we stretch the data, and then we just find the z-norm euclidean distance between subseq. I have the same opinion.

> I really don't know what should be covered. So these explaination is required, I'll add it to below non-normalize formula.

I think everything is good! Just add explanation about z-norm case in uniform scaling technique.

ken-maeda:
I fixed what i have notived.

seanlaw:
Thanks @ken-maeda. @NimaSarajpoor is taking a break for now but I will try to find some time to review and provide feedback.

ken-maeda:
of course, no rush, i just wanted to let you know that i have fixed.

ken-maeda:
it seems I should reconstruct notebook by creating new one based on current notebook. The flow of notebook was followed by paper's, althought I'm going to assemble minimum requirement to meet this your review.
1-1. what is uniform scaling
1-2. scaling factor calculation
2-1. uniform scaling with Matrix profile
(different example with different strech ratio?)

seanlaw:
> it seems I should reconstruct notebook by creating new one based on current notebook. The flow of notebook was followed by paper's, althought I'm going to assemble minimum requirement to meet this your review.
1-1. what is uniform scaling
1-2. scaling factor calculation
2-1. uniform scaling with Matrix profile

@ken-maeda I think that it may be sufficient to only use the MALLAT example and carry it from end-to-end. Does that make sense?

ken-maeda:
>  think that it may be sufficient to only use the MALLAT example and carry it from end-to-end. Does that make sense?
I' going to use Q and C signal dataset also in formula explanation.  But it is now quite simple. Except for this one, it's the MALAT. Rest of notebook should be more simple. I'll upload another draft soon.

ken-maeda:
I upload shorter version of notebook: `Tutorial_Matrix_Profile_Top_Ten_short.ipynb`.
@seanlaw  Please cheack the flow of new one.

seanlaw:
Thanks @ken-maeda! I will find some time to review

docs/Tutorial_Matrix_Profile_Top_Ten_short.ipynb
```diff
@@ -0,0 +1,499 @@
+{
```
seanlaw, line 1: <p>I don't think the average person will understand this as it is too technical.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

seanlaw, line 1: <p>This language is too technical and not suitable for a tutorial</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

seanlaw, line 1: <p>0.242 to 0.240 seems negligible</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanlaw, line 1: <p>Can you please provide a description of how to interpret the plots? What is the green subsequence?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

seanlaw, line 1: <p>Please explain the plots</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

ken-maeda:
@seanlaw  I changed introduction completly to understand how usefull uniform scaling is. How do you think?

docs/Tutorial_Matrix_Profile_Top_Ten_short.ipynb
```diff
@@ -0,0 +1,665 @@
...
```
seanlaw, line 1: <p>Let's start off by showing <code>A</code> and <code>B</code> (side-by-side) and asking the  reader, "are A and B the same? are they similar?  how would you quantify their similarity?" Then plot <code>A</code> and <code>B</code> on top of each other to show that they are actually different lengths but that <code>B</code> is stretched (or <code>A</code>  is compressed)  relative to the other. In other words, in this case, <code>A</code> and <code>B</code> have very similar shapes but, due to the stretching/compressing, they are said to be "out-of-phase" with each other. If you compute the Euclidean distance between <code>A</code> and <code>B</code> as is, then they'd have a large Euclidean distance between them. However, if you could figure out the best way to stretch/compress <code>A</code> or <code>B</code> so as to best match the locations of the peaks/valleys (i.e., bring them "in-phase"), then the Euclidean distance between would become smaller. One obvious way to accomplish this is by uniformly scaling (i.e., stretching/compressing) <code>A</code> or <code>B</code> by some constant scaling factor...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

seanlaw, line 1: <p>Technically, we don't need the blue curve. We only need <code>A</code> and <code>B</code> to support our point.    </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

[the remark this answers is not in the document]
ken-maeda: <p>In unifrom scaling, data have to be located in same interval. Therefore scaled data should include non-original data point (which is at least on interpolated line). Its validity depends on domain and purpose of procesing. We can adjust scaled data by choosing interpolation function. I thought such kind of topics can be too detaild.</p>

ken-maeda: <p>I remove it, and follow your above guide.</p>

ken-maeda: <p>I'll adjust the notebook according to this guide.</p><p> </p>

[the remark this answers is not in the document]
seanlaw: > In unifrom scaling, data have to be located in same interval. 

This is fine.

> Therefore scaled data should include non-original data point (which is at least on interpolated line). 

This is also fine and I do not dispute any of this.

> Its validity depends on domain and purpose of procesing. We can adjust scaled data by choosing interpolation function. I thought such kind of topics can be too detaild.

So, when I asked:

> Will I find the original (exact) values in the interpolated/fitted curve? Is it guaranteed?

I meant, given a dataset with points, say, `original = [0.0, 1.0, 2.0, 3.0, 4.0]` and we interpolate a dataset by adding a few points to `interpolated = [0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0]`. Notice that this contains new points (not a problem) but it also contains all of the original points as well. Is it always `True` that  `len(set(original).difference(set(interpolated))) == 0`?

Instead, if `interpolated = [0.0001, 0.5, 1.0001, 1.5, 2.0001, 2.5, 3.0001, 3.5, 4.0001]`, then `len(set(original).difference(set(interpolated))) == 5` (i.e., none of the original points are guaranteed to be found in `interpolated` then this may be undesirable.

Does that make sense?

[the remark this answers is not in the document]
ken-maeda: I understand what you meant. But I'm not still clear which point is undesirable.

> none of the original points are guaranteed to be found in `interpolated` 

if `interpolated = [0.0001, 0.5, 1.0001, 1.5, 2.0001, 2.5, 3.0001, 3.5, 4.0001]`, Indeed it doesn't include original points. But `interpolated`  should detect `original` because it has only very small error.
In aspect of finding motif, `interpolated` should be accurately "assumed sereis" rather than having exact orignal data point.
In calculating error, all point have same calculation weight for it . Majority point isn't on original data interval.  If I adjust `interpolated` to only have original data point, the data point not on `original` (like 0.5, 1.5) can be out of my "assummed line", and those point should cause unexpected error.

I feel at least in calculating error  `Interpolated ` should be adjusted to be "assummed line" in all point, but "assummed line" depends on the situation. So I don't know comprehensive way to check `Interpolated `.
In this `original = [0.0, 1.0, 2.0, 3.0, 4.0]`, as we can assume linearity.

The point you mention maybe is different from mine...

Do you think some kind of validation for `interpolated` is required?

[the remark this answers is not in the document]
seanlaw: When I look at the paper, they perform `uniform scaling` in Matlab with:

```
T2 = T(1: 100/164: end); % Unofficial Matlab way to resample
```

Unfortunately, I do not have access to Matlab. When I use Octave (which is suppose to be "free-Matlab") with:

```
x = [1.01, 2.01, 3.01, 3.3, 3.5, 4.01, 5.01, 6.01]
x(1:100/150:end)  # stretch to 150%
```

It gives:

```
1.0100   2.0100   2.0100   3.0100   3.3000   3.3000   3.5000   4.0100  4.0100   5.0100   6.0100
```

Again, I don't know if this is the exact same as Matlab but, assuming that it is, this is what the paper is referring to for "uniform scaling". It appears that it is stretching out the length to 150% of the original length by repeating the original numbers. Notice that, unlike interpolation where completely new values are created, there are no new values in this time series (i.e., all values are derived from the original time input series). 

I'm not saying that this is great approach but we should try to recreate the paper where possible even if their approach is more naive. Otherwise, we'll end up with more questions that distract from the point. Additionally, we can always choose to add a "Bonus Section" at the end that offers suggestions for more "advance" scaling via interpolation.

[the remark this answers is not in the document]
seanlaw: [This post](https://stackoverflow.com/questions/53904765/stretch-numpy-vector-to-arbitrary-length-without-interpolation) might offer some reasonable suggestions that are closer to the spirit of the Matlab code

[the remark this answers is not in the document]
ken-maeda: OK, I'm clear what you meant. But I don't know paper can repduce example with that code results. I have to check in "matlab" way. And why did author use unofficial resmapling code.. Anyway I'll try to get matlab enviroment.

ken-maeda:
I'm taking a while to get matlab enviroment, anyway I simulate resmapling in matlab-like as you did.
I'm not still sure about how matlab resmaple works, however It seems paper didn't use "that resmapling".
As paper example, data is streached smoothly. Matlab-like way make weird noise in especially "rising edge" and "falling edge".

> Paper example.

image

> Result

image

I'm going to replace uniform scaling with "naively stretching/shrinking", after confirming how matlab works.

ken-maeda:
I got matlab, and try resampling with official(`T2 = resample(T, 164, 100)`) and unofficial(`T2 = T(1: 100/164: end);`) way.
Paper example is more similar than unofficial one to official one.
image

Unofficial way also make weird noise in especially "rising edge" and "falling edge".
But official sampling function is implemented with interpolation and filter according to matlab source code.

> RESAMPLE(X,Tx,Fs) uses interpolation and an anti-aliasing
> filter to resample the signal at a uniform sample rate, Fs, expressed in hertz (Hz).

I think we have 3 options for scaling.
1.Unofficial way
It is maybe easy way to implement, 
but It seems it's not reproduction of paper, and I don't know how to search how it works since that is not a function.

2.Official way
It is better in paper reproduction.  But the logic is a little bit complicated.

3.interpolation
Close to paper reproduction.

seanlaw:
> Unofficial way also make weird noise in especially "rising edge" and "falling edge".

I am okay with this because it's basically repeating some value multiple times before repeating the next value multiple times. What is more important is "ultimately, does the unofficial way (option 1) affect finding the motif?" If we are still able to find the motif using the unofficial way then I would do that first because it is straightforward/obvious to explain.

Then, at the end of the tutorial, I would add a "Bonus Section" (like our other tutorials) that explains how to use interpolation but that there is some precautions to know.

Does that make sense @ken-maeda?

ken-maeda:
OK, I understand. I searched what happend in unooficial resmapling. It is actually the easiest interpolation "nearest neighbor interpolation". Here is result, I can reproduce in python.
image
"Nearest neighbor intepolation" is recognized easiest and fastest calculation. It can be done in scipy also.

How should I do?

seanlaw:
@ken-maeda "nearest neighbor interpolation" seems fine but does it affect our ability to use `stumpy` to find the motif?

ken-maeda:
<p>Yes, it isn't, Here shows uniform scaling with matrix profile can find another "motif"(long-term motif and short-time motif). First of this chapter, I'd like to introduce normal matrix profile result with new dataset "MALLAT". Then move to unifrom scaling with matrix profile.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

seanlaw:
Thanks @ken-maeda! I will find some time to review it

seanlaw:
@ken-maeda  For the matrix profile examples, I think that we ALWAYS want to find the embedded/imbedded motif and not some other motif. So, we'd start with a stretched time series and demonstrate that we can shrink it in a systematic way and eventually recuperate the original embedded motif. And then demonstrate the same thing starting with a shruken time series that you stretch and then still be able to re-discover the same embedded motif.

ken-maeda:
<p>Once scaling is done, it can't be self-math? Because they are not same anymore. First of all, I show motif of each scaling in above fig. Then scaled motif is recuperated in below fig.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/667/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

seanlaw:
> it should be done with another below figure.?

No, I think the matrix profile section should start with Figure 3 bottom left (stretched) and comment that it has an embedded motif but, since it has been stretched, the matrix profile is "bad". So, we can shrink it systematically and recover the original time series through compression and basically stop when the distance between the motif pair is smallest.

ken-maeda:
Figure 3 bottom and Figure 3 upper should be independent. I thought the example is about finding different motif from same series by "streached".

>  comment that it has an embedded motif but, since it has been stretched, the matrix profile is "bad".

Why is it bad? Those motif is not real, however it is more similar than upper fig. We can shrink after that to fit real.

> Fig. 3 bottom clearly shows that there may be some
> situations in which there is a need to find motifs with invariance to uniform scaling. (From paper)

I might have miss the guide.

seanlaw:
> Why is it bad? Those motif is not real, however it is more similar than upper fig. We can shrink after that to fit real.

Maybe I'm wrong but I think that the average user won't be able to follow the logic of a changing motif. In other words, it is too complicated and we aren't discussing the relevance of the other motifs. Instead, we should focus our examples around being able to find the same embedded motif even if the time series were stretched/shrunken

ken-maeda:
I still can't understand how notebook should be by communicating in text.

> Instead, we should focus our examples around being able to find the same embedded motif even if the time series were stretched/shrunken

Dose it have to be "uniform scaled" to find same motif?

My notebook is poor and may not be conveying the intent well, so I'll make it simple.

To illustrate current notebook case with a real world example, A signal inculde "go up stairs(A)" and "go up stairs **slowly**(B)".
A and B have same signal structure, however they're not same. Therefore Matrix profile isn't able to find A and B as motif.

On the other hand, if B was shrunken to the length A by unifrom scaling, A and B become really "similar signal". So Matrix profile can find A and B(shrunken). We can now exstract A and the same action signal even if series length is different.

I think this kind of situation happens often in many fields.

seanlaw:
> To illustrate current notebook case with a real world example, A signal inculde "go up stairs(A)" and "go up stairs slowly(B)".
A and B have same signal structure, however they're not same. Therefore Matrix profile isn't able to find A and B as motif.
> On the other hand, if B was shrunken to the length A by unifrom scaling, A and B become really "similar signal". So Matrix profile can find A and B(shrunken). We can now exstract A and the same action signal even if series length is different.
> I think this kind of situation happens often in many fields.

Yes, exactly! I believe that most users will "look" (visualize) at their time series and see a motif pair and expect that the matrix profile should be able to capture the motif pair. And this motif pair should be super obvious except that one is "slow". So, to address this, we need "uniform scaling" to resolve it and then we can apply the matrix profile technique to find the motif pair. I think we can shorten the current example to match this. I think we are close except that we should always focus on finding the same embedded motif. We can still use the same example but we just need to make the point clear that "you can still find the same embedded motif if you shrink/stretch the time series"

seanlaw:
@ken-maeda  Please pull the latest changes/commits into your branch

ken-maeda:
Let me confirm, how should I fix notebook from here?

seanlaw:
> Let me confirm, how should I fix notebook from here?

I think what we may be missing is a storyline that helps to logically tie everything together rather than sounding too technical. My opinion is that we should tell the user:

If your dominant/major motif has the same window size then `stumpy.stump` works pretty well out of the box. However, if your pattern is either compressed/stretched (relative to the first time a similar subsequence is observed) then `stumpy.stump` might miss it. Then, we show examples (you don't need new examples, the existing ones work) and describe how to recover the motif. Note that we should stay with the same motif throughout and only find that motif through uniform scaling (whether we stretch or compress). This will require starting with a stretched time series and then demonstrating how to compress it and vice versa. Again, what matters is how easy/logical is the story to understand for somebody who wants to move quickly?

Unfortunately, with the matrix profile section, it doesn't seem like we could/should compress/stretch the entire time series. I think it is a little confusing in the paper that they only stretched/compressed the latter half of the time series as this implies you already know which sections need to be stretched/compressed. It almost feels like we'd need exclusion zones for an AB-join so that we can avoid the stretched/compressed self-match

ken-maeda:
> I think what we may be missing is a storyline that helps to logically

It seems a major probem is missing  storyline. The scaling rate and scaling area is defined previously. Data itself doesn't have the context. So reader might miss the porpose of notebook. However It's really hard to  generate storyline for this dataset.
So it might be better to use modified example of below example.
31_4_a
31_4_b

The purpose of this case is really obvious to find the motif on another day.
This data was pointed out unrealistic dataset, but it might be usable if modified.

> Unfortunately, with the matrix profile section, it doesn't seem like we could/should compress/stretch the entire time series.

I think several way to avlid self-match exist in uniform scaling with MP, I don't believe that the entire time series necessarily shouldn't be compressed/stretched. However explaing itself, it might blur the story line. I think we should put the storyline as the first priority for now.

Or, as minimum necessary element, it may be placed somewhere as a minimum adaptive example notebook....

NimaSarajpoor:
> @NimaSarajpoor Do you have any thoughts?

Sorry for my late response. I just noticed it when I was wandering here :) I will think about it and share with you if I have something :)

NimaSarajpoor:
@seanlaw  said:
> Unfortunately, with the matrix profile section, it doesn't seem like we could/should compress/stretch the entire time series. I think it is a little confusing in the paper that they only stretched/compressed the latter half of the time series as this implies you already know which sections need to be stretched/compressed. It almost feels like we'd need exclusion zones for an AB-join so that we can avoid the stretched/compressed self-match.

@ken-maeda said:
> I don't believe that the entire time series necessarily shouldn't be compressed/stretched.

So, I am going to talk about two scenarios:
(1) We have `T_A` and `T_B` that are separate from each other. (Note that `T_A` and `T_B` can be from the same time series `T` but still separated. For instance: `T=np.r_[T_A, T_B]`)

Here, we can search for motifs by calculating `stumpy.stump(T_A, T_B_scaled)`

(2) We want to find motifs in `T` while considering the `scaling` technique. Let's say I scale `T` by the factor `1.000001`. In such case, if length of time series is small, we may assume `i` in `T` is probably very close to its corresponding index in `T_scaled`. Also, I probably prefer to exclude trivial matches because our scaling factor is very small. In other words, `T_scaled[i_prime + m]`  is very similar to `T[i + m]`, where `i_prime` is the index corresponds to `i` after the scaling process. (Is this true? I do not know! Let's think about it!) But, what if I increase the scaling factor. Should I still exclude self match? At what point, I can say `T[i + m]` and `T_scaled[i_prime + m]` are different.


I haven't thought this through completely. So, I may have said something that is meaningless :) But, I think this is what one should think about.

seanlaw:
> So, I am going to talk about two scenarios:

Yes! I was thinking the same thing @NimaSarajpoor 

> (1) We have `T_A` and `T_B` that are separate from each other. (Note that `T_A` and `T_B` can be from the same time series `T` but still separated. For instance: `T=np.r_[T_A, T_B]`)
> 
> Here, we can search for motifs by calculating `stumpy.stump(T_A, T_B_scaled)`

This is the happy path where you somehow got lucky and:

i) You have a roughly isolated pattern contained in T_A and
ii) the pattern is already separated from the time series T_B you'd like to search in

However, it feels somewhat contrived to start with this premise. I don't think it is clear to users how one actually gets to this state since, usually, one starts with a single time series And if the inputs don't look their their inputs then they will either not use the tutorial or they will ask us "how should I isolate/separate my time series into T_A and T_B?". I have no good answer for this and, again, it feels like a bad idea to publish a tutorial without that answer.

> (2) We want to find motifs in `T` while considering the `scaling` technique. Let's say I scale `T` by 0.1%. In such case, if length of time series is small, we may assume `i` in `T` is probably very close to its corresponding index in `T_scaled`. Also, I probably prefer to exclude trivial matches because our scaling factor is very small. In other words, `T_scaled[i_prime + m]`  is very similar to `T[i + m]`, where `i_prime` is the index corresponds to `i`. (Is this true? I mean...is it similar to self-match case? I do not know! Let's think about it!) But, what if I increase the scaling factor? Should I still exclude self match? At what point, I can say `T[i + m]` and `T_scaled[i_prime + m]` are different.

I agree 100%! I have no idea how to avoid or exclude the self-match/trivial-match when the scaling is small. 

@ken-maeda I feel that unless we are able to address these two points rationally, the tutorial will be confusing as it will invite too many questions from users and that we will not have an answer to.

ken-maeda:
About 2nd topics(self-match),
As @seanlaw and @NimaSarajpoor mentioned, scaling factor is important parameter. So search range of scaling in my notebook is not calculated on around 100% scaling rate to avoid it. But I can show how large/small scaling rate is required to break self-match in calculation space.

On the other hand, if we chose above fig for more clear storyline, We don't have self-match problem anymore, do we?
(They are separeted series)

> usually, the user starts with a single time series

Do you prioritize this? and think of another way?

seanlaw:
> On the other hand, if we chose above fig for more clear storyline, We don't have self-match problem anymore, do we?
(They are separeted series)

Yes, this is the same as case 1 that @NimaSarajpoor was talking about. However, how does one get to the point of starting with `T_A` (red) and `T_B` (green)?

seanlaw:
> Yes, this is the same as case 1 that @NimaSarajpoor was talking about. However, how does one get to the point of starting with T_A (red) and T_B (green)?

@ken-maeda Just to complete the point. Instead of `T_A` being the red pattern, what if `T_A = np.random.rand(n)` (i.e., `T_A` was not an obvious pattern that could be found in `T_B`)? My point is that most users will not know what pattern to use in `T_A`. So what is the workflow that they should use to eventually get to your `T_A` (red) and `T_B` (green) situation?

ken-maeda:
Which kind of info should I add to fit in storyline?
1.[T_A, T_B] normal MP => can't be motif
2.We know both signal has common signal structure by domain knowlege. => streach series 
3.[T_A, T_B] scaling MP => can be motif

> Instead of T_A being the red pattern, what if T_A = np.random.rand(n)
> i.e., T_A was not an obvious pattern that could be found in T_B

The example in this case has motivation to streach series by domain knowledge. Also user at leaset should have reasonable "clue" to streach series. We may need reproducing failures tha are likely to occur in real life if a failure example is needed.

> So what is the workflow that they should use to eventually get to your T_A (red) and T_B (green) situation?

You meant both series are removed except motif parts, and it is not realistic?
If so, I have to modify dataset or find better one.

seanlaw:
> Which kind of info should I add to fit in storyline?
1.[T_A, T_B] normal MP => can't be motif
2.We know both signal has common signal structure by domain knowlege. => streach series
3.[T_A, T_B] scaling MP => can be motif

I think this is mostly sufficient. Perhaps, the only thing remaining is how do you map/relate the stretched motif back onto the original raw `T_B`. In step 3, you would get the index, `idx_stretched`, from `T_B_stretched` but what `idx` does `idx_stretched` correspond to in the original raw `T_B`?

Maybe I've been overthinking it and your comment about "domain knowledge" is relevant. Initially, in my brain, I was hoping that we could guide the user to find a "better motif" through uniform scaling WITHOUT any domain knowledge. Even in this case, `T_A` itself must be derived from some domain knowledge and I am hoping that we can remove the need for domain knowledge so:

1. `T_A` is simply historical data starting from time `t=0` to some later time `t=j`
2. And we want to know if any subsequence within `T_A` has roughly repeated itself after time `t > j` and we call this "future" data `T_B`
3. If we try to compute the matrix profile directly for `T_A` and `T_B` then we'll get some motif but it isn't too strong 
4. However, if we uniformly scale `T_B` then we may be able to account for any stretching/compressing that would/could lead to a motif with a smaller Euclidean distance

I think you basically have this and it's just a matter explaining this to the user in a simple manner. The only challenge is that the user needs to choose an appropriate `j` and they

ken-maeda:
> The only challenge is that the user needs to choose an appropriate j and they

I agree with you idea, but When I use above figure for it. T_A and T_B are too obvious to search which part of signal should be compared. I'm getting more and more confused now. I have tried some examples, 
although It's really hard to create  notebook without good dataset with storyline.  I'll pend until finding good dataset.

seanlaw:
@ken-maeda Could we simply go back to using the [steamgen dataset](https://stumpy.readthedocs.io/en/latest/Tutorial_STUMPY_Basics.html#Visualizing-the-Steamgen-Dataset) but modify it so that we stretch all values `j > 2000` (i.e., we'll need to shrink it)?

ken-maeda:
I tried other category datasets, it works well to find the motif. But it's really hard to set the reason why I'm going to use uniform scaling and why best-motif(unifrom scaled) are detected(expalin the motifs relationship). 

> Could we simply go back to using the [steamgen dataset](https://stumpy.readthedocs.io/en/latest/Tutorial_STUMPY_Basics.html#Visualizing-the-Steamgen-Dataset)
but modify it so that we stretch all values j > 2000 (i.e., we'll need to shrink it)?

It is almost like fig3(on the paper) example below. It had problem important things are defined by domain knowledge.
image
This example also isn't obvious about how to decide scaling area and relationship of best motif(uniform scaled).

So I was tring to get public dataset for it.

@seanlaw steamgen example is ok for you?

seanlaw:
@ken-maeda I agree with you. It's not too different from fig3. I'm wondering if it may not be a good use case to convey to the user and maybe we should abandon it and avoid attracting more questions from users :)

Having said that, it is nice that we can show how to perform uniform scaling. Maybe I can accept the PR notebook and direct users to the notebook when they ask but we don't publish the notebook to our tutorials. How do you feel about that? We've done that with some of our other notebooks as well that aren't "ready" for a wider audience or that are a work-in-progress.

ken-maeda:
Sounds good, Let you know, when I get better dataset.
I'm using two dataset converted from paper figure, should I store those somewhere to  run the notebook?

seanlaw:
> I'm using two dataset converted from paper figure, should I store those somewhere to run the notebook?

Can you try dragging and dropping those dataset files here in a comment and then update the notebook to point to the files here?

ken-maeda:
What else can I do?

seanlaw:
@ken-maeda One final thing. Can you please rename the notebook from `docs/Tutorial_Matrix_Profile_Top_Ten_short.ipynb` to `docs/Tutorial_Uniform_Scaling.ipynb`? 

After that, I think it should be ready to merge

ken-maeda:
@seanlaw I fixed it.

seanlaw:
Thank you for this contribution @ken-maeda!

PR #674, opened 2022-09-13
Fix #673 Fix two minor issues in motifs and aamp_motifs

### LINKED ISSUE #673: A few minor issues in stumpy.motifs
I investigated `stumpy.motifs` for the data provided in Discussion #670 (also attached here: [your_time_seriesa.csv](https://github.com/TDAmeritrade/stumpy/files/9546933/your_time_series.csv)) and, throughout my investigation, I noticed there are a few things that could/should be taken care of.

(1) The default value of `cutoff` can be set to `0`. While this is not a problem, it is an extreme condition that only results in finding perfect motif (i.e. `distance-to-nn = 0`). So, I think the following change might be good:
```
if cutoff is None:  # pragma: no cover
        P_copy = P.copy().astype(np.float64)
        P_copy[np.isinf(P_copy)] = np.nan
        
        cutoff = np.nanmax(
            [np.nanmean(P_copy) - 2.0 * np.nanstd(P_copy), A]
        )
       
        if cutoff == 0:
              cutoff = np.sort(np.unique(P_copy))[1]  # np.partition is a better choice 
```
In fact `cutoff=0` is the main reason behind the problem discussed in Discussion #670 .

(2) `query_matches` in:

```
# in stumpy/motifs.py

# in function _motifs(...)

query_matches = match(  
            Q,
            T,
            M_T=M_T,
            Σ_T=Σ_T,
            max_matches=None,
            max_distance=max_distance,
            atol=atol,
            query_idx=candidate_idx,
        )

if len(query_matches) > min_neighbors:
        motif_distances.append(query_matches[:max_matches, 0])
        motif_indices.append(query_matches[:max_matches, 1])

      
for idx in query_matches[:, 1]:
    core.apply_exclusion_zone(P, int(idx), excl_zone, np.inf)
```
might be empty. In such case, `for idx in query_matches[:, 1]:` can result in error. Hence, we should do:

```
if len(query_matches) > 0:
      for idx in query_matches[:, 1]:
          core.apply_exclusion_zone(P, int(idx), excl_zone, np.inf)
else:
    core.apply_exclusion_zone(P, candidate_idx, excl_zone, np.inf) 
```

(3) Let's say a motif at index `idx` has matrix profile value of `v`. So, we expect to see `match_distances=[v, ..]` and `match_indices=[idx, ...]`. However, if we use the attached data, we can see that:

```
m = 5
normalize = True

mp = stumpy.stump(T, m, normalize=normalize)
P = mp[:, 0].astype(np.float64)

motif_dist, motif_indices = stumpy.motifs(
    T, 
    P, 
    min_neighbors=1,
    max_distance=None,
    cutoff=None,
    max_motifs=2,
)
```
And,
```
>>> motif_dist[1]
[1.10509969e-07, ...]

>>> motif_indices[1]
[151, ...]
```

However, `mp[151, 0]` is `0.0`. So, we should have seen `0.0` instead of `1.10509969e-07`. I am not sure if I missed anything here or not. Note that the `cutoff` is set  to `0` in this case. This means that the result might sound strange because when `cutoff` is `zero`, we should expect `0` as the first element in `motif_dist[1]`

NimaSarajpoor:
This PR resolves the issues (1) and (2) discussed in #673.

Fix (1): Throwing warning when `cutoff` is set to `0.0`
Fix (2): Replace empty array with a fake 2D array to allow traversing over its second column (and avoid choosing the already selected `candidate_idx` in the next iteration)

### DIFF
--- a/stumpy/aamp_motifs.py
+++ b/stumpy/aamp_motifs.py
@@ -272,6 +272,17 @@ def aamp_motifs(
             [np.nanmean(P_copy) - 2.0 * np.nanstd(P_copy), np.nanmin(P_copy)]
         )
 
+    if cutoff == 0.0:  # pragma: no cover
+        suggested_cutoff = np.partition(P, 1)[1]
+        logger.warn(
+            "The `cutoff` has been set to 0.0 and may result in little/no candidate "
+            "motifs being identified."
+        )
+        logger.warn(
+            "You may consider relaxing the constraint by increasing the `cutoff` "
+            f"(e.g., cutoff={suggested_cutoff}."
+        )
+
     T, T_subseq_isfinite = core.preprocess_non_normalized(T[np.newaxis, :], m)
     P = P[np.newaxis, :].astype(np.float64)
 
--- a/stumpy/aamp_motifs.py
+++ b/stumpy/aamp_motifs.py
@@ -129,6 +129,10 @@ def _aamp_motifs(
             motif_distances.append(query_matches[:max_matches, 0])
             motif_indices.append(query_matches[:max_matches, 1])
 
+        if len(query_matches) == 0:  # pragma: no cover
+            # create fake `query_matches`
+            query_matches = np.array([[np.nan, candidate_idx]])
+
         for idx in query_matches[:, 1]:
             core.apply_exclusion_zone(P, int(idx), excl_zone, np.inf)
 
--- a/stumpy/aamp_motifs.py
+++ b/stumpy/aamp_motifs.py
@@ -284,7 +284,7 @@ def aamp_motifs(
         )
         logger.warn(
             "You may consider relaxing the constraint by increasing the `cutoff` "
-            f"(e.g., cutoff={suggested_cutoff}."
+            f"(e.g., cutoff={suggested_cutoff})."
         )
 
     T, T_subseq_isfinite = core.preprocess_non_normalized(T[np.newaxis, :], m)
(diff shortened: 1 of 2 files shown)


stumpy/aamp_motifs.py
```diff
@@ -129,6 +129,10 @@ def _aamp_motifs(
             motif_distances.append(query_matches[:max_matches, 0])
             motif_indices.append(query_matches[:max_matches, 1])
 
+        if len(query_matches) == 0:  # pragma: no cover
+            # create fake `query_matches`
```
seanlaw, line 133: Maybe "create empty `query_matches`" is slightly better?

NimaSarajpoor: @seanlaw 
Just to make sure you did not mistype anything... so I should use the term "empty"? (because the array being empty was the main issue). But, if that is okay with you, I can go ahead and change it for sure. (I had the term `pseudo` in mind but I think it is not suitable here)

seanlaw: Or maybe just omit the comment altogether?

NimaSarajpoor: Sure.

seanlaw:
@NimaSarajpoor Looks like the tests have all passed and we are ready to merge. May I?

NimaSarajpoor:
@seanlaw 
> @NimaSarajpoor Looks like the tests have all passed and we are ready to merge. May I?

Sure. It seems everything is okay here.

seanlaw:
@NimaSarajpoor Thanks again for this PR!

Issue #677, opened 2022-09-16
bug in `stumpy/snippet.py`


NimaSarajpoor:
I think there is a ~minor~ bug in `stumpy/snippet.py`. one can see the bug by setting param `s` to a value less than or equal to `m / 2`.

To reproduce the error:
```
T = np.random.rand(20)
m = 5
k = 2
s=2
stumpy.snippets(T, m, k=k, s=s)
```

seanlaw:
[This](https://github.com/TDAmeritrade/stumpy/pull/878#issuecomment-1632235605) may possibly be related

NimaSarajpoor:
Note:
I think the bug mentioned in the first post, i.e.

> I think there is a minor bug in stumpy/snippet.py. one can see the bug by setting param s to a value less than or equal to m / 2.

is now resolved: see #823 in which the slicing (of arrays) is changed from `start : start + s` to `start : stop - s + 1` (see the changes in the snippet module)

Having said that, a recent error is observed:
 
> [This](https://github.com/TDAmeritrade/stumpy/pull/878#issuecomment-1632235605) may possibly be related:
> 
> https://github.com/TDAmeritrade/stumpy/actions/runs/5520192550/jobs/10066477926

I will check that out.

NimaSarajpoor:
> Having said that, a recent error is observed:
> 
> > [This](https://github.com/TDAmeritrade/stumpy/pull/878#issuecomment-1632235605) may possibly be related:
> > https://github.com/TDAmeritrade/stumpy/actions/runs/5520192550/jobs/10066477926
> 
> I will check that out.

[update] 
As discussed in #881, the issue in #881 is neither `self-match` nor `s < m/2`. Hence, I think it should be safe to close this issue as it has been addressed already.

PR #668, opened 2022-09-07
Fix precision

NimaSarajpoor:
This PR is a replacement for PR #657, which tackles issue #610. 

Regarding PR #657: There was a discrepancy between the files in the local repo and the remote repo, but I couldn't figure that out. That is why I am submitting a new PR.

PR #678, opened 2022-09-16
Fix 677 fix bug in snippet module

NimaSarajpoor:
This PR addresses issue #677. It resolves the bug. 

After solving the issue, I realized there exist another problem. In some cases, there is a (small?) difference between the output of the performant and naive version. This might be related to the slight imprecision in calculation. 

I added a new test function to `tests/test_snippet.py` with a specified random seed **to reproduce the error when I push the commits here.**


tests/naive.py
```diff
@@ -1087,18 +1087,18 @@ def mpdist_vect(T_A, T_B, m, percentage=0.05, k=None):
...
 
-    k = min(int(k), P_ABBA.shape[0] - 1)
-
-    for i in range(n_B - n_A + 1):
+    for i in range(MPdist_vect.shape[0]):
         P_ABBA[:j] = stump(T_A, m, T_B[i : i + n_A])[:, 0]
         P_ABBA[j:] = stump(T_B[i : i + n_A], m, T_A)[:, 0]
         P_ABBA.sort()
-        MPdist_vect[i] = P_ABBA[min(k, P_ABBA.shape[0] - 1)]
+        MPdist_vect[i] = P_ABBA[k]
```
seanlaw, line 1101: Why did you change this?

stumpy/snippets.py
```diff
@@ -107,8 +107,8 @@ def _get_all_profiles(
             S_i,
             T,
             s,
-            M_T[start : start + s],
-            Σ_T[start : start + s],
+            M_T[start : start + m],
```
seanlaw, line 110: Why should this be `m` instead of `s`?

seanlaw (commented):
Do we need to update `aamp_snippets.py` as well?

NimaSarajpoor:
> Do we need to update `aamp_snippets.py` as well?

I think we are good in the non-normalized version.

tests/naive.py
NimaSarajpoor: **just to clean it the code**

I think because we already had  `k = min(int(k), P_ABBA.shape[0] - 1)` before this for-loop, in line 1095 (in `tests/naive.py` of the `main` branch). So, I think it was redundant.

seanlaw: Cool. Thanks

seanlaw:
> I think we are good in the non-normalized version.

Why wouldn't we need to also do:

```
M_T[start : start + m],
```

In the non-normalized version?

stumpy/snippets.py
NimaSarajpoor: `M_T[start : start + ....]` should be the sliding mean of subseqs (of length `s`) in `S_i`. `S_i` itself has length `m`. So, I think a more accurate value for the sliding mean is `M_T[start : start + m - s + 1]` (I just realized this latter part). But since we iterate over subseqs of length `s` in `S_i`, `M_T[start : start + m]` works. 

However, it might be better to do `M_T[start : start + m - s + 1]` as I think it is clearer and more readable in this case.

seanlaw: I see. Thank you for explaining it

NimaSarajpoor:
> > I think we are good in the non-normalized version.
> 
> Why wouldn't we need to also do:
> 
> ```
> M_T[start : start + m],
> ```
> 
> In the non-normalized version?

I did not go through the code thoroughly. But, if I remember correctly, I did not find `M_T` and then I remembered non-normalized version does not have `M_T`. (or maybe it has in snippts and I missed it)

NimaSarajpoor:
@seanlaw 
btw, for `k=2`, we got error (PROBABLY due to imprecision). I changed it to `k=1` as my goal was to fix this issue for now.

I think you may want to check the test for `k=2` and see the error. And, if you think it is imprecision, we can just close this PR with `k=1` for the new test functions.

stumpy/snippets.py
NimaSarajpoor: > However, it might be better to do `M_T[start : start + m - s + 1]` as I think it is clearer and more readable in this case.

If that is okay with you, I am going to use `M_T[start : start + m - s + 1]` as I think it forces reader to think (and hopefully remember the reason behind doing `+ m -  s  + 1`)

seanlaw: Let's do whatever is the best option

tests/test_aampdist_snippets.py
```diff
@@ -7,8 +7,8 @@
 
 
 test_data = [np.random.uniform(-1000, 1000, [64]).astype(np.float64)]
-s = [6, 7, 8]
-percentage = [0.7, 0.8, 0.9]
+s = [3, 6, 7, 8]
```
seanlaw, line 10: Why are these added in `test_aampdist_snippets.py` rather than `test_snippets.py`?

stumpy/snippets.py
seanlaw: Perfect

tests/test_aampdist_snippets.py
NimaSarajpoor: I wanted to make sure we do not have the problem in `test_aampdist_snippets.py`. I also added them in `test_snippets.py`, but then I got failure, so I decided to create two new test functions for them. So, in the new test function, I am testing the function with `s` that has low value. Same goes for `percentage`. However, now that the test suite is passing, I think I should better try putting the value `3` back into the list and see how it geos.

NimaSarajpoor: @seanlaw 
So, I tested `snippets.py` after adding `3` to list `s`, i.e. `s=[3, 6, 7, 8]`. I did the same for percentege: `percentage = [0.3, ...]`
And, I got failure. I think it is related to precision. 

Do you want me to push it so you can see  it here?

seanlaw: Sure

NimaSarajpoor:
I think we also need to revise this line of code in both normalized and non-normalized version:

```
# last for-loop in function snippets / aampdist_snippets
total_min = total_min - mask.astype(np.float64)
```

I think the whole idea is to just make `total_min` negative so they cannot be considered in the next iteration of the for-loop when we do `mask = snippets_profiles[i] <= total_min`. I think a better /reasonable approach is to do:

```
total_min[mask] = np.NINF
```
I think `total_min = total_min - mask.astype(np.float64)` is effective ONLY IF `total_min` is less than 1.

seanlaw:
> I think `total_min = total_min - mask.astype(np.float64)` is effective ONLY IF `total_min` is less than 1. However, this is not gauranteed. So, `np.NINF` is better.

Honestly, I don't remember this part and how it works. I'll trust your judgement

NimaSarajpoor:
@seanlaw 
Please feel free to check out the last test function I added to the snippet unit test. As you can see, the test fails.

seanlaw:
> Please feel free to check out the last test function I added to the snippet unit test. As you can see, the test fails.

The difference in value seems far too big. Have you been able to confirm the root cause for the difference?

NimaSarajpoor:
@seanlaw 
> Have you been able to confirm the root cause for the difference?

So, after some investigation, I noticed that both naive and performant version have close values in `snippets_profiles` (i.e.  `npt.assert_almost_equal(...)` does not fail for this output). However, we can see naive and performant have different results in  `snippets_fractions`. Note that `snippets_fractions` is calculated as follows: 

```
# total_min = np.min(snippets_profiles, axis=0)

mask = snippets_profiles[i] <= total_min
snippets_fractions[i] = np.sum(mask) / total_min.shape[0]
```
Therefore, a small imprecision in `float` values in `snippets_profiles` can change a boolean value in `mask`, which means the numerator `np.sum(mask)` can be changed by `1` (which is a big number). So, if you remove the constant denominator `total_min.shape[0]` and checkout  the values `snippets_fractions` again (so basically just `np.sum(mask)`), you can see the difference is just 1 unit! I do not have the values right now, but if I remember correctly, the `ref`(naive) was `28, 19, 10` and `comp` (performant) was `27, 20, 10`. So, we can see there is `1` unit of difference for the `k=3` snippets. 

I may investigate further to find the exact element in `snippets_profiles` that gives different boolean value when we do `snippets_profiles[i] <= total_min`

seanlaw:
> So, to wrap up: imprecision

Technically speaking, which one is actually "off"? Is it the naive version or the performant version?

NimaSarajpoor:
> which one is actually "off"? Is it the naive version or the performant version?

It is the performant version. To be more precise, the distance profile between query `T[40:43]` and `T` must be  `0` at index `40` (because they are the same!!) However, this value, obtained through `core._mass(...)`, is `1.9 * 10e-6` instead. And then, I investigated further and I realized the calculated pearson, in `stumpy/core.py::_calculate_squared_distance`, between `T[40:43]` and `T[40:43]` becomes `0.9999999999994555` (should have been `1.0`). 

So, what can we do?
* Add threshold for pearson in `stumpy/core.py::_calculate_squared_distance`.
* Replace numba `fastmath=True` flag by a less restrictive set of flags: e.g. avoid using `arcp`, `reassoc` (but then we need to compare the computation time of the new version and old version)

---

**side note**
In `stumpy/core.py::_calculate_squared_distance`, we have the following line:
```
D_squared = np.abs(2 * m * (1.0 - (QT - m * μ_Q * M_T) / denom))
```
However, I think we should better do this:
```
pearson = QT - m * μ_Q * M_T) / denom
if pearson > 1.0:
    pearson = 1.0
D_squared = np.abs(2 * m * (1.0 - pearson))
```
Because, now, if `pearson` becomes bigger than 1.0, say `1.00001`, we reset it to 1.0, which makes `D_squared` zero. However, previous code gives non-zero value for `D_squared`.

NimaSarajpoor:
@seanlaw 


> > which one is actually "off"? Is it the naive version or the performant version?
> 
> It is the performant version. To be more precise, the distance profile between query `T[40:43]` and `T` must be `0` at index `40` (because they are the same!!)

The query `T[40:43]` is `Q=np.array([-840.66778922 -886.33759718 -843.33818273])`. Now, if we use the formula:
```
denom = m * σ_Q * Σ_T
pearson = (QT - m * μ_Q * M_T) / denom  # in this case, QT can be calculated as follows np.dot(Q, Q) 
```
to calculate `pearson`, we can see that it is not 1.0.  Hence, the `pearson-distance` relationship itself causes imprecision.

seanlaw:
> Because, now, if pearson becomes bigger than 1.0, say 1.00001, we reset it to 1.0, which makes D_squared zero. However, previous code gives non-zero value for D_squared.

Awesome. I like that! Please add it. 

Btw, should there be an equivalent lower bound? Could we use `np.clip`?

NimaSarajpoor:
> It should be this, right:
> 
> ```
> if pearson > config.STUMPY_CORRELATION_THRESHOLD:
>     pearson = 1.0
> ```

Correct. It can now resolve the following issue as well
> and I realized the calculated pearson, in stumpy/core.py::_calculate_squared_distance, between T[40:43] and T[40:43] becomes 0.9999999999994555 (should have been 1.0).



> Btw, should there be an equivalent lower bound? Could we use `np.clip`?

Technically yes. But, practically speaking, we should be okay if we do not use it. For instance, let's assume the subsequence `S_i` and its 1NN has pearson correlation `-1`. And, let's say there is another subsequence, `S_j`, whose pearson correlation with its 1NN is `-1.002`. If we do not replace `-1.002` with `-1.0`, we will get a distance that is bigger than the possible maximum value and we may choose it as the first discord before considering `S_i`. If you still think we should consider lower bound, it should be okay to just do:

```
if pearson < -1.0:
     pearson = -1.0
```

seanlaw:
> The problem in this PR is in mass and the problem in PR https://github.com/TDAmeritrade/stumpy/pull/668 is in _stump. However, they are somehow related because the main cause is the pearson correlation in both cases. so, I can modify both functions _calculate_squared_distance and _stump in PR https://github.com/TDAmeritrade/stumpy/pull/668. (and then add the last test function in test_snippets.py in the PR https://github.com/TDAmeritrade/stumpy/pull/668 .


I'd say please add modify both functions _calculate_squared_distance and _stump in PR #668 but add the last test function here in #677.

Issue #685, opened 2022-09-24
Add Novelet Notebook Reproducer


seanlaw:
[Novelets paper](https://static1.squarespace.com/static/56dca322f8baf30cf9752a85/t/631900e16f7b7e792701943a/1662583010347/Novelets_V029_CameraReady.pdf)

NimaSarajpoor:
**Just an idea:**
It seems this paper is inspired by the contrast profile paper (see #465). So, it might be beneficial to read that paper first and provide a tutorial notebook for it, and then continue working on this one.

PR #692, opened 2022-10-14
[WIP] #685 Tutorial on Novelets

NimaSarajpoor:
This addresses issue #685, Novelet Notebook. I was talking about this paper with one of my friends, and I decided to implement it.  And, then I decided to create a notebook and push it here. 

I have not reproduced the figures yet. For now, I just tried to provide the gist of paper and implement the algorithm and show the result.

Next steps:
* Check out MATLAB code  to ensure the authors remain faithful to the algo provided in the paper
* Reproduce figures


NimaSarajpoor:
@seanlaw 
side note: the error raised [here](https://github.com/TDAmeritrade/stumpy/actions/runs/3252262194/jobs/5338274531) might be one of those (easy-to-resolve?) cases that might be resolved by the following:

* setting config variable `STUMPY_STDDEV_THRESHOLD` to smaller value: `1e-20`
* refine std calculation if it is less than 1.0 (recall that this can mean not taking advantage of rolling std when we have T = `np.random.rand()` as std will be definitely less than 1.0

seanlaw:
Thanks @NimaSarajpoor. Looks good

PR #714, opened 2022-11-11
[WIP] Fix #639 Top-K Nearest Neighbors to Matrix Profile (normalize=False)

### LINKED ISSUE #639: Add Top-K Nearest Neighbors to Matrix Profile (normalize=False)
For non-normalized methods, add top-k matrix profile capability.

This is related to #592 and #640

NimaSarajpoor:
This PR addresses #639 . This is a follow up to PR #595.

We are going to add top-k support for the following modules:

- [x] aamp
- [x] aamped
- [x] gpu_aamp
- [x] scraamp
- [x] aampi

----

**Issues Tracker (so that we do not forget)**
- [ ] In `scraamp` (and `scrump`), the following if-block is (probably?) never executed
```
if P_NORM[thread_idx, i, 0] == np.inf:  # pragma: no cover
            I[thread_idx, i, 0] = -1
            continue
```

- [ ] In `scraamp`, we may do `if-continue` in the if block below to avoid computing the rest of code
```
if not T_A_subseq_isfinite[i]:  # pragma: no cover
            p_norm_profile[:] = np.inf
            continue # NEW
```

### DIFF
--- a/stumpy/aamp.py
+++ b/stumpy/aamp.py
@@ -255,7 +255,8 @@ def _aamp(
     return np.power(P[0, :, :], 1.0 / p), I[0, :, :]
 
 
-def aamp(T_A, m, T_B=None, ignore_trivial=True, p=2.0):
+def aamp(T_A, m, T_B=None, ignore_trivial=True, p=2.0, k=1):
+    # function needs to be changed to return top-k matrix profile
     """
     Compute the non-normalized (i.e., without z-normalization) matrix profile
 
@@ -281,6 +282,11 @@ def aamp(T_A, m, T_B=None, ignore_trivial=True, p=2.0):
 
     p : float, default 2.0
         The p-norm to apply for computing the Minkowski distance.
+
+    k : int, default 1
+        The number of top `k` smallest distances used to construct the matrix profile.
+        Note that this will increase the total computational time and memory usage
+        when k > 1.
 
     Returns
     -------
--- a/stumpy/aamp.py
+++ b/stumpy/aamp.py
@@ -88,6 +88,8 @@ def _compute_diagonal(
     """
     n_A = T_A.shape[0]
     n_B = T_B.shape[0]
+    uint64_m = np.uint64(m)
+    uint64_1 = np.uint64(1)
 
     for diag_idx in range(diags_start_idx, diags_stop_idx):
         k = diags[diag_idx]
@@ -98,41 +100,54 @@ def _compute_diagonal(
             iter_range = range(-k, min(n_A - m + 1, n_B - m + 1 - k))
 
         for i in iter_range:
-            if i == 0 or (k < 0 and i == -k):
+            uint64_i = np.uint64(i)
+            uint64_j = np.uint64(i + k)
+
+            if uint64_i == 0 or uint64_j == 0:
                 p_norm = (
-                    np.linalg.norm(T_B[i + k : i + k + m] - T_A[i : i + m], ord=p) ** p
+                    np.linalg.norm(
+                        T_B[uint64_j : uint64_j + uint64_m]
+                        - T_A[uint64_i : uint64_i + uint64_m],
+                        ord=p,
+                    )
+                    ** p
                 )
             else:
                 p_norm = np.abs(
                     p_norm
-                    - np.absolute(T_B[i + k - 1] - T_A[i - 1]) ** p
-                    + np.absolute(T_B[i + k + m - 1] - T_A[i + m - 1]) ** p
+                    - np.absolute(T_B[uint64_j - uint64_1] - T_A[uint64_i - uint64_1])
+                    ** p
+                    + np.absolute(
+                        T_B[uint64_j + uint64_m - uint64_1]
+                        - T_A[uint64_i + uint64_m - uint64_1]
+                    )
+                    ** p
                 )
 
             if p_norm < config.STUMPY_P_NORM_THRESHOLD:
                 p_norm = 0.0
 
-            if T_A_subseq_isfinite[i] and T_B_subseq_isfinite[i + k]:
+            if T_A_subseq_isfinite[uint64_i] and T_B_subseq_isfinite[uint64_j]:
                 # Neither subsequence contains NaNs
-                if p_norm < P[thread_idx, i, 0]:
-                    P[thread_idx, i, 0] = p_norm
-                    I[thread_idx, i, 0] = i + k
+                if p_norm < P[thread_idx, uint64_i, 0]:
+                    P[thread_idx, uint64_i, 0] = p_norm
+                    I[thread_idx, uint64_i, 0] = uint64_j
 
                 if ignore_trivial:
-                    if p_norm < P[thread_idx, i + k, 0]:
-                        P[thread_idx, i + k, 0] = p_norm
-                        I[thread_idx, i + k, 0] = i
+                    if p_norm < P[thread_idx, uint64_j, 0]:
+                        P[thread_idx, uint64_j, 0] = p_norm
+                        I[thread_idx, uint64_j, 0] = uint64_i
 
-                    if i < i + k:
+                    if uint64_i < uint64_j:
                         # left matrix profile and left matrix profile index
-                        if p_norm < P[thread_idx, i + k, 1]:
-                            P[thread_idx, i + k, 1] = p_norm
-                            I[thread_idx, i + k, 1] = i
+                        if p_norm < P[thread_idx, uint64_j, 1]:
+                            P[thread_idx, uint64_j, 1] = p_norm
+                            I[thread_idx, uint64_j, 1] = uint64_i
 
                         # right matrix profile and right matrix profile index
-                        if p_norm < P[thread_idx, i, 2]:
-                            P[thread_idx, i, 2] = p_norm
-                            I[thread_idx, i, 2] = i + k
+                        if p_norm < P[thread_idx, uint64_i, 2]:
+                            P[thread_idx, uint64_i, 2] = p_norm
+                            I[thread_idx, uint64_i, 2] = uint64_j
 
     return
 
--- a/stumpy/aamp.py
+++ b/stumpy/aamp.py
@@ -241,22 +241,22 @@ def _aamp(
 
     Returns
     -------
-    profile : numpy.ndarray
+    out1 : numpy.ndarray
         The (top-k) matrix profile
 
-    indices : numpy.ndarray
-        The (top-k) matrix profile indices
-
-    left profile : numpy.ndarray
+    out2 : numpy.ndarray
         The (top-1) left matrix profile
 
-    left indices : numpy.ndarray
-        The (top-1) left matrix profile indices
-
-    right profile : numpy.ndarray
+    out3 : numpy.ndarray
         The (top-1) right matrix profile
 
-    right indices : numpy.ndarray
+    out4 : numpy.ndarray
+        The (top-k) matrix profile indices
+
+    out5 : numpy.ndarray
+        The (top-1) left matrix profile indices
+
+    out6 : numpy.ndarray
         The (top-1) right matrix profile indices
 
     Notes
--- a/stumpy/aamp.py
+++ b/stumpy/aamp.py
@@ -30,12 +30,18 @@ def _compute_diagonal(
     diags_stop_idx,
     thread_idx,
     P,
+    PL,
+    PR,
     I,
+    IL,
+    IR,
     ignore_trivial,
+    k,
 ):
     """
-    Compute (Numba JIT-compiled) and update P, I along a single diagonal using a single
-    thread and avoiding race conditions
+    Compute (Numba JIT-compiled) and update the (top-k) matrix profile P,
+    PL, PR, I, IL, and IR sequentially along individual diagonals using a single
+    thread and avoiding race conditions.
 
     Parameters
     ----------
@@ -48,12 +54,6 @@ def _compute_diagonal(
 
     m : int
         Window size
-
-    P : numpy.ndarray
-        Matrix profile
-
-    I : numpy.ndarray
-        Matrix profile indices
 
     T_A_subseq_isfinite : numpy.ndarray
         A boolean array that indicates whether a subsequence in `T_A` contains a
@@ -78,9 +78,32 @@ def _compute_diagonal(
     thread_idx : int
         The thread index
 
+    P : numpy.ndarray
+        The (top-k) matrix profile, sorted in ascending order per row
+
+    PL : numpy.ndarray
+        The top-1 left marix profile
+
+    PR : numpy.ndarray
+        The top-1 right marix profile
+
+    I : numpy.ndarray
+        The (top-k) matrix profile indices
+
+    IL : numpy.ndarray
+        The top-1 left matrix profile indices
+
+    IR : numpy.ndarray
+        The top-1 right matrix profile indices
+
     ignore_trivial : bool
         Set to `True` if this is a self-join. Otherwise, for AB-join, set this to
         `False`. Default is `True`.
+
+    k : int
+        The number of top `k` smallest distances used to construct the matrix profile.
+        Note that this will increase the total computational time and memory usage
+        when k > 1.
 
     Returns
     -------
@@ -92,16 +115,16 @@ def _compute_diagonal(
     uint64_1 = np.uint64(1)
 
     for diag_idx in range(diags_start_idx, diags_stop_idx):
-        k = diags[diag_idx]
+        g = diags[diag_idx]
 
-        if k >= 0:
-            iter_range = range(0, min(n_A - m + 1, n_B - m + 1 - k))
+        if g >= 0:
+            iter_range = range(0, min(n_A - m + 1, n_B - m + 1 - g))
         else:
-            iter_range = range(-k, min(n_A - m + 1, n_B - m + 1 - k))
+            iter_range = range(-g, min(n_A - m + 1, n_B - m + 1 - g))
 
         for i in iter_range:
             uint64_i = np.uint64(i)
-            uint64_j = np.uint64(i + k)
+            uint64_j = np.uint64(i + g)
 
             if uint64_i == 0 or uint64_j == 0:
                 p_norm = (
@@ -129,25 +152,40 @@ def _compute_diagonal(
 
             if T_A_subseq_isfinite[uint64_i] and T_B_subseq_isfinite[uint64_j]:
                 # Neither subsequence contains NaNs
-                if p_norm < P[thread_idx, uint64_i, 0]:
-                    P[thread_idx, uint64_i, 0] = p_norm
-                    I[thread_idx, uint64_i, 0] = uint64_j
 
-                if ignore_trivial:
-                    if p_norm < P[thread_idx, uint64_j, 0]:
-                        P[thread_idx, uint64_j, 0] = p_norm
-                        I[thread_idx, uint64_j, 0] = uint64_i
+                # `P[thread_idx, i, :]` is sorted ascendingly and MUST be updated
+                # when the newly-calculated `p_norm` value becomes smaller than the
+                # last (i.e. greatest) element in this array. Note that the goal
+                # is to have top-k smallest distancs for each subsequence.
+                if p_norm < P[thread_idx, uint64_i, -1]:
+                    idx = np.searchsorted(P[thread_idx, uint64_i], p_norm)
+                    core._shift_insert_at_index(
+                        P[thread_idx, uint64_i], idx, p_norm, shift="right"
+                    )
+                    core._shift_insert_at_index(
+                        I[thread_idx, uint64_i], idx, uint64_j, shift="right"
+                    )
+
+                if ignore_trivial:  # self-joins only
+                    if p_norm < P[thread_idx, uint64_j, -1]:
+                        idx = np.searchsorted(P[thread_idx, uint64_j], p_norm)
+                        core._shift_insert_at_index(
+                            P[thread_idx, uint64_j], idx, p_norm, shift="right"
+                        )
+                        core._shift_insert_at_index(
+                            I[thread_idx, uint64_j], idx, uint64_i, shift="right"
+                        )
 
                     if uint64_i < uint64_j:
                         # left matrix profile and left matrix profile index
-                        if p_norm < P[thread_idx, uint64_j, 1]:
-                            P[thread_idx, uint64_j, 1] = p_norm
-                            I[thread_idx, uint64_j, 1] = uint64_i
+                        if p_norm < PL[thread_idx, uint64_j]:
+                            PL[thread_idx, uint64_j] = p_norm
+                            IL[thread_idx, uint64_j] = uint64_i
 
                         # right matrix profile and right matrix profile index
-                        if p_norm < P[thread_idx, uint64_i, 2]:
-                            P[thread_idx, uint64_i, 2] = p_norm
-                            I[thread_idx, uint64_i, 2] = uint64_j
+                        if p_norm < PR[thread_idx, uint64_i]:
+                            PR[thread_idx, uint64_i] = p_norm
+                            IR[thread_idx, uint64_i] = uint64_j
 
     return
 
@@ -158,7 +196,15 @@ def _compute_diagonal(
     fastmath=True,
 )
 def _aamp(
-    T_A, T_B, m, T_A_subseq_isfinite, T_B_subseq_isfinite, p, diags, ignore_trivial
+    T_A,
+    T_B,
+    m,
+    T_A_subseq_isfinite,
+    T_B_subseq_isfinite,
+    p,
+    diags,
+    ignore_trivial,
+    k,
 ):
     """
     A Numba JIT-compiled version of AAMP for parallel computation of the matrix
@@ -194,13 +240,30 @@ def _aamp(
         Set to `True` if this is a self-join. Otherwise, for AB-join, set this to
         `False`. Default is `True`.
 
+    k : int
+        The number of top `k` smallest distances used to construct the matrix profile.
+        Note that this will increase the total computational time and memory usage
+        when k > 1.
+
     Returns
     -------
-    P : numpy.ndarray
-        Matrix profile
+    profile : numpy.ndarray
+        The (top-k) matrix profile
 
-    I : numpy.ndarray
-        Matrix profile indices
+    indices : numpy.ndarray
+        The (top-k) matrix profile indices
+
+    left profile : numpy.ndarray
+        The (top-1) left matrix profile
+
+    left indices : numpy.ndarray
+        The (top-1) left matrix profile indices
+
+    right profile : numpy.ndarray
+        The (top-1) right matrix profile
+
+    right indices : numpy.ndarray
+        The (top-1) right matrix profile indices
 
     Notes
     -----
@@ -213,8 +276,15 @@ def _aamp(
     n_B = T_B.shape[0]
     l = n_A - m + 1
     n_threads = numba.config.NUMBA_NUM_THREADS
-    P = np.full((n_threads, l, 3), np.inf, dtype=np.float64)
-    I = np.full((n_threads, l, 3), -1, dtype=np.int64)
+
+    P = np.full((n_threads, l, k), np.inf, dtype=np.float64)
+    I = np.full((n_threads, l, k), -1, dtype=np.int64)
+
+    PL = np.full((n_threads, l), np.inf, dtype=np.float64)
+    IL = np.full((n_threads, l), -1, dtype=np.int64)
+
+    PR = np.full((n_threads, l), np.inf, dtype=np.float64)
+    IR = np.full((n_threads, l), -1, dtype=np.int64)
 
     ndist_counts = core._count_diagonal_ndist(diags, m, n_A, n_B)
     diags_ranges = core._get_array_ranges(ndist_counts, n_threads, False)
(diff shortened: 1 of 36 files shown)


NimaSarajpoor:
@seanlaw 
I realized that `aamped` was almost already changed when I was changing `aamp`. So, I tried to finalize both of them. (Nothing is new here! The changes are basically based on what we did in the Top-K normalized version). 

After you review these changes and show the greenlight, I will push the next set of commits for another module.

seanlaw (approved):
@NimaSarajpoor I quickly glanced over it but didn't see anything that stood out. I trust that there isn't much that will be tricky here but please draw my attention to specific parts if you want me to really turn on my "eagle eyes". Otherwise, I'm going to rely on the unit tests to guide me (of course, your changes to the `naive.py` could affect things).

NimaSarajpoor:
@seanlaw 
I've added top-k feature to `scraamp`. Again, all stuff is the same as the normalized matrix profile. 

---

**There is one thing that I would like to share with you...**

In some of the new test functions (related to top-k) in `test_scraamp`, we have three for-loops that are nested: 
* one for-loop changes param `k` (NEW)
* one for-loop changes param `p`, p-norm 
* one for-loop changes param `s`

(e.g. see `tests/test_scraamp::test_prescraamp_self_join_KNN`)

Note that in the normalized version (i.e. `test_scrump`), we do not have param `p`, and thus the nested loop consists of two for-loop there. 

I was wondering if this is okay with you? As an alternative approach, I can break this down and make it two for-loops, as follows:
```
k = 2
# two for-loops to change `p` and `s`

k = 3
# two for-loops to change `p` and `s`
```

seanlaw (approved):
@NimaSarajpoor While the triple for-loop isn't my favorite, it seems okay. Please proceed

NimaSarajpoor:
@seanlaw 
FYI: This PC does not have GPU. I need to push to check if it passes the tests. Sorry for the inconvenience.

NimaSarajpoor:
@seanlaw 
Something strange is going on here:
https://github.com/TDAmeritrade/stumpy/actions/runs/3453517947/jobs/5764176028

It exceeds 2hr!

NimaSarajpoor:
I think the existing `naive.aampi` (and its performant version) are incorrect in handling nan value.

```
# Example
m = 3
seed = 10844
n = 30
T = np.random.rand(n)
T[-1] = np.nan

ref_mp = naive.aampi_egress(T, m)
t = np.random.rand()
ref_mp.update(t) # egress

>>> ref_mp.left_I_
array([-1,  0,  1,  0,  3,  4,  0,  1,  1,  2,  3,  4, 11, 12, 13,  8,  9,
       15, 17,  3, 13,  0,  1, 19, 15, 14, -1,  1], dtype=int64)
```
However, I think the last value in `ref_mp.left_I_` should be `-1` because the value-before-last is `np.nan`. 

---

It seems the problem lies in the fact that 
```
D = core.mass_absolute(self._T[-self._m :], self._T)   # btw, this line is redundant in the script
D = cdist(
            core.rolling_window(self._T[-self._m :], self._m),
            core.rolling_window(self._T, self._m),
            metric="minkowski",
            p=self._p,
        )[0]
```

gives `D` with `nan` values. Therefore, later in the code, there is the following if-block, which will not executed...
```
I_last = np.argmin(D)
if np.isinf(D[I_last]):
    self.I_[-1] = -1
    self.P_[-1] = np.inf
```

Solutions:
* we can do `if not np.isfinite(D[I_last])`
* we can just preprocess `T` right from the beginning and get rid of `np.inf` / `np.nan`

And, we should fix the performant version `stumpy/aampi.py` as well.

NimaSarajpoor:
I think the existing `naive.aampi` (and its performant version) are incorrect in handling nan value.

```
# Example
m = 3
seed = 10844
n = 30
T = np.random.rand(n)
T[-1] = np.nan

ref_mp = naive.aampi_egress(T, m)
t = np.random.rand()
ref_mp.update(t) # egress

>>> ref_mp.left_I_
array([-1,  0,  1,  0,  3,  4,  0,  1,  1,  2,  3,  4, 11, 12, 13,  8,  9,
       15, 17,  3, 13,  0,  1, 19, 15, 14, -1,  1], dtype=int64)
```
However, I think the last value in `ref_mp.left_I_` should be `-1` because the value-before-last is `np.nan`. 

---

It seems the problem lies in the fact that 
```
D = core.mass_absolute(self._T[-self._m :], self._T)   # btw, this line is redundant in the script
D = cdist(
            core.rolling_window(self._T[-self._m :], self._m),
            core.rolling_window(self._T, self._m),
            metric="minkowski",
            p=self._p,
        )[0]
```

gives `D` with `nan` values. Therefore, later in the code, there is the following if-block, which will not executed...
```
I_last = np.argmin(D)
if np.isinf(D[I_last]):
    self.I_[-1] = -1
    self.P_[-1] = np.inf
```

Solutions:
* we can do `if not np.isfinite(D[I_last])`
* we can just preprocess `T` right from the beginning and get rid of `np.inf` / `np.nan`

And, we should fix the performant version `stumpy/aampi.py` as well.

seanlaw:
> I am pushing an empty commit to see if the same issue occurs again.

It looks okay (1 h 30 min). I am also re-running the one that took a long time. Not sure what's happening but it appears to be happening in the coverage testing. You can see some things taking over 10 minutes

NimaSarajpoor:
> happening in the coverage testing. You can see some things taking over 10 minutes
> 
> **Update** After re-running, it completed in 1 h 26 m

Yeah....I noticed that. I will think about it to see if I can something.

seanlaw:
@NimaSarajpoor Please resolve the conflicts above and let me know when it is ready for me to review. Thanks!

stumpy/aampi.py
```diff
@@ -202,40 +203,48 @@ def _update_egress(self, t):
         self._p_norm_new[0] = (
             np.linalg.norm(self._T[: self._m] - S[: self._m], ord=self._p) ** self._p
         )
-        self._p_norm_new[:] = np.where(
-            self._p_norm_new < config.STUMPY_P_NORM_THRESHOLD, 0, self._p_norm_new
-        )
+
+        mask = self._p_norm_new < config.STUMPY_P_NORM_THRESHOLD
+        self._p_norm_new[mask] = 0
```
NimaSarajpoor, line 208: is "using mask" better than `np.where`?

stumpy/gpu_aamp.py
```diff
@@ -9,14 +9,15 @@
...
+from . import core, config, gpu_stump
 
 logger = logging.getLogger(__name__)
 
 
 @cuda.jit(
     "(i8, f8[:], f8[:], i8, f8, f8[:], f8[:], f8[:], b1[:], b1[:],"
-    "i8, b1, i8, f8[:, :], i8[:, :], b1)"
+    "i8, b1, i8, f8[:, :], f8[:], f8[:], i8[:, :], i8[:], i8[:], b1,"
+    "i8[:], i8, i2)"
```
NimaSarajpoor, line 20: here, we used `i2` (not `i8`) as signature of `k`.

tests/naive.py
```diff
@@ -267,71 +267,107 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False, k=1):
...
-    distance_matrix = cdist(rolling_T_A, rolling_T_B, metric="minkowski", p=p)
+    P = np.full((l, k + 2), np.inf, dtype=np.float64)
+    I = np.full((l, k + 2), -1, dtype=np.int64)  # two more columns are to store
+    # ... left and right top-1 matrix profile indices
 
-    if ignore_trivial:
-        diags = np.arange(exclusion_zone + 1, n_A - m + 1)
-    else:
-        diags = np.arange(-(n_A - m + 1) + 1, n_B - m + 1)
+    if row_wise:
```
NimaSarajpoor, line 305: To be consistent with `naive.stump`, I added param `row_wise` to `naive.aamp`. Then, I used `row_wise==True` for testing `gpu_aamp`

tests/test_gpu_aamp.py
```diff
@@ -45,7 +45,7 @@ def test_gpu_aamp_self_join(T_A, T_B):
     m = 3
     zone = int(np.ceil(m / 4))
     for p in [1.0, 2.0, 3.0]:
-        ref_mp = naive.aamp(T_B, m, exclusion_zone=zone, p=p)
+        ref_mp = naive.aamp(T_B, m, exclusion_zone=zone, p=p, row_wise=True)
```
NimaSarajpoor, line 48: we now traverse row by row. (we are now consistent with `test_gpu_stump_self_join`)

NimaSarajpoor (commented):
@seanlaw 
I caught a few things. I was wondering if you could check them out. I will then fix them. Then, I think it will be ready for your review.

stumpy/aamp.py
```diff
@@ -129,36 +146,59 @@ def _compute_diagonal(
...
+                        if p_norm < PR[thread_idx, uint64_i]:
+                            PR[thread_idx, uint64_i] = p_norm
+                            IR[thread_idx, uint64_i] = uint64_j
 
     return
 
 
 @njit(
-    # "(f8[:], f8[:], i8, b1[:], b1[:], i8[:], b1)",
+    # "(f8[:], f8[:], i8, b1[:], b1[:], i8[:], b1, i8)",
```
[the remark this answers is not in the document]
seanlaw: I don't think that is wise to use `i2` because in order to do math (i.e., `j + k` where `j` is `i8`) with it will likely need some conversion to similar types. Let's keep it as `i8` please. It also doesn't make sense to save insignificant space here.

stumpy/aampi.py
```diff
@@ -101,12 +101,13 @@ def __init__(self, T, m, egress=True, p=2.0, k=1):
...
+        self._k = k
 
-        mp = aamp(self._T, self._m, p=self._p)
-        self._P = mp[:, 0].astype(np.float64)
-        self._I = mp[:, 1].astype(np.int64)
-        self._left_I = mp[:, 2].astype(np.int64)
-        self._left_P = np.empty(self._P.shape, dtype=np.float64)
+        mp = aamp(self._T, self._m, p=self._p, k=self._k)
+        self._P = mp[:, :k].astype(np.float64)
+        self._I = mp[:, k : 2 * k].astype(np.int64)
```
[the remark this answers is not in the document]
seanlaw: My preference for a class is to store the parameter ASAP (i.e., `self._k = k` and then use `self._k` everywhere without delay and treat the parameter `k` as being "dead". This way, we are consistent. Thus, I would propose to also change it to `self._k` in `stumpi` as well

stumpy/aampi.py
seanlaw: I find it more straightforward to reason with in my head as `mask` is simply a boolean array. So, it is more appropriate IMHO. `np.where` would be much more complicated I think

stumpy/gpu_aamp.py
seanlaw: `i8` please (see comment above)

tests/test_gpu_aamp.py
seanlaw: Perfect!

stumpy/aampi.py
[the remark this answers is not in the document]
NimaSarajpoor: > My preference for a class is to store the parameter ASAP (i.e., self._k = k and then use self._k everywhere without delay and treat the parameter k as being "dead". This way, we are consistent. T

Thank you for the explanation. I will take care of it.

stumpy/aamp.py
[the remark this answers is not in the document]
NimaSarajpoor: I see your point! I will revise it.

seanlaw (commented):
So far, so good! Keep up the great work!

tests/test_scraamp.py
```diff
@@ -741,3 +715,211 @@ def test_prescraamp_A_B_join_larger_window(T_A, T_B):
...
+@pytest.mark.parametrize("T_A, T_B", test_data)
+@pytest.mark.parametrize("percentages", percentages)
+def test_scraamp_plus_plus_self_join_KNN(T_A, T_B, percentages):
+    m = 3
+    zone = int(np.ceil(m / 4))
+    for k in range(2, 4):
+        for p in [1.0, 2.0, 3.0]:
+            for s in range(1, zone + 1):
+                for percentage in percentages:
+                    seed = np.random.randint(100000)
```
NimaSarajpoor, line 849: We have nested 4 for-loops here! Note that when `m==3`, `zone=1` and thus we can remove the `for s in range(1, zone+1)`.  The downside is that we are making the code less flexible (I mean.....if we change `m` later, it does not affect `s`).

seanlaw: Obviously, I don't love this. However, if this unit test doesn't take long to complete then it might not matter at all. So let's be pragmatic as nested for-loops only really matter when there are many, many thousands+ things to iterate over in each loop level. In this case, there are probably 10's - 100's of variations, which might not be that bad from a time perspective. Let's start with compute time and then decide if we need to do anything.

I would feel very differently if this were in the actual codebase (and not the unit tests)!

NimaSarajpoor: > I would feel very differently if this were in the actual codebase (and not the unit tests)!

Yeah...I know :)

---

So, I measured the computing time for different values of `k` (note that we are iterating over  `k in range(2, 4)`). And, I got this:

```
k=1 --> 0.95  
k=2 --> 1.30
k=3 --> 1.61
```
Recall that `k=1` is default, and thus this case is already covered by one of the test function in branch `main`. The new test function covers `k=2` and `k=3`. So, I think it is okay to keep the `for-loop` over `k` as its computing time is small.

seanlaw: I agree! Let's keep it

seanlaw:
@NimaSarajpoor There have been some new commits that you may want to merge with this branch that (somehow?) makes the tests run faster

NimaSarajpoor:
@seanlaw @alvii147 
The test suite has become fast :) Cool !!!

NimaSarajpoor:
I am trying to measure the computing time. I first ran:
```
./test.sh custom 1 tests/test_gpu_aamp.py
```

But, I got this error:
```
E   numba.core.errors.TypingError: Failed in cuda mode pipeline (step: nopython frontend)
E   Internal error at resolving type of attribute "_gpu_searchsorted_right" of "$366load_global.1".
E   partially initialized module 'stumpy.gpu_stump' has no attribute '_gpu_searchsorted_right' (most likely due to a circular import)
E   During: typing of get attribute at E:\+Machine_Learning_Journey\contributions\stumpy\stumpy\gpu_aamp.py (181)
E   Enable logging at debug level for details.
E
E   File "stumpy\gpu_aamp.py", line 181:
E   def _compute_and_update_PI_kernel(
E       <source elided>
E           if p_norm < profile[j, -1]:
E               idx = gpu_stump._gpu_searchsorted_right(profile[j], p_norm, bfs, nlevel)
E               ^
=========================== short test summary info ===========================
ERROR tests/test_gpu_aamp.py - numba.core.errors.TypingError: Failed in cuda ...
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
============================== 1 error in 5.67s ===============================
Error: pytest encountered exit code 2
```

**But, why do we get no error when I push the commits here?**

---

This is how I call the device function `_gpu_searchsorted_right` in `gpu_aamp`

```
# in stumpy/gpu_aamp.py

from stumpy import gpu_stump

# some code

gpu_stump._gpu_searchsorted_right(...)
```

seanlaw:
> But, why do we get no error when I push the commits here?

@NimaSarajpoor I am guessing it's because you actually have a GPU and Github Actions doesn't so it's likely skipping tests. We need to fix the code for your GPU-enabled environment!

NimaSarajpoor:
Unfortunately, that didn't work...

```
=================================== ERRORS ====================================
___________________ ERROR collecting tests/test_gpu_aamp.py ___________________
ImportError while importing test module 'E:\+Machine_Learning_Journey\contributions\stumpy\tests\test_gpu_aamp.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
C:\Users\nimas\anaconda3\envs\stumpy\lib\importlib\__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests\test_gpu_aamp.py:4: in <module>
    from stumpy import gpu_aamp
stumpy\__init__.py:33: in <module>
    from .gpu_stump import gpu_stump  # noqa: F401
stumpy\gpu_stump.py:13: in <module>
    from .gpu_aamp import gpu_aamp
stumpy\gpu_aamp.py:13: in <module>
    from stumpy.gpu_stump import _gpu_searchsorted_right
E   ImportError: cannot import name '_gpu_searchsorted_right' from partially initialized module 'stumpy.gpu_stump' (most likely due to a circular import) (E:\+Machine_Learning_Journey\contributions\stumpy\stumpy\gpu_stump.py)
=========================== short test summary info ===========================
ERROR tests/test_gpu_aamp.py
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
============================== 1 error in 4.05s ===============================
```

seanlaw:
> UPDATE:
I noticed that we have the following line in gpu_stump:
from .gpu_aamp import gpu_aamp
I am trying to see if this causes confusion for python!

Sorry. This should be the right answer :)

seanlaw:
> Sorry does this work?

If not, please move `_gpu_searchsorted_right` into `core.py` instead. It's likely that you are seeing a circular dependency.

NimaSarajpoor:
@seanlaw 
Yeah. `from .gpu_stump import ...` did not work. So, I will move  `_gpu_searchsorted_left` and `_gpu_searchsorted_right` to `core.py`. 

---

Should I wrap a function around these device functions and write a test functions for them in `test_core.py`

seanlaw:
Up to you. What do you think?

NimaSarajpoor:
@seanlaw 
I prefer to write a test to be consistent with other functions in core.py. It can also help us make sure that that these two functions work properly (so, we do not need to be worried about them later if we face a bug)

NimaSarajpoor:
> I prefer to write a test to be consistent with other functions in core.py. It can also help us make sure that that these two functions work properly (so, we do not need to be worried about them later if we face a bug)

Apparently , we already had the test for `_gpu_searchsorted`, located in `test_gpu_stump` 😃. I just updated the test functions.  

I will now start comparing the computing time.

NimaSarajpoor:
@seanlaw 
Before I forget, it might be worthwhile to improve top-k related docstrings by mentioning that: 

"Each of the top-k neighbors of a sequence `s` is not the trivial match of `s` when `ignore_trivial==True` . However, it should be noted that the top-k neighbors themselves might be trivial match of each other regardless of the value of `ignore_trivial`".

Please ignore this suggestion if it makes the docstrings more confusing :)

seanlaw:
> Before I forget, it might be worthwhile to improve top-k related docstrings by mentioning that:

I get your point. Maybe:

```
Note that for self-joins (i.e., `ignore_trivial==True`) the top-k neighbors for each subsequence is guaranteed to be located outside of its corresponding exclusion zone. However, it should be noted that the top-k neighbors themselves may be located directly adjacent to/near each other so as long as they are all located outside of the reference subsequence's exclusion zone.
```

seanlaw:
> Apparently , we already had the test for _gpu_searchsorted, located in test_gpu_stump 😃. I just updated the test functions.

Yeah, I was thinking about that the other day but wasn't in front of my computer to check. It makes sense though that we'll simply need to move the tests over. This is good. It means that we've been proactive and "Future Nima thanks past Nima for thinking ahead and making his life easier" 😄

NimaSarajpoor:
> I get your point. How about this:
> 
> > Note that for self-joins (i.e., `ignore_trivial==True`) the top-k neighbors for each subsequence is guaranteed to be located outside of the reference subsequence's exclusion zone. However, the top-k neighbors themselves may be located directly adjacent to/near each other so as long as they are all located outside of the reference subsequence's exclusion zone.

Yes... your suggestion is cleaner as it specifically talks about the case `ignore_trivial==True`

>  This is good. It means that we've been proactive and "Future Nima thanks past Nima for thinking ahead and making his life easier" 😄 

😄 



> You are developing good habits and you'll have confidence in your own development approach and you'll find that it is more and more rare for there to be a huge mistake/bug if you've protected yourself sufficiently. More importantly, it means that it'll be harder for other people to break things.

True! Since we take care of things in the first place, we will face less issue. Even if we face one, it will be easier to resolve it :) 👍

NimaSarajpoor:
sneak peek on computing time :)

| aamp                | n=10\_000, m=50 | n=20\_000, m=50 | n=50\_000, m=50 |
| ------------------- | --------------- | --------------- | --------------- |
| MAIN\_BRANCH        | 0.8             | 3.1             | 20.2            |
| TOP-K BRANCH (k==1) | 0.84            | 3.2             | 19.9            |

NimaSarajpoor:
@seanlaw

In the following tables, length of subsequence, `m`, has  been set  to 50.

# aamp: ok
| aamp, p=2(integer) | n = 1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| ------------------ | -------- | --------- | --------- | ---------- |
| main branch        | 0.002    | 0.13      | 3.03      | 13.05      |
| top-k, k==1        | 0.013    | 0.21      | 3.37      | 13.5       |


# aamped: ok
| aamped, p=2(integer) | n = 1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| -------------------- | -------- | --------- | --------- | ---------- |
| main branch          | 0.032    | 0.16      | 3.11      | 13.7       |
| top-k, k==1          | 0.05     | 0.27      | 3.73      | 14.6       |


# (!!!) gpu_aamp: NOT ok (see n=50k, 100k)
| gpu\_aamp, p=2(integer) | n = 1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| ----------------------- | -------- | --------- | --------- | ---------- |
| main branch             | 0.27     | 2.02      | 9.28      | 35.02      |
| top-k, k==1             | 0.34     | 3.15      | 67.33     | 267.2      |

# prescraamp: ok
| prescraamp, p=2(integer) | n = 1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| ------------------------ | -------- | --------- | --------- | ---------- |
| main branch              | 0.002    | 0.07      | 1.99      | 8.63       |
| top-k, k==1              | 0.006    | 0.12      | 2.22      | 9.39       |

# scraamp_update: Not ok ?!
| scraamp with two updates, p=2(integer) | n = 1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| -------------------------------------- | -------- | --------- | --------- | ---------- |
| main branch                            | 0.005    | 0.05      | 0.34      | 0.88       |
| top-k, k==1                            | 0.014    | 0.15      | 0.62      | 1.36       |


# aampi: ok
| aampi with one updates, p=2(integer) | n = 1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| ------------------------------------ | -------- | --------- | --------- | ---------- |
| main branch                          | 0.011    | 0.21      | 3.43      | 14.12      |
| top-k, k==1                          | 0.022    | 0.3       | 3.87      | 14         |

----

I will investigate `gpu_aamp`.

NimaSarajpoor:
@seanlaw 
I fixed `gpu_aamp`. I ran all the modules again and measured their computing time. It seems everything is okay. 

---

# aamp
| aamp; p=2 (int) | n=1000  | n=10\_000 | n=50\_000 | n=100\_000 |
| ---------- | ----- | ------- | ------- | -------- |
| main       | 0.002 | 0.12    | 3.18    | 14.14    |
| top-k  (k == 1)    | 0.008 | 0.17    | 3.18    | 13.43    |


```
# performance  code
seed = 0
np.random.seed(seed)
m = 50
p = 2
k = 1
n_iter = 3

n = 1000 # options: 1000, 10k, 50k, 100k
T = np.random.rand(n)
stumpy.aamp(T, m, p=p, k=k)
lst = []
for _ in range(n_iter):
    tic = time.time()
    stumpy.aamp(T, m, p=p, k=k)
    toc = time.time()
    lst.append(toc - tic)

np.mean(lst)
```

---

# aamped

| aamped; p=2 (int) | n=1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| ------------ | ---- | ------- | ------- | -------- |
| main         | 0.03 | 0.16    | 3.21    | 14.47    |
| top-k  (k==1)     | 0.04 | 0.21    | 3.96    | 14.96    |

```
# performance  code
import numba
from dask.distributed import Client

seed = 0
np.random.seed(seed)
m = 50
p = 2
k = 1
n_iter = 3

n = 1000 # options: 1000, 10k, 50k, 100k
T = np.random.rand(n)

with Client(n_workers=1, threads_per_worker=numba.config.NUMBA_NUM_THREADS) as client:
    stumpy.aamped(client, T, m, p=p, k=k)
    
    lst = []
    for _ in range(n_iter):
        tic = time.time()
        stumpy.aamped(client, T, m, p=p, k=k)
        toc = time.time()
        
        lst.append(toc - tic)

np.mean(lst)
```

---

# gpu_aamp

| gpu\_aamp; p=2.0 (float) | n=1000 | n=10\_000 | n=50\_000 | n=100\_000 |
| ---------------- | ---- | ------- | ------- | -------- |
| main             | 0.28 | 1.91    | 9.84    | 34.02    |
| top-k   (k==1)    | 0.36 | 2.59    | 12.53   | 34.03    |

```
# performance  code
seed = 0
np.random.seed(seed)
m = 50
p = 2.0
k = 1
n_iter = 3

n = 1000 # options: 1000, 10k, 50k, 100k
T = np.random.rand(n)
stumpy.gpu_aamp(T, m, p=p, k=k)
lst = []
for _ in range(n_iter):
    tic = time.time()
    stumpy.gpu_aamp(T, m, p=p, k=k)
    toc = time.time()
    lst.append(toc - tic)

np.mean(lst)
```

---

# aampi(egress=True) with 5 updates

| aampi; p=2 (int) | n=1000  | n=10\_000 | n=50\_000 | n=100\_000 |
| ----------------------------------- | ----- | ------- | ------- | -------- |
| main                                | 0.12  | 0.21    | 4       | 14.82    |
| top-k  (k==1)                        | 0.019 | 0.27    | 3.65    | 13.54    |


```
# performance  code
seed = 0
np.random.seed(seed)
m = 50
p = 2
k = 1
n_iter = 3

n = 1000 # options: 1000, 10k, 50k, 100k
T = np.random.rand(n)

seed = 0
np.random.seed(seed)
t_vals = np.random.rand(5) # for 5 updates 

stream = stumpy.aampi(T, m, egress=True, p=p, k=k)
stream.update(t_vals[0])

lst = []
for _ in range(n_iter):
    tic = time.time()
    stream = stumpy.aampi(T, m, egress=True, p=p, k=k)
    for t in t_vals:
        stream.update(t)
    toc = time.time()

    lst.append(toc - tic)

np.mean(lst)
```

---

# aampi(egress=False) with 5 updates

| aampi; p=2 (int) | n=1000  | n=10\_000 | n=50\_000 | n=100\_000 |
| ------------------------------------ | ----- | ------- | ------- | -------- |
| main                                 | 0.012 | 0.21    | 3.56    | 15.14    |
| top-k   (k==1)                   | 0.024 | 0.33    | 3.9     | 14.89    |

```
# performance  code
seed = 0
np.random.seed(seed)
m = 50
p = 2
k = 1
n_iter = 3

n = 1000 # options: 1000, 10k, 50k, 100k
T = np.random.rand(n)

seed = 0
np.random.seed(seed)
t_vals = np.random.rand(5) # for 5 updates 

stream = stumpy.aampi(T, m, egress=False, p=p, k=k)
stream.update(t_vals[0])

lst = []
for _ in range(n_iter):
    tic = time.time()
    stream = stumpy.aampi(T, m, egress=False, p=p, k=k)
    for t in t_vals:
        stream.update(t)
    toc = time.time()

    lst.append(toc - tic)

np.mean(lst)
```

---

# prescraamp

| prescraamp; p=2 (int) | n=1000  | n=10\_000 | n=50\_000 | n=100\_000 |
| ---------------- | ----- | ------- | ------- | -------- |
| main             | 0.002 | 0.07    | 2.06    | 9.31     |
| top-k  (k==1)       | 0.006 | 0.12    | 2.21    | 9.19     |

```
# performance  code
seed = 0
np.random.seed(seed)
m = 50
p = 2
k = 1
n_iter = 3

n = 1000 # options: 1000, 10k, 50k, 100k
T = np.random.rand(n)

seed = 1
np.random.seed(seed)
prescraamp(T, m, p=p, k=k)

lst = []
for _ in range(n_iter):
    np.random.seed(seed)
    tic = time.time()
    prescraamp(T, m, p=p, k=k)
    toc = time.time()
    
    lst.append(toc - tic)

np.mean(lst)
```

---

# scraamp (prescraamp==False) with 5 updates

| scraamp (p=2)                                     | 1000 | 10\_000 | 50\_000 | 100\_000 |
| ------------------------------------------------- | ---- | ------- | ------- | -------- |
| main                                              | 0.01 | 0.11    | 0.73    | 1.96     |
| top-k                                             | 0.03 | 0.3     | 1.6     | 3.54     |
| top-k (with prange in \_aamp)<br>\*see note below | 0.01 | 0.07    | 0.47    | 1.27     |

```
# performance  code
seed = 0
np.random.seed(seed)
m = 50
p = 2
k = 1
n_iter = 3

n = 1000 # options: 1000, 10k, 50k, 100k
T = np.random.rand(n)

seed = 1
np.random.seed(seed)
approx = stumpy.scraamp(T, m, p=p, k=k)
approx.update()

lst = []
for _ in range(n_iter):
    np.random.seed(seed)
    tic = time.time()
    approx = stumpy.scraamp(T, m, p=p, k=k)
    for _ in range(5): # 5 updates
        approx.update()
    toc = time.time()
    
    lst.append(toc - tic)

np.mean(lst)
```

**NOTE:**
The method `.update()` in `scraamp` calls `_aamp`. To detect the source of overhead in top-k version. I replace `core._merge_topk_PI` with `prange` in function `_aamp`. As you can see, the computing time has decreased. We can work on it in another PR. 

---

I think everything is good :)

seanlaw:
> I fixed gpu_aamp

Can you please describe what the issue was? I am curious if it was simply a bug or did we have to do something drastically different from the normalized version. My main concern is maintaining consistency across normalized and non-normalized methods.

seanlaw:
@NimaSarajpoor So far, everything looks good to me and nothing obvious stood out. Is there anything else left to be done?

stumpy/gpu_aamp.py
```diff
@@ -263,7 +304,7 @@ def _gpu_aamp(
     See Table II, Figure 5, and Figure 6
     """
     threads_per_block = config.STUMPY_THREADS_PER_BLOCK
-    blocks_per_grid = math.ceil(k / threads_per_block)
+    blocks_per_grid = math.ceil(w / threads_per_block)
```
NimaSarajpoor, line 307: This was the issue! It is just a silly mistake. I forgot to update this line. Since we now use `k` for `top-k`, I had to change the name of variable but I forgot to do so for this line.

stumpy/gpu_stump.py
```diff
@@ -854,12 +764,12 @@ def gpu_stump(
         core._merge_topk_PI(profile[0], profile[i], indices[0], indices[i])
 
         # Update (top-1) left matrix profile and matrix profile indices
-        mask = profile_L[0] < profile_L[i]
+        mask = profile_L[0] > profile_L[i]
```
NimaSarajpoor, line 767: This is another issue that I caught it myself. I think the test suite runs on 1 GPU (?) and that is why it couldn't  catch this mistake. I just noticed it and fixed it.

stumpy/gpu_aamp.py
seanlaw: Excellent. I'm glad that it makes logical sense and things are still consistent between normalized and non-norm

NimaSarajpoor:
> I fixed gpu_aamp

Can you please describe what the issue was? I am curious if it was simply a bug or did we have to do something drastically different from the normalized version. My main concern is maintaining consistency across normalized and non-normalized methods.

I reviewed the code to just show you the parts that are related to `gpu` and fixed by my recent commits.

> @NimaSarajpoor So far, everything looks good to me and nothing obvious stood out. Is there anything else left to be done?

I think everything is good. The only thing you may want to do is to just scan the docstrings to make sure everything is correct.

seanlaw:
> The only thing you may want to do is to just scan the docstrings to make sure everything is correct.

I can do that!

seanlaw (changes requested):
Just a minor suggestion

stumpy/aamp.py
```diff
@@ -194,13 +234,30 @@ def _aamp(
...
+    k : int
+        The number of top `k` smallest distances used to construct the matrix profile.
+        Note that this will increase the total computational time and memory usage
+        when k > 1.
+
     Returns
     -------
-    P : numpy.ndarray
-        Matrix profile
+    profile : numpy.ndarray
```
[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
Yeah...I missed it! Btw, I think we should use the names `out1, out2, ...`. This is because the final output is calculated inside the return statement.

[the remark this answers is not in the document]
seanlaw: I'm good with that

seanlaw:
@NimaSarajpoor Thanks again for this contribution!

PR #726, opened 2022-11-15
Fixed #721 aampi: updating left matrix profile (index)

### LINKED ISSUE #721: In `aampi`: wrong left matrix profile index when its corresponding value is np.inf
I think the existing `naive.aampi` (and its performant version) are incorrect in handling nan value.

```
# Example
m = 3
seed = 10844
n = 30
T = np.random.rand(n)
T[-1] = np.nan

ref_mp = naive.aampi_egress(T, m)
t = np.random.rand()
ref_mp.update(t) # egress

>>> ref_mp.left_I_[-1]
1
```
However, I think the last value in `ref_mp.left_I_` should be `-1` because the value-before-last in time series is `np.nan`. 
(and hence the last subsequence contains non-finite value)

---

It seems the problem lies in the fact that 
```
# in tests/naive.py
# in aampi

D = core.mass_absolute(self._T[-self._m :], self._T)   # btw, this line is redundant in the script
D = cdist(
            core.rolling_window(self._T[-self._m :], self._m),
            core.rolling_window(self._T, self._m),
            metric="minkowski",
            p=self._p,
        )[0]
```

gives `D` with `nan` values. Therefore, later in the code, there is the following if-block, which will not be effective in handing `nan`:

```
I_last = np.argmin(D)
if np.isinf(D[I_last]):
    self.I_[-1] = -1
    self.P_[-1] = np.inf
```

Solutions:
* we can do `if not np.isfinite(D[I_last])`
* Or, we can just preprocess `T` right from the beginning and get rid of `np.inf` / `np.nan`

And, we should fix the performant version `stumpy/aampi.py` as well.

---

I can handle it in PR #714 if that is okay.

NimaSarajpoor:
This PR addresses issue #721.

### DIFF
--- a/stumpy/aampi.py
+++ b/stumpy/aampi.py
@@ -225,8 +225,10 @@ class aampi:
             self._I[-1] = I_last + self._n_appended
             self._P[-1] = D[I_last]
 
-        self._left_I[-1] = I_last + self._n_appended
-        self._left_P[-1] = D[I_last]
+        # Regarding the last subsequence, the left profile (index) value is the
+        # same as the profile (index) value.
+        self._left_I[-1] = self._I[-1]
+        self._left_P[-1] = self._P[-1]
 
         self._p_norm[:] = self._p_norm_new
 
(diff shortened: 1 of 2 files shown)


seanlaw:
Thanks @NimaSarajpoor! Looks good to me. Are we ready to merge?

NimaSarajpoor:
Yes. I think it is ready :)

seanlaw:
Awesome! Thanks for the fix

PR #728, opened 2022-11-15
Fixed #727: missing coverage

### LINKED ISSUE #727: Missing Coverage in `naive.py`
In this [coverage test](https://github.com/TDAmeritrade/stumpy/actions/runs/3467770021/jobs/5792942355), I am seeing:

```
============================= 18 passed in 20.48s ==============================
Name             Stmts   Miss  Cover   Missing
tests/naive.py    1118      2    99%   1892-1893
TOTAL            11796      2    99%

79 files skipped due to complete coverage.
Cleaning Up
```

These lines corresponds to:

```
# in merge_topk_ρI

    if ρA.ndim == 1:
        for i in range(ρA.shape[0]):
            if ρB[i] > ρA[i] and IB[i] != IA[i]:  # This isn't always tested
                ρA[i] = ρB[i]
                IA[i] = IB[i]
        return
```

It looks like we need a unit tests that explicitly triggers this `if` condition.

NimaSarajpoor:
This PR fixes issue #727

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1134,20 +1134,12 @@ def test_merge_topk_PI_with_overlap():
         npt.assert_almost_equal(ref_I, comp_I)
 
 
-def test_merge_topk_PI_with_1D_input_with_overlap():
-    # including some overlaps randomly
-    n = 50
-    PA = np.random.rand(n)
-    PB = np.random.rand(n)
+def test_merge_topk_PI_with_1D_input():
+    PA = np.array([0.1, 0.3, 0.5, 0.7, 0.9])
+    PB = np.array([0.2, 0.3, 0.6, 0.8, 1.0])
 
-    IA = np.arange(n)
-    IB = IA + n
-
-    n_overlaps = np.random.randint(1, n + 1)
-    IDX_rows_with_overlaps = np.random.choice(np.arange(n), n_overlaps, replace=False)
-    imprecision = np.random.uniform(low=-1e-06, high=1e-06, size=n_overlaps)
-    PB[IDX_rows_with_overlaps] = PA[IDX_rows_with_overlaps] + imprecision
-    IB[IDX_rows_with_overlaps] = IA[IDX_rows_with_overlaps]
+    IA = np.array([0, 1, 2, 3, 4])
+    IB = np.array([10, 1, 12, 13, 14])
 
     ref_P = PA.copy()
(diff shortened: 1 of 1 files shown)


NimaSarajpoor:
Should I be worried about this?
```
Name                 Stmts   Miss  Cover   Missing
--------------------------------------------------
tests/test_core.py     866     10    99%   1175-1179, 1293-1297
--------------------------------------------------
TOTAL                12243     10    99%
```

The missing lines are just a part of test functions that is added to ensure the full coverage on `core.py`

seanlaw:
> Should I be worried about this?

Yes, you should be. Instead of a random input array, can you provide a hardcoded array for that unit test instead where you have full control of the inputs and know that it will precisely trigger the `if` block?

NimaSarajpoor:
> Yes, you should be. Instead of a random input array, can you provide a hardcoded array for that unit test instead where you have full control of the inputs and know that it will precisely trigger the `if` block? I'm looking for a super simple unit test without any `if` statements and likely around 10 lines of code

Got it! I will try it.

seanlaw:
Thanks @NimaSarajpoor. Sorry to be a nuisance but instead of replacing the contents of `test_merge_topk_PI_with_1D_input`, `test_merge_topk_ρI_with_overlap`, `test_merge_topk_ρI_with_1D_input_with_overlap`, and `test_merge_topk_ρI_with_1D_input_without_overlap`, can you simply add one or more additional unit tests that has hardcoded values? 

I like having the original randomly generated arrays for unit tests but for very specific cases where we need something to trigger, I think adding a separate unit test makes it unambiguous as to what we are testing for when our typical tests don't catch/trigger a condition.

NimaSarajpoor:
@seanlaw 
> I like having the original randomly generated arrays for unit tests but for very specific cases where we need something to trigger, I think adding a separate unit test makes it unambiguous as to what we are testing for when our typical tests don't catch/trigger a condition.

Okay...I think I understand your point. Please let me try one more time and push the commits, and then see if that is what you are looking for or not.

NimaSarajpoor:
@seanlaw 
I revised the test function. Now, I am keeping the original test function with randomly-generated values. And, I've added a new test function with hardcoded values. Please let me know if that works or not.

seanlaw:
Thanks @NimaSarajpoor! Much appreciated

Issue #466, opened 2021-09-14
Add DTW Motif Discovery - SWAMP


seanlaw:
The paper can be found [here](https://www.cs.ucr.edu/~eamonn/DTWMotifsSWAMP_Expanded.pdf)

The supplementary webpage is [here](https://sites.google.com/site/dtwmotifdiscovery/)

The code is 
[here](https://github.com/TDAmeritrade/stumpy/files/7161674/DTWMotifDiscovery.zip)

The data is [here](https://sites.google.com/site/dtwmotifdiscovery/#h.p_R5YBI2_GHL6_)

Ninimama:
@seanlaw 
I think the  package [tslearn](https://github.com/tslearn-team/tslearn) might help those who want to work on SWAMP. For instance, it provides the implementation of [LB_Keogh](https://tslearn.readthedocs.io/en/stable/gen_modules/metrics/tslearn.metrics.lb_keogh.html). 

That might give us some idea that can be useful for pythonic implementation of SWAMP.

ken-maeda:
I reproduced dtwMP by dtw calculation(from paper) and stump(k-top) MP.

SWAMP itself is from combination dtw and Lower bound for reducing calculation cost.
Could you share with your idea how SWAMP is implemented in stumpy.  

To DO SWAMP(original):  

- [x] : dtw MP  

- [ ] : LB MP

- [ ] : Early abandon

- [ ] : together as one

To DO(improvement):
- [ ] : dtw supports "itakura" band also.  
    (paper code itself is only for "sakoe-chiba", but paper mentioned both)
- [ ] : MP calculation optimaization.  
    (It don't have to be like MASS algorithm?)
- [ ] : z-normalized on off  
    (paper code only support z-normalized.)
- [ ] : pre/post process for inf, nan

Basically dtw calculation speed should be key factor, therefore I would consider improvement of the speed.

[Paper result]
![image](https://user-images.githubusercontent.com/42758592/204138488-dc3b2e78-78d4-4e5f-b9d8-3176ea9c42d9.png)

[Python implement]
![image](https://user-images.githubusercontent.com/42758592/204138541-6197ad3c-6d90-495a-bd94-f3148a6d2a8d.png)

seanlaw:
[...]

Issue #46, opened 2019-07-01
Add MPdist


seanlaw:
According to the [ICDM publication](https://docs.google.com/viewer?a=v&pid=sites&srcid=ZGVmYXVsdGRvbWFpbnxtcGRpc3RpbmZvfGd4OjNjMGZjOTBmYWEzNjM0ZWY), the calculation of MPdist is pretty straightforward. For two time series, `A` and `B`, of identical length and a window size, `m`, equal to, say, 50:

1. Compute the matrix profile AB for `stumpy.stump(A, m=50, T_B=B, ignore_trivial=False)`
2. Compute the matrix profile BA for `stumpy.stump(B, m=50, T_B=A, ignore_trivial=False)`
3. Concatenate both matrix profiles into PABBA
4. Choose some `k` that is 5 percent of `2 * n`

As mentioned on on page 3 of the above paper, "section, this data structure PABBA
contains all the information we need to compute the MPdist."

The pseudocode can be found [here](https://docs.google.com/viewer?a=v&pid=sites&srcid=ZGVmYXVsdGRvbWFpbnxtcGRpc3RpbmZvfGd4OjJmZGFkZWEwMTcxODYyMDM).

The supporting site can be found [here](https://sites.google.com/site/mpdistinfo/)

Deepakgthomas:
I believe after coding the MPDist measure, the development of the Snippet finding algorithm should be relatively simple. I was wondering if your team has any plans to develop it in Python.

seanlaw:
@Deepakgthomas I think you are correct and, to be clear, there is only one person currently supporting the continued developments so we certainly welcome PRs and we can work with you on this. We’d want to keep the MPdist PR separate from the Snippets PR and consider if it makes sense considering that they’ll be light wrappers around the matrix profile calculation. Currently, MPdist seems like a nice to have but most of it is possible with STUMPY and a little bit of code. What do you think?

With constrained resources, we have to decide whether it is worth supporting the functionality. Also, the key part is ensuring that all new features have 100% test coverage.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Deepakgthomas:
I agree with you that there is no immediate need for any additional functions for MPdist, due to the functionality that STUMPY provides. 
I would also love to help with the Snippet discovery algorithm, depending upon the time that I have :)

seanlaw:
[...]

Issue #20, opened 2019-05-25
Semantic segmentation on multi-dimensional Time Series 


ywy9876:
Hello, 
I was looking at the example of Multi-dimensional time series data with MSTUMP. I tried to run the code and found that the returned values (matrix_profile, matrix_profile_indices) are both 3d (1 for each Time series I guess). My doubt is, when the indices for a segment i point to different nearest  segments (say 847, 237, 847), how do we decide which is the nearest neighbour? In a document of Prof. Eamonn, it says that we simply add the distance profiles together and find the minimum. But with your implementation, I'm not sure how to do it.
Also I wonder if the [FLOSS](https://www.cs.ucr.edu/~eamonn/Segmentation_ICDM.pdf) or similar algorithm for semantic segmentation is already implemented.

Thank you in advance.

seanlaw:
@ywy9876 Thank you for your question. Our implementation of multi-dimensional STOMP (MSTUMP) produces identical results to the original author's open source reference implementation that can be found [here](https://github.com/mcyeh/mstamp).

To answer your question, each row of the returned matrix profile actually corresponds to using one time series up to D-time series. So, if you have three time series then the **first row** of the matrix profile chooses the **one** time series (out of three) that produces the smallest (minimum) distance value for that window. The second row of the matrix profile chooses two time series (out of three) that produces the smallest (minimum) average distance value for that window. Finally, the third row of the matrix profile gives the (average) distance value for when all three time series are used.

The matrix profile indices tell you where along the time series you'd find matching subsequence. However, it doesn't tell you **which** of the time series is chosen as you increase k, the number of dimensions to include. 

We have not implemented FLOSS but I suspect that you can do this fairly easily since the hard part is having an efficient way to compute the full matrix profile (which STUMPY does for you).

Let me know if this answers your question.

ywy9876:
@seanlaw ,
Thank you for your reply. 
Just to be clear with your sentence: 

> The matrix profile indices tell you where along the time series you'd find matching subsequence. 

If I have the following results:
```
matrix_profile_indices[0][4] 
Out[22]: 867

matrix_profile_indices[1][4]
Out[23]: 205

matrix_profile_indices[2][4]
Out[24]: 867
```
What I should interpret is that for the subsequence starting from position 4 of the original TS, the nearest neighbour is the subsequence starting from position 867 if we take into account all the dimensions (due to matrix_profile_indices[2][4]), am I wrong? 

[...]

PR #733, opened 2022-11-28
Swamp discussion

ken-maeda:
# SWAMP/ DTW MP discussion
This is a pull-request for information exchange for #466.
About SWAMP, DTW MP, LB MP and related algorithm to reduce calculation cost.

# Pull Request Checklist

- [ ] Create a new branch


seanlaw:
Thank you @ken-maeda. Please allow me some time to review

docs/DTW_MP_trial.ipynb
```diff
@@ -0,0 +1,810 @@
+{
```
seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #43.        return np.sqrt(cost[k_])
</pre><p>I recommend returning the squared cost rather than the actual cost as performing the sqrt here is very expensive. You can still compare squared costs against each other</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/733/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

seanlaw, line 1: <p>So, the slowness likely comes in these three lines:</p><p> </p><pre class="ql-syntax" spellcheck="false">a&nbsp;= (T_A[uint64_i:uint64_i+uint64_m] - μ_Q[uint64_i])*σ_Q_inverse[uint64_i]
b = (T_B[uint64_j:uint64_j+uint64_m] - M_T[uint64_j])*Σ_T_inverse[uint64_j]
dist = _njit_dtw(a, b, r)# ------------ update
</pre><p> </p><p>because you are not benefitting from traversing the diagonal and re-using the last covariance. Instead, you are computing the individual variances from scratch during each iteration and this is computationally very expensive.</p><p> </p><p>I think what we need to do is express the variance in a "rolling window" approach. See <code>core._welford_nanvar</code> for some possible inspiration </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/733/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #20.                    if bj &gt;= 0 and ρB[i, bj] &lt; ρA[i, aj]:  # ------------ update
</pre><p>Should this be <code>if bj &lt; k and ρB[i, bj] &lt; ρA[i, aj]]:</code>?</p><p> </p><p>I wonder if we can simply re-use <code>core._merge_topk_PI</code> ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/733/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #46.        r,
</pre><p>What is <code>r</code>?? Is <code>r &lt;= m</code>  always <code>True</code> or can <code>r</code> be greater than <code>m</code>?     </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/733/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

ken-maeda:
I skipped these preprocess/nan-inf process, because it doesn't have no infulence to the result in this example. Coding of these preprocess and calculation optimization depends on how SWAMP is coded.
At the first, I would ask how SWAMP should  be coded . 
As the paper, SWAMP is combination of several algorithm(DTW MP is a part of them). However DTW MP itself can be standalone function.

For the discussion, maybe should I code a whole SWAMP(including LB MP, some calculation time algo)? 

But I got good feedback like dtw can be imported from "tslearn".

docs/DTW_MP_trial.ipynb
ken-maeda: It should works.

ken-maeda: We need some preprocess for limitation of `r`.

seanlaw:
> But I got good feedback like dtw can be imported from "tslearn".

So, based on your notebook, it looks like it would not be a good idea to call DTW from `tslearn` (which accepts two arrays as input) in this case because you would not benefit from the sliding variances for the two subsequences that can be computed using the current `stump` approach. Right now, the variance of subsequences `a` and `b` are computed from scratch every single time rather. This is very slow. Luckily, your DTW function isn't too complicated and accepts two variances (`a` and `b`) as input so you'll need to figure out the math for using the sliding window approach.

> At the first, I would ask how SWAMP should be coded .
As the paper, SWAMP is combination of several algorithm(DTW MP is a part of them). However DTW MP itself can be standalone function.

I think I understand the `DTW+MP` part that you have in your notebook. Basically, instead of computing the z-normalized Euclidean distance, you are computing the z-normalized DTW. Is that correct?

Can you explain to me what is SWAMP? How does it relate to `DTW+MP`? It would be great if we could use this notebook to teach us what SWAMP is.

ken-maeda:
I'm sorry for the delay. I got numba error. I have almost done walkthrogh for swamp. But I have calculation speed problem of matrix profile for simple loop(dtwMP, LBMP don't have to be diagonal.). Do you have any resource of code I should refer to?

seanlaw:
> Do you have any resource of code I should refer to?

Would [this example](https://gist.github.com/seanlaw/c95126ee9969202d25a839771be229d9) of computing the matrix profile row-by-row (instead of diagonally) be helpful?

ken-maeda:
@seanlaw Thanks, I'll check this, and try out.

ken-maeda:
I reproduced swamp in DTW_MP_trial5_SWAMP step by step.

What is SWAMP?
Object: Finding top-"dtw"motifs.
1.But DTW MP calculation is high.
2.DTW and Lower Bound(LB) are correlated.
3.Lower Bound and Down Sampled LB(DSLB) are also correlated.
4.So DTW and DSLB are correlated, it is far faster calculation.
5.By calculating DSLB from higher sampling rate to lower rate, it helps to limit only menalingfull area for calculating dtw.
6.Calculate DTW with a part of pairs.

I hope the note can help understanding what swamp is.

seanlaw:
@ken-maeda Thank you for pulling all of this together. I can see that it was a lot of work and I appreciate you taking the time to work on it. Given the complexity of DTW and all of the nuanced variants, I think it is best to keep `swamp` as a separate package from STUMPY as it is beyond the scope of STUMPY. 

If it is okay with you, I would like close this PR and keep it here for future reference. How does that sound?

ken-maeda:
@seanlaw No problem, I'll find other things I can do. Thanks

PR #752, opened 2022-12-03
fix skip in `test_core` when no GPU available

NimaSarajpoor:
After moving the two functions `_gpu_searchsorted_left` and `_gpu_searchsorted_right` to `core.py` in PR #714 , I modified `test_core.py`. **However, my modification resulted in skipping the whole unit test `test_core.py` when there is no GPU.** 

This PR fixes this issue.

### DIFF
--- a/stumpy/__init__.py
+++ b/stumpy/__init__.py
@@ -50,6 +50,7 @@ else:  # pragma: no cover
     from .core import _gpu_aampdist_driver_not_found as gpu_aampdist  # noqa: F401
     from .core import _gpu_stimp_driver_not_found as gpu_stimp  # noqa: F401
     from .core import _gpu_aamp_stimp_driver_not_found as gpu_aamp_stimp  # noqa: F401
+
     core._gpu_searchsorted_left = core._gpu_searchsorted_left_driver_not_found
     core._gpu_searchsorted_right = core._gpu_searchsorted_right_driver_not_found
 
@@ -179,7 +180,7 @@ else:  # pragma: no cover
             gpu_aamp_stimp.__doc__ = ast.get_docstring(cd)
 
     # Fix _gpu_searchsorted_left Docs
-    _gpu_searchsorted_left.__doc__ = ""
+    core._gpu_searchsorted_left.__doc__ = ""
     filepath = pathlib.Path(__file__).parent / "core.py"
 
     file_contents = ""
@@ -191,10 +192,10 @@ else:  # pragma: no cover
     ]
     for fd in function_definitions:
         if fd.name == "_gpu_searchsorted_left":
-            _gpu_searchsorted_left.__doc__ = ast.get_docstring(fd)
+            core._gpu_searchsorted_left.__doc__ = ast.get_docstring(fd)
 
     # Fix _gpu_searchsorted_right Docs
-    _gpu_searchsorted_right.__doc__ = ""
+    core._gpu_searchsorted_right.__doc__ = ""
     filepath = pathlib.Path(__file__).parent / "core.py"
 
     file_contents = ""
@@ -206,7 +207,7 @@ else:  # pragma: no cover
     ]
     for fd in function_definitions:
         if fd.name == "_gpu_searchsorted_right":
-            _gpu_searchsorted_right.__doc__ = ast.get_docstring(fd)
+            core._gpu_searchsorted_right.__doc__ = ast.get_docstring(fd)
 
 try:
     _dist = get_distribution("stumpy")
--- a/stumpy/__init__.py
+++ b/stumpy/__init__.py
@@ -38,9 +38,8 @@ if cuda.is_available():
     from .gpu_aampdist import gpu_aampdist  # noqa: F401
     from .gpu_stimp import gpu_stimp  # noqa: F401
     from .gpu_aamp_stimp import gpu_aamp_stimp  # noqa: F401
-    from .core import _gpu_searchsorted_left  # noqa: F401
-    from .core import _gpu_searchsorted_right  # noqa: F401
 else:  # pragma: no cover
+    from . import core
     from .core import _gpu_stump_driver_not_found as gpu_stump  # noqa: F401
     from .core import _gpu_aamp_driver_not_found as gpu_aamp  # noqa: F401
     from .core import _gpu_ostinato_driver_not_found as gpu_ostinato  # noqa: F401
@@ -51,12 +50,8 @@ else:  # pragma: no cover
     from .core import _gpu_aampdist_driver_not_found as gpu_aampdist  # noqa: F401
     from .core import _gpu_stimp_driver_not_found as gpu_stimp  # noqa: F401
     from .core import _gpu_aamp_stimp_driver_not_found as gpu_aamp_stimp  # noqa: F401
-    from .core import (
-        _gpu_searchsorted_left_driver_not_found as _gpu_searchsorted_left,
-    )  # noqa: F401
-    from .core import (
-        _gpu_searchsorted_right_driver_not_found as _gpu_searchsorted_right,
-    )  # noqa: F401
+    core._gpu_searchsorted_left = core._gpu_searchsorted_left_driver_not_found
+    core._gpu_searchsorted_right = core._gpu_searchsorted_right_driver_not_found
 
     import ast
     import pathlib
--- a/stumpy/__init__.py
+++ b/stumpy/__init__.py
@@ -38,6 +38,8 @@ if cuda.is_available():
     from .gpu_aampdist import gpu_aampdist  # noqa: F401
     from .gpu_stimp import gpu_stimp  # noqa: F401
     from .gpu_aamp_stimp import gpu_aamp_stimp  # noqa: F401
+    from .core import _gpu_searchsorted_left  # noqa: F401
+    from .core import _gpu_searchsorted_right  # noqa: F401
 else:  # pragma: no cover
     from .core import _gpu_stump_driver_not_found as gpu_stump  # noqa: F401
     from .core import _gpu_aamp_driver_not_found as gpu_aamp  # noqa: F401
@@ -49,6 +51,13 @@ else:  # pragma: no cover
     from .core import _gpu_aampdist_driver_not_found as gpu_aampdist  # noqa: F401
     from .core import _gpu_stimp_driver_not_found as gpu_stimp  # noqa: F401
     from .core import _gpu_aamp_stimp_driver_not_found as gpu_aamp_stimp  # noqa: F401
+    from .core import (
+        _gpu_searchsorted_left_driver_not_found as _gpu_searchsorted_left,
+    )  # noqa: F401
+    from .core import (
+        _gpu_searchsorted_right_driver_not_found as _gpu_searchsorted_right,
+    )  # noqa: F401
+
     import ast
     import pathlib
 
@@ -173,6 +182,36 @@ else:  # pragma: no cover
     for cd in class_definitions:
         if cd.name == "gpu_aamp_stimp":
             gpu_aamp_stimp.__doc__ = ast.get_docstring(cd)
+
+    # Fix _gpu_searchsorted_left Docs
+    _gpu_searchsorted_left.__doc__ = ""
+    filepath = pathlib.Path(__file__).parent / "core.py"
+
+    file_contents = ""
+    with open(filepath, encoding="utf8") as f:
+        file_contents = f.read()
+    module = ast.parse(file_contents)
+    function_definitions = [
+        node for node in module.body if isinstance(node, ast.FunctionDef)
+    ]
+    for fd in function_definitions:
+        if fd.name == "_gpu_searchsorted_left":
+            _gpu_searchsorted_left.__doc__ = ast.get_docstring(fd)
+
+    # Fix _gpu_searchsorted_right Docs
+    _gpu_searchsorted_right.__doc__ = ""
+    filepath = pathlib.Path(__file__).parent / "core.py"
+
+    file_contents = ""
+    with open(filepath, encoding="utf8") as f:
+        file_contents = f.read()
+    module = ast.parse(file_contents)
+    function_definitions = [
+        node for node in module.body if isinstance(node, ast.FunctionDef)
+    ]
+    for fd in function_definitions:
+        if fd.name == "_gpu_searchsorted_right":
+            _gpu_searchsorted_right.__doc__ = ast.get_docstring(fd)
 
 try:
     _dist = get_distribution("stumpy")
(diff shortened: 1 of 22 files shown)


NimaSarajpoor:
@seanlaw 
I was wondering if we could report the changes in the skipped tests. If we have something similar to coverage report but for `skipped tests`, then we could have caught this sooner.

seanlaw:
Hmm, perhaps, I think we should move the functions into `gpu_core.py`. This way, indeed, the whole file can be skipped and we NEVER put any GPU stuff in `core.py`. Would that be a reasonable alternative?

> I was wondering if we could generate report that shows the changes in the number of skipped tests. If we have something similar to coverage report but for skipped tests, then we could have caught this earlier. 

I don't see an easy way to know what the difference in the number of skipped tests would be without re-running all of the before and after tests. Hmmm....

NimaSarajpoor:
> Hmm, perhaps, I think we should move the functions into `gpu_core.py`. This way, indeed, the whole file can be skipped and we NEVER put any GPU stuff in `core.py`. Would that be a reasonable alternative? I think keeping CPU code separate from GPU code will make things much easier to maintain

I totally agree with this! I will create `gpu_core.py`.


> I don't see an easy way to know what the difference in the number of skipped tests would be without re-running all of the before and after tests. Hmmm....

If I find something regarding this matter, I will share it with you.

seanlaw:
> pytest.skip("Skipping Tests No GPUs Available", allow_module_level=True)

I probably should've paid closer attention to this when I was reviewing but `allow_module_level=True` skips the entire module (i.e., all tests in the `.py` file). It is okay to skip single tests within a test file by omitting `allow_module_level=True`:

https://github.com/TDAmeritrade/stumpy/blob/ac238b61897d53621f8e4607f88c7517f77ecf52/tests/test_non_normalized_decorator.py#L431-L435

I see that you've fixed that now in the latest commit.

However, now that we've fixed that, instead of splitting it off into `core.py` as I've proposed, let's keep the GPU functions where they are in `core.py` and simply ensure that we only deal with skipping at the local function level (i.e., remove `allow_module_level=True`).

=====

I understand what you are trying to do here but this section seems out of place within the test file:

```
if cuda.is_available():
    from stumpy.core import _gpu_searchsorted_left, _gpu_searchsorted_right

    @cuda.jit("(f8[:, :], f8[:], i8[:], i8, b1, i8[:])")
    def _gpu_searchsorted_kernel(a, v, bfs, nlevel, is_left, idx):
        # A wrapper kernel for calling device function _gpu_searchsorted_left/right.
        i = cuda.grid(1)
        if i < a.shape[0]:
            if is_left:
                idx[i] = _gpu_searchsorted_left(a[i], v[i], bfs, nlevel)
            else:
                idx[i] = _gpu_searchsorted_right(a[i], v[i], bfs, nlevel)

else:  # pragma: no cover
    from stumpy.core import (
        _gpu_searchsorted_left_driver_not_found as _gpu_searchsorted_left,
    )
    from stumpy.core import (
        _gpu_searchsorted_right_driver_not_found as _gpu_searchsorted_right,
    )
```

Specifically, the imports using `if/else` looks weird. Maybe we can simplify things by moving this logic into `stumpy/__init__.py`:

https://github.com/TDAmeritrade/stumpy/blob/ac238b61897d53621f8e4607f88c7517f77ecf52/stumpy/__init__.py#L32-L51

Then, you can simply do:

```
from stumpy import core, config, _gpu_searchsorted_left, _gpu_searchsorted_right

if cuda.is_available():
    @cuda.jit("(f8[:, :], f8[:], i8[:], i8, b1, i8[:])")
    def _gpu_searchsorted_kernel(a, v, bfs, nlevel, is_left, idx):
        # A wrapper kernel for calling device function _gpu_searchsorted_left/right.
        i = cuda.grid(1)
        if i < a.shape[0]:
            if is_left:
                idx[i] = _gpu_searchsorted_left(a[i], v[i], bfs, nlevel)
            else:
                idx[i] = _gpu_searchsorted_right(a[i], v[i], bfs, nlevel)


@pytest.mark.filterwarnings("ignore", category=NumbaPerformanceWarning)
@patch("stumpy.config.STUMPY_THREADS_PER_BLOCK", TEST_THREADS_PER_BLOCK)
def test_gpu_searchsorted():
    if not cuda.is_available():  # pragma: no cover
        pytest.skip("Skipping Tests No GPUs Available")

    .
    .
    .
    _gpu_searchsorted_kernel[blocks_per_grid, threads_per_block](
            device_A, device_V, device_bfs, nlevel, is_left, device_comp_IDX
        )
```

=====

Unrelated: What is concerning is that our reporting didn't explicitly tell us that the entire `test_core.py` was skipped:

```
========================================= test session starts =========================================
platform darwin -- Python 3.10.8, pytest-7.2.0, pluggy-1.0.0
rootdir: /Users/slaw/Git/stumpy.git, configfile: pytest.ini
plugins: check-links-0.7.1, cov-4.0.0, anyio-3.6.2
collected 0 items / 1 skipped

======================================= short test summary info =======================================
SKIPPED [1] tests/test_gpu_stump.py:19: Skipping Tests No GPUs Available
========================================= 1 skipped in 0.74s ==========================================
========================================= test session starts =========================================
platform darwin -- Python 3.10.8, pytest-7.2.0, pluggy-1.0.0
rootdir: /Users/slaw/Git/stumpy.git, configfile: pytest.ini
plugins: check-links-0.7.1, cov-4.0.0, anyio-3.6.2
collected 0 items / 1 skipped
```

So, I've changed the `pytest` reporting in `test.sh` (see commit 92e4100c043e622cdb617b81346a6a3bb0a2b2e7) and updated the flags to `py.test -rsx`, which will show extra summaries (`-r`) on skipped tests (`-s`) and failed tests (`-x`)

NimaSarajpoor:
@seanlaw 
Thank you for your thorough explanation! 

> However, now that we've fixed that, instead of splitting it off into `core.py` as I've proposed, let's keep the GPU functions where they are in `core.py` and simply ensure that we only deal with skipping at the local function level (i.e., remove `allow_module_level=True`).

Sure. I will keep it. 

I think you are trying to do some sort of trade-off in your mind to see if it is worth it to put a few `gpu`-related code to a separate module.

> Specifically, the imports using `if/else` looks weird. Maybe we can simplify things by moving this logic into `stumpy/__init__.py`:

Thank you for the suggestion! I've moved the if-else logic to the `__init__.py`.


> So, I've changed the `pytest` reporting in `test.sh` (see commit [92e4100](https://github.com/TDAmeritrade/stumpy/commit/92e4100c043e622cdb617b81346a6a3bb0a2b2e7)) and updated the flags to `py.test -rsx`, which will show extra summaries (`-r`) on skipped tests (`-s`) and failed tests (`-x`). It's still now what you were asking for but at least we would've seen that `test_core.py` was skipped

I merged the changes and ran tests locally. The pytest result's report is good 👍 

---

**A couple of notes:**
1) Please wait till I test modules in my other PC that has GPU so we can make sure everything is fine.

2) Could you please checkout the changed files to see if they make sense or not?

seanlaw:
>Please wait till I test modules in my other PC that has GPU so we can make sure everything is fine.

Sounds good

>Could you please checkout the changed files to see if they make sense or not?

I will find some time to do that

stumpy/gpu_aamp.py
```diff
@@ -10,6 +10,7 @@
 from numba import cuda
 
 from . import core, config
+from .core import _gpu_searchsorted_right
```
seanlaw, line 13: I think this will be a problem. If you directly import from `.core` then your python environment will have no concept/idea of `_gpu_searchsorted_right_driver_not_found`! Instead, If there are no GPUs, it will still be trying to execute the real `core._gpu_searchsorted_right` 

The only way it will detect the "correct" `_gpu_searchsorted_right` is if you import it directly from `stumpy._gpu_searchsorted_right` (which inherits from `__init__.py`)

I will say that I don't love having a private function in the top-level `stumpy` namespace. I would actually prefer to continue being able to `from .core import _gpu_searchsorted_right` as you've shown and somehow trick the python environment into figuring out which function to use (either the GPU available one OR the `driver_not_found` version).

stumpy/gpu_stump.py
```diff
@@ -10,6 +10,7 @@
 from numba import cuda
 
 from . import core, config
+from .core import _gpu_searchsorted_right
```
seanlaw, line 13: See comment above

stumpy/gpu_aamp.py
seanlaw: Ohh! I think found a way!

In `__init__.py`, we do:

```
else:
    from . import core
    core._gpu_searchsorted_left = core._gpu_searchsorted_left_driver_not_found
```

Then, we might be able to keep our import as `from .core import _gpu_searchsorted_right`

seanlaw: So, we'd remove `from .core import _gpu_searchsorted_right` in `__init__.py` and then simply replace the function when there is no GPU present

seanlaw (changes requested):

stumpy/gpu_aamp.py
seanlaw: This way, `_gpu_searchsorted_right` is NOT at the top-level import and can only be accessed via importing from `core`

NimaSarajpoor: > If you directly import from `.core` then your python environment will have no concept/idea of `_gpu_searchsorted_right_driver_not_found`! Instead, If there are no GPUs, it will still be trying to execute the real `core._gpu_searchsorted_right`

Right! 

> I will say that I don't love having a private function in the top-level stumpy namespace. I would actually prefer to continue being able to from .core import _gpu_searchsorted_right as you've shown and somehow trick the python environment into figuring out which function to use (either the GPU available one OR the driver_not_found version).

Yeah... better to keep private functions in a private mode :)



> Ohh! I think found a way!
> 
> In `__init__.py`, we do:
> 
> ```
> else:
>     from . import core
>     core._gpu_searchsorted_left = core._gpu_searchsorted_left_driver_not_found
> ```
> 
> Then, we might be able to keep our import as `from .core import _gpu_searchsorted_right`

> This way, _gpu_searchsorted_right is NOT at the top-level import and can only be accessed via importing from core

That is a very very smart move! Cool!

stumpy/__init__.py
```diff
@@ -174,6 +179,36 @@
         if cd.name == "gpu_aamp_stimp":
             gpu_aamp_stimp.__doc__ = ast.get_docstring(cd)
 
+    # Fix _gpu_searchsorted_left Docs
```
seanlaw, line 182: Btw, I don't think docstring replacement is necessary here. We only needed to replace docstrings for GPU functions that are at the top-level and really needed for display in our API documentation (at `stumpy.readthedocs.org`). Without it, our documentation on readthedocs doesn't show up correctly. However, since these functions are private, they are not displayed on readthedocs.

Please remove

NimaSarajpoor: @seanlaw 
Ahhh. I did not know about it! 

I thought the reason for replacing the docstring was to show it to the user in a pop-up mode. (like...in jupyter, if I do `shift+tab`, I can see the docstring) I assume some of editors have this capability. If I remove it, then user will see the docstring of the dummy function. 

Is that okay?

seanlaw: I don't know. Given that it's a private function, users wouldn't even know that it exists, right?

NimaSarajpoor: > Given that it's a private function, users wouldn't even know that it exists, right?

Right :) cool...I will remove it.

stumpy/__init__.py
```diff
@@ -49,6 +50,10 @@
     from .core import _gpu_aampdist_driver_not_found as gpu_aampdist  # noqa: F401
     from .core import _gpu_stimp_driver_not_found as gpu_stimp  # noqa: F401
     from .core import _gpu_aamp_stimp_driver_not_found as gpu_aamp_stimp  # noqa: F401
+
+    core._gpu_searchsorted_left = core._gpu_searchsorted_left_driver_not_found
```
seanlaw, line 54: This feels disconnected from the `from . import core` above

stumpy/gpu_aamp.py
seanlaw, line 13: It almost feels like you don't need this line and then later, you'd do

`core._gpu_searchsorted_right`

stumpy/gpu_aamp.py
```diff
@@ -186,7 +187,7 @@ def _compute_and_update_PI_kernel(
                 indices_R[j] = i
 
         if p_norm < profile[j, -1]:
-            idx = core._gpu_searchsorted_right(profile[j], p_norm, bfs, nlevel)
+            idx = _gpu_searchsorted_right(profile[j], p_norm, bfs, nlevel)
```
seanlaw, line 190: See comment above

stumpy/gpu_stump.py
seanlaw, line 13: See comment above

stumpy/gpu_stump.py
```diff
@@ -194,7 +195,7 @@ def _compute_and_update_PI_kernel(
                 indices_R[j] = i
 
         if p_norm < profile[j, -1]:
-            idx = core._gpu_searchsorted_right(profile[j], p_norm, bfs, nlevel)
+            idx = _gpu_searchsorted_right(profile[j], p_norm, bfs, nlevel)
```
seanlaw, line 198: see comment above

seanlaw (changes requested):

seanlaw:
It looks cleaner already!

stumpy/__init__.py
NimaSarajpoor: Yeah...I noticed that! I am waiting to see how the test suite goes. Then, I will test gpu modules on my other PC tomorrow, and update my main and merge it to this branch and take care of these small things. Then, I will push again :)

stumpy/gpu_aamp.py
NimaSarajpoor: Yeah...the only reason I did that was to avoid the very little time needed to look up the function  `_gpu_searchsorted_right` in the `core.py`. Since `_gpu_searchsorted_right` can be called a lot of times throughout the process, I thought it might improve performance a little bit 😄 

I will do `core._gpu_searchsorted_right`

tests/test_core.py
```diff
@@ -12,36 +12,23 @@
...
+            else:
+                idx[i] = core._gpu_searchsorted_right(a[i], v[i], bfs, nlevel)
+
 
 try:
     from numba.errors import NumbaPerformanceWarning
 except ModuleNotFoundError:
     from numba.core.errors import NumbaPerformanceWarning
 
-TEST_THREADS_PER_BLOCK = 10
```
seanlaw, line 29: For consistency, I would keep this line and later do:

```
@patch("stumpy.config.STUMPY_THREADS_PER_BLOCK", TEST_THREADS_PER_BLOCK)
```

PR #754, opened 2022-12-05
Fixed #693 Bug in `is False` by converting to bool

### LINKED ISSUE #693: Replace `is False` with `== False`  and Refactor warning of `ignore_trivial`
(1) As discussed in #691, `is` checks if two objects have the same `id`s or not. The outcome of an expression can be `False` but it can have different `id` than the object `False`. Hence, we need to replace `is False` with `== False`

(2) As suggested by @seanlaw in #691, it is better to refactor the warning messages that check the param `ignore_trivial` and add test function(s) to check its functionality

seanlaw:
See #693 

# Pull Request Checklist

- [x] Create a new branch

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -383,7 +383,7 @@ def are_arrays_equal(a, b):  # pragma: no cover
     if a.shape != b.shape:
         return False
 
-    return ((a == b) | (np.isnan(a) & np.isnan(b))).all()
+    return bool(((a == b) | (np.isnan(a) & np.isnan(b))).all())
 
 
 def are_distances_too_small(a, threshold=10e-6):  # pragma: no cover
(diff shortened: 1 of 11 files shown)


stumpy/core.py
```diff
@@ -383,7 +383,7 @@ def are_arrays_equal(a, b):  # pragma: no cover
     if a.shape != b.shape:
         return False
 
-    return ((a == b) | (np.isnan(a) & np.isnan(b))).all()
+    return bool(((a == b) | (np.isnan(a) & np.isnan(b))).all())
```
NimaSarajpoor, line 386: Cool. I think `bool(...)` should return `False` that points to the same memory where the Boolean value `False` is located.

stumpy/core.py
```diff
@@ -2989,3 +2989,60 @@ def _gpu_searchsorted_right(a, v, bfs, nlevel):
         idx = next_idx
 
     return idx
+
+
+def check_ignore_trivial(T_A, T_B, ignore_trivial):
+    """
+    Check inputs and verify the appropriateness for self-joins vs AB-joins and
+    provides relevant warnings.
```
NimaSarajpoor, line 2997: just a minor comment here: 
I think it should be "provide"

PR #759, opened 2022-12-10
Parallelize rolling std

NimaSarajpoor:
Using parallelization to compute the rolling std. So, instead of using welford, we calculate the "std"  of each subsequence on its own. The time complexity of this approach is `O(n.m)`, where `n` is the length of the time series input `T` (if 1D), and `m` is the window size.

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -3104,8 +3104,8 @@ def _rolling_nanstd_2d(T, m):
         is the std value of T[i, j : j + m]
     """
     out = np.empty((T.shape[0], T.shape[1] - m + 1), dtype=np.float64)
-    for i in range(out.shape[0]):
-        for j in prange(out.shape[1]):
+    for i in range(T.shape[0]):
+        for j in prange(T.shape[1] - m + 1):
             out[i, j] = np.nanstd(T[i, j : j + m])
 
     return out
@@ -3116,7 +3116,7 @@ def rolling_nanstd(T, m):
     Compute the rolling standard deviation for 1D and 2D arrays while ignoring
     NaNs.
 
-    This a convenience wrapper around `_rolling_nanstd`.
+    This a convenience wrapper around `_rolling_nanstd_2d`.
 
     This essentially replaces:
 
@@ -3139,7 +3139,7 @@ def rolling_nanstd(T, m):
         raise ValueError("The input array `T` must be 1D or 2D.")
 
     out = _rolling_nanstd_2d(np.atleast_2d(T), m)
-    if T.ndim == 1:
-        return np.squeeze(out)
-    else:
+    if T.ndim == 2:
         return out
+    else:  # T.ndim is 1
+        return np.squeeze(out)
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -659,13 +659,14 @@ def welford_nanstd(a, w=None):
     return np.sqrt(np.clip(welford_nanvar(a, w), a_min=0, a_max=None))
 
 
-def rolling_nanstd_fast(a, w):
+def rolling_nanstd_fast(a, w):  # pragma nocover
     """
     Compute the rolling standard deviation for 1-D and 2-D arrays while ignoring NaNs
     using a modified version of Welford's algorithm but is much faster than using
     `np.nanstd` with stride tricks.
 
-    This a convenience wrapper around `welford_nanstd`.
+    This a convenience wrapper around `welford_nanstd`. This is left for future
+    reference if needed.
 
     This essentially replaces:
 
@@ -3135,7 +3136,7 @@ def rolling_nanstd(T, m):
     output : numpy.ndarray
         Rolling window nanstd.
     """
-    if T.ndim > 2:
+    if T.ndim > 2:  # pragma nocover
         raise ValueError("The input array `T` must be 1D or 2D.")
 
     out = _rolling_nanstd_2d(np.atleast_2d(T), m)
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -693,7 +693,7 @@ def rolling_nanstd(a, w, welford=False):
     NaNs.
 
     This essentially replaces:
-        `np.nanstd(rolling_window(a, w), axis=a.ndim)`
+        `np.nanstd(rolling_window(a[..., start:stop], w), axis=a.ndim)`
 
     Parameters
     ----------
@@ -731,7 +731,7 @@ def _rolling_nanmin_1d(a, w=None):
 
     This essentially replaces:
 
-        `np.nanmin(rolling_window(T[..., start:stop], m), axis=T.ndim)`
+        `np.nanmin(rolling_window(a[..., start:stop], w), axis=a.ndim)`
 
     Parameters
     ----------
@@ -761,7 +761,7 @@ def _rolling_nanmax_1d(a, w=None):
 
     This essentially replaces:
 
-        `np.nanmax(rolling_window(T[..., start:stop], m), axis=T.ndim)`
+        `np.nanmax(rolling_window(a[..., start:stop], w), axis=a.ndim)`
 
     Parameters
     ----------
@@ -793,7 +793,7 @@ def rolling_nanmin(a, w):
 
     This essentially replaces:
 
-        `np.nanmin(rolling_window(T[..., start:stop], m), axis=T.ndim)`
+        `np.nanmin(rolling_window(a[..., start:stop], w), axis=a.ndim)`
 
     Parameters
     ----------
@@ -822,7 +822,7 @@ def rolling_nanmax(a, w):
 
     This essentially replaces:
 
-        `np.nanmax(rolling_window(T[..., start:stop], m), axis=T.ndim)`
+        `np.nanmax(rolling_window(a[..., start:stop], w), axis=a.ndim)`
 
     Parameters
     ----------
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -3083,28 +3083,32 @@ def check_ignore_trivial(T_A, T_B, ignore_trivial):
 
 
 @njit(parallel=True, fastmath={"nsz", "arcp", "contract", "afn", "reassoc"})
-def _rolling_nanstd(T, m, out):
+def _rolling_nanstd_2d(T, m):
     """
     A Numba JIT-compiled and parallelized function for computing the std of
-    subsequences of length `m` in `T`, which is 1D array.
+    subsequences of length `m` in `T`, which must be 2D array. It is user's
+    responsibility to make sure that the inputs are valid.
 
     Parameters
     ----------
     T : numpy.ndarray
-        The input array
+        The input array with two dimensions.
 
     m : int
         The rolling window size
 
-    out : numpy.ndarray
-        The variable to store the std values, with length len(T) - m + 1
-
     Returns
     -------
-    None
+    out : numpy.ndarray
+        This 2D array has the shape of `(T.shape[0], T.shape[1]-m+1)`. out[i,j]
+        is the std value of T[i, j : j + m]
     """
-    for i in prange(len(out)):
-        out[i] = np.nanstd(T[i : i + m])
+    out = np.empty((T.shape[0], T.shape[1] - m + 1), dtype=np.float64)
+    for i in range(out.shape[0]):
+        for j in prange(out.shape[1]):
+            out[i, j] = np.nanstd(T[i, j : j + m])
+
+    return out
 
 
 def rolling_nanstd(T, m):
@@ -3134,13 +3138,7 @@ def rolling_nanstd(T, m):
     if T.ndim > 2:
         raise ValueError("The input array `T` must be 1D or 2D.")
 
-    T_2D = np.atleast_2d(T)
-    l = T_2D.shape[1] - m + 1
-
-    out = np.empty((T_2D.shape[0], l), dtype=np.float64)
-    for i, T_row in enumerate(T_2D):
-        _rolling_nanstd(T_row, m, out[i])
-
+    out = _rolling_nanstd_2d(np.atleast_2d(T), m)
     if T.ndim == 1:
         return np.squeeze(out)
     else:
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -7,7 +7,7 @@ import functools
 import inspect
 
 import numpy as np
-from numba import njit, cuda
+from numba import njit, cuda, prange
 from scipy.signal import convolve
 from scipy.ndimage import maximum_filter1d, minimum_filter1d
 from scipy import linalg
@@ -659,7 +659,7 @@ def welford_nanstd(a, w=None):
     return np.sqrt(np.clip(welford_nanvar(a, w), a_min=0, a_max=None))
 
 
-def rolling_nanstd(a, w):
+def rolling_nanstd_fast(a, w):
     """
     Compute the rolling standard deviation for 1-D and 2-D arrays while ignoring NaNs
     using a modified version of Welford's algorithm but is much faster than using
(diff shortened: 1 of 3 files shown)


NimaSarajpoor:
@seanlaw 
Could you please check out the changes and let me know how you feel?

seanlaw:
@NimaSarajpoor I think what you have is too complicated. Here's what I would like to suggest when thinking about the structure:

1. Starting with the interface:

```
def rolling_nanstd(a, w, welford=False):
    axis = a.ndim - 1  # Account for rolling
    if welford:
        return np.apply_along_axis(
            lambda a_row, w: welford_nanstd(a_row, w), axis=axis, arr=a, w=w
        )
    else:
        return np.apply_along_axis(
            lambda a_row, w: _rolling_nanstd(a_row, w), axis=axis, arr=a, w=w
        )
```

2. Then, add a `_rolling_nanstd` function

```
@njit(parallel=True, fastmath={"nsz", "arcp", "contract", "afn", "reassoc"})
def _rolling_nanstd(a, w):
    out = np.empty(a.shape[0] - m + 1), dtype=np.float64)
    for i in prange(a.shape[0] - m + 1):
        out[i] = np.nanstd(a[i : i + m])

    return out
```

Note that the inputs should not be `(T, m)` and should be `(a, w)` instead to be consistent with `numpy`. So, by default, we use the higher precision version. However, power users can use `welford=True` if it is ever needed (rarely) and the interface is the same.

NimaSarajpoor:
> Note that the inputs should not be `(T, m)` and should be `(a, w)` instead to be consistent with `numpy`

Ahh.. I thought you used `a, w` before and then switched to `T, m` but forgot to revise the functions with `a` and `w` :) So, I was trying to be consistent 😄 .... I will fix it.



> What do you think? Hopefully I didn't miss anything

It seems everything is good :) I will revise the code.

stumpy/core.py
```diff
@@ -659,16 +659,43 @@ def welford_nanstd(a, w=None):
...
 
 
-def rolling_nanstd(a, w):
+@njit(parallel=True, fastmath={"nsz", "arcp", "contract", "afn", "reassoc"})
+def _rolling_nanstd_1d(a, w):
     """
-    Compute the rolling standard deviation for 1-D and 2-D arrays while ignoring NaNs
-    using a modified version of Welford's algorithm but is much faster than using
-    `np.nanstd` with stride tricks.
+    A Numba JIT-compiled and parallelized function for computing the std of
```
seanlaw, line 665: A Numba JIT-compiled and parallelized function for computing the rolling standard deviation for a 1-D array while ignoring NaN

stumpy/core.py
```diff
@@ -659,16 +659,43 @@ def welford_nanstd(a, w=None):
...
+    each subsequence with length `w` in `a`, which is a 1D array.
 
-    This a convenience wrapper around `welford_nanstd`.
+    Parameters
+    ----------
+    a : numpy.ndarray
+        The input 1D array
```
seanlaw, line 671: The input array

stumpy/core.py
```diff
@@ -659,16 +659,43 @@ def welford_nanstd(a, w=None):
...
+        The rolling window size
+
+    Returns
+    -------
+    out : numpy.ndarray
+        This 1D array has the length of `a.shape[0]-w+1`. `out[i]`
+        contains the std value of `a[i : i + w]`
+    """
+    l = a.shape[0] - w + 1
+    out = np.empty(l, dtype=np.float64)
```
seanlaw, line 683: I think it's fine to combine the two lines into `out = np.empty(a.shape[0] - w + 1, dtype=np.float64)` as it is pretty straightforward

stumpy/core.py
```diff
@@ -659,16 +659,43 @@ def welford_nanstd(a, w=None):
...
+    for i in prange(l):
+        out[i] = np.nanstd(a[i : i + w])
+
+    return out
 
-    This essentially replaces:
 
+def rolling_nanstd(a, w, welford=False):
+    """
+    Compute the rolling standard deviation for 1D and 2D arrays while ignoring
```
seanlaw, line 692: Compute the rolling standard deviation for 1D and 2D arrays while ignoring NaNs.

stumpy/core.py
```diff
@@ -679,15 +706,30 @@ def rolling_nanstd(a, w):
     w : numpy.ndarray
         The rolling window size
 
+    welford : bool, default False
+        When False (default), the computation is parallelized and the std of
```
seanlaw, line 710: replace `std` with `stddev`

stumpy/core.py
```diff
@@ -679,15 +706,30 @@ def rolling_nanstd(a, w):
...
+        each subsequence is calculated on its own. When `welford==True`, the
+        welford method is used to reduce the computing time at the cost of slight
```
seanlaw, line 712: at the cost of slightly reduced precision

stumpy/core.py
```diff
@@ -679,15 +706,30 @@ def rolling_nanstd(a, w):
...
+        loss of precision
+
     Returns
     -------
-    output : numpy.ndarray
-        Rolling window nanstd.
+    out : numpy.ndarray
+        Rolling window nanstd. When `a` is 1D, `out[i]` is the std of `a[i, i + w]`.
```
seanlaw, line 718: Rolling window nanstd

stumpy/core.py
```diff
@@ -679,15 +706,30 @@ def rolling_nanstd(a, w):
...
+        When `a` is 2D, `out[i, j]` is the std of `a[i, j : j + w]`.
     """
+    if a.ndim > 2:  # pragma nocover
```
seanlaw, line 721: Wouldn't this still work for for 3D or 4D or ND arrays?

seanlaw (changes requested):

stumpy/core.py
NimaSarajpoor: @seanlaw 
Could you please checkout the docstring of this function and your suggestion again? I am trying to see what change I need to apply to the docstring. 
(Did you mean I should have tried to put the word `NaNs` in the first line?)

stumpy/core.py
NimaSarajpoor: I will revise this.

stumpy/core.py
NimaSarajpoor: Thank you for the suggestion. I will fix this.

stumpy/core.py
NimaSarajpoor: Did you mean I should remove the rest of description for `out`?

stumpy/core.py
NimaSarajpoor: I think it works but I do not know if it would be meaningful for this function since `np.nanstd(..)` will be the nanstd of flattened array.

stumpy/core.py
NimaSarajpoor: Just to make sure we are on the same page, I assume you already noticed `l` in the next line in `prange(l)`, and you are okay with replacing that with `a.shape[0] - w + 1.`

Is that correct?

seanlaw: Ahh, I didn't notice that. In that case, please ignore this comment

stumpy/core.py
seanlaw: I meant that you should simply shorten it to what I suggested above and remove everything else

stumpy/core.py
seanlaw: Correct. I meant that you should simply shorten it to what I suggested above and remove everything else

stumpy/core.py
seanlaw: In that case, shouldn't we change `_rolling_nanstd_1d` to:

```
out[i] = np.nanstd(a[..., i : i + w], axis=a.ndim)
```

seanlaw: Danggg, `numba` doesn't support `axis` argument for `np.nanstd`

seanlaw: This is going to get dirty as the behavior is going to be different form the other rolling functions

seanlaw: As mentioned in other functions, this function should be compared against the output of its numpy equivalent of:

```
np.nanstd(rolling_window(a[..., :], w), axis=a.ndim)
```

stumpy/core.py
NimaSarajpoor: Ahh.. got it! I will revise it.

stumpy/core.py
NimaSarajpoor: > Danggg, `numba` doesn't support `axis` argument for `np.nanstd`. Hmmmm

Yeah...it only supports the first argument.


> As mentioned in other functions, this `rolling_nanstd` function should be compared against the output of its numpy equivalent of:
> 
> ```
> np.nanstd(rolling_window(a[..., start:stop], w), axis=a.ndim)
> ```
> 
> This should be tested for multiple dimensions

I have no solution for it as of now. I am going to think about it...

seanlaw: Okay, let's not worry about generalizing it. However, is it correct for the 2D case? Something feels wrong

NimaSarajpoor: > Okay, let's not worry about generalizing it

Sure.

> However, is it correct for the 2D case? Something feels wrong

I just created the following naive function. 
```
def naive_std(T, m):
    return np.nanstd(core.rolling_window(T, m), axis=T.ndim)
```
I compared `ref` with `comp` for 2D array `T` using `npt.assert_almost_equal`. It raises no error and everything is good. 


---

personally speaking, it was hard for me to quickly understand how `numpy.apply_along_axis` works. Because, for a 2D input T, where each row is a time series, `numpy.apply_along_axis(...., axis=1, ...)` confused me a little bit. After reading the examples provided in the numpy doc, I can better understand how this function works.


Side note: 
The [numpy doc](https://numpy.org/doc/stable/reference/generated/numpy.apply_along_axis.html) says:
> Parameters
func1dfunction (M,) -> (Nj…)
This function should accept 1-D arrays. It is applied to 1-D slices of arr along the specified axis.

However, when I tried:
```
def my_func(a):
    return (a[0] + a[-1]) * 0.5

A = np.random.randint(1, 100, size=(3, 4, 5))
np.apply_along_axis(my_func, axis=0, arr=A)
```

It works. I think the slice of A along the specified axis is not 1D in this example.

NimaSarajpoor:
> @NimaSarajpoor Can you please add some tests for `test_core.py` and specifically to test `core.rolling_nanstd` directly for:
> 
> ```
> a = np.random.rand(60)
> w = 5
> 
> core.rolling_nanstd(a, w)
> core.rolling_nanstd(a.reshape(3, 20), w)
> ```

Sure. That would be a good idea to add such tests to `test_core.py`. I tested in jupyter notebook. I am going to put them in the `test_core.py`

stumpy/core.py
NimaSarajpoor: @seanlaw 
**So, I have been working on this, and I found the following:**

It seems that the new function works correctly for `a` with `ndim > 2`. I do not know why/how :) I am trying to understand how `core._rolling_nanstd_1d` works when input is not 1D and no axis is specified in `np.nanstd(..)`!!!


```
Example

# input
a = np.random.rand(4 * 3 * 17).reshape(4, 3, 17)
w = 5

# naive
start = 0
stop = 18
A = core.rolling_window(a[..., start:stop], w) # shape:  (4, 3, 13, 5)
ref_std = np.nanstd(A, axis=a.ndim) # shape: (4, 3, 13)

 
# our new func when `welford==False`
axis = a.ndim - 1
comp_std = np.apply_along_axis(
            lambda a_row, w: core._rolling_nanstd_1d(a_row, w), axis=axis, arr=a, w=w
        ) 

# comp_std.shape: (4, 3, 13)

>>> npt.assert_almost_equal(ref_std, comp_std)
True
```

NimaSarajpoor: I think this is how `numpy.apply_along_axis(func1d, ...)` works: 

Let's say `arr` has shape of `(2, 3, 4)` and `axis=1`, then it applies `func1d` on all of the following slices:
```
arr[0, :, 0]
arr[0, :, 1]
arr[0, :, 2]
arr[0, :, 3]
arr[0, :, 4]
...
arr[1, :, 1]
arr[1, :, 2]
arr[1, :, 3]
arr[1, :, 4]
```

So, we do not need to specify `axis` for `np.nanstd`.

NimaSarajpoor:
@seanlaw 
Please feel free to check out the changes.

stumpy/core.py
seanlaw: Ahhh, interesting! I just took a look at it and I am beginning to understand it better. So, each function must still handle the "rolling" calculation along a 1D array on its own but the `np.apply_along_axis` slices everything up for us. And we still benefit from `parallel=True` and `prange` since the function still needs to roll along the 1D array slice.

I guess we got lucky! This is great

NimaSarajpoor: > Should we remove the `ValueError` then?

Yeah...I will fix it now.

stumpy/core.py
```diff
@@ -659,17 +659,41 @@ def welford_nanstd(a, w=None):
...
+        out[i] = np.nanstd(a[i : i + w])
+
+    return out
 
-        `np.nanstd(rolling_window(T[..., start:stop], m), axis=T.ndim)`
+
+def rolling_nanstd(a, w, welford=False):
+    """
+    Compute the rolling standard deviation over the last axis of `a` while ignoring
+    NaNs.
```
NimaSarajpoor, line 693: I added "over the last axis of `a`"

stumpy/core.py
```diff
@@ -659,17 +659,41 @@ def welford_nanstd(a, w=None):
...
+
+    This essentially replaces:
+        `np.nanstd(rolling_window(a, w), axis=a.ndim)`
```
NimaSarajpoor, line 696: I removed `..., start:stop`

tests/naive.py
```diff
@@ -6,6 +6,11 @@
 from stumpy import core, config
 
 
+def rolling_nanstd(a, w):
+    # a can be 1D, 2D, or more. The rolling occurs on last axis.
```
NimaSarajpoor, line 10: Is this comment good enough for the naive function?

NimaSarajpoor (commented):
@seanlaw 
I left some comments to just bring your attention to them. If you think you are okay with them, just give it a 👍 :)

stumpy/core.py
```diff
@@ -659,17 +659,41 @@ def welford_nanstd(a, w=None):
...
+    w : int
+        The rolling window size
+
+    Returns
+    -------
+    out : numpy.ndarray
+        This 1D array has the length of `a.shape[0]-w+1`. `out[i]`
+        contains the stddev value of `a[i : i + w]`
+    """
+    l = a.shape[0] - w + 1
```
seanlaw, line 682: Can we use `n` instead of `l`?

NimaSarajpoor: Sure. I will revise it.

stumpy/core.py
seanlaw: I would keep this so that it is consistent with the other similar function

seanlaw: Ultimately, this will be used in `compute_mean_std` and in there we actually use:

```
rolling_nanstd(T[..., start:stop], m)
```

NimaSarajpoor: I see your point  now :) I will retrieve it.

NimaSarajpoor: Should I use `T, m` (instead of `a, w`) for the sake of consistency?

seanlaw: If you don't mind, can you change all of the docstrings (where applicable) to use `a` and `w` instead?

NimaSarajpoor: Sure. I will revise it.

seanlaw:
Alright @NimaSarajpoor, everything checks out for me. Thank you for making all the changes. Please let me know if I can merge

NimaSarajpoor:
@seanlaw 

seanlaw:
@NimaSarajpoor Thank you for contributing this!

PR #789, opened 2023-01-28
Fix #695 Add "subseq_isconstant" param to API

### LINKED ISSUE #695: Add `T_A_subseq_isconstant` and `T_B_subseq_isconstant` Parameters to API
Currently, STUMPY has trouble cleanly differentiating between a "constant subsequence" and a subsequence with a very, very small standard deviation. By adding `T_A_subseq_isconstant` and `T_B_subseq_isconstant` parameters to our API, we can ignore/override the standard deviation completely when computing distances with:

```
if T_B_subseq_isconstant[j] or T_A_subseq_isconstant[i]:
     pearson = 0.5  # Equivalent to a distance of `m`

if T_B_subseq_isconstant[j] and T_A_subseq_isconstant[i]:
    pearson = 1.0  # Equivalent to a distance of `0.0`
```

Doing this would disambiguate between small stddevs and user defined "constant subsequences" and all stddevs can be taken at face value. The only downside is needing to pass around one or two additional vectors but this is a price worth paying as we try to address [other precision-related issues](https://github.com/TDAmeritrade/stumpy/issues/610).

NimaSarajpoor:


stumpy/stump.py
```diff
@@ -551,6 +564,12 @@ def stump(T_A, m, T_B=None, ignore_trivial=True, normalize=True, p=2.0, k=1):
         when k > 1. If you have access to a GPU device, then you may be able to
         leverage `gpu_stump` for better performance and scalability.
 
+    T_A_subseq_isconstant : numpy.ndarray, default None
```
seanlaw, line 567: It's probably a good idea to remind the user that this parameter is ignored when `normalize=False`.

NimaSarajpoor:
@seanlaw 
For the sake of consistency, I think we should add the "subseq_isconstant" param to the following modules as well.

- [ ] stumped
- [ ] gpu_stump
- [ ] floss
- [ ] mpdist
- [ ] stimp
- [ ] mmotifs
- [ ] `mstump.py::multi_distance_profile`
- [ ] scrump (and prescrump)
- [ ] snippets
- [ ] stimp?
- [ ] ....

stumpy/stump.py
NimaSarajpoor: Right! Thanks for pointing this out. I will revise it.

seanlaw:
> For the sake of consistency, I think we should add the "subseq_isconstant" param to the following modules as well.

What about `stumpi`, `mstumped`, `ostinato`, `ostinatoed`, `gpu_ostinato`, `mpdisted`, `gpu_mpdist`, `stimped`, `gpu_stimp`?

NimaSarajpoor:
> What about `stumpi`, `mstumped`, `ostinato`, `ostinatoed`, `gpu_ostinato`, `mpdisted`, `gpu_mpdist`, `stimped`, `gpu_stimp`?

Well... I haven't explored all modules yet but we should defintely check them. I also need to check  the ones I mentioned in my previous comment to make sure the implementation is doable/ reasonable. 

For instance: I have some difficulty in understanding how users can get benefit from this feature when data is updated dynamically. In `stumpi`, matrix profile is computed in the context of an streaming data. While users may provide their own input for `subseq_isconstant` for the initial input `T`, I do not understand how this should be updated as new data being inserted to the time series.

What about PAN matrix profile? I haven't studied its module yet but it seems it computes matrix profile for different window length. So, in that case, I think we should avoid allowing user to insert their own "subseq_isconstant" array for just one window size. Or, we should allow them to provide this array for each window size.

I will try to explore modules one by one to see if we can add this new support for them. Please let me know if you have any suggestion.

seanlaw:
> Please let me know if you have any suggestion.

So, I'm wondering if we could do something like:

```
import inspect

def rolling_isconstant(a, w, custom_func=None):
    """
    """
    axis = a.ndim - 1
    rolling_isconstant_func = _rolling_isconstant

    if custom_func is not None:
        custom_func_args = set(inspect.signature(some_func).parameters.keys()
        if len(custom_func_args.difference(set(['a', 'w']))):
            rolling_isconstant_func = custom_func
        else:
            msg = "Incompatible parameters found in custom function (in `rolling_isconstant`)"
            warnings.warn(msg)

    return np.apply_along_axis(
        lambda a_row, w: rolling_isconstant_func(a_row, w), axis=axis, arr=a, w=w
    )
```

And then, in a function like `stumpy.stump`, we can do something like:

```
if callable(T_subseq_isconstant):
    isconstant_func = T_subseq_isconstant  # save the function in case we need it for later??
    T_subseq_isconstant = core.rolling_isconstant(T, m, isconstant_func)
if T_subseq_isconstant is None:
    T_subseq_isconstant = core.rolling_isconstant(T, m)    
```

I haven't thought this through and it is still somewhat convoluted but some variation of this might work after we clean it up. It should even be usable for `stimp` since the user's custom function would always be applied in place of our `_rolling_isconstant` and it would be dynamic for each window size (i.e., `stimp` would only accept a custom function for `T_subseq_isconstant` and not a numpy array).

Again, just a soft proposal for you to consider.

NimaSarajpoor:
> It should even be usable for `stimp` since the user's custom function would always be applied in place of our `_rolling_isconstant` and it would be dynamic for each window size (i.e., `stimp` would only accept a custom function for `T_subseq_isconstant` and not a numpy array).

This actually sounds great! We can also let user know that subsequences with at least one nan or inf will be treated as "not constant" regardless of the provided custom  function. (Otherwise, we need to modify `T_subseq_isfinite`)

environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - pandas>=0.20.0
   - flake8>=3.7.7
   - flake8-docstrings>=1.5.0
-  - black>=22.8.0
+  - black>=22.1.0
```
seanlaw, line 11: I believe this should be `23.1.0` rather than `22.1.0`, right?

NimaSarajpoor: @seanlaw 
I think the most updated one is `23.1.0`. I was doing some experiment and I realized that I had to install the most updated version to handle [skip-trailing-comma](https://github.com/psf/black/issues/2135). Otherwise, there was a discrepency between the behavior of black between the local and remote branch. I should have chagned this yml file as well since stumpy install `black 23.1.0` in git actions.

stumpy/core.py
```diff
@@ -1843,6 +1888,13 @@ def preprocess(T, m, copy=True, M_T=None, Σ_T=None, T_subseq_isconstant=None):
         A boolean array that indicates whether a subsequence in `T`
         is constant (True)
 
+    isconstant_custom_func : object, default None
+        A custom, user-defined function that returns boolean numpy ndarray that indicate
+        if a subsequence is constant or not. It takes two arguments, `a`, a 1-D array,
+        and `w`, the window size, and may have default arguments if needed. When `None`,
+        this will be default to the function `core._rolling_isconstant`. This parameter
+        does not recompute `T_subseq_isconstant`.
+
```
NimaSarajpoor, line 1897: Is this good?

stumpy/core.py
```diff
@@ -2254,14 +2318,29 @@ def rolling_isconstant(a, w):
     w : numpy.ndarray
         The rolling window size
 
+    custom_func : object, default None
+        A custom, user-defined function that returns boolean numpy ndarray that indicate
+        if a subsequence is constant or not. It takes two arguments, `a`, a 1-D array,
+        and `w`, the window size, and may have default arguments if needed. When `None`,
+        this will be default to the function `core._rolling_isconstant`.
+
```
NimaSarajpoor, line 2326: I think it would be much more useful if users can easily provide a user-defined threshold for stddev. As @seanlaw mentioned before, we may consider type `float` as well. so, we may rename this parameter to something like `isconstant_criteria` that can have type `float` or be a callable object. Since calculating `Σ` is time consumting, we may add `Σ=None` and we can use it to speed up calculation when it is provied.

tests/naive.py
```diff
@@ -6,9 +6,19 @@
...
 
 
-def rolling_isconstant(a, w):
+def ptp_1d(a, w):  # `a` is 1-D
+    return np.ptp(core.rolling_window(a, w), axis=1) == 0
+
+
+def rolling_isconstant(a, w, custom_func=None):
+    if custom_func is None:
+        custom_func = ptp_1d
```
NimaSarajpoor, line 15: maybe I should use lambda for function `ptp_1d` and remove `def ptp_1d(...): ...`

tests/test_core.py
```diff
@@ -1035,6 +1036,19 @@ def test_rolling_isconstant():
     npt.assert_almost_equal(ref, comp)
 
 
+def test_rolling_isconstant_custom_func():
```
NimaSarajpoor, line 1039: Is this test function enough?

NimaSarajpoor (commented):
@seanlaw 
Please feel free to do a review.

stumpy/core.py
seanlaw: What do you mean "This parameter does not recompute `T_subseq_isconstant`"? If it's a function that needs to be called, it should compute `T_subseq_isconstant`, right?

tests/naive.py
seanlaw: It's fine to have `ptp_1d` but I would prefer for it to be more `naive`:

```
def ptp_1d(a, w):
    n = a - w + 1
    out = np.empty(n)
    for i in range(n):
        out[i] = np.max(a[i: i + w]) - np.min(a[i:i + w])
    return out 
```

seanlaw (changes requested):
@NimaSarajpoor I think we're off the mark with regards to using `isconstant_custom_func`. I'd prefer to stick with `T_subseq_isconstant` in the API and allow that argument to be either a ndarray OR a function

NimaSarajpoor:
> I think we're off the mark with regards to using `isconstant_custom_func`. I'd prefer to stick with `T_subseq_isconstant` in the API and allow that argument to be either a ndarray OR a function

Ah...I thought we wanted to avoid allowing the new param accept array as its type. My bad. I will fix it.

tests/naive.py
NimaSarajpoor: Sure. I will revise it.

environment.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - pandas>=0.20.0
   - flake8>=3.7.7
   - flake8-docstrings>=1.5.0
-  - black>=22.1.0
```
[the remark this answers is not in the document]
NimaSarajpoor: Sorry for the incovenience. I will be more careful next time.

stumpy/core.py
```diff
@@ -1488,6 +1490,13 @@ def mass(
     T_subseq_isconstant : numpy.ndarray, default None
         A boolean array that indicates whether a subsequence in `T` is constant (True)
 
+    isconstant_custom_func : object, default None
```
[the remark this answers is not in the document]
NimaSarajpoor: > Is it necessary?

Good question! I think it is necessary. Suppose that I have one `T` and several `Q`. To calculate their distance profile via mass, we can compute the four parameters ` M_T // Σ_T // T_subseq_isfinite // T_subseq_isconstant` first and then pass them to `mass` to speed up the computation of distance profile between `T` and each of those queries. If we use a custom function to compute array `T_subseq_isconstant`, we should then find a way to allow user to use custom function to determine if Q is constant or not. 

What do you think? If you agree, what name do you suggest for this param? how about `Q_subseq_isconstant` (that can be array, which has length one, or function) ?

stumpy/stump.py
```diff
@@ -510,8 +510,20 @@ def _stump(
...
+)
+def stump(
+    T_A,
+    m,
+    T_B=None,
+    ignore_trivial=True,
+    normalize=True,
+    p=2.0,
+    k=1,
+    isconstant_custom_func=None,
```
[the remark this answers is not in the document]
NimaSarajpoor: Sounds great!  I will revise docstings accordingly.

seanlaw:
> Ah...I thought we wanted to avoid allowing the new param to have "array" type. My bad. I will fix it.

In my brain, when it comes to `T_subseq_isconstant`, I want to force the user to do as much of the work/thinking OUTSIDE of STUMPY. The most naive thing they can do is annotate their subsequences by hand and then provide a boolean array as input. I like the array approach the most because it is the less ambiguous and the most explicit. If they don't want to annotate by hand then they should compute the boolean array themselves and then pass it into `stump`. Alternatively, they can hand over that custom computation (a function) into `stump` and have it be executed by `stump` but it is much more limited as their function MUST have a specific function signature. Now, we might not advertise that `T_subseq_isconstant` can be a function and only start off with `np.ndarray` in the docstrings but I think it would be smart to allow it since, as you pointed out, there are places (other functions) that may require applying a function rather than handing over a list of `T_subseq_isconstant` for a list of time series (i.e., ostinato?).

Now that I think about it, I wonder if there is a way to check the type and shape (i.e., bool array) of the output that is returned by their custom function just as a safety measure.

stumpy/core.py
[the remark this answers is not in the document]
seanlaw: Ahh, got it. I failed to realize that this was for `Q`. Yes, please use `Q_subseq_isconstant` then

seanlaw:
@NimaSarajpoor FYI, I updated the black formatting in some of the other files in this commit: 

https://github.com/TDAmeritrade/stumpy/commit/78d3c1249180434b18675e436c2f5fada1920a73

Perhaps, you are able to pull those changes into your branch?

NimaSarajpoor (commented):
@seanlaw 
I addressed the comments, and I left some new comments. I am still waiting to first finalize the changes to `core` and `stump`. After, I will add the new feature to other modules. please feel free to do a review.

NimaSarajpoor:
> I wonder if there is a way to check the type and shape (i.e., bool array) of the output that is returned by their custom function just as a safety measure.

I haven't added this to the code yet; however, I was wondering why not simply do the following? 

```
def rolling_isconstant(a, w, custom_func=None):
     # out = some code
    out.shape = # like a.shape but the last axis now must have the size `a.shape[-1] - w + 1`
```

Regarding the type, I think we can just be more specific and ask user to provide a function that returns numpy boolean array. I think there should be a way to check type and dtype.

NimaSarajpoor:
> Now that I think about it, I wonder if there is a way to check the type and shape (i.e., bool array) of the output that is returned by their custom function just as a safety measure.

Can we wrap the user_defined custom_func as follows?

```

def our_custom_func(a, w, user_custom_func):
       out = user_custom_func(a, w)
       if type(out) != np.ndarry or out.shape != len(a) - w + 1 or ...:
           raise ValueError("msg")
       return out
```

seanlaw:
> @seanlaw I addressed the comments, and I left some new comments. I am still waiting to first finalize the changes to `core` and `stump`. After, I will add the new feature to other modules. please feel free to do a review.

Please allow me some time to review

seanlaw:
> @seanlaw I addressed the comments, and I left some new comments. I am still waiting to first finalize the changes to `core` and `stump`. After, I will add the new feature to other modules. please feel free to do a review.

Please allow me some time to review

stumpy/core.py
```diff
@@ -1859,8 +1941,28 @@ def preprocess(T, m, copy=True, M_T=None, Σ_T=None, T_subseq_isconstant=None):
...
+            msg = (
+                "The acceptable types for `T_subseq_isconstant`"
+                + " are np.ndarray, function, or functools.partial"
+            )
+            raise ValueError(msg)
+
+        if callable(T_subseq_isconstant):
+            isconstant_custom_func = T_subseq_isconstant
+
+    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
```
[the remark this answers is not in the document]
seanlaw: I would keep `T_subseq_isconstant = np.logical_and(T_subseq_isconstant, T_subseq_isfinite)` outside of `core.rolling_isconstant` just to keep `core.rolling_isconstant` "pure". I think switching the boolean array to `False` when a subsequence is `np.inf`/`np.nan` has nothing to do with `rolling_isconsant` and should be handled outside of that function

stumpy/core.py
```diff
@@ -1952,11 +2064,20 @@ def preprocess_diagonal(T, m):
...
     T[np.isnan(T)] = 0
 
     M_T, Σ_T = compute_mean_std(T, m)
     Σ_T[T_subseq_isconstant] = 1.0  # Avoid divide by zero in next inversion step
+    if np.any(Σ_T == 0.0):  # pragma nocover
+        raise ValueError(
+            "The sliding standard deviation of input contains 0.0 at indices"
+            "where T_subseq_isconstant is False. Try to set those indices to"
+            "True in `T_subseq_isconstant`."
+        )
```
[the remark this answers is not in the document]
seanlaw: Yes, warning is better and alert the user as to precisely WHICH subsequence indices were flipped to `True`.

Having said that, I don't understand something. `if np.any(Σ_T == 0.0):` only checks for if any `Σ_T == 0.0` but this does not check that `T_subseq_isconsant == False` at those same indices. So, your error message seems inconsistent with what you are checking. If the user supplies `T = [1.0, 1.0, 1.0, 1.0, 2.0]`, `T_subseq_isconstant=[True, False]`, and `m = 4`, the error message of "where `T_subseq_isconstant` is False" is incorrect.

stumpy/core.py
```diff
@@ -2254,14 +2375,29 @@ def rolling_isconstant(a, w):
...
     Returns
     -------
     output : numpy.ndarray
         Rolling window isconstant.
     """
+    rolling_isconstant_func = _rolling_isconstant
+    if custom_func is not None:
+        custom_func_args = set(inspect.signature(custom_func).parameters.keys())
+        if len(set(["a", "w"]).difference(custom_func_args)):  # pragma: no cover
+            msg = "Incompatible parameters found in `custom_func`"
```
seanlaw, line 2393: We should probably tell the user what the problem is and how to fix it:

```
incomp_args = list(custom_func_args.difference(set("a", "w")))
msg = f"Incompatible arguments {incomp_args} found in `custom_func`. "
msg += "Please provide a `custom_func` with arguments `a`, a 1-D array, and `w`, the window size. "
msg += "The `custom_func` has been ignored and the default `core._rolling_iscontant` function has been applied."
```

stumpy/core.py
```diff
@@ -2254,14 +2375,29 @@ def rolling_isconstant(a, w):
...
+            warnings.warn(msg)
+        else:
+            rolling_isconstant_func = custom_func
+
     axis = a.ndim - 1
     return np.apply_along_axis(
-        lambda a_row, w: _rolling_isconstant(a_row, w), axis=axis, arr=a, w=w
+        lambda a_row, w: rolling_isconstant_func(a_row, w), axis=axis, arr=a, w=w
```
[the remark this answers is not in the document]
seanlaw: I understand. However, I would prefer to keep `rolling_isconstant` pure because the assumption is that the responsibility of this function is to look for constant subsequences. 

Outside of `core.rolling_isconstant`, I would simply add another function called `core.check_isconstant`, which simply does np.logical_and(`T_subseq_isconstant`, `T_subseq_isfinite`) raises a ValueError (describing the precise index where things are bad). In this case, I would rather not save the user and force/train them to fix the problem. I think this is fine since the error is caught very early in the execution process. It would really suck to raise an error after 10 minutes of computing the matrix profile.

So, `core.rolling_isconstant` will blindly return an array if an array is provided or apply a custom function if one is provided (i.e., blindly trust the user). Otherwise, it will apply `core._rolling_isconstant`.
Then, we check it `T_subseq_isconstant` separately.

stumpy/stump.py
```diff
@@ -551,6 +564,26 @@ def stump(T_A, m, T_B=None, ignore_trivial=True, normalize=True, p=2.0, k=1):
...
+        the corresponding value `False` in this boolean array.
+
+    T_B_subseq_isconstant : numpy.ndarray or function, default None
+        A boolean array that indicates whether a subsequence in `T_B` is constant
+        (True). Alternatively, a custom, user-defined function that returns a
+        boolean array that indicates whether a subsequence in `T_B` is constant
+        (True). The function must only take two arguments, `a`, a 1-D array,
+        and `w`, the window size, while additional arguments may be specified
+        by currying the user-defined function using `functools.partial`. Any
+        subsequence with at least one nan/inf value will be enforced to have
```
seanlaw, line 584: See comment above

stumpy/stump.py
```diff
@@ -629,8 +662,9 @@ def stump(T_A, m, T_B=None, ignore_trivial=True, normalize=True, p=2.0, k=1):
            [0.11633857113691416, 0, 0, -1]], dtype=object)
     """
     if T_B is None:
```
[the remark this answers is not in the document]
seanlaw: I wouldn't worry about it. The cost is correlated with the length of the time series but, with longer time series, the majority of the cost is elsewhere.

seanlaw (changes requested):
I only reviewed `core.py` and `stump.py` for now

seanlaw:
> > Now that I think about it, I wonder if there is a way to check the type and shape (i.e., bool array) of the output that is returned by their custom function just as a safety measure.
> 
> Can we wrap the user-defined custom_func as follows?
> 
> ```
> # the code below is to just give us an idea
> 
> def our_custom_func(a, w, user_custom_func):
>        out = user_custom_func(a, w)
>        if type(out) != np.ndarry or out.shape != len(a) - w + 1 or ... :
>            msg = "bad function"
>            raise ValueError("msg")
>        return out
> ```
> 
> Note: I haven't implelemented it yet. However, I think it should be okay to be passed to the `np.apply_along_axis`

That seems too complicated. After `T_subseq_isconstant` is computed, all we need is a simple:

```
if not issubclass(T_subseq_isconstant.dtype.type, np.bool_):  # Check if the type is boolean
    raise ValueError
    # Message should clearly state that the dtype of `T_subseq_isconstant` is {T_subseq_isconstant.dtype} but dtype `np.bool` was expected
```

stumpy/core.py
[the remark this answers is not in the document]
seanlaw: On second thought, I would accept a function called `core.fix_isconstant_isfinite_conflicts` that fixes the conflicts between `T_subseq_isconstant` and `T_subseq_isfinite` and maybe it has the following signature:

```
def fix_isconstant_isfinite_conflicts(T, m, T_subseq_isconstant, T_subseq_isfinite=None):
    if T_subseq_isfinite is None:
        T_subseq_isfinite = rolling_isfinite(T, m)

    fixed = np.logical_and(T_subseq_isconstant, T_subseq_isfinite)
    # Add warning where `fixed != T_subseq_isconstant` and provide indices
    return fixed
```

Something like that would be fine too

stumpy/core.py
```diff
@@ -1659,17 +1704,41 @@ def mass_distance_matrix(
...
-    Q, μ_Q, σ_Q, Q_subseq_isconstant = preprocess(Q, m)
+    Q, μ_Q, σ_Q, Q_subseq_isconstant = preprocess(
+        Q, m, T_subseq_isconstant=Q_subseq_isconstant
+    )
 
     T, M_T, Σ_T, T_subseq_isconstant = preprocess(
-        T, m, copy=True, M_T=M_T, Σ_T=Σ_T, T_subseq_isconstant=T_subseq_isconstant
+        T,
+        m,
+        copy=True,  # TEMP comment: to be consitent with Q, remove this.
```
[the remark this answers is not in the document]
NimaSarajpoor: > "explicit over implicit" is the "Zen of Python"

Cool :) I see your point.

stumpy/core.py
```diff
@@ -1859,8 +1941,28 @@ def preprocess(T, m, copy=True, M_T=None, Σ_T=None, T_subseq_isconstant=None):
     check_window_size(m, max_size=T.shape[-1])
 
     T[np.isinf(T)] = np.nan
-    if T_subseq_isconstant is None:
-        T_subseq_isconstant = rolling_isconstant(T, m)
+
+    isconstant_custom_func = None
+    if T_subseq_isconstant is not None:
+        if type(T_subseq_isconstant) not in {
```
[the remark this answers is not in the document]
NimaSarajpoor: > Note that we should NOT put these checks in the private functions as the private functions should assume that the inputs are correct.

Got it. I will revise the functions and will ask you to check them out.

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: > Having said that, I don't understand something. `if np.any(Σ_T == 0.0):` only checks for if any `Σ_T == 0.0` but this does not check that `T_subseq_isconsant == False` at those same indices. So, your error message seems inconsistent with what you are checking. If the user supplies `T = [1.0, 1.0, 1.0, 1.0, 2.0]`, `T_subseq_isconstant=[True, False]`, and `m = 4`, the error message of "where `T_subseq_isconstant` is False" is incorrect.

```
Σ_T[T_subseq_isconstant] = 1.0  # Avoid divide by zero in next inversion step
if np.any(Σ_T == 0.0):  # pragma nocover
    raise ValueError(...)
```

Note that we first do `Σ_T[T_subseq_isconstant] = 1.0`. So, if there exists a subseq with stddev of zero, but with corresponding value `False` in `T_subseq_isconstant`, its stddev will remain 0. 

I confess that the code should have been more clear. Maybe this:

```
mask_conflicts = np.logical_and(Σ_T==0, T_subseq_isconstant==False) 
if np.any(mask_conflicts):
       # throw warning
       T_subseq_isconstant[mask_conflicts] = True
Σ_T[T_subseq_isconstant] = 1.0  
```

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: Thanks for the suggestion. I think I got the point.

stumpy/core.py
NimaSarajpoor: @seanlaw 
Sure. Two things:
(1) Just to make sure we are on the same page, I think we should do: `set("a", "w").difference(custom_func_args)`. right?

(2) Since the user is (probably) advanced and wants to use a `custom_func`, shouldn't we just throw an error instead of warning? I am just curious and want to understand how you see this situation.

stumpy/core.py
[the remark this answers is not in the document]
seanlaw: Maybe it's just me but if I walk away and look at this again in 6 months, I won't remember what either of those pieces of code are doing and what it is trying to achieve. It's just not obvious. I think the original:

```
Σ_T[T_subseq_isconstant] = 1.0  # Avoid divide by zero in next inversion step
Σ_T_inverse = 1.0 / Σ_T
M_T_m_1, _ = compute_mean_std(T, m - 1)
```

is pretty obvious but the new lines are confusing for me and hard to read

stumpy/core.py
seanlaw: > 1) Just to make sure we are on the same page, I think we should do: set("a", "w").difference(custom_func_args). right?

No, because I actually want to warn the user about the EXTRA unrecognized arguments.

> (2) Since the user is (probably) advanced and wants to use a custom_func, shouldn't we just throw an error instead of warning? I am just curious and want to understand how you see this situation.

Yes, you should throw an ValueError

seanlaw (changes requested):

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
I have been thinking about this and, as you said, the original version is obvious. How about adding a new line to fix `subseq_isconstant` when it has conflict with stddev=0? (This is similar to what you proposed for the conflict between `subseq_isconstant` and `std`)

```
T_subseq_isconstant = fix_isconstant_stddev_zero_conflict(T, m, T_subseq_isconstant, Σ_T)  # NEW
Σ_T[T_subseq_isconstant] = 1.0
Σ_T_inverse = 1.0 / Σ_T
M_T_m_1, _ = compute_mean_std(T, m - 1)
```

And, inside function `fix_isconstant_stddev_zero_conflict`, we check and make sure that `T_subseq_isconstant` is True when `Σ_T==0`.

NimaSarajpoor:
@seanlaw 
I addressed the comments. Please let me know your thoughts regarding this: https://github.com/TDAmeritrade/stumpy/pull/789#discussion_r1118020468

Also, please feel free to checkout the changes. I think we are close to start enhancing other modules by adding the new param. If you are happy with the current changes, I can start working on other modules.

stumpy/core.py
```diff
@@ -2254,15 +2349,108 @@ def rolling_isconstant(a, w):
...
+    """
+    if T_subseq_isfinite is None:
+        T_subseq_isfinite = rolling_isfinite(T, m)
+
+    fixed = np.logical_and(T_subseq_isconstant, T_subseq_isfinite)
+
+    conflicts = fixed != T_subseq_isconstant
+    if np.any(conflicts):
+        msg = (
+            "Found indices where T_subseq_isconstant is True for subsequence "
```
seanlaw, line 2446: I think this message can be more concise with something like:

f"Subsequences located at indices {np.nonzero(conflicts)} contain one or more `np.nan`/`np.inf` and so their corresponding values in `T_subseq_isconstant` have been automatically switched from `True` to `False`.

seanlaw (changes requested):
@NimaSarajpoor Let's start with these suggestions/comments

stumpy/core.py
```diff
@@ -2254,15 +2349,108 @@ def rolling_isconstant(a, w):
     w : numpy.ndarray
         The rolling window size
 
+    custom : np.ndarray or function, default None
```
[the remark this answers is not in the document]
NimaSarajpoor: I agree. I avoided `custom_func` to reflect that the param is not necessarily a funciton. But, your suggestion together with what you proposed below sound better :+1: 

I will revise the name and the code.

stumpy/core.py
NimaSarajpoor: Thanks for the feedback. I will revise it,

stumpy/core.py
```diff
@@ -2254,15 +2349,108 @@ def rolling_isconstant(a, w):
...
+        None, this defaults to `_rolling_isconstant`.
+
     Returns
     -------
-    output : numpy.ndarray
-        Rolling window isconstant.
+    T_subseq_isconstant : numpy.ndarray
+        Rolling window isconstant
+    """
+    if custom is None:
```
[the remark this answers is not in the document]
NimaSarajpoor: I think I can understand your mindset. You are trying to first do some sort of preprocessing, and then start computing the ouput if needed (i.e. when the input is not `np.ndarray`). This sounds great. 

I have a couple of comments:

(1) This very first line in the code, i.e. 

> isconstant_func = _rolling_isconstant

should be `T_subseq_isconstant = _rolling_isconstant`. Right? Because, if input `T_subseq_isconstant` is None, then the current code raises an error.

(2) I couldn't find an easy way to find the arguments of function with no default value. Currently, as you can see in the code, I am doing:

```
custom_args = []
        for arg_name, arg in inspect.signature(custom).parameters.items():
            if arg.default == inspect.Parameter.empty:
                custom_args.append(arg_name)
```

Is there an easier way?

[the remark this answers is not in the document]
seanlaw: > I think I can understand your mindset. You are trying to first do some sort of preprocessing, and then start computing the ouput if needed (i.e. when the input is not np.ndarray). This sounds great.

Yes, I think one can get too "smart" by trying to fit everything into nested if/else statements. However, this becomes very hard to read/follow. Instead, what I've done is split things up into logical parts. By analogy, when you are an expert chef, you can certainly boil the water, chop the veggies while the water is boiling, and then cook the pasta at the same time and be super efficient in order to save yourself 3 minutes. However, when you're providing instructions for somebody else to read, it's better to chop the veggies first, then boil the water, and then finally cook the pasta all in separate steps as it will be easier to follow and, more importantly, easy for ANYBODY to read the recipe in the future.

> should be T_subseq_isconstant = _rolling_isconstant. Right? Because, if input T_subseq_isconstant is None, then the current code raises an error.

You're right, I missed the `T_subseq_isconstant is None` case. Perhaps, it should start with:

```
if T_subseq_isconstant is None:
    T_subseq_isconstant = _rolling_isconstant
```

> I couldn't find an easy way to find the arguments of function with no default value.

I'm not sure why you are looking "for arguments of function with no default value". What is wrong with what I proposed?

```
func_args = list(inspect.signature(T_subseq_isconstant).parameters.items())
incomp_args = set(func_args).difference({"a", "w"})
```

The goal is simply to check whether the signature of the supplied function has exactly two arguments (no more, no less) and that they are named `a` and `w`. I think my proposed code does exactly that. I'm not sure what that has to do with `default value`. Can you please elaborate?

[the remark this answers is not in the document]
seanlaw: Using `.keys` should work for all of the following functions

```
def some_func(a, w):
    return


def some_func(a=10, w):
    return

def some_func(a=10, w=20):
    return
```

In all three cases, `list(inspect.signature(T_subseq_isconstant).parameters.keys())` would return `[a, w]`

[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
Unfortunately, this does not work when `functools.partial` is used. Maybe I am doing something wrong, but this is what I get:

```
import functools
import inspect

def custom_func(a, w, extra_param):
    return 

f = functools.partial(custom_func, extra_param=None)

params_name = list(inspect.signature(f).parameters.keys())
print(params_name)

# output
['a', 'w', 'extra_param']
```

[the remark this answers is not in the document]
seanlaw: @NimaSarajpoor Thanks for the explanation and it looks like your proposal solves the problem. I think it would help to explain the functools.partial case as a comment.

This is basically the same as your solution and I'm not sure if it's better/worse:

```
func_args = [k for k, v in inspect.signature(f).parameters.items() if v.default == inspect.Parameter.empty]
``` 

I think your approach is a little more readable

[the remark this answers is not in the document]
seanlaw: Let's make sure to add a `functools.partial` example in our unit test. This will protect us in the future if we accidentally change the argument checking logic (i.e., the `functools.partial` test would fail)

stumpy/core.py
NimaSarajpoor: It is already added to the unit test. In fact, that resulted in raising an error, which in turn helped me with solving it.

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: It is already added to the unit test. In fact, that resulted in raising an error, which in turn helped me with solving it.

stumpy/core.py
```diff
@@ -2254,15 +2349,114 @@ def rolling_isconstant(a, w):
...
-    axis = a.ndim - 1
-    return np.apply_along_axis(
-        lambda a_row, w: _rolling_isconstant(a_row, w), axis=axis, arr=a, w=w
-    )
+    if T_subseq_isconstant is None:
+        T_subseq_isconstant = _rolling_isconstant
+
+    isconstant_func = None
+    if callable(T_subseq_isconstant):
+        non_default_args = []
```
seanlaw, line 2371: I don't know if we need/want to do it now but it feels like this argument comparison code would be potentially useful for other `core.rolling_func` functions and maybe it would be nice to pull it out into it's own function:

```
def is_signature_compatible(func, required_args):
    if not isinstance(required_args, list):
        required_args = list(required_args)

    non_default_args = []
    for arg_name, arg in inspect.signature(T_subseq_isconstant).parameters.items():
        # inspect.signature(functools.partial(func)) returns all arguments
        # including the ones with default values. the following if block
        # is to find non-default arguments.
        if arg.default == inspect.Parameter.empty:
            non_default_args.append(arg_name)

        incomp_args = set(non_default_args).difference(set(required_args))

    return len(incomp_args) > 0
```

seanlaw (changes requested):
@NimaSarajpoor Left one comment but everything looks good for the most part and likely ready to move forward.

stumpy/core.py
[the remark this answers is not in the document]
seanlaw: Awesome! I had missed that since I had not reviewed the tests yet. You're a pro!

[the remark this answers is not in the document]
NimaSarajpoor: Thank you for your kind words. Still having a long way to get there :) but I am pro in enjoying the journey 😀

stumpy/core.py
NimaSarajpoor: Right! I think that would be a good idea!! I will add it to the `core.py`

NimaSarajpoor:
@seanlaw 

> @NimaSarajpoor Left one comment but everything looks good for the most part and likely ready to move forward.

Awesome...I will start working on the updating the other modules.

stumpy/core.py
```diff
@@ -3343,3 +3529,36 @@ def _client_to_func(client):
     func = getattr(module, prefix + calling_func)
 
     return func
+
+
+def find_incompatible_args(func, required_args):
```
seanlaw, line 3534: Maybe make this a private function?

seanlaw (commented):
Added one minor suggestion. Also, I wonder if it makes sense to break this up into smaller individual PRs rather than one giant one? The current size of this PR is okay but maybe we merge this one when it is ready and then add other files in a separate PR?

stumpy/core.py
NimaSarajpoor: Right! I think this should be private. I will revise it. Thank you for the suggestion!

NimaSarajpoor:
@seanlaw 
> I wonder if it makes sense to break this up into smaller individual PRs rather than one giant one? The current size of this PR is okay but maybe we merge this one when it is ready and then add other files in a separate PR?

According to our experience in top-k PR, I think what you are suggesting is reasonable. I checked out the changed files and I think this PR is ready. We already added the param to `stump`, `stumped`, and `gpu_stump`. So, I think it is good to be merged. Please allow me to address your comment and take a look at the changes.

NimaSarajpoor:
@seanlaw 
> Please allow me to address your comment and take a look at the changes for one last time.

**[Update]**
I addressed your comment, and checked the changed files. They look good to me. Please feel free to merge.

seanlaw:
@NimaSarajpoor It looks like we are missing some code coverage:

```
Name                 Stmts   Miss  Cover   Missing
--------------------------------------------------
tests/naive.py        1216      1    99%   243
tests/test_core.py     993      7    99%   89, 1577, 1583, 1589, 1595, 1601, 1607
--------------------------------------------------
TOTAL                13037      8    99%
```

Note that even though these are are `naive.py` and `test_core.py`, this implies that some paths are not traversed within these functions, which is a problem (i.e., please do not simply do `pragma no cover`)

NimaSarajpoor:
@seanlaw 
> It looks like we are missing some code coverage

I really need to understand the importance of checking code coverage by heart :)

tests/naive.py
```diff
@@ -190,13 +221,26 @@ def stump(T_A, m, T_B=None, exclusion_zone=None, row_wise=False, k=1):
...
     distance_matrix[np.isnan(distance_matrix)] = np.inf
+    for i in range(distance_matrix.shape[0]):
+        for j in range(distance_matrix.shape[1]):
+            if np.isfinite(distance_matrix[i, j]):
+                if T_A_subseq_isconstant[i] and T_B_subseq_isconstant[j]:
+                    distance_matrix[i, j] = 0.0
+                elif T_A_subseq_isconstant[i] or T_B_subseq_isconstant[j]:
+                    distance_matrix[i, j] = np.sqrt(m)
+                else:
+                    pass
```
NimaSarajpoor, line 243: It seems that `else: pass` is treated as skipped coverage in spite of the fact that the code goes through these lines. I wanted to avoid `if / elif` structure and wanted to include `else` as well. What do you think @seanlaw ?

tests/test_core.py
```diff
@@ -1536,3 +1550,65 @@ def test_gpu_searchsorted():
...
+
+    comp = core.fix_isconstant_isfinite_conflicts(T, m, T_subseq_isconstant)
+
+    npt.assert_almost_equal(ref, comp)
+
+
+def test_find_incompatible_args():
+    # case1: having exact required argument
+    def func_case1(x, y):
+        return
```
NimaSarajpoor, line 1577: This function and other functions below it are designed to test the functionality of `core._find_incompatible_args`. However, since we do not call these functions (right?), these functions are skipped according to the result shown in code coverage. Any suggestion @seanlaw ?

seanlaw:
> I really need to understand the importance of checking code coverage by heart :)

First, I recommend running the tests locally first for non-trivial PRs. Having said that, I'm going to add something to our coverage reporting to force it to fail if the coverage is below 100%. Hopefully, that'll help. I should've done it a long time ago

NimaSarajpoor:
> First, I recommend running the tests locally first for non-trivial PRs.

Right! Need to keep that in mind!

> I'm going to add something to our coverage reporting to force it to fail if the coverage is below 100%.

Cool!! I think that would be a great idea!

seanlaw:
@NimaSarajpoor I just pushed a new commit that I think/hope will cause a failure. Would you mind pulling it into this branch?

NimaSarajpoor:
That was quick :) I will update my branch.

seanlaw:
Please pull the latest commit (the last one wasn't enough).

NimaSarajpoor:
@seanlaw 
I ran the test on google colab, and I got this:

```
Name                 Stmts   Miss  Cover   Missing
--------------------------------------------------
tests/naive.py        1216      1    99%   243
tests/test_core.py     993      7    99%   89, 1577, 1583, 1589, 1595, 1601, 1607
--------------------------------------------------
TOTAL                13037      8    99%

78 files skipped due to complete coverage.
Cleaning Up
```
I am going to push the commits...

NimaSarajpoor:
We got error. That is good. I will wait till you handle the error occured in your last commit (You may want to see this: https://github.com/nedbat/coveragepy/issues/198)

tests/naive.py
seanlaw: In this case, I accept `# pragma: no cover`

seanlaw:
> We got error. That is good. I will wait till you handle the error occured in your last commit (You may want to see this: [nedbat/coveragepy#198](https://github.com/nedbat/coveragepy/issues/198))

It should be fixed now.

tests/test_core.py
NimaSarajpoor: @seanlaw 
FYI: To fix code coverage, I added `# pragma: no cover` here and for the next few functions.

NimaSarajpoor:
@seanlaw 
Please let me know if I should take care of anything else for this PR.

seanlaw:
@NimaSarajpoor Everything looks good here. Merging now. Thanks!

Issue #465, opened 2021-09-08
Add Contrast Profile Tutorial


seanlaw:
This paper is [here](https://www.cs.ucr.edu/~eamonn/ContrastProfile_V048_names.pdf)

The supporting site is [here](https://sites.google.com/view/contrastprofile)

The main goal of a contrast profile is to annotate each subsequence of a time series so that you can discover if there are "good" subsequences that can be used for time series classification (see Definition 6 in the paper above). In order to extract these subsequences, you need to have 2 time series. One time series, <img src="https://render.githubusercontent.com/render/math?math=T^{(%2B)}">, and another time series, <img src="https://render.githubusercontent.com/render/math?math=T^{(%2D)}">. It is recommended that you compute the the Contrast Profile only when you believe that the two following assumptions are likely to be true:

* <img src="https://render.githubusercontent.com/render/math?math=T^{(%2B)}"> contains at least two behaviors that are unique to the phenomena of interest.
* <img src="https://render.githubusercontent.com/render/math?math=T^{(%2D)}"> contains zero behaviors of interest.

JaKasb:
The following notebook reproduces the PDF which is contained in "ContrastProfile_GettingStarted.zip"
[contrast_profile_notebook.pdf](https://github.com/TDAmeritrade/stumpy/files/8825896/contrast_profile_notebook.pdf)

seanlaw:
Thanks @JaKasb! This is a great start. Would you care to contribute a full tutorial for this?

JaKasb:
After re-reading the paper, I have come to the conclusion that I can't imagine a specific topic that needs an extra tutorial.
The paper already explains online-updates, k-Platos, and comparisons of Contrast-Profile vs Shapelets/HMM/LSTM.

seanlaw:
In case it matters, our tutorials generally reproduce one or more of the figures in the paper and tries to elaborate on any insights/thoughts beyond what was provided in the paper that might not immediately be obvious to non-experts. Sometimes, reading the same concepts but explained in different ways may help others digest the information. That's usually the goal of our STUMPY tutorials. Code is one component but educating users on how to think about and interpret their results is equally as important. It may help provide more perspective.

ken-maeda:
Is there still demand to reproduce this paper? Basic idea looks not complicated.

seanlaw:
> Is there still demand to reproduce this paper? Basic idea looks not complicated.

[...]

PR #800, opened 2023-03-02
Add Contrast Profile Tutorial

ken-maeda:
# Pull Request Checklist

#465 reproducing the paper for tutorial.

- [x] Create a new branch


seanlaw:
@ken-maeda Thank you for this contribution. Please allow me some time to review

ken-maeda:
<p>v_query is typical norml ecg query we can see everywhere in dataset. so it is picked up one randomly.</p><p> </p><p>The purpose of this distance profile is finding desired behavior by just comparing v_query(typical ecg signal) with desired behavior(rare signal). As assumption, it could be highest in distance profile. But it didn't happen. </p><p> </p><p> </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/#comment-1457997133'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ken-maeda:
<p>I add more descriptino to </p><h2><br></h2><h2>Contrast Profile </h2><p> </p><p>The subsequence in 𝐓(+) corresponding to the highest point in the Contrast Profile is called the <strong>Plato</strong>.</p><p> </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/#comment-1457997158'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ken-maeda:
I appriciate your feedback, I fixed those.

NimaSarajpoor:
@ken-maeda 
I have a suggestion for you. 

I think it is better to develop the notebook section by section. So, for each section that you add, you can wait to get some feedback and then apply those, and then after getting the green light, you can move forward. Right now, you may know what is going on in your notebook, however, the main goal is to make sure the reader can understand what is going on! 

You can keep the current notebook somewhere in your local pc. Then, you can start again by just providing the first section or a couple of sections. Currently, if I do not understand a part of your notebook, I try to read other parts to better understand the concept. However, this is not desirable. I think this is a red flag. There should be a flow in your tutorial and I believe each segment should be understandable on its own.

NimaSarajpoor:
**Regarding contrast profile, this is how I see it:**

we can see each subsequence of length `m` as a data point in $R^{m}$ space. For the sake of visualization, let's illustrate the problem in 2D space. 

First, let's review the definitions of `T(+)` and `T(-)`.

```
T(+) : contains at least two instances that are unique to the phenomena of interest.
T(-) : contains no instances of interest (and instead, I think we should say it contains the regular, obvious patterns in `T(+)`)
```

Before we talk about `T(-)` and `T(+)`, it is better to just talk about `T`. Let's assume the figure below shows the subsequences of `T`. 


image


If I look at this data, I can see that the regular, obvious pattern is where the crowded part is. However, the motif we might be interested in can be the motif pair `(A, B)`. Note that this motif pair may not be easily captured as their distance is greater than the distance of any other point and its nearest neighbour. 

So, what can we do? We can create `T(-)` which just contains the regular behaviour of our the Data, and left the remaining ones as `T(+)`.


image


Now, we can see that the `d = dist(A, A_nn_in_Tneg) - dist(A, A_nn_in_Tpos)` has a high value. Let's call it contrast distance. The contrast profile, `cp`, is an array where `cp[I]` is the contrast distance that corresponds to the `i-th` subsequence in `T(+)`. The peak of this contrast profile can reveal the motif pair `(A, B)`.

---

Question: Can we see it as twin-freak problem? In other words, this might be an anomaly that appears more than once. So, we can easily find the motif pair `(A, B)` by finding the  subsequence that has the greatest distance to its 2nd nearest neighbour. 

Answer: I do not know! I think a good way to investigate this is to get some data and find a pair of subsets using each of these two approaches: (1) twin-freak (2) contrast profile, and see if they result in different outcomes.

ken-maeda:
@NimaSarajpoor 
I'm sorry for causing trouble, and greatful your kind guidance. I uploaded new notebook first section of `Tutorial_Contrast_Profile2.ipynb`. I should have considered the contrast profile concept more.

I added its scatter plot should best to understand the "constrat" concept to the notebook. 

> T(+) : contains at least two instances that are unique to the phenomena of interest.
> T(-) : contains no instances of interest (and instead, I think we should say it contains the regular, obvious patterns in `T(+)`)

This might be tricky precondition, the robustness for this precondition also is argued. I thought the notebook should be enoguh only for expalining contrast profile conceptl.

docs/Tutorial_Contrast_Profile2.ipynb
```diff
@@ -0,0 +1,249 @@
+{
```
NimaSarajpoor, line 1: <blockquote>Small values in the Matrix Profile are called <em>motifs</em>, and large values are called <em>discords</em>.</blockquote><p> </p><p>I can see that this sentence is from the introduction part of the paper; however, I think it is not meaningful to call small value in matrix profile, a motif. Because a<strong> motif is a subsequence</strong> that has small value to its nearest neighbour.</p><p> </p><p>I think the authors provided a better definition in the abstract of the paper:</p><p>"<strong>Time series motifs refer to two particularly close subsequences, whereas time series discords indicate subsequences that are far from their nearest neighbors.</strong>"</p><p> </p><p> </p><blockquote>It may be usefull to score subsequences with</blockquote><p>I think the paper is clearer here. In the paper, it says: "....score subsequences with a <strong>meta-data that reflects that ...</strong>"</p><p> </p><p>Also, I think it would be a good idea to introduce <code>T(+)</code> and <code>T(-)</code> to the reader.</p><p> </p><blockquote>This is exactly the property we desire.</blockquote><p> </p><p>Why? According to my understanding and what I read in the paper, this property can be used to find a subsequence that can uniquely identify a class. In other words, this can be used in classification. Is that correct? I think it would be nice to provide an example to explain the significance of contrast profile. I know there is an example in paper (see Fig. 1), but I think it is a little bit complicated. So, let's try and see if we can come up with a good example to show the importance of contrast profile. Think about it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

NimaSarajpoor, line 1: <blockquote>Let's walk through by running example of a noisy electrocardiogram(ECG).  </blockquote><p> </p><p>Maybe remove this line. </p><p> </p><blockquote>We proposed to compute the Contrast Profile only when we belive that the two following assumptions are likely to be true: </blockquote><p> </p><p>Before we start talking about the contrast profile, I think it would be a nice idea to show the data and talk about the problem. What is the problem? and why should we care about that problem?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

NimaSarajpoor, line 1: <p>Do we use column 0 later? If not, then it might be better to just read the csv for that column only. So, we can do:</p><pre class="ql-syntax" spellcheck="false">T = pd.read_csv(..., usecols=["1"]).to_numpy(np.float64)
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

NimaSarajpoor, line 1: <p>Is there a particular reason behind using the name <code>v</code> here? Can we use <code>T</code> instead? Like... <code>T_pos</code> and <code>T_neg</code> ?</p><p> </p><p> </p><p>ALSO: How do we know the indices 63630, 68129, etc? This can create a confusion for readers. Reading those indices made me think that using contrast profile requires us having some prior knowledge about those indices. If that is not the case, then can you provide a brief explanation on how one can create the datasets <code>T(+)</code> and <code>T(-)</code> in a real-world problem?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

NimaSarajpoor, line 1: <p>(1) Why are they desirable? Can we add some explanation as why one might be interested in finding those patterns?</p><p> </p><p>(2) The code says <code>label="desired instances")</code> However, the figure's legend says "desired behaviour"</p><p> </p><p>(3) Right after this block, it might be useful to compute matrix profile, and discover motif and discord, and show that the discovered motifs/discords do not reveal the patterns we are looking for. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

NimaSarajpoor, line 1: <blockquote>T(+) looks noisy, 𝐓(+) has to include contains  at  least  two  instances</blockquote><p> </p><p>Please rewrite it as follows: <code>T(+) looks noisy. It contains at least two instances that are of our interest.</code> </p><p> </p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

NimaSarajpoor:
> @NimaSarajpoor I'm sorry for causing trouble, and greatful your kind guidance.

No need to be sorry. I provided a few comments. Let's start with those. Please do not add any new section. Let's take care of the current sections first. Please feel free to discuss something if you feel there is a need for that.

seanlaw:
@ken-maeda Instead of uploading a `.png` file, I would prefer if you could add the code that could create/recreate the image and have it inside the notebook

docs/Tutorial_Contrast_Profile2.ipynb
ken-maeda: <p>FIx</p>

ken-maeda: <p>1) add discription </p><p>2) fixed</p><p>3) add matrix profile section</p>

ken-maeda: <blockquote>Is there a particular reason behind using the name&nbsp;v&nbsp;here? Can we use&nbsp;T&nbsp;instead? Like...&nbsp;T_pos&nbsp;and&nbsp;T_neg&nbsp;</blockquote><p>Fixed</p>

ken-maeda: <p>I modified the dataset itself, it was originally exstracted dataset.</p>

ken-maeda: <blockquote>Before we start talking about the contrast profile, I think it would be a nice idea to show the data and talk about the problem. What is the problem? What are we trying to find here?</blockquote><p>Add simple example in the beginning</p>

docs/Tutorial_Contrast_Profile2.ipynb
```diff
@@ -0,0 +1,474 @@
...
```
NimaSarajpoor, line 1: <blockquote><strong>If two or more discords in same dataset are, what can we do?</strong></blockquote><p> </p><p>I think this sentence is ungrammatical. The verb of the <code>if clause</code> is "are". Hence, you may revise it as follows: "If two or more discords <strong>are</strong> in a dataset, what can we do?"</p><p> </p><p>But, still, this feels weird to me. we may re-write it as:</p><p>"If a dataset has more than one discord, what can we do?"</p><p> </p><p><strong>When you ask the question "what can we do?", it is still <em>not</em> clear. </strong>Because I do not know what the goal is at this point. For instance: "If weather is not good, what can we do?" is not clear because it does not show the main goal and its challenge here. It can be anything... for example (1) Do we have a flight and because of weather our flights got cancelled? In this case "what can we do?" means "how can we adjust our trip?" (2) Do we have a plan to play soccer outside? Then "what can we do?" can mean what other games we can play in this weather (maybe indoor games). (3) etc.</p><p> </p><p>As you can see, a story should be clear. So, I think you should provide further explanation here.</p><p> </p><p> </p><blockquote>Let's demonstrate it in ECG data with anormaly behavior.</blockquote><p> </p><p>can you help me understand what <code>it</code> is referring to here? is that challenge?</p><p> </p><p>Also: I think <code>anomaly</code> is not adjective.  It is a noun. So, please remove the term behavior.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

NimaSarajpoor, line 1: <p>Please put the first six lines in a seperate cell.</p><p> </p><blockquote>type(flsoat)</blockquote><p> </p><p>Does this work? I think there is a typo here, and it should have been <code>float</code> . right? I think you should run the cells one last time before pushing them to this PR to make sure everthing is working properly.</p><p> </p><blockquote>ecg_demo_df</blockquote><p> </p><p>Are we using this later? If not, then a simple name like <code>df</code> should suffice.</p><pre class="ql-syntax" spellcheck="false">df = pd.read_csv(...).astype(...)
T_m = df.values[:,0]
T_p = df.values[:,1] 
</pre><p> </p><p>What is <code>m</code> in <code>T_m </code> ? Did you mean <code>T_n</code> (because of the word <code>negative</code>) If yes, then please use <code>T_neg</code> and <code>T_pos</code> . Having said that, I think we should avoid positive / negative thing and just use <code>T </code> since we haven't discussed pos/neg yet. So, based on what I can see in the next few cells, it seems that we are not using <code>T_m </code>. So, maybe we should just do:</p><pre class="ql-syntax" spellcheck="false">T = df.values[:,1]  
</pre><p> </p><p>Then, discuss the problem and the challenge on this data <code>T</code>. </p><p> </p><blockquote>MP_PP1 = stumpy.stump(T_p[:2000], m, normalize=False)</blockquote><p> </p><p>Can we just use the name <code>mp</code> ?    </p><p> </p><blockquote>discord_idx1 = np.argmax(MP_PP1[:, 0])&nbsp;# disccords</blockquote><p>Can we just say: <code>discord_idx </code> ? Also, there is a typo in your comment: <code>disccords</code> has one <code>c</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

NimaSarajpoor, line 1: <p>Earlier, you mentioned: </p><p> </p><blockquote>Matrix Profile on whole dataset which has&nbsp;<strong>2 similar anomaly behaviors</strong>.</blockquote><p> </p><p>However, you haven't shown result for whole dataset yet. And, then, here, you start explaining <code>T(+)</code> and <code>T(-)</code> . If you think defining T(+) and T(-) is related to the first part, i.e. "Matrix Profile on the first half of dataset which has&nbsp;<strong>1 anomaly behavior</strong>.", then you may create a subsection with an appropriate title. If you think you should talk about 1 and 2 first, then you should have first provided the  results for <code>Matrix Profile on whole dataset</code> , before the definiton of T(+) and T(-).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

NimaSarajpoor:
@ken-maeda 
I haven't reviewed the whole notebook yet. Let's start with the current comments and some of the previous comments if you haven't addressed them yet.

Again, if you are in doubt, or you think there is a need to discuss something, please feel free to discuss it. When you are done, please let me know. 

---
Also:
Please pull the latest changes to your branch.

ken-maeda:
I fixed these typos and  variables name. Then I rearranged the order I explain in introduction.

docs/Tutorial_Contrast_Profile2.ipynb
```diff
@@ -0,0 +1,511 @@
...
```
NimaSarajpoor, line 1: <blockquote>... patterns(PVCs) ...</blockquote><p> </p><p>... patterns (PVCs) ... [Notice the space I added]</p><p> </p><p> </p><p>Also, what is PVC? should reader know about this? If not, then you can/should remove it. If yes, then you should avoid using abbreviation unless you are going to use it again later. Even in such case, you should first provide the original phrase. </p><p> </p><blockquote>Those are target we try to find. </blockquote><p> </p><p>Those are the targets we try to find.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    discord_nn_idx = mp[discord_idx, 1]  # discords nearest neighborhood index
</pre><p> </p><p>Regarding  the comment: <code>discord nearest neighbor index</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

NimaSarajpoor, line 1: <p>Great job in providing a clear plot!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

NimaSarajpoor, line 1: <blockquote>We can find clealy the discord by Matrix profile. It indicate the discord in Matrix profile by far. </blockquote><p> </p><p>Maybe: In this case, the discord, indicated by matrix profile, is what we are looking for.</p><p> </p><p>What do you think? Because, any matrix profile with a unique global maxima can reveal a unique discord. Whether or not that discord is what we are looking for is a different story. Here, we use matrix profile to find discord, and we are lucky that the discovered discord is what we are looking for.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

NimaSarajpoor, line 1: <blockquote>Matrix profile indicates slided wrong place as discords.</blockquote><p> </p><p>As I explained in my comment above, I think we should rephrase it. We may say: "The discord discovered by matrix profile is not what are looking for."</p><p> </p><p>I am just trying to avoid using the word "wrong" here. In my head, contrast matrix profile is another tool that can be used to reveal something. Wether that thing is what we are loooking for or not depends on our perspective and our goal. In this case, we will probably see that contrast abstract profile can reveal something that is not captured by matrix profile. But, is that always useful? Well, that depends on the domain. In other words, a user does not know if the discord discovered by matrix profile is useful. The same goes for contrast matrix profile. A user can compute both and then investigate their outcome to see which reveals better information about the data.</p><p> </p><blockquote>ost important thing is Matrix profile also doesn't indicate remarkable distance for the discord.</blockquote><p> </p><p>How can one know if it is a remarkable distance for discord or not? I think this sentence is a little bit biased. This is because we already saw the matrix profile computed for the first half of <code>T</code>and we now feel the maximum distance shown in figure above is not large enough. So, I think we should remove this sentence as we prefer to not be subjective in our analysis.</p><p> </p><p> </p><blockquote>=&gt; Because similar anomalies are nearest neighbor relationship.</blockquote><p> </p><p>Maybe we should say this first:</p><p> </p><p>"Because the targets we are looking for are similar to each other, they can be nearest neighbor of each other and hence their corresponding distance in matrix profile will be small. Therefore, none of them will be detected via matrix profile"</p><p> </p><blockquote>We find those similar anomalies from the dataset with those following specification.</blockquote><p> </p><p>Now, I think this is the place that we should do our best and provide a clear explanation. We can say: "To find the desirable anomalies, we should first understand what properties distinguish them from the other patterns. Note that these two anomalies are similar to each other but they are dissimmilar to other (regular) patterns. This is the main concept behind the contrast matrix profile."</p><p> </p><p> </p><blockquote>This is exactly the property we desire.&nbsp;In other words, we need to prepare two following data.</blockquote><p> </p><p>Then you can create a section here and name it "Contrast matrix profile". Then, we say:</p><p>"To compute the contrast matrix profile, two time series data are needed as follows: "</p><p> </p><p> </p><blockquote>Take a look at the plot below to confirm the condictions</blockquote><p> </p><p>Are we showing any important information here? If not, then we can just remove this line and the code you have provided for plotting the two time series <code>T_p</code> and <code>T_n</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

NimaSarajpoor:
@ken-maeda 
I reviewed up to the section `Loading the ECG data for Contrast Profile`. Whenever you address a comment, you can go to `ReviewNB` (see the top of this PR) and then click on "Resolve Conversation" whenever you are done with that comment.

docs/Tutorial_Contrast_Profile2.ipynb
ken-maeda: I removed the abbreviation.

ken-maeda: <p>I agree with you</p>

ken-maeda: <p>I replaced sententce with your suggested contents.</p>

ken-maeda:
@NimaSarajpoor I changed to separate last plot(maybe redundant?), as you mentioned, it was crowded it is hard to recognized where is indicated. I fixed the last sentence also.

docs/Tutorial_Contrast_Profile2.ipynb
```diff
@@ -0,0 +1,498 @@
...
```
NimaSarajpoor, line 1: <p>Please change</p><p> </p><p> </p><blockquote>If a dataset has similar <strong>more than one discords</strong>, we should know how those discords are calculated in Matrix Profile.  </blockquote><p> </p><p>to:</p><p> </p><p>If a dataset has two similar subsequences that are far from the rest of subsequences, they cannot be easily discovered by matrixprofile.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

NimaSarajpoor, line 1: <p>Let's see the entire dataset first.</p><p> </p><p> </p><p> </p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

NimaSarajpoor, line 1: <p>Maybe remove the line <code>Define index of desired instances on the purpose</code>?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    desired0_idx, desired1_idx = 550, 2030
</pre><p>Please move this line to the top of the next code cell, where you show the desired instances. Also, please add one blank line after it for sake of readability.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

NimaSarajpoor, line 1: <p>(1) Is there extra space between the first two sentences?</p><p>(2)<code>According to...,</code> note that I replace the dot with comma</p><p>(3) please modify the second sentence as follows:</p><pre class="ql-syntax" spellcheck="false">.... of premature venticular contractions, with start index of 580 and 2030 (see figure below).
</pre><p> </p><p>(4) please modify the third sentence as follows:</p><pre class="ql-syntax" spellcheck="false">Those are the two instances we would like to discover.
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

NimaSarajpoor, line 1: <p>please replace upper with <code>top</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

NimaSarajpoor, line 1: <p>I think you should replace the window with a vertical line in the matrix profile figure. For example, see the section  <code>Find potential anomalies (discords) using stump</code> in this page: https://stumpy.readthedocs.io/en/latest/Tutorial_STUMPY_Basics.html</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

NimaSarajpoor, line 1: <p>Please replace this sentence with:</p><p> </p><p>"As shown in this figure, the motifs discovered by the matrix profile (of <code>T_p</code>) does not reveal our two desirble subsequences, the ones that were shown in the previous figure. "</p><p> </p><p> </p><p>I think this can help readers understand why we need to use contrast profile technique later to discover the two desirbale subsequences.</p><p> </p><p>Please revise other figures accordingly if needed.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

NimaSarajpoor, line 1: <p>Cool. I can better understand this figure. I think what you are trying to say is that the discord discovered by matrix profile is not what we are looking for. I suggest to do the following instead:</p><p> </p><p>1 - compute the discord index and its nearest neighbor using matrix profile.</p><pre class="ql-syntax" spellcheck="false">idx = np.argmax(MP_PP[:,0])  # start index of discord
nn_idx = MP_PP[:,1]  # start index of nearest neighbor of discord
</pre><p> </p><p> </p><p>2- show two figures only,  and both of them should be <code>T(+)</code>. In one, you plot <code>T(+)</code>   and show the discord and its nearest neighbor subsequences, according to the computed <code>idx</code>  and<code>nn_idx</code>. In another figure, you plot <code>T(+)</code> again but you show  the desirbale subsequences instead. </p><p> </p><p>Please choose a proper title for each figure. For the first figure, we may say: "the discord and its nearest neighbor discovered by the matrix profile", and for the second figure, we may say: "the desirbale subsequences of our interest"</p><p> </p><p>Now, instead of four figures, we only show two figrues, and I think it should be easily understandable. Note that the figure of matrix profile itself (the last two figures shown above) does not help that much. The third figure only shows where the matrix profile is maximum . and the fourth figure might be confusing. So, I think using two lines of code is better than showing two more figures here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

NimaSarajpoor:
@ken-maeda 
I provided my final touch on the notebook. After addressing the comments, we should see what @seanlaw thinks regarding this notebook. 

If everything is okay, we can then move forward and add the part where you compute the contrast profile and show that it can discover the subsequences of our interest.

ken-maeda:
@NimaSarajpoor 
I'm sorry for the delay, I have fixed the notebook in the point you mentioned. I hope the notebook I created is fine now.

NimaSarajpoor:
@ken-maeda 
Thank you for addressing the comments. While there is still some room for improvement, we can do it later. I think @seanlaw  can take a look at the notebook now and see how if he has any opinion / suggestion.

@seanlaw 
Do you have any comment on the second notebook, i.e. `docs/Tutorial_Contrast_Profile2.ipynb` ?

seanlaw:
Let me find some time to provide some comments

docs/Tutorial_Contrast_Profile2.ipynb
```diff
@@ -0,0 +1,485 @@
...
```
seanlaw, line 1: <ol><li>There seems to be an extra space at the beginning of the title " Novel Time Series...". Please remove the space</li><li>"matrix profile" are two separate words, not one. Please add a space</li><li><code>If a dataset has two similar subsequences that are far from the rest of subsequences</code> Can you please elaborate on this sentence? I am not getting your point. If they are similar enough then they may not be the top motif but it is possible to be, say, in the top 50 motifs?</li><li>Usually, in the opening paragraph, we want to try an explain what the problem is that contrast profiles can help solve. Can you clearly explain the problem? It is okay to borrow from the published paper</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

seanlaw, line 1: <p>Can you please place the ECG data into one of the comments of the PR and then link to that uploaded CSV?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

seanlaw, line 1: <p>I don't think it is correct to call the grey areas "anomalies". Are they ever referred to as anomalies in the paper?</p><p> </p><p>Edit: Below, you use the term "phenomena" and I think this is reasonable and suitable instead of "anomaly". </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

seanlaw, line 1: <p>Why do we set <code>normalize=False</code>? Did they do that in the original paper?</p><p> </p><p>Why are we looking for the discord index first? It seems like the wrong order of steps. I feel like the first thing that somebody would do is to:</p><p> </p><ol><li>Apply <code>stump</code> to the entire time series</li><li>Then they realize that the "interesting" subsequence pair isn't in their top motif(s)</li><li>Finally, we explain why a naive approach is insufficient to find the "interesting" subsequence pair</li></ol><p> </p><p>Then, we can motivate the question, "How would/could we find the "interesting" subsequence pair?" and then provide clear and concrete steps on how to do it. Otherwise, this section seems out of place or presented too soon.</p><p> </p><p>I think we have to think about "what is the simplest thing that the user would have tried?" and then, when that fails, we build upon that knowledge and move past it with some better suggestion(s)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

seanlaw, line 1: <p>This section is the most important and I think we should move all of this to the top after you plot the ECG time series for the first time.</p><p> </p><p>@nimasarajpoor I wonder what would happen if we did a top-k nearest neighbor (with <code>k = 5</code>)and then plot the distances only for the 5th nearest neighbor (rather than the first 4 nearest neighbors). Perhaps, for periodic data with a single interesting pair of subsequences like this one, it would/could be a nice tutorial for understand top-k nearest neighbor matrix profiles. It is too much for this tutorial though.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

seanlaw, line 1: <p>Why are we only showing the first 2 rows? What is the point?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

NimaSarajpoor: @seanlaw 
we can try top-k matrix profile and see its output for this data. I am having a hard time to picture the outcome in my mind in advance. So, what I am about to say is just based on some vague images in my mind. I have not thought this through. Please proceed with caution. 

We know that the top-k neighbors of a sequence, computed by `stumpy.stump`, are out of the trivial zone of the sequence, but the k neighbors might be in trivial neighbor of each other. So, for sequence `S_i`, we may find out that `S_j` is its 1NN, which is outside of the trivial zone of `i`, and say their distance is small (similar to the case provided in the notebook) It MIGHT BE possible that the 2nd, 3rd, ..., k-th NN of `S_i` are very close to `S_j`. For instanc, they might be `S_{j+1}` and `S_{j+2}`, .... So, the distance between `S_i` and its other top-k neighbors might be small as well. Therefore, if matrix profile `P` has `k` columns that correspond to top-k NN, `np.argmax(P[:, x])` may not necessarily give us the answer. Ofcourse, what I just said depends on the value of `k`. For a large enough (?) "k", we might find the desirable patterns (of the phenomena) via `np.argmax(P[:, x])`.  

We can try removing the trivial zone of 1NN, and find the next NN, out of that top-k NN. Then, `np.argmax(P')` might give us the answer, where `P'` is the distance to second nn AFTER IGNORING THE TRIVIAL ZONE OF 1NN of a subsequence (in additon to ignoring the trivial zone of the subsequence itself). The problem may become more challenging if we have more than two almost-similar-but-distinct-from-other patterns since now you have to choose a larger `k` if you want to consider the removal of trivial zone after computing the top-k neighbors. Note that for large window size `m`, the exclusion zone becomes bigger, and we may be forced to choose greater value for `k`.

To me, contrast profile can be seen as  a tool that allows user to put a filter on usual, regular patterns. So, for instance, say I have temperature data of a device and it is mostly as usual. But, there might be some strange behavior from time to time. How to find it? If it is a really unique, strange behavior, it may be discovered by discord, i.e. `np.argmax(P) `. But, what if it is not unique? One way might be to start from left and as you move to the right on the time axis, you only consider left matrix profile, and you detect an anomally as soon as you find one. Then, you can just remove it from your usual data, So, when you get to the next anomly, it can be detected (otherwise, it may not be detectable if it is similar/close to previous anomaly). So, what is the point of contrast profile? Maybe one use contrast profile to see if the new anomaly is something that appears before. In other words: repeated anomaly.

Again, it is better to do the experiment on the data and see what the data / result tell us.

seanlaw: @ken-maeda To be clear, I am NOT suggesting that we do top-k here. It was simply an observation that this data set might be a good example for a separate top-k tutorial. Sorry for the distraction!

ken-maeda:
> Apply stump to the entire time series
> what is the simplest thing that the user would have tried?

Regarding the point whether it is natural to calculate the Matrix Profile for the entire series first, the characteristic we are trying to find this time is neither motif nor discord, so I feel there is no motivation to calculate stump directly. Therefore, even if you simply calculate it for the entire series, I don't think there is much to say from the event itself, so I compared when there is a single discord and when there are multiple discords, and built on what can be said from there.
It might be better to make it easy to understand what I'm trying to do from the very beginning.

seanlaw:
> Regarding the point whether it is natural to calculate the Matrix Profile for the entire series first, the characteristic we are trying to find this time is neither motif nor discord, so I feel there is no motivation to calculate stump directly. Therefore, even if you simply calculate it for the entire series, I don't think there is much to say from the event itself, so I compared when there is a single discord and when there are multiple discords, and built on what can be said from there.
It might be better to make it easy to understand what I'm trying to do from the very beginning.

I think the point is that `stump` will not be able to help you here precisely because the subsequence is neither a top motif or a top discord. Certainly, if you traverse down to the top-N motifs then you might eventually find it. I think it is important to motivate "why" computing the full matrix profile is not enough and also demonstrate its ineffectiveness for this particular problem (i.e., when the subsequence of interest does not have a nearest neighbor that is as close as other motifs)

seanlaw:
> 3-1. If a discord is included once, it can be found.
3-2. If a discord is included twice, it cannot be found.
3-3. So, what should we do?
Regarding this flow, I thought that it would be better to write more concisely at the beginning about what the Contrast Profile brings. However, what do you think should be written in the introduction?

If you look at the comments, I don't think "discord" or "anomaly" is the right word here as a discord is referring to "a subsequence that has a one-nearest neighbor that is very far away". In our example, the one nearest neighbor may not necessarily be very far away and, instead, it is that the subsequence of interest isn't discovered in the first few motifs. The paper refers to them as "phenomena of interest" and not "discord"

When I look at the contrast profile paper, it presents the problem as:

1. Imagine that you have a time series that usually has a reasonably well defined set of one or more repeating patterns and other subsequences are not expected
2. Then, something occurs that induces a new subsequence (a phenomena) that has never been observed before AND it isn't quite a discord because the phenomena has repeated itself at least once more and while this nearest neighbor might be "close" to the first manifestation, it isn't as "close" to the nearest neighbor of previously known motifs (i.e., the repeating patterns in 1.) 

So, you have some historical data where everything is well known and then you encounter an event that causes a new, never seen before subsequence AND it happens a second time. And so it becomes obvious to ask the question, "given how rare the phenomena is, how might we go about detecting it in a systematic way?"

The first (naive) thing that you might try is to compute the matrix profile for the entire time series and then iterate through the top-N motifs via `stumpy.motif` and then consider how similar the `ith` motif is to the first `i-1` motifs. However, there is a "better" way... Then introduce the contrast profile (including its assumptions, what it aims to do, and what it isn't/can't do)

ken-maeda:
What I'm trying to say is that it doesn't make sense to me why we're calculating the matrix profile for the entire series when there's not enough reason to expect that the phenomena we're trying to discover will be detected as motifs. It feels like we're just saying, 'We tried calculating it and this is what happened.' 

> So, you have some historical data where everything is well known and then you encounter an event that causes a new, never seen before subsequence AND it happens a second time. And so it becomes obvious to ask the question, "given how rare the phenomena is, how might we go about detecting it in a systematic way?"

In this way, the phenomenon we wish to find this time is mentioned, but I believe no mention is made of the relationship with other intervals or motifs that occur in other intervals. Under these circumstances, would a user try to find  it as a motif?

Particularly in this case, where other parts clearly have repetitive subsequence, it feels unnatural to calculate motifs to find the phenomena.

.

seanlaw:
> Particularly in this case, where other parts clearly have repetitive subsequence, it feels unnatural to calculate motifs to find the phenomena.

@ken-maeda If this is the case, you really need to explain this fact about the data at the beginning when you present the data. I don't think it is enough to assume that the reader will notice the repetition. Instead, you must point out the facts even if it is obvious. Then, you should explicitly explain that computing the full matrix profile won't really help you find the phenomena and why that is. Better yet, I was advocating that you just compute the matrix profile and show its limitations in that the phenomena won't be captured easily by the matrix profile (or using the `motifs` function).

ken-maeda:
As you suggested, I tried it. Does this seem to be okay?

top-motif location by motifs
image

image

We were able to find something close to the motif we want to find as the top 10th motif.

seanlaw:
> As you suggested, I tried it. Does this seem to be okay? We were able to find something close to the motif we want to find as the top 10th motif.

Yes! What code did you use to get that information? Was it `stumpy.stump` followed by `stumpy.motifs`?

ken-maeda:
```
mp = stumpy.stump(T, m, normalize=False)
motifs = stumpy.motifs(T, mp[:, 0])
```
I simply calculated the above example from the `motifs`.
Considering the following flow for the introduction, how far do you think it's appropriate to explain?

1."By naively using `stump`, you can discover motifs and discords.
2.In this case, the signal we want to find are similar, so it should be possible to findthem as some top-motif when we compute `stump`.
3.We are calculating `stump`, but the signal we want to find have clear characteristics in their size(amplitude), so we calculate with normalization set to False, which is True as the default in `stump`.
4.It could be calculated as the 10th top-motif.
5.However, in reality, we don't know as which top-motif it will be calculated. There is a more reliable way to find signal with these characteristics. This can be achieved with a concept called Contrast Profile, which I would like to explain in detail."

seanlaw:
@ken-maeda I think this level of detail is more than fine. However, I think it is important to point out that there is repetition in the data and so finding the "phenomena" (only repeated once) is challenging since the majority of the top motifs will look like the repeated pattern.

docs/Tutorial_Contrast_Profile3_introduction.ipynb
```diff
@@ -0,0 +1,226 @@
+{
```
seanlaw: <p>In all of our tutorials, we try to reproduce specific figures from the original paper (without alteration). Which figure is this reproducing from the paper? This looks different Figure 4. Did you alter the time series? If yes, can you please use the same time series as provided in the original paper? The goal here is reproducibility of the work.</p><p> </p><p>Also, the original paper does not say anything about setting <code>normalize=False</code> and so we probably shouldn't confuse the reader by doing it here either.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/800/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

ken-maeda: The introduction I'm currently creating is a reproduction of Figure 2, not Figure 4. We have discussed up to this point that the original figure is abstract and hard to understand, so it's necessary to explain what it means with actual signal. Therefore, I'm using parts from the original data that were not used in the paper but are suitable for explanation this time. So, the normalization issue is not something to be referenced from the paper. If that is a problem, I think it could have been pointed out before creation.I checked the flow of the introduction above to prevent such confusion... and  inconsistent with past disscussion.

Issue #778, opened 2023-01-14
Reducing computing time in floss


NimaSarajpoor:
We can see the following lines in method `update`:

https://github.com/TDAmeritrade/stumpy/blob/59ee63a5cc2d9fb24e195e1141f23b92ec0aab5b/stumpy/floss.py#L560-L565

As shown above, we are computing the full arrays `M_T` and `Σ_T` in each call of this method. However, this might be time consuming, and a better way is to just update these arrays in each call. It seems that this part of code is similar to:

https://github.com/TDAmeritrade/stumpy/blob/59ee63a5cc2d9fb24e195e1141f23b92ec0aab5b/stumpy/stumpi.py#L231-L242 

which is a part of method `_update_egress` in class `stumpi`. Here, we take advantage of the previously-computed  `M_T` and `Σ_T`, and we just update it.

In `floss`, we need to create attributes `M_T` and `Σ_T` when we create an instance of this class, and then update them.

Issue #808, opened 2023-03-09
minor issue in `floss`


NimaSarajpoor:
It seems that there exists a minor issue in the `floss` module.

https://github.com/TDAmeritrade/stumpy/blob/b6aba2b3f9b7d5cf8256f633c5ae0b0d2ada1d2a/stumpy/floss.py#L519-L521

If I understand correctly, the last line should be ` self._mp[inf_indices, 3] = -1 `. In fact, the last line is not necessary because `self._mp[:, 3]` is already set at `-1`, and we use this column to find `inf_indices` in line 519. So, a good fix is to just remove the last line.

NimaSarajpoor:
**Refactoring**

In `stumpi`, we also use indices to calculate matrix profile value:

https://github.com/TDAmeritrade/stumpy/blob/b6aba2b3f9b7d5cf8256f633c5ae0b0d2ada1d2a/stumpy/stumpi.py#L148-L170

And, in `core.py`, we have this function:
https://github.com/TDAmeritrade/stumpy/blob/b6aba2b3f9b7d5cf8256f633c5ae0b0d2ada1d2a/stumpy/core.py#L2374
that does the same thing.

Hence, I think there is an opportunity to do refactoring here.

NimaSarajpoor:
As a follow up to my previous comment, the following line needs to be fixed as well:

https://github.com/TDAmeritrade/stumpy/blob/b6aba2b3f9b7d5cf8256f633c5ae0b0d2ada1d2a/stumpy/floss.py#L566

I think it should have been: `self._mp[-1, 3] = -1 `

seanlaw:
@NimaSarajpoor I don't know if I agree. It's been a long time since I've written that code and I don't remember the details but I'm guessing that there's an issue with actually allowing the right index (i.e., `self._mp[:, 3]`) to have negative values. Technically, a negative index value has no meaning (i.e., it is an invalid index). However, in `numpy`, a negative value corresponds to the last position of an array and so we can't "trust" that negative index value anymore and we need to replace it with something else that is not `-1`.

NimaSarajpoor:
@seanlaw 
> so we can't "trust" that negative index value anymore

Right! I missed that. Now that you mentioned it, I checked out the `core._idx_to_mp` code again:

https://github.com/TDAmeritrade/stumpy/blob/b6aba2b3f9b7d5cf8256f633c5ae0b0d2ada1d2a/stumpy/core.py#L2417-L2432

[...]

PR #810, opened 2023-03-12
Add param subseq_isconstant to floss

NimaSarajpoor:
This PR adds subseq_isconstant param to `floss` module

### DIFF
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -697,6 +697,11 @@ class floss:
         The indices stored in `self.I_` reflect the starting index of
         subsequneces with respect to the original time series.
         """
+        # Comparing the right matrix profile index value with the self index
+        # position (i.e., self._mp[:, 3] == np.arange(len(self._mp)) is avoided
+        # because things are constantly being shifted (ingress+egress) to the left
+        # and so the aforementioned check is all `False` as soon as we start using
+        # `.update()`.
         I = self._mp[:, 3].astype(np.int64).copy()
         I[self._mp[:, 0] == np.inf] = -1
 
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -647,10 +647,6 @@ class floss:
                 T_subseq_isconstant=self._T_subseq_isconstant,
                 Q_subseq_isconstant=self._Q_subseq_isconstant,
             )
-            D[self._Q_subseq_isconstant & self._T_subseq_isconstant] = 0
-            D[
-                np.logical_xor(self._Q_subseq_isconstant, self._T_subseq_isconstant)
-            ] = np.sqrt(self._m)
         else:
             D = core.mass_absolute(self._T[-self._m :], self._T, p=self._p)
 
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -496,7 +496,7 @@ class floss:
         self._p = p
         if T_subseq_isconstant is None:
             T_subseq_isconstant = core._rolling_isconstant
-        if not callable(T_subseq_isconstant):
+        if not callable(T_subseq_isconstant):  # pragma: no cover
             msg = (
                 "The parameter `T_subseq_isconstant` must be "
                 + "callabel. For details, see the docstring."
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -695,7 +695,8 @@ class floss:
         Get the updated (right) matrix profile indices
 
         The indices stored in `self.I_` reflect the starting index of
-        subsequneces with respect to the original time series.
+        subsequneces with respect to the full time series (i.e., including
+        all egressed data points).
         """
         # Comparing the right matrix profile index value with the self index
         # position (i.e., self._mp[:, 3] == np.arange(len(self._mp)) is avoided
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -347,6 +347,14 @@ class floss:
     normalize : bool, default True
         When set to `True`, this z-normalizes subsequences prior to computing distances
 
+    T_subseq_isconstant : function, default None
+        A custom, user-defined function that returns a boolean array that indicates
+        whether a subsequence in `T` is constant (True). The function must only take
+        two arguments, `a`, a 1-D array, and `w`, the window size, while additional
+        arguments may be specified by currying the user-defined function using
+        `functools.partial`. Any subsequence with at least one np.nan/np.inf will
+        automatically have its corresponding value set to False in this boolean array.
+
     Attributes
     ----------
     cac_1d_ : numpy.ndarray
@@ -412,6 +420,7 @@ class floss:
         custom_iac=None,
         normalize=True,
         p=2.0,
+        T_subseq_isconstant=None,
     ):
         """
         Initialize the FLOSS object
@@ -465,6 +474,15 @@ class floss:
         p : float, default 2.0
             The p-norm to apply for computing the Minkowski distance. This parameter is
             ignored when `normalize == True`.
+
+        T_subseq_isconstant : function, default None
+            A custom, user-defined function that returns a boolean array that indicates
+            whether a subsequence in `T` is constant (True). The function must only take
+            two arguments, `a`, a 1-D array, and `w`, the window size, while additional
+            arguments may be specified by currying the user-defined function using
+            `functools.partial`. Any subsequence with at least one np.nan/np.inf will
+            automatically have its corresponding value set to False in this boolean
+            array.
         """
         self._mp = copy.deepcopy(np.asarray(mp))
         self._T = copy.deepcopy(np.asarray(T))
@@ -476,6 +494,16 @@ class floss:
         self._custom_iac = custom_iac
         self._normalize = normalize
         self._p = p
+        if T_subseq_isconstant is None:
+            T_subseq_isconstant = core._rolling_isconstant
+        if not callable(T_subseq_isconstant):
+            msg = (
+                "The parameter `T_subseq_isconstant` must be "
+                + "callabel. For details, see the docstring."
+            )
+            raise ValueError(msg)
+        else:
+            self._isconstant_func = T_subseq_isconstant
 
         self._k = self._mp.shape[0]
         self._n = self._T.shape[0]
@@ -485,7 +513,9 @@ class floss:
         self._finite_T = self._T.copy()
         self._finite_T[~np.isfinite(self._finite_T)] = 0.0
         self._finite_Q = self._finite_T[-self._m :].copy()
-
+        self._T_subseq_isconstant = core.rolling_isconstant(
+            self._T, self._m, self._isconstant_func
+        )
         if self._custom_iac is None:  # pragma: no cover
             self._custom_iac = _iac(
                 self._k,
@@ -514,6 +544,15 @@ class floss:
                 - core.z_norm(right_nn, 1),
                 axis=1,
             )
+            nn_subseq_isconstant = self._T_subseq_isconstant[self._mp[:, 3]]
+            # subseq and its nn are both constant
+            mask = T_subseq_isconstant & nn_subseq_isconstant
+            self._mp[mask, 0] = 0
+
+            # Of a subseq and its nn, only one is constant.
+            mask = np.logical_xor(T_subseq_isconstant & nn_subseq_isconstant)
+            self._mp[mask, 0] = np.sqrt(m)
+
         else:
             self._mp[:, 0] = np.linalg.norm(
                 core.rolling_window(self._T, self._m) - right_nn,
@@ -551,12 +590,17 @@ class floss:
         self._T_isfinite[:-1] = self._T_isfinite[1:]
         self._finite_T[:-1] = self._finite_T[1:]
         self._finite_Q[:-1] = self._finite_Q[1:]
+        self._T_subseq_isconstant[:-1] = self._T_subseq_isconstant[1:]
+
         self._T[-1] = t
         self._T_isfinite[-1] = np.isfinite(t)
         self._finite_T[-1] = t
         if not np.isfinite(t):
             self._finite_T[-1] = 0.0
         self._finite_Q[-1] = self._finite_T[-1]
+        self._Q_isconstant = self._isconstant_func(self._T[: -self._m], self._m)
+        self._T_subseq_isconstant[:-1] = self._Q_isconstant
+
         excl_zone = int(np.ceil(self._m / config.STUMPY_EXCL_ZONE_DENOM))
         # Note that the start of the exclusion zone is relative to
         # the unchanging length of the matrix profile index
@@ -571,15 +615,16 @@ class floss:
 
         # Ingress
         if self._normalize:
-            T_subseq_isconstant = core.rolling_isconstant(self._T, self._m)
             M_T, Σ_T = core.compute_mean_std(self._T, self._m)
             D = core.mass(
                 self._finite_Q,
                 self._finite_T,
                 M_T,
                 Σ_T,
-                T_subseq_isconstant=T_subseq_isconstant,
+                T_subseq_isconstant=self._T_subseq_isconstant,
             )
+            D[self._Q_isconstant & self._T_subseq_isconstant] = 0
+            D[np.logical_xor(self._Q_isconstant, self._T_subseq_isconstant)] = 0
         else:
             D = core.mass_absolute(self._T[-self._m :], self._T, p=self._p)
 
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -568,6 +568,15 @@ class floss:
                 axis=1,
                 ord=self._p,
             )
+        # Technically, a negative index value has no meaning (i.e., it is an
+        # invalid index). However, in numpy, a negative value corresponds to the last
+        # position of an array and so we can't "trust" that negative index value anymore
+        # and we need to replace it with something else that is not -1. In the following
+        # lines, if subsequence `S_i` has no valid right nearest neigbor
+        # (i.e. self._mp[i, 3]==3), then `self._mp[i, 3]` will be set to `i`. This
+        # modification results in having distance 0 (instead of np.inf) between
+        # `S_i` and its nearest neighbor. This can be resolveed by applying a
+        # post-processing step after computing distances.
         inf_indices = np.argwhere(self._mp[:, 3] < 0).flatten()
         self._mp[inf_indices, 0] = np.inf
         self._mp[inf_indices, 3] = inf_indices
@@ -690,6 +699,12 @@ class floss:
     def I_(self):
         """
         Get the updated (right) matrix profile indices
+
+        Note1: In contrast to the output of `stump`, any invalid index `-1`
+        in this (right) matrix profile indices is replaced with the start
+        index of the corresponding subsequence.
+        Note2: The indices stored in `self.I_` reflect the starting index of
+        subsequneces with respect to the original time series.
         """
         return self._mp[:, 3].astype(np.int64)
 
--- a/stumpy/floss.py
+++ b/stumpy/floss.py
@@ -347,7 +347,7 @@ class floss:
     normalize : bool, default True
         When set to `True`, this z-normalizes subsequences prior to computing distances
 
-    T_subseq_isconstant : function, default None
+    T_subseq_isconstant_func : function, default None
         A custom, user-defined function that returns a boolean array that indicates
         whether a subsequence in `T` is constant (True). The function must only take
         two arguments, `a`, a 1-D array, and `w`, the window size, while additional
@@ -420,7 +420,7 @@ class floss:
         custom_iac=None,
         normalize=True,
         p=2.0,
-        T_subseq_isconstant=None,
+        T_subseq_isconstant_func=None,
     ):
         """
         Initialize the FLOSS object
@@ -475,7 +475,7 @@ class floss:
             The p-norm to apply for computing the Minkowski distance. This parameter is
             ignored when `normalize == True`.
 
-        T_subseq_isconstant : function, default None
+        T_subseq_isconstant_func : function, default None
             A custom, user-defined function that returns a boolean array that indicates
             whether a subsequence in `T` is constant (True). The function must only take
             two arguments, `a`, a 1-D array, and `w`, the window size, while additional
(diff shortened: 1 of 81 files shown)


NimaSarajpoor (commented):
@seanlaw 
I think `floss` is almost ready. Please let me know what you think about the comments I provided here.

stumpy/floss.py
```diff
@@ -476,6 +494,16 @@ def __init__(
...
+        if T_subseq_isconstant is None:
+            T_subseq_isconstant = core._rolling_isconstant
+        if not callable(T_subseq_isconstant):  # pragma: no cover
+            msg = (
+                "The parameter `T_subseq_isconstant` must be "
+                + "callabel. For details, see the docstring."
+            )
+            raise ValueError(msg)
+        else:
+            self._isconstant_func = T_subseq_isconstant
```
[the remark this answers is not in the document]
NimaSarajpoor: I think your proposal is clearer as the suggested name follows the structure of the original `T_subseq_isconstant`. I will revise the name of variables.

stumpy/floss.py
```diff
@@ -347,6 +347,14 @@ class floss:
     normalize : bool, default True
         When set to `True`, this z-normalizes subsequences prior to computing distances
```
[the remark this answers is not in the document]
NimaSarajpoor: Sure. I will fix it.

stumpy/floss.py
```diff
@@ -476,6 +494,16 @@ def __init__(
         self._custom_iac = custom_iac
         self._normalize = normalize
         self._p = p
+        if T_subseq_isconstant is None:
+            T_subseq_isconstant = core._rolling_isconstant
+        if not callable(T_subseq_isconstant):  # pragma: no cover
+            msg = (
+                "The parameter `T_subseq_isconstant` must be "
+                + "callabel. For details, see the docstring."
```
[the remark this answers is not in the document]
NimaSarajpoor: Thank you for catching the typo. I will fix it and revies the message.

stumpy/floss.py
```diff
@@ -476,6 +494,16 @@ def __init__(
...
+            )
+            raise ValueError(msg)
+        else:
```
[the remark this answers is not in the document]
NimaSarajpoor: Right! I will remove it!

tests/test_floss.py
```diff
@@ -411,3 +419,85 @@ def test_aamp_floss_inf_nan(substitute, substitution_locations):
             npt.assert_almost_equal(ref_P, comp_P)
             npt.assert_almost_equal(ref_I, comp_I)
             npt.assert_almost_equal(ref_T, comp_T)
+
+
+def test_floss_with_isconstant():
+    def custom_func(a, w):
+        _, Σ_a = core.compute_mean_std(a, w)
+
+        return Σ_a < 0.5
```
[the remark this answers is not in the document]
NimaSarajpoor: The function `isconstant_func_stddev_threshold(T, m, ...)` decides a subsequence is constant or not based on its standard deviation and a threshold, denoted as `stddev_threshold` in the code. However, the value of this threshold depends on the full timeseries `T` (and not just one subsequence); See line below:

https://github.com/TDAmeritrade/stumpy/blob/203aa3f7721509d1139e451c43bb3aaccc63d713/tests/naive.py#L2088

A subequence `S` (in the timeseries `T`) has a certain value for its standard deviation; however, the threshold `stddev_threshold`, in the line shown above, DEPENDS on the stddev of ALL subseqs in `T`. So, when I update `T`, the threshold might change. In other words, `S` might be constant in `T`, but may not be constant in `np.append(T[1:], new_value)`. 


Now, let's see what is happening in `floss`:

```
# when `floss` object is initialized, we compute `self._T_subseq_isconstant`
self._T_subseq_isconstant = core.rolling_isconstant(self._T, self._m, self._T_subseq_isconstant_func)

# Then when we call `update()`, we have:
self._T_subseq_isconstant[:-1] = self._T_subseq_isconstant[1:]
self._T_subseq_isconstant[-1] = core.rolling_isconstant(self._T[-self._m :], self._m, self._T_subseq_isconstant_func)
```
Here, we shift the array so as to avoid computing the full array everytime. Note that the last line applies the function `self._T_subseq_isconstant_func` on ONLY ONE SUBSEQUENCE as opposed to the whole time series. Also note that the program DOES NOT MEMORIZE / LEARN the threshold `stddev_threshold` and instead it just compute it whenever the function `self._T_subseq_isconstant_func` is called. So, a different result can be obtained if, instead of shifting `self._T_subseq_isconstant`, we compute it for the whole data (and not just the last subsequence). In other words, the outcome of
```
self._T_subseq_isconstant = core.rolling_isconstant(self._T, self._m, self._T_subseq_isconstant_func)
```
can be different. Again, just to be clear, this is only because of the way the function `isconstant_func_stddev_threshold` is designed, where the stddev threshold depends on the sliding_stddev of the subsequences in the first argument of this function.

---

Now, if we look at the test function `test_floss.py`, we can see the following line:
```
ref_T_subseq_isconstant = custom_func(ref_T, m)
```
Here, we compute "subseq_isconstant " for the full array `ref_T`. (`ref_T` is a shifted version of data[:n], see the for loop ` for i, ref_T in enumerate(rolling_Ts):` in `the test_floss.py`) 

To wrap up, if I want to use the function `isconstant_func_stddev_threshold(T, m, ...)` then, I need to avoid shifiting `self._T_subseq_isconstant `, and instead just recompute it for the array as follows:
```
# in .update()
T_subseq_isconstant = core.rolling_isconstant(self._T, self._m, func)
```
note that `self._T` is the updated version of `T`. Also, I do not need to store `T_subseq_isconstant ` as attribute since we need to recompute the full array everytime. The downside is we may increase the computing time of `.update()`. Also, I am not sure even if it makes any sense to have a function to say a subsequence `S` is constant in `T` but then it says it is not constant in the shifted version of `T`. So, maybe it may not be meaningful to have a funciton like `isconstant_func_stddev_threshold`, where its output regarding a particular subsequence can be different between `T` and `np.append(T[1:], new_value)`.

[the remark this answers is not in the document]
seanlaw: Thanks for the explanation. So, the issue seems to be with `isconstant_func_stddev_threshold` in that it requires knowledge of the FULL time series in order to compute the stddev threshold. Here are some thoughts/ideas:

Move `stddev_threshold` into the list of parameters of `isconstant_func_stddev_threshold` so that it looks like:
```
def isconstant_func_stddev_threshold(a, w, quantile_threshold=0, stddev_threshold=None):
    sliding_stddev = rolling_nanstd(a, w)
    if stddev_threshold is not None:
        return sliding_stddev <= stddev_threshold
    elif quantile_threshold == 0:  # pragma: no cover
        return sliding_stddev == 0
    else:
        stddev_threshold = np.quantile(sliding_stddev, quantile_threshold)
        return sliding_stddev <= stddev_threshold
```

Then, instead of computing `stddev_threshold` for the full time series, I would simply compute `stddev_threshold` ONCE AND ONLY ONCE for the initial `T` chunk and then, for each new data point and corresponding new subsequence, use the same pre-computed `stddev_threshold` from the first chunk. In other words, you make an assumption that all future, unseen data will have the same `stddev_threshold`.

To me, this is somewhat similar to ML feature engineering when you use [MinMaxScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html) to force your training data to have values between `0` and `1`. However, it is completely possible that your test set may have values that are larger than your `max` training data and therefore causes you to have test values that are larger than `1`. It's an assumption and I think it's okay especially for unit testing.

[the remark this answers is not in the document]
NimaSarajpoor: > Then, instead of computing `stddev_threshold` for the full time series, I would simply compute `stddev_threshold` ONCE AND ONLY ONCE for the initial `T` chunk and then, for each new data point and corresponding new subsequence, use the same pre-computed `stddev_threshold` from the first chunk

Cool! I will revise it.

> To me, this is somewhat similar to ML feature engineering when you use [MinMaxScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html) to force your training data to have values between `0` and `1`. However, it is completely possible that your test set may have values that are larger than your `max` training data and therefore causes you to have test values that are larger than `1`. It's an assumption and I think it's okay especially for unit testing.

Exactly! We assume the initial `T` chunk represents the distribution of stddev for FULL data.

NimaSarajpoor:
@seanlaw 
Could you please checkout the error? and re-run the checks if you think that should resolve the issue.

stumpy/floss.py
```diff
@@ -514,6 +551,17 @@ def __init__(
                 - core.z_norm(right_nn, 1),
                 axis=1,
             )
+            nn_subseq_isconstant = self._T_subseq_isconstant[
+                self._mp[:, 3].astype(np.int64)
+            ]
```
NimaSarajpoor, line 556: As discussed in #808, we may add a comment to inform users of the fact that some of the arrays in `right_nn` and some booean values in `nn_subseq_isconstant` are meaningless as `self._mp[:, 3]` contains `-1`, which is not the start index of  a subsequence.  Howver, the postprocessing at the end of this "if/else block" takes care of this matter.

@seanlaw : Any suggestion?

seanlaw: Is it to inform users or to inform contributors? I think we should be consistent and follow what we said earlier by adding a comment/reminder

NimaSarajpoor: @seanlaw 
Oops! I meant `contributors`. I am going to use what you said in #808 

```
"Technically, a negative index value has no meaning (i.e., it is an invalid index). "
"However, in numpy, a negative value corresponds to the last position of an array "
"and so we can't "trust" that negative index value anymore and we need to replace it "
"with something else that is not -1."
```

I may do some minor changes if need. But, the main message will follow yours.

stumpy/floss.py
```diff
@@ -514,12 +551,32 @@ def __init__(
...
                 ord=self._p,
             )
+        # Technically, a negative index value has no meaning (i.e., it is an
+        # invalid index). However, in numpy, a negative value corresponds to the last
+        # position of an array and so we can't "trust" that negative index value anymore
+        # and we need to replace it with something else that is not -1. In the following
+        # lines, if subsequence `S_i` has no valid right nearest neigbor
+        # (i.e. self._mp[i, 3]==3), then `self._mp[i, 3]` will be set to `i`. This
+        # modification results in having distance 0 (instead of np.inf) between
+        # `S_i` and its nearest neighbor. This can be resolveed by applying a
```
NimaSarajpoor, line 578:
```suggestion
        # `S_i` and its invalid nearest neighbor. This can be resolved by applying a
```

stumpy/floss.py
```diff
@@ -514,12 +551,32 @@ def __init__(
...
+            # Only the subseq OR its nn is constant but not both
+            mask = np.logical_xor(self._T_subseq_isconstant, nn_subseq_isconstant)
+            self._mp[mask, 0] = np.sqrt(m)
+
         else:
             self._mp[:, 0] = np.linalg.norm(
                 core.rolling_window(self._T, self._m) - right_nn,
                 axis=1,
                 ord=self._p,
             )
```
NimaSarajpoor, line 570: Below I have added the comment based on what you mentioned before. I expanded it a little. But now I think it may be too verbose. If you feel the same way, I can prune it.

stumpy/floss.py
```diff
@@ -626,6 +699,12 @@ def P_(self):
     def I_(self):
         """
         Get the updated (right) matrix profile indices
+
+        Note1: In contrast to the output of `stump`, any invalid index `-1`
+        in this (right) matrix profile indices is replaced with the start
+        index of the corresponding subsequence.
+        Note2: The indices stored in `self.I_` reflect the starting index of
+        subsequneces with respect to the original time series.
```
NimaSarajpoor, line 707: I have also added two notes here that can help user better understand the information stored in `self._I`

stumpy/floss.py
```diff
@@ -486,6 +517,12 @@ def __init__(
         self._finite_T[~np.isfinite(self._finite_T)] = 0.0
         self._finite_Q = self._finite_T[-self._m :].copy()
 
+        if self._normalize:
+            self._T_subseq_isconstant = core.rolling_isconstant(
+                self._T, self._m, self._T_subseq_isconstant_func
+            )
+            self._M_T, self._Σ_T = core.compute_mean_std(self._T, self._m)
```
seanlaw, line 524: When we close this PR, we should also close #778  as well, right?

stumpy/floss.py
seanlaw: Yes, it is too long. How about:

> Note that a negative matrix profile index (e.g. `-1`) represents a null index. However, in numpy, negative indices are interpreted as counting from the end of the array. To resolve this, we temporarily replace a negative index with a self index. So, if subsequence `S_i` has no valid right nearest neighbor, then `self._mp[i, 3]` will be set to `i` and its corresponding matrix profile distance is set to 0.0. This temporary resolution is later resolved (by setting them back to `-1` and `np.inf`) in a post-processing step.

seanlaw: I kinda like that as it provides just enough of a reminder that is clear and concise

stumpy/floss.py
```diff
@@ -571,15 +629,30 @@ def update(self, t):
...
                 self._finite_T,
-                M_T,
-                Σ_T,
-                T_subseq_isconstant=T_subseq_isconstant,
+                self._M_T,
+                self._Σ_T,
+                T_subseq_isconstant=self._T_subseq_isconstant,
+                Q_subseq_isconstant=self._Q_subseq_isconstant,
             )
+            D[self._Q_subseq_isconstant & self._T_subseq_isconstant] = 0
```
seanlaw, line 652: Maybe apply the same `mask` here as above for consistency? Or simply add the same comment as above

seanlaw (changes requested):

stumpy/floss.py
seanlaw: You don't need `Note1` and `Note2`:

> The right matrix profile indices reflect the starting index of subsequences with respect to the original time series and all invalid matrix profile indices are replaced with the start index of the corresponding subsequence (rather than with `-1`).

I can't remember if we should/ned to keep bad indices as the "self index". I think we can/should convert the index back to `-1` as a post-processing step (and then we can omit the second half of the note above). However, we should make a copy first and then hand it back to the user

stumpy/floss.py
NimaSarajpoor: > clear and (mostly) concise

Right! I will revise the comment. Thank you for the feedback :)

stumpy/floss.py
NimaSarajpoor: Yes, we can close that issue.

stumpy/floss.py
NimaSarajpoor: I will. 

---

Now that I am looking at this again, I think we should be able to avoid this post-processing part. Because, `mass` is a public API, and we are already passing these two arguments `Q_subseq_isconstant` and `T_subseq_isconstant` to this function. So, this `mass` function itself is taking care of this. In fact, if we follow the codes, we can see we pass these two params all the way to function `_calculate_squared_distance` (in `core.py`). .

stumpy/floss.py
NimaSarajpoor: @seanlaw 
> I can't remember if we should/ned to keep bad indices as the "self index".

Me neither :) I think I wanted to do that but after realizing the concept behind it, I did not think about it again.

> I think we can/should convert the index back to `-1` as a post-processing step

As you suggested in #808, we can do:
```
I = self._mp[:, 3].astype(np.int64).copy()
I[self._mp[:, 0] == np.inf] = -1
```

I might have some thoughts regarding the `self._I` that I can share with you in the future when I finish my investigation. 

For now, I think we should be good. I will revise the code.

stumpy/floss.py
```diff
@@ -626,8 +693,14 @@ def P_(self):
     def I_(self):
         """
         Get the updated (right) matrix profile indices
+
+        The indices stored in `self.I_` reflect the starting index of
+        subsequneces with respect to the original time series.
         """
-        return self._mp[:, 3].astype(np.int64)
+        I = self._mp[:, 3].astype(np.int64).copy()
+        I[self._mp[:, 0] == np.inf] = -1
```
seanlaw, line 701: I would feel more comfortable checking both condition with something like:

```
mask = np.logical_or(self._mp[:, 0] == np.inf, self._mp[:, 3] == np.arange(len(self._mp)))
I[mask] = -1
```

tests/naive.py
```diff
@@ -2080,10 +2080,12 @@ def find_matches(D, excl_zone, max_distance, max_matches=None):
     return np.array(result[:max_matches], dtype=object)
 
 
-def isconstant_func_stddev_threshold(a, w, quantile_threshold=0):
+def isconstant_func_stddev_threshold(a, w, quantile_threshold=0, stddev_threshold=None):
     sliding_stddev = rolling_nanstd(a, w)
```
seanlaw, line 2084: call it `rolling_stddev` instead

tests/naive.py
```diff
@@ -2080,10 +2080,12 @@ def find_matches(D, excl_zone, max_distance, max_matches=None):
...
-    if quantile_threshold == 0:  # pragma: no cover
-        return sliding_stddev == 0
-    else:
-        stddev_threshold = np.quantile(sliding_stddev, quantile_threshold)
-        return sliding_stddev <= stddev_threshold
+    if stddev_threshold is None:
```
seanlaw, line 2085:
```
if stddev_threshold is None:
    stddev_threshold = np.quantile(sliding_stddev, quantile_threshold)
    if quantile_threshold == 0:
        stddev_threshold = 0
```

I didn't love the `else` as it was a bit harder to read due to it being nested

stumpy/floss.py
NimaSarajpoor: I understand your concern as I think you want to be safer here. However, you may want to reconsider it as the new check, i.e `self._mp[:, 3] == np.arange(len(self._mp)`, can be only useful at the initial step where floss object is initialized. Even in that case, the main check `self._mp[:, 0] == np.inf` should be clear and less complicated. Note that with each call of method `.update`, we shift the indices to the left and add a new value `last_idx` at the end which is bigger than `len(mp)`. In other words, `elf._mp[:, 3] == np.arange(len(self._mp))` becomes useless as soon as we start using `.update()`

tests/naive.py
NimaSarajpoor: @seanlaw 
I totally agree with you. What do you suggest @seanlaw ? I can go with what you initially suggested and `return` an ouput `if stddev_threshold is not None` 

```
rolling_stddev = rolling_nanstd(a, w)
if stddev_threshold is not None:
     return  rolling_stddev <  stddev_threshold

# else, meaning `stddev_threshold` is None
 if quantile_threshold == 0:
       stddev_threshold = 0  # reset to 0 
else:
       stddev_threshold = np.quantile(sliding_stddev, quantile_threshold)
```

I personally think we can replace the if/else block with this:

```
# else, meaning `stddev_threshold` is None
 stddev_threshold = np.quantile(sliding_stddev, quantile_threshold)
 if quantile_threshold == 0:
       stddev_threshold = 0  # replce computed `stddev_threshold` with 0
```

stumpy/floss.py
seanlaw: > In other words, self._mp[:, 3] == np.arange(len(self._mp)) becomes useless as soon as we start using .update()

Ahh, good point! Can you please add a note here with something like:

```
# Note that we avoid comparing the right matrix profile index value with the # self index position (i.e., self._mp[:, 3] == np.arange(len(self._mp))
# because things are constantly being shifted (ingress+egress) to the left
# and so the self._mp[:, 3] == np.arange(len(self._mp)) becomes useless as # soon as we start using .update().
```

Maybe something like that?

tests/naive.py
seanlaw: Hmm, is there a problem with what I proposed above?

```
rolling_stddev = rolling_nanstd(a, w)

if stddev_threshold is None:
    stddev_threshold = np.quantile(rolling_stddev, quantile_threshold)
    if quantile_threshold == 0:
        stddev_threshold = 0

return rolling_stddev < stddev_threshold
```

Am I overlooking some logic here? It seems reasonably clear and okay to keep it all in my head. What I also don't love is returning a value in the middle of a function rather than at the end.

seanlaw (changes requested):

tests/naive.py
NimaSarajpoor: I thought you were just quoting a part of my code and you just asked me to find a way to avoid nested if. I now noticed that the code in your initial comment was actually your suggestion :) All good! 

I knew about your concern about having return in the middle of the function but I thought you may allow this time :) I will revise the code accordingly.

stumpy/floss.py
NimaSarajpoor: Sure. I will add a comment regarding this matter.

tests/test_floss.py
```diff
@@ -411,3 +429,91 @@ def test_aamp_floss_inf_nan(substitute, substitution_locations):
             npt.assert_almost_equal(ref_P, comp_P)
             npt.assert_almost_equal(ref_I, comp_I)
             npt.assert_almost_equal(ref_T, comp_T)
+            npt.assert_almost_equal(ref_mp[:, 3], stream._mp[:, 3])
```
NimaSarajpoor, line 432: I added this test but I now think it might not be needed. what do you think @seanlaw ?

stumpy/floss.py
seanlaw: I think this should say "with respect to the full time series (i.e., including all egressed data points)"

NimaSarajpoor:
@seanlaw 

> Aside from one comment, everything looks good on my end

I addressed your comment. I think everything should be good now. Please feel free to merge.

stumpy/floss.py
```diff
@@ -476,6 +498,15 @@ def __init__(
         self._custom_iac = custom_iac
         self._normalize = normalize
         self._p = p
```
seanlaw, line 500: I think it's important to initialize some variables:

```
self._T_subseq_isconstant = None
self._M_T = None
self._Σ_T = None
```

seanlaw (changes requested):

seanlaw:
Apologies. I just noticed one minor thing. Once you fix that then I will merge

stumpy/floss.py
NimaSarajpoor: Fixed!

seanlaw:

PR #812, opened 2023-03-13
Fixed #811 addressing issues in floss

### LINKED ISSUE #811: Potential issues in `floss`
Before moving forward with #810, I think the `floss` module needs to be investigated regarding the following issues:

1)

https://github.com/TDAmeritrade/stumpy/blob/c06f0e998d28e7b717bab1e4e32a2aeb897ddfc3/stumpy/floss.py#L561-L566

`self._mp[-1,3]` should be set to the last index. But, `self._last_idx` is set as:
https://github.com/TDAmeritrade/stumpy/blob/c06f0e998d28e7b717bab1e4e32a2aeb897ddfc3/stumpy/floss.py#L478

So, I was wondering if we should change the code above to `self._last_idx = self._n - self._m`

2)

https://github.com/TDAmeritrade/stumpy/blob/c06f0e998d28e7b717bab1e4e32a2aeb897ddfc3/stumpy/floss.py#L584

In code above, `self._T_isfinite` is a boolean array. I think this line should be changed to:
```
self._T_isfinite = core.rolling_isfinite(self._T, self._m)
```

Also, if we are just shifting or updating, why should we recompute the full array everytime?

3) 

https://github.com/TDAmeritrade/stumpy/blob/c06f0e998d28e7b717bab1e4e32a2aeb897ddfc3/stumpy/floss.py#L568-L580

If I understand correctly, we are not removing the very first point in the time series in "ingress" mode. But, we already shift `self._T`, see below:

https://github.com/TDAmeritrade/stumpy/blob/c06f0e998d28e7b717bab1e4e32a2aeb897ddfc3/stumpy/floss.py#L546

So, shouldn't we seperate ingress / egress, like the code in `stumpy/stumpi.py` module.

---

I think it is worth it to investigate the aforementioned items and clean up this module if possible. I have not read the `floss` paper. So, my observation is purely based on the code itself.

NimaSarajpoor:


NimaSarajpoor:
[Update]
I noticed that, in test function, we already handle the non-finite values correctly.  However, I think the performant version does not handle it correctly. It is interesting that this hasn't been caught by the unit test.

stumpy/floss.py
```diff
@@ -581,7 +581,7 @@ def update(self, t):
 
         D[zone_start:] = np.inf
 
-        T_subseq_isfinite = core.rolling_isfinite(self._T_isfinite, self._m)
+        T_subseq_isfinite = core.rolling_isfinite(self._T, self._m)
```
NimaSarajpoor: Thanks to your code coverage, I realized that the `core.rolling_isfinite` accepts boolean array as well ! (see `core._rolling_isfinite_1d`)

So, everything was fine in the first place!

**side note:**
The description of the output of `core.rolling_isfinite` needs to be corrected in the docstring.

https://github.com/TDAmeritrade/stumpy/blob/c06f0e998d28e7b717bab1e4e32a2aeb897ddfc3/stumpy/core.py#L2297-L2298

stumpy/floss.py
```diff
@@ -570,8 +570,8 @@ def update(self, t):
             T_subseq_isconstant = core.rolling_isconstant(self._T, self._m)
             M_T, Σ_T = core.compute_mean_std(self._T, self._m)
             D = core.mass(
-                self._finite_Q,
-                self._finite_T,
+                self._T[-self._m :],
+                self._T,
                 M_T,
```
NimaSarajpoor, line 575: I revise the code to avoid the postprocessing after the `if/else` block. However, if you think we should keep the postprocessing to increase the safety, I can undo the changes.

stumpy/floss.py
seanlaw: Ahh, you know what? I mis-read `self._T_isfinite` as `self._T_subseq_isfinite`. The former is for `isfinite` for each element of `T` while the latter is the `isfinite` for each subsequence of `T`. Now that makes much more sense why it never failed.

seanlaw: > The description of the output of core.rolling_isfinite needs to be corrected in the docstring.

Sounds good. It was likely copied over from `rolling_nanmax`

stumpy/floss.py
seanlaw: I kind of like how explicit the original version was. I say we revert back to the original

stumpy/floss.py
NimaSarajpoor: @seanlaw 
Feel free to close the PR. I will address it in another PR.

Issue #814, opened 2023-03-18
Discrepency between performant and naive computation in `test_stumpi`


NimaSarajpoor:
It seems that there is a slight discrepency between the performant version of stumpi and its naive computation in `test_stumpi`. This issue is exposed after adding the  `T_subseq_isconstant_func` parameter.

In `stumpy/stumpi.py`, we can see the following line of code:
https://github.com/TDAmeritrade/stumpy/blob/a4bb1e1836e76562974d6b611d1dd71f36eb3202/stumpy/stumpi.py#L139

which computes matrix profile using the performant version `stump`. Note that the approach of performant stump is the same as `naive.stump(T, row_wise=False)`. Also, note that when we call the method `update()`, we update matrix profile in a way that is similar to `naive.stump(row_wise=True)` Because, each time that we update, we calculate the distance profile, and update the matrix profile value/indices accordingly. Hence, we have a mixure of  `row_wise=False` and `row_wise=True` here. However, in `test_stumpi`, we just do `naive.stump(T_updated, row_wise=True)`, as shown below:

https://github.com/TDAmeritrade/stumpy/blob/a4bb1e1836e76562974d6b611d1dd71f36eb3202/tests/test_stumpi.py#L36

Note that `stream._T` is the updated version of `T`. In other words, we apply function `naive.stump(row_wise=True)` on full data whereas in the performant version, we first compute `mp` for the initial chunk via `row_wise==False` approach and then update it in a manner that is similar to `row_wise==True`.

---

To make sure this is the reason behind getting failure in my local test, I am going to add param `mp=None` to stumpi, and then compute`mp` using `naive.stump(T, row_wise=True)` and pass it to the `stumpi`. I think this should resolve the issue and `    npt.assert_almost_equal(ref_left_I, comp_left_I)` will pass.

NimaSarajpoor:
[Update]
Adding the param `mp` to `stumpi`  and passing `naive.stump(T, m, row_wise=True)` as `mp` to the function `stumpi` fixed the issue. So, as I mentioned in my previous comment, we should avoid comparing the `ref_I` and `comp_I` as the way the datapoints are traversed are different. I think we should avoid comparing `ref_I` and `comp_I` in all test functions in `test_stumpi.py`. Logically speaking, these two arrays can be compared only when the datapoints are being travesed in the same manner in both performant and naive version in  a test funcion.

seanlaw:
[...]

PR #815, opened 2023-03-19
Add isconstant param to stumpi

NimaSarajpoor:
To resolve the issue in this PR, I think we should first address #814 (I can do it in this PR if that is okay).  Or, we can just avoid comparing matrix profile indices obtained, in uni testing, via naive and performant version.

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -126,7 +126,6 @@ def non_normalized(non_norm, exclude=None, replace=None):
             "p",
             "T_A_subseq_isconstant",
             "T_B_subseq_isconstant",
-            "T_subseq_isconstant_func",
         ]
 
     @functools.wraps(non_norm)
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -121,7 +121,13 @@ def non_normalized(non_norm, exclude=None, replace=None):
         The desired z-normalized/non-normalized function (or class)
     """
     if exclude is None:
-        exclude = ["normalize", "p", "T_A_subseq_isconstant", "T_B_subseq_isconstant"]
+        exclude = [
+            "normalize",
+            "p",
+            "T_A_subseq_isconstant",
+            "T_B_subseq_isconstant",
+            "T_subseq_isconstant_func",
+        ]
 
     @functools.wraps(non_norm)
     def outer_wrapper(norm):
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -105,7 +105,8 @@ def non_normalized(non_norm, exclude=None, replace=None):
     exclude : list, default None
         A list of function (or class) parameter names to exclude when comparing the
         function (or class) signatures. When `exlcude is None`, this parameter is
-        automatically set to `exclude = ["normalize", "p"]` by default.
+        automatically set to `exclude = ["normalize", "p", "T_A_subseq_isconstant",
+        T_B_subseq_isconstant]` by default.
 
     replace : dict, default None
         A dictionary of function (or class) parameter key-value pairs. Each key that
--- a/stumpy/stumpi.py
+++ b/stumpy/stumpi.py
@@ -42,6 +42,14 @@ class stumpi:
         Note that this will increase the total computational time and memory usage
         when k > 1.
 
+     T_subseq_isconstant_func : function, default None
+        A custom, user-defined function that returns a boolean array that indicates
+        whether a subsequence in `T` is constant (True). The function must only take
+        two arguments, `a`, a 1-D array, and `w`, the window size, while additional
+        arguments may be specified by currying the user-defined function using
+        `functools.partial`. Any subsequence with at least one np.nan/np.inf will
+        automatically have its corresponding value set to False in this boolean array.
+
     Attributes
     ----------
     P_ : numpy.ndarray
@@ -95,7 +103,16 @@ class stumpi:
     array([-1,  0,  1,  2])
     """
 
-    def __init__(self, T, m, egress=True, normalize=True, p=2.0, k=1):
+    def __init__(
+        self,
+        T,
+        m,
+        egress=True,
+        normalize=True,
+        p=2.0,
+        k=1,
+        T_subseq_isconstant_func=None,
+    ):
         """
         Initialize the `stumpi` object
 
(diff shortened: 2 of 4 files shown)


NimaSarajpoor:
@seanlaw 
I resolved #814 locally and the test functions of this module were passing. I decided, however, to not push its corresponding commit and let you see the error and get a chance to discuss #814 first.

stumpy/stumpi.py
```diff
@@ -125,26 +142,54 @@ def __init__(self, T, m, egress=True, normalize=True, p=2.0, k=1):
...
         self._T_isfinite = np.isfinite(self._T)
         self._egress = egress
 
-        mp = stump(self._T, self._m, k=self._k)
+        self._T_subseq_isconstant = core.rolling_isconstant(
+            self._T, self._m, self._T_subseq_isconstant_func
+        )
+
+        mp = stump(
+            T_A=self._T,
```
seanlaw, line 180: `T_A` and `m` are required arguments and so we do not need to specify the `T_A=` or `m=`. While it is nice to be explicit, I think it is pretty consistent throughout the codebase that `T_A` and `m` are required arguments and may be omitted. It's a little bit different if we are calling a function that is rarely used and we need a reminder what the first two arguments are but this is not one of those cases.

seanlaw:
> I resolved https://github.com/TDAmeritrade/stumpy/issues/814 locally and the test functions of this module were passing. I decided, however, to not push its corresponding commit and let you see the error and get a chance to discuss https://github.com/TDAmeritrade/stumpy/issues/814 first.

It looks like the matrix profile distances are not the same. Is that the error that you are referring to? That seems unexpected

stumpy/stumpi.py
NimaSarajpoor: > While it is nice to be explicit, I think it is pretty consistent throughout the codebase that `T_A` and `m` are required arguments and may be omitted. It's a little bit different if we are calling a function that is rarely used and we need a reminder what the first two arguments are but this is not one of those cases.

I agree. The reason that I tried to be explicit here is that we see `T_A_subseq_isconstant=self._T_subseq_isconstant` when we call function `stump`, and I thought it might confuse reader as where that `T_A` comes from.

> it is pretty consistent throughout the codebase that T_A and m are required arguments

Now that you mentioned it, I think the developer probably knows that the first two arguments are `T_A` and  `m`. So, it should be clear. 


I will revise this part.

NimaSarajpoor:
> Also, please see my comments in https://github.com/TDAmeritrade/stumpy/issues/814 regarding how we might proceed next

Sure.

> It looks like the matrix profile distances are not the same. Is that the error that you are referring to? That seems unexpected

I noticed the error is related to the signature of partial function. Maybe you saw a different error (?) I will fix this particular error and will let you know when things are ready.

NimaSarajpoor:
@seanlaw 
Okay...I found it. I locally used the new version of function `isconstant_func_stddev_threshold` based on what you suggested in #810. However, since that PR has not been merged yet, this function in this PR was changed back to its original version when I undo the local changes. I will wait till we merge #810, then I will update this PR and then you can see the error I was talking about.

seanlaw:
> @seanlaw Okay...I found it. I locally used the new version of function `isconstant_func_stddev_threshold` based on what you suggested in #810. However, since that PR has not been merged yet, this function in this PR was changed back to its original version when I undo the local changes. But, I used the new version of this function in the unit test, that is why I got error regarding the signature of partial function.
> 
> I will wait till we merge #810, then I will update this PR and then you can see the error I was talking about.

Right, it looks like `stddev_threshold` isn't available here yet

seanlaw:
@NimaSarajpoor Is this ready to be merged? I took a quick look and it seems okay to me

stumpy/core.py
```diff
@@ -121,7 +121,13 @@ def norm_func(Q, T, A_norm=None, other_norm=None, normalize=True, p=2.0):
         The desired z-normalized/non-normalized function (or class)
     """
     if exclude is None:
-        exclude = ["normalize", "p", "T_A_subseq_isconstant", "T_B_subseq_isconstant"]
+        exclude = [
+            "normalize",
+            "p",
+            "T_A_subseq_isconstant",
+            "T_B_subseq_isconstant",
+            "T_subseq_isconstant_func",
```
NimaSarajpoor, line 129: I added `T_subseq_isconstant_func` here since, in some functions, we have param `T_subseq_isconstant_func`

tests/naive.py
```diff
@@ -950,14 +974,25 @@ def update(self, t):
...
         self._I = np.roll(self._I, -1, axis=0)
         self._left_P[:] = np.roll(self._left_P, -1)
         self._left_I[:] = np.roll(self._left_I, -1)
 
-        D = core.mass(self._T[-self._m :], self._T)
+        D = core.mass(
+            self._T[-self._m :],
+            self._T,
+            T_subseq_isconstant=self._T_subseq_isconstant,
+            Q_subseq_isconstant=self._T_subseq_isconstant[[-1]],
```
NimaSarajpoor, line 994: I am using [-1] as index so that it returns a numpy array rather than just a boolean value. 

Alternative approach: we might add support for `bool`. In other words, if I pass a bool, `core.rolling_isconstant`  converts it to an array of size 1 internally.

tests/test_stumpi.py
```diff
@@ -1119,3 +1121,188 @@ def test_stumpi_self_join_egress_passing_mp():
...
+    naive.replace_inf(ref_P)
+    naive.replace_inf(ref_left_P)
+    naive.replace_inf(comp_P)
+    naive.replace_inf(comp_left_P)
+
+    # comparing matrix profile indices are avoided as, in this case,
+    # the performant version computes matrix profile in a diagonal
+    # manner, which is different than how the naive version computes
+    # the full data (i.e. original `T` including egressed points).
+    # which is in a `row_wise==True` manner
```
NimaSarajpoor, line 1183: I added an expanation as why we are avoiding comparing indices.

tests/test_stumpi.py
```diff
@@ -1119,3 +1121,188 @@ def test_stumpi_self_join_egress_passing_mp():
...
+
+    npt.assert_almost_equal(ref_P, comp_P)
+    # npt.assert_almost_equal(ref_I, comp_I)
+    npt.assert_almost_equal(ref_left_P, comp_left_P)
+    # npt.assert_almost_equal(ref_left_I, comp_left_I)
+
+    # with passing `mp`
```
NimaSarajpoor, line 1190: Instead of adding a new test function, I am doing a new test here where I am passing `mp` to the stumpi so that I can compare matrix profile indices.

tests/test_stumpi.py
```diff
@@ -1119,3 +1121,188 @@ def test_stumpi_self_join_egress_passing_mp():
...
+    naive.replace_inf(comp_left_P)
+
+    npt.assert_almost_equal(ref_P, comp_P)
+    npt.assert_almost_equal(ref_I, comp_I)
+    npt.assert_almost_equal(ref_left_P, comp_left_P)
+    npt.assert_almost_equal(ref_left_I, comp_left_I)
+
+
+def test_stumpi_self_join_egress_with_isconstant():
+    m = 3
```
NimaSarajpoor, line 1247: In this function, I am NOT passing `mp` to the stumpi; however, I can compare matrix profile indices. The reason is that I am updating both performant and naive in the same manner through the for-loop. I did not add any explanation regarding this matter. If needed, I can add a comment.

NimaSarajpoor (commented):
@seanlaw 
Hi Sean. I think everything is good. I just want to bring your attention to a few things.

stumpy/core.py
seanlaw: hmmm, I don't think that is great. I think it should be added to the individual function decorators that have `T_subseq_isconstant_func`

The parameters that I have here are the ubiquitous ones

tests/naive.py
seanlaw: This is probably fine. Thanks for asking

tests/test_stumpi.py
seanlaw: Thank you

tests/test_stumpi.py
seanlaw: Yes, please add a note

stumpy/core.py
NimaSarajpoor: > The parameters that I have here are the ubiquitous ones

Exactly. That was my concern too. Okay...I will revise it. Thanks for the input!

seanlaw:
@NimaSarajpoor Do you think we are good to merge now?

NimaSarajpoor:
I believe everything is good now.

seanlaw:
Thank you @NimaSarajpoor

PR #817, opened 2023-03-22
Fix #621 Add param mp to stumpi

### LINKED ISSUE #621: Allow user to use pre-calculated `mp` in `stumpy.stumpi.py`
It seems that  `stumpy.stumpi.py` needs to calculate `matrix profile (mp)` in the beginning of the algorithm.

https://github.com/TDAmeritrade/stumpy/blob/576de379994896063be71d264e48a08ae17476cf/stumpy/stumpi.py#L118

And, then it will be updated quickly for each new individual, scaler value `t` (because, I think it just needs to calculate a new distance profile ~for~ considering `t` and update matrix profile.

But, what if the initial `T` is big? In that case, user may want to use `stumped` or `gpu_stump` to get `mp`, and then pass it to `stumpy.stumpi.py`. Or, maybe they store it somewhere and decide to use it later.

So, I was wondering if it would be reasonable to do this:
```
def __init__(..., mp=None):
        """
        mp: numpy.ndarray, default None
        the matrix profile of `T`. If user haven't calculated it yet, set it to None. Otherwise, it is user's responsibility to make sure the input 
        `mp` is  correct.
        """
        if mp is None:
             mp = stump(self._T, self._m) # user can now calculate the `mp` using `gpu_stump`
```

NimaSarajpoor:


stumpy/aampi.py
```diff
@@ -35,6 +35,13 @@ class aampi:
         Note that this will increase the total computational time and memory usage
         when k > 1.
 
+    mp : numpy.ndarry, default None
+        A 2D array of shape `(len(T) - m + 1, 2 * k + 2)`, where the first k columns
```
seanlaw, line 39: You've jumped right into the description but you should really start off with "A pre-computed matrix profile (and corresponding matrix profile indices). This is a 2D array..."

tests/naive.py
```diff
@@ -919,7 +922,10 @@ def __init__(self, T, m, excl_zone=None, k=1):
             self._excl_zone = int(np.ceil(self._m / config.STUMPY_EXCL_ZONE_DENOM))
 
         self._l = self._T.shape[0] - m + 1
-        mp = stump(T, m, exclusion_zone=self._excl_zone, k=self._k)
+        if mp is None:
+            mp = stump(T, m, exclusion_zone=self._excl_zone, k=self._k)
```
seanlaw, line 926: use `self._T, self._m`

stumpy/aampi.py
```diff
@@ -104,7 +117,21 @@ def __init__(self, T, m, egress=True, p=2.0, k=1):
...
+        if mp is None:
+            mp = aamp(self._T, self._m, p=self._p, k=self._k)
+        else:
+            mp = mp.copy()
+
+        if mp.shape != (
+            len(self._T) - self._m + 1,
+            2 * self._k + 2,
+        ):  # pragma: no cover
+            msg = (
```
seanlaw, line 129: See comment below

stumpy/aampi.py
NimaSarajpoor: Great point! Thanks for the input. I will revise the docstrings.

stumpy/stumpi.py
```diff
@@ -136,7 +150,21 @@ def __init__(self, T, m, egress=True, normalize=True, p=2.0, k=1):
...
+            mp = stump(self._T, self._m, k=self._k)
+        else:
+            mp = mp.copy()
+
+        if mp.shape != (
+            len(self._T) - self._m + 1,
+            2 * self._k + 2,
+        ):  # pragma: no cover
+            msg = (
+                "The shape of `mp` must match `(len(T)-m+1, 2 * k + 2)`"
```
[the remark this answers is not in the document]
NimaSarajpoor: I was trying to help user understand the components of shape. But...this is not a place for tutorial :) I will revise it.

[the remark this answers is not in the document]
seanlaw: It's not a bad idea but the immediate message should really be that this `mp` doesn't appear to be for this `T`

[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
>  but the immediate message should really be that this mp doesn't appear to be for this T.

Just to make sure that I am not missing anything here, are you thinking of adding a clearer explanation to the beginning of the error message, like..
```
msg = (
     "The provided `mp` does not appear to be the matrix profile of `T`, `m`, and `k` due to the shape mismatch. The shape of `mp` must match ..."
)
```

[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
> are you thinking of adding a clearer explanation to the beginning of the error message

If needed, I can revise and push again.

tests/test_aampi.py
```diff
@@ -1146,3 +1146,64 @@ def test_aampi_self_join_egress_KNN():
...
+        mp = naive.aamp(T, m, p=p)
+
+        ref_mp = naive.aampi_egress(T, m, p=p, mp=mp)
+        ref_P = ref_mp.P_.copy()
+        ref_I = ref_mp.I_
+        ref_left_P = ref_mp.left_P_.copy()
+        ref_left_I = ref_mp.left_I_
+
+        stream = aampi(T, m, egress=True, p=p, mp=mp)
+
```
NimaSarajpoor, line 1169: No sure if this is a good way to test the new param `mp`. I pass `mp` to both naive and performant `floss`. Alternatively, we can just pass it to performant `floss`, and then compare it with naive version when `mp` is computed internally.

NimaSarajpoor (commented):
@seanlaw 
Do you think I should revise the test functions? So, as we discussed before, I can remove any comparision between matrix profile indices unless they both follow `row_wise==True` approach in computing/updating matrix profile

seanlaw:
> Do you think I should revise the test functions? So, as we discussed before, I can remove any comparision between matrix profile indices unless they both follow row_wise==True approach in computing/updating matrix profile

Whenever possible, I like to keep the comparison of matrix profile indices. For the existing tests, I don't think anything should be affected. It is only when we have an array that actually contains constant subsequences where comparisons of the indices may be a problem. Only in those cases (where there is a constant subsequence), then maybe you just pass in the pre-computed `mp`. So we test BOTH pre-computed `mp` along with `T_subseq_isconstant` at the same time for the same unit tests. It's not ideal but I am okay with it as long as we get full coverage

NimaSarajpoor:
> Only in those cases (where there is a constant subsequence), then maybe you just pass in the pre-computed `mp`. So we test BOTH pre-computed `mp` along with `T_subseq_isconstant` at the same time for the same unit tests. It's not ideal but I am okay with it as long as we get full coverage

Correct! As long as there is only one index with global minima, we should be fine when we compare matrix profile indices.

Okay....so, for the sake of this PR only, I believe the added test functions should suffice as we get full coverage for now. Then, after merging this PR, we can resume the work on PR #817, where we add test functions for `T_subseq_isconstant`

stumpy/aampi.py
```diff
@@ -104,7 +119,21 @@ def __init__(self, T, m, egress=True, p=2.0, k=1):
...
+                f"The shape of `mp` must match ({len(T)-m+1}, {2 * k + 2})"
```
seanlaw, line 132: Can you please add " but " (note the space after "but")

stumpy/stumpi.py
```diff
@@ -136,7 +152,21 @@ def __init__(self, T, m, egress=True, normalize=True, p=2.0, k=1):
...
+            mp = stump(self._T, self._m, k=self._k)
+        else:
+            mp = mp.copy()
+
+        if mp.shape != (
+            len(self._T) - self._m + 1,
+            2 * self._k + 2,
+        ):  # pragma: no cover
+            msg = (
+                f"The shape of `mp` must match ({len(T)-m+1}, {2 * k + 2})"
```
seanlaw, line 165: Please add " but " at the end of line 165

tests/test_aampi.py
seanlaw: Passing `mp` to both seems reasonable

seanlaw (changes requested):
@NimaSarajpoor A couple of minor comments and then I think we are ready to merge when you agree.

stumpy/aampi.py
seanlaw:
```
f"The shape of `mp` must match ({len(T)-m+1}, {2 * k + 2}) but "
```

Otherwise, there is no space between this line and the next line

NimaSarajpoor: I forgot adding dot or space. Thanks for pointing this out! This should be fixed now.

PR #820, opened 2023-03-25
Add support for "isconstant function" in stamp

NimaSarajpoor:


stumpy/stamp.py
```diff
@@ -12,11 +12,12 @@ def _mass_PI(
     T,
     M_T,
     Σ_T,
-    T_subseq_isconstant,
     trivial_idx=None,
     excl_zone=0,
     left=False,
     right=False,
+    T_subseq_isconstant=None,
+    Q_subseq_isconstant=None,
```
NimaSarajpoor, line 20: `Q_subseq_isconstant` was missing here. So, I added it. In addition, I put the isconstant params at the end so that the order of arguments becomes more similar to what we have in `core.mass`.

tests/test_stamp.py
```diff
@@ -26,13 +28,12 @@ def test_stamp_mass_PI(T_A, T_B):
     trivial_idx = 2
     zone = int(np.ceil(m / 2))
     Q = T_B[trivial_idx : trivial_idx + m]
-    T_subseq_isconstant = core.rolling_isconstant(T_B, m)
```
NimaSarajpoor, line 29: This line is removed as the purpose of this test function is not to test the support for isconstant param. Instead, I added new test functions at the end of the `tests/test_stamp.py`

tests/test_stamp.py
```diff
@@ -149,3 +148,141 @@ def test_stamp_nan_zero_mean_self_join():
...
+    )
+
+    m = 3
+    zone = int(np.ceil(m / 2))
+
+    T_B_subseq_isconstant = naive.rolling_isconstant(T_B, m, isconstant_custom_func)
+    M_T, Σ_T = core.compute_mean_std(T_B, m)
+
+    trivial_idx = np.random.choice(np.flatnonzero(~T_B_subseq_isconstant))
+    Q = T_B[trivial_idx : trivial_idx + m]
```
NimaSarajpoor, line 167: In this test function, I am choosing a Q that is not constant. And, in the next test function, I am choosing a Q that is constant.

NimaSarajpoor (commented):
I provided a few comments that you may want to take a look at.

stumpy/stomp.py
```diff
@@ -100,13 +100,35 @@ def _stomp(T_A, m, T_B=None, ignore_trivial=True):
...
             P, I = stamp._mass_PI(
-                T_A[:m], T_B, M_T, Σ_T, T_subseq_isconstant, 0, excl_zone
+                T_A[:m],
+                T_B,
+                M_T,
+                Σ_T,
+                0,
+                excl_zone,
+                T_subseq_isconstant=T_subseq_isconstant,
+                Q_subseq_isconstant=Q_subseq_isconstant[[0]],
```
NimaSarajpoor, line 110: I haven't added support for `subseq_isconstant` to `stomp.py` yet. According to its docstring, this function is not supported and is only here for reference. So, should we add `subseq_isconstant` param to it? If not, then we can just avoid using `T_subseq_isconstant` and `Q_subseq_isconstant` within this function and do not pass them to `stamp._mass_PI`.

tests/test_stamp.py
seanlaw: Maybe add a note

NimaSarajpoor: @seanlaw 
There is actually a comment just in the first line of this test function. 
```
def test_stamp_mass_PI_with_isconstant_case1():
    # case1: The query `Q` is not constant
```

Does that suffice?

(There is a similar comment for the test function "test_stamp_mass_PI_with_isconstant_case2")

stumpy/stomp.py
seanlaw: Yes, please skip stomp

NimaSarajpoor:
@seanlaw 
I checked the changes. I think everything is good. 

There is just one thing I would like to discuss. In some cases, we have support for both `self-join` and `AB-join`. Also, we would like param `subseq_isconstant` to accept function or a numpy array in some APIs. Hence, ideally we should add four test functions. However, I have not been doing that so far. In `tests/test_stump.py` , I only checked "self-join" (not "AB-join") I tested it for two different cases (one case is when `subseq_isconstant` is numpy array and the other case is when it is a function). In this PR, two new test functions in `tests/test_stamp.py` cover "self-join" and "AB-join"; however, in both cases, param `subseq_isconstant` is a numpy array (not a function). I know this is not ideal. But, from a practical point of view, do you think it is good enough for now?

NimaSarajpoor:
@seanlaw 
Cool. Please allow me to revise the test functions.

NimaSarajpoor:
@seanlaw 
I think it is good to be merged.

seanlaw:
Thanks @NimaSarajpoor for this contribution!

PR #821, opened 2023-03-27
Add param isconstant to mpdist

NimaSarajpoor:
`tests/test_mpdist.py` contains this test function 

https://github.com/TDAmeritrade/stumpy/blob/f8a8e937cfb78b58ab5cbbce2d13d2a4a0b8b4ff/tests/test_mpdist.py#L50-L62

which tests `core._compute_P_ABBA`. I think we should move this test function to `tests/test_core.py`. `test_mpdist` is another test function that should also be moved to test_core.py.

(Alternatively, we may move the performant functions from `stumpy/core.py` to `stumpy/mpdist.py`)

### DIFF
--- a/stumpy/aampdist.py
+++ b/stumpy/aampdist.py
@@ -127,8 +127,8 @@ def aampdist(T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    mp_func = functools.partial(aamp, p=p)
-    return core._mpdist(T_A, T_B, m, mp_func, percentage, k)
+    partial_mp_func = functools.partial(aamp, p=p)
+    return core._mpdist(T_A, T_B, m, partial_mp_func, percentage, k)
 
 
 def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
@@ -184,5 +184,5 @@ def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    mp_func = functools.partial(aamped, p=p)
-    return core._mpdist(T_A, T_B, m, mp_func, percentage, k, client=client)
+    partial_mp_func = functools.partial(aamped, p=p)
+    return core._mpdist(T_A, T_B, m, partial_mp_func, percentage, k, client=client)
--- a/stumpy/aampdist.py
+++ b/stumpy/aampdist.py
@@ -3,6 +3,7 @@
 # STUMPY is a trademark of TD Ameritrade IP Company, Inc. All rights reserved.
 
 import math
+from functools import partial
 
 import numpy as np
 
@@ -126,7 +127,8 @@ def aampdist(T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    return core._mpdist(T_A, T_B, m, aamp, percentage, k, p=p)
+    mp_func = partial(aamp, p=p)
+    return core._mpdist(T_A, T_B, m, mp_func, percentage, k)
 
 
 def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
@@ -182,4 +184,5 @@ def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    return core._mpdist(T_A, T_B, m, aamped, percentage, k, client=client, p=p)
+    mp_func = partial(aamped, p=p)
+    return core._mpdist(T_A, T_B, m, mp_func, percentage, k, client=client)
--- a/stumpy/aampdist.py
+++ b/stumpy/aampdist.py
@@ -126,7 +126,7 @@ def aampdist(T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    return core._aampdist(T_A, T_B, m, aamp, percentage, k, p=p)
+    return core._mpdist(T_A, T_B, m, aamp, percentage, k, p=p)
 
 
 def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
@@ -182,4 +182,4 @@ def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    return core._aampdist(T_A, T_B, m, aamped, percentage, k, client=client, p=p)
+    return core._mpdist(T_A, T_B, m, aamped, percentage, k, client=client, p=p)
--- a/stumpy/aampdist.py
+++ b/stumpy/aampdist.py
@@ -2,7 +2,6 @@
 # Copyright 2019 TD Ameritrade. Released under the terms of the 3-Clause BSD license.
 # STUMPY is a trademark of TD Ameritrade IP Company, Inc. All rights reserved.
 
-import functools
 import math
 
 import numpy as np
@@ -127,8 +126,7 @@ def aampdist(T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    partial_mp_func = functools.partial(aamp, p=p)
-    return core._mpdist(T_A, T_B, m, partial_mp_func, percentage, k)
+    return core._aampdist(T_A, T_B, m, aamp, percentage, k, p=p)
 
 
 def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
@@ -184,5 +182,4 @@ def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    partial_mp_func = functools.partial(aamped, p=p)
-    return core._mpdist(T_A, T_B, m, partial_mp_func, percentage, k, client=client)
+    return core._aampdist(T_A, T_B, m, aamped, percentage, k, client=client, p=p)
--- a/stumpy/aampdist.py
+++ b/stumpy/aampdist.py
@@ -2,8 +2,8 @@
 # Copyright 2019 TD Ameritrade. Released under the terms of the 3-Clause BSD license.
 # STUMPY is a trademark of TD Ameritrade IP Company, Inc. All rights reserved.
 
+import functools
 import math
-from functools import partial
 
 import numpy as np
 
@@ -127,7 +127,7 @@ def aampdist(T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    mp_func = partial(aamp, p=p)
+    mp_func = functools.partial(aamp, p=p)
     return core._mpdist(T_A, T_B, m, mp_func, percentage, k)
 
 
@@ -184,5 +184,5 @@ def aampdisted(client, T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    mp_func = partial(aamped, p=p)
+    mp_func = functools.partial(aamped, p=p)
     return core._mpdist(T_A, T_B, m, mp_func, percentage, k, client=client)
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -3910,6 +3910,8 @@ def _mpdist(
     client=None,
     device_id=None,
     custom_func=None,
+    T_A_subseq_isconstant=None,
+    T_B_subseq_isconstant=None,
 ):
     """
     A convenience function for computing the matrix profile distance (MPdist) measure
@@ -3964,6 +3966,26 @@ def _mpdist(
         `MPdist` value. The `percentage` and `k` parameters are ignored when
         `custom_func` is not None.
 
+    T_A_subseq_isconstant : numpy.ndarray or function, default None
+        A boolean array that indicates whether a subsequence in `T_A` is constant
+        (True). Alternatively, a custom, user-defined function that returns a
+        boolean array that indicates whether a subsequence in `T_A` is constant
+        (True). The function must only take two arguments, `a`, a 1-D array,
+        and `w`, the window size, while additional arguments may be specified
+        by currying the user-defined function using `functools.partial`. Any
+        subsequence with at least one np.nan/np.inf will automatically have its
+        corresponding value set to False in this boolean array.
+
+    T_B_subseq_isconstant : numpy.ndarray or function, default None
+        A boolean array that indicates whether a subsequence in `T_B` is constant
+        (True). Alternatively, a custom, user-defined function that returns a
+        boolean array that indicates whether a subsequence in `T_B` is constant
+        (True). The function must only take two arguments, `a`, a 1-D array,
+        and `w`, the window size, while additional arguments may be specified
+        by currying the user-defined function using `functools.partial`. Any
+        subsequence with at least one np.nan/np.inf will automatically have its
+        corresponding value set to False in this boolean array.
+
     Returns
     -------
     MPdist : float
@@ -3980,7 +4002,17 @@ def _mpdist(
     n_B = T_B.shape[0]
     P_ABBA = np.empty(n_A - m + 1 + n_B - m + 1, dtype=np.float64)
 
-    _compute_P_ABBA(T_A, T_B, m, P_ABBA, mp_func, client, device_id)
+    _compute_P_ABBA(
+        T_A,
+        T_B,
+        m,
+        P_ABBA,
+        mp_func,
+        client,
+        device_id,
+        T_A_subseq_isconstant=T_A_subseq_isconstant,
+        T_B_subseq_isconstant=T_B_subseq_isconstant,
+    )
 
     if k is not None:
         k = min(int(k), P_ABBA.shape[0] - 1)
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -2492,7 +2492,7 @@ def _get_partial_mp_func(mp_func, client=None, device_id=None):
     elif device_id is not None:
         partial_mp_func = functools.partial(mp_func, device_id=device_id)
     else:
-        partial_mp_func = mp_func
+        partial_mp_func = functools.partial(mp_func)
 
     return partial_mp_func
 
@@ -3805,10 +3805,6 @@ def _compute_P_ABBA(
     mp_func,
     client=None,
     device_id=None,
-    normalize=True,
-    p=2.0,
-    T_A_subseq_isconstant=None,
-    T_B_subseq_isconstant=None,
 ):
     """
     A convenience function for computing the (unsorted) concatenated matrix profiles
@@ -3850,35 +3846,6 @@ def _compute_P_ABBA(
         computation. A list of all valid device ids can be obtained by
         executing `[device.id for device in numba.cuda.list_devices()]`.
 
-    normalize : bool, default True
-        When set to `True`, this z-normalizes subsequences prior to computing distances.
-        Otherwise, this function gets re-routed to its complementary non-normalized
-        equivalent set in the `@core.non_normalized` function decorator.
-
-    p : float, default 2.0
-        The p-norm to apply for computing the Minkowski distance. This parameter is
-        ignored when `normalize == True`.
-
-    T_A_subseq_isconstant : numpy.ndarray or function, default None
-        A boolean array that indicates whether a subsequence in `T_A` is constant
-        (True). Alternatively, a custom, user-defined function that returns a
-        boolean array that indicates whether a subsequence in `T_A` is constant
-        (True). The function must only take two arguments, `a`, a 1-D array,
-        and `w`, the window size, while additional arguments may be specified
-        by currying the user-defined function using `functools.partial`. Any
-        subsequence with at least one np.nan/np.inf will automatically have its
-        corresponding value set to False in this boolean array.
-
-    T_B_subseq_isconstant : numpy.ndarray or function, default None
-        A boolean array that indicates whether a subsequence in `T_B` is constant
-        (True). Alternatively, a custom, user-defined function that returns a
-        boolean array that indicates whether a subsequence in `T_B` is constant
-        (True). The function must only take two arguments, `a`, a 1-D array,
-        and `w`, the window size, while additional arguments may be specified
-        by currying the user-defined function using `functools.partial`. Any
-        subsequence with at least one np.nan/np.inf will automatically have its
-        corresponding value set to False in this boolean array.
-
     Returns
     -------
     None
@@ -3893,7 +3860,10 @@ def _compute_P_ABBA(
     n_A = T_A.shape[0]
     partial_mp_func = _get_partial_mp_func(mp_func, client=client, device_id=device_id)
 
-    if inspect.signature(mp_func).parameters.get("normalize") is not None:
+    if inspect.signature(partial_mp_func).parameters.get("normalize") is not None:
+        params = partial_mp_func.keywords
+        T_A_subseq_isconstant = params.get("T_A_subseq_isconstant")
+        T_B_subseq_isconstant = params.get("T_B_subseq_isconstant")
         P_ABBA[: n_A - m + 1] = partial_mp_func(
             T_A,
             m,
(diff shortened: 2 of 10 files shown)


seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

seanlaw:
> which tests core._compute_P_ABBA. I think we should move this test function to tests/test_core.py. test_mpdist is another test function that should also be moved to test_core.py.
(Alternatively, we may move the performant functions from stumpy/core.py to stumpy/mpdist.py)

Yes, can you please move them over to `test_core.py`. I moved the `_compute_P_ABBA` from `mpdist.py` into `core.py` when I was resolving the circular dependencies and left the tests. However, it makes sense to move them accordingly

NimaSarajpoor:
@seanlaw 
Before moving the aformentioned test functions to `tests/test_core.py`, I wanted to fix the code to make sure all unit tests are passing. I couldn't find an easy way to add param "subseq_isconstant" to `core._mpdist`. This is because this function is also called by a non-normalized function `stumpy/aampdist.py`. Note that we cannot use partial function to pass `subseq_isconstant`. In other words, the following code does not work:

```
def mpdist(T_A, T_B, m, ..., T_A_subseq_isconstnat=None, T_B_subseq_isconstnat=None):
    stump_partial = functools.partial(
                                       stump, 
                                       T_A_subseq_isconstnat=T_A_subseq_isconstnat, 
                                       T_B_subseq_isconstnat=T_B_subseq_isconstnat,
                                      )
    
    return core._mpdist(T_A, T_B, m, stump_partial, ....)
```

The reason is that the function `core._mpdist` calls the function `core._compute_P_ABBA`, which computes both `stump(T_A, m, T_B, ...)`  and `stump(T_B, m, T_A, ...)`. Therefore, we need to pass the params `T_A_subseq_isconstnat` and `T_B_subseq_isconstnat` all the way to `core._compute_P_ABBA` so that they can be passed properly to both calls of `stumpy.stump`.

---

I resolved the issue by adding the non-normalized version of `core._mpdist` and `core._compute_P_ABBA`.

seanlaw:
> I resolved the issue by adding the non-normalized version of core._mpdist and core._compute_P_ABBA.

@NimaSarajpoor I don't really like doing it this way as it increases the complexity and reduces the maintainability. Instead, is it possible to:

1. add `T_A_subseq_isconstant` and `T_B_subseq_isconsant` to `core._mpdist` and `core._compute_P_ABBA` with default values of `None`? In other words, create partial functions (for normalized functions) but OMIT `T_A_subseq_isconstant` and `T_B_subseq_isconsant` in the partial function and pass these two params directly into `core._mpdist`
2. Inside of `core._compute_P_ABBA`, check whether the partial function is normalized/non-normalized by doing something like:

```
if inspect.signature(mp_func).parameters.get("normalize") is not None:
    # Only non-normalized mp funcs can have a "noramlize" parameter in its function signature
    P_ABBA[: n_A - m + 1] = partial_mp_func(T_A, m, T_B, ignore_trivial=False, T_A_subseq_isconstant=T_A_isconstant, T_B_subseq_isconstant=T_B_subseq_isconstant)[:, 0]
    P_ABBA[n_A - m + 1 :] = partial_mp_func(T_B, m, T_A, ignore_trivial=False, T_B_subseq_isconstant=T_A_isconstant, T_B_subseq_isconstant=T_A_subseq_isconstant)[:, 0]
else:
    # Ignore `T_A_subseq_isconstant` and `T_B_subseq_isconstant` parameters for all non-normalized mp funcs 
    P_ABBA[: n_A - m + 1] = partial_mp_func(T_A, m, T_B, ignore_trivial=False)[:, 0]
    P_ABBA[n_A - m + 1 :] = partial_mp_func(T_B, m, T_A, ignore_trivial=False)[:, 0]
```

NimaSarajpoor:
> I am trying to avoid managing separate `_mpdist` and `_compute_P_ABBA` functions for norm/non-norm

I see your point. Cool. As you said, your approach is more maintable.  I will try it.

stumpy/aampdist.py
```diff
@@ -127,8 +126,7 @@ def aampdist(T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    partial_mp_func = functools.partial(aamp, p=p)
-    return core._mpdist(T_A, T_B, m, partial_mp_func, percentage, k)
+    return core._mpdist(T_A, T_B, m, aamp, percentage, k, p=p)
```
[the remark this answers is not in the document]
NimaSarajpoor: I wanted to be consistent. In constrast to `stump`, where we switch to `aamp` when `normalize==False` (and it has its own private functions),  we do not do such thing here. I mean...we use `aamp` here but we pass argument "subseq_isconstant" all the way to "_compute_P_ABBA" and then we decide whether we should use it or not depending on the existence of param `normalize` in the `mp_func`. 

Using `functools.partial` gives me this feeling that we are trying to do the same thing but in two different places: when `normalize==False`, we handle `p` (the arg that is only supported when `normalize==False`) using partial, and then when `normalize==True`, we handle "subseq_isconstans" (the arg that is only supported when `normalize==True`) in `_compute_P_ABBA`.  

Therefore, I thought we could just pass all args and then if normalize, we just use the ones that are only supported when normalize==True, and when normalize==False, we use the arguments that are supported only if `normalize==False`. I wanted to handle both at one place. So, later, if we decide to add more and more arguemtns, we can just pass them all the way to `_compute_P_ABBA` and then separate the normalized / non-normalized paths there at the end.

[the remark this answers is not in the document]
seanlaw: Hmm, I've considered and understand your response but I don't think it is wise to open up `core._mpdist` or `core._compute_P_ABBA` and force those functions to require `normalize`, `p`. At the end of the day, neither of those functions are actually using those parameters themselves (i.e., it's the underlying `mp_func` that would use those parameters) and I think it's better to abstract that away so that those functions are very general. One thing to keep in mind is that `_mpdist` and `_compute_P_ABBA` are both private functions. The more we add parameters like `normalize`, `p`, and other parameters to these functions then we are forced to remember to add them all the way down the chain of function calls. By wrapping them up in `functools.partial` at the highest layer, we simply package it up once and avoid a bunch of `if/else` statements. Now that I think about it more, we could still wrap everything using `functools.partial` and check for the existence of `T_A_subseq_isconstant`/`T_B_subseq_isconstant` inside of `core._compute_P_ABBA` by checking the `keywords` that are wrapped by `functools.partial`:

```
from functools import partial
import inspect
from stumpy import stump

partial_mp_func = partial(stump, T_A_subseq_isconstant=T_A_subseq_isconstant, T_B_subseq_isconstant=T_B_subseq_isconstant)    
```

Then, inside of `core._compute_P_ABBA` (note that we'd remove/omit `T_A_subseq_isconstant`/`T_B_subseq_isconstant` parameters from BOTH `core._mpdist` and `core._compute_P_ABBA`), you'd do:

```
if inspect.signature(mp_func).parameters.get("normalize") is not None:
    # Non-normalized (stump-like)
    # Only non-normalized mp funcs can have a "normalize" parameter in its function signature
    params = partial_mp_func.keywords
    T_A_subseq_isconstant = params.get("T_A_subseq_isconstant")
    T_B_subseq_isconstant = params.get("T_B_subseq_isconstant")
    P_ABBA[: n_A - m + 1] = partial_mp_func(T_A, m, T_B, ignore_trivial=False, T_A_subseq_isconstant=T_A_isconstant, T_B_subseq_isconstant=T_B_subseq_isconstant)[:, 0]
    P_ABBA[n_A - m + 1 :] = partial_mp_func(T_B, m, T_A, ignore_trivial=False, T_B_subseq_isconstant=T_A_isconstant, T_B_subseq_isconstant=T_A_subseq_isconstant)[:, 0]
else:
    # Normalized (aamp-like)
    # Ignore/omit `T_A_subseq_isconstant` and `T_B_subseq_isconstant` parameters for all non-normalized mp funcs 
    P_ABBA[: n_A - m + 1] = partial_mp_func(T_A, m, T_B, ignore_trivial=False)[:, 0]
    P_ABBA[n_A - m + 1 :] = partial_mp_func(T_B, m, T_A, ignore_trivial=False)[:, 0]
``` 

Here we auto-discover the necessary `T_A_subseq_isconstant`/`T_B_subseq_isconstant` if they exist. If not, the default Dict `get` would/should return `None`. And that should be it.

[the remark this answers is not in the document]
NimaSarajpoor: Thank you for the thorough explanation. 

> One thing to keep in mind is that _mpdist and _compute_P_ABBA are both private functions. The more we add parameters like normalize, p, and other parameters to these functions then we are forced to remember to add them all the way down the chain of function calls. By wrapping them up in functools.partial at the highest layer, we simply package it up once and avoid a bunch of if/else statements.

Great point! I should keep that in mind! 

> by checking the keywords that are wrapped by functools.partial

Cool! I will revise the code.

stumpy/core.py
```diff
@@ -2492,7 +2492,7 @@ def _get_partial_mp_func(mp_func, client=None, device_id=None):
     elif device_id is not None:
         partial_mp_func = functools.partial(mp_func, device_id=device_id)
     else:
-        partial_mp_func = mp_func
+        partial_mp_func = functools.partial(mp_func)
```
NimaSarajpoor, line 2495: I made the change above so that the output of this functions always has type `functools.partial`. Otherwise, in function `core._compute_P_ABBA`, where we have:

```
# in core._compute_P_ABBA
...
partial_mp_func = _get_partial_mp_func(mp_func, client=client, device_id=device_id)
if inspect.signature(partial_mp_func).parameters.get("normalize") is not None:
        params = partial_mp_func.keywords
```

we first need to check if `partial_mp_func` is of type `function` or `functools.partial`, and if it is the former, we cannot use the attribute `.keywords `. 

What do you think @seanlaw ? If you are not okay with this, then we can just add a code (like this one: https://stackoverflow.com/questions/12627118/get-a-function-arguments-default-value) to get default value of argument "subseq_isconstant". Or, we can initialize them with `None`, and get their value only if `partial_mp_func` is of type `functools.partial`.

seanlaw (changes requested):

stumpy/core.py
```diff
@@ -3852,8 +3860,29 @@ def _compute_P_ABBA(T_A, T_B, m, P_ABBA, mp_func, client=None, device_id=None):
     n_A = T_A.shape[0]
     partial_mp_func = _get_partial_mp_func(mp_func, client=client, device_id=device_id)
 
-    P_ABBA[: n_A - m + 1] = partial_mp_func(T_A, m, T_B, ignore_trivial=False)[:, 0]
-    P_ABBA[n_A - m + 1 :] = partial_mp_func(T_B, m, T_A, ignore_trivial=False)[:, 0]
+    if inspect.signature(partial_mp_func).parameters.get("normalize") is not None:
```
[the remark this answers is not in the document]
NimaSarajpoor: I will add them.

stumpy/aampdist.py
```diff
@@ -127,8 +127,8 @@ def aampdist(T_A, T_B, m, percentage=0.05, k=None, p=2.0):
 
     See Section III
     """
-    partial_mp_func = functools.partial(aamp, p=p)
-    return core._mpdist(T_A, T_B, m, partial_mp_func, percentage, k)
+    mp_func = partial(aamp, p=p)
```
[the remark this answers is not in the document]
NimaSarajpoor: Sure. I will revise the name of variable.

stumpy/aampdist.py
```diff
@@ -2,8 +2,8 @@
 # Copyright 2019 TD Ameritrade. Released under the terms of the 3-Clause BSD license.
 # STUMPY is a trademark of TD Ameritrade IP Company, Inc. All rights reserved.
 
-import functools
 import math
+from functools import partial
```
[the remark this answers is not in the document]
NimaSarajpoor: Got it! I will revise it.

stumpy/core.py
NimaSarajpoor: @seanlaw 
We are calling `_get_partial_mp_func` in `stumpy.stimp`, `stumpy.ostinato`, and their non-normalize versions. 

https://github.com/TDAmeritrade/stumpy/blob/814977f110403bcfd1629bcf1ba4f90938394bfe/stumpy/stimp.py#L188-L191

we may just change `self._mp_func` to `self._partial_mp_func`

---

https://github.com/TDAmeritrade/stumpy/blob/814977f110403bcfd1629bcf1ba4f90938394bfe/stumpy/ostinato.py#L228-L239

seanlaw: > we may just change self._mp_func to self._partial_mp_func

Yeah, I think that would be a good idea as it would be slightly more consistent

stumpy/core.py
```diff
@@ -3852,8 +3860,29 @@ def _compute_P_ABBA(T_A, T_B, m, P_ABBA, mp_func, client=None, device_id=None):
...
+            ignore_trivial=False,
+            T_A_subseq_isconstant=T_B_subseq_isconstant,
+            T_B_subseq_isconstant=T_A_subseq_isconstant,
+        )[:, 0]
+    else:
+        P_ABBA[: n_A - m + 1] = partial_mp_func(T_A, m, T_B, ignore_trivial=False)[:, 0]
+        P_ABBA[n_A - m + 1 :] = partial_mp_func(T_B, m, T_A, ignore_trivial=False)[:, 0]
 
 
 def _mpdist(
```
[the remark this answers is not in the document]
seanlaw: 1. I think the `mp_func` currently for `_mpdist` should be a `functools.partial` because we want to wrap all of the `T_x_subseq_isconstant` inside of that `mp_func`. Thus, I think we should just call the `_mpdist` input parameter as `partial_mp_func` instead of `mp_func`. For example, regardless of whether it is `stump`, `stumped`, or `gpu_stump`, we want all `T_x_subseq_isconstant` to be wrapped inside of it as a `functools.partial` and then used as input for `_mpdist`
2. Inside of `_mpdist`, we'd pass this partial function into `_compute_P_ABBA`, which then further wraps the partial function a second time so that the function adds additional relevant parameters (e.g., `client`, `device_id`). 
3. Inside of `_get_partial_mp_func`, I would check if the `mp_func` is already a partial function. If not, then wrap it. Otherwise, `partial_mp_func = mp_func`:

```
    if client is not None:
        partial_mp_func = functools.partial(mp_func, client)
    elif device_id is not None:
        partial_mp_func = functools.partial(mp_func, device_id=device_id)
    elif type(mp_func) != functools.partial:
        partial_mp_func = functools.partial(mp_func)
    else:
        partial_mp_func = mp_func

    return partial_mp_func
```

seanlaw (changes requested):

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: > 1. I think the `mp_func` currently for `_mpdist` should be a `functools.partial` because we want to wrap all of the `T_x_subseq_isconstant` inside of that `mp_func`. Thus, I think we should just call the `_mpdist` input parameter as `partial_mp_func`

Good point! I will revise the code.

stumpy/gpu_mpdist.py
```diff
@@ -85,6 +87,9 @@ def gpu_mpdist(
     ...         device_id=all_gpu_devices)
     0.00019935236191097894
     """
-    MPdist = core._mpdist(T_A, T_B, m, gpu_stump, percentage, k, device_id=device_id)
+    partial_gpu_stump = functools.partial(gpu_stump)
```
seanlaw, line 90: Don't we need to add `T_X_subseq_isconstant` to this function?

seanlaw (changes requested):

stumpy/gpu_mpdist.py
NimaSarajpoor: We do. I am planning to do it in this PR after taking care of the current issues and moving some test functions to test_core.

NimaSarajpoor:
@seanlaw 
* a function is moved to `tests/naive.py`. 
* A couple of test functions were moved to `tests/test_core.py`
* param `T_x_subseq_isconstant` were added to `gpu_stump` and a test function is added to `tests/test_gpu_stump.py` 
* the unit testing is conducted on google colab with GPU and all tests are passing with 100% coverage.

tests/naive.py
```diff
@@ -2129,3 +2149,16 @@ def isconstant_func_stddev_threshold(a, w, quantile_threshold=0, stddev_threshol
             stddev_threshold = 0
 
     return sliding_stddev <= stddev_threshold
+
+
+def select_mpdist(P_ABBA, m, percentage, n_A, n_B):
```
seanlaw, line 2154: what is this? why is it called "select"? later, why do we compare core._mpdist with this function?

NimaSarajpoor: I moved the following function

https://github.com/TDAmeritrade/stumpy/blob/814977f110403bcfd1629bcf1ba4f90938394bfe/tests/test_mpdist.py#L153-L166

to `tests/test_core.py`, and I just made slight changes. But before that, I would like to bring your attention to one thing here. I am not sure if this is a correct way of testing `custom_func` since the `custom_func` we are passing here is just for `percentage = 0.05`, and note that the naive version is being computed only based on `T_A`, `T_B`, and `m`. In other words, we are just testing the DEFAULT value of `custom_func`. 

---

I just moved `some_func` to `tests/naive.py` and tried to just rename it as `some_func` was a bit strange for me.

seanlaw: >  I am not sure if this is a correct way of testing custom_func since the custom_func we are passing here is just for percentage = 0.05, and note that the naive version is being computed only based on T_A, T_B, and m. In other words, we are just testing the DEFAULT value of custom_func.

1. Since we don't actually use `k` anywhere in this test, we should remove `@pytest.mark.parametrize("k", k)` and update the function definition to `def test_mpdist_custom_func(T_A, T_B):` (note the removal of `k`)
2. The goal of this test is simply to see if `_mpdist` understands how to accept a custom function and to infer `k` from it. From what I can see, there is no default `percentage` value for `custom_func` but we are testing `custom_func` against the default `percentage` value of `_mpdist`, which is perfectly fine. Again, the goal is to simply check if the `custom_func`, which is a wrapped partial function, can produce the same outputs via `_mpdist` as the naive version when neither `k` nor `percentage` is provided as parameters to `_mpdist`. This is the most basic test but, I think, a reasonable one as-is

seanlaw: > I just moved some_func to tests/naive.py and renamed it to select_mpdist since the name some_func was a bit strange for me. I think we are just using this function (i.e. some_func, which is now: select_mpdist) in one test function. So, we may do:

Thank you. Moving it is fine. Instead of doing a closure, I think simply renaming it from the current `select_mpdist` to `mpdist_custom_func` would be fine

NimaSarajpoor: > 1. Since we don't actually use `k` anywhere in this test, we should remove `@pytest.mark.parametrize("k", k)` and update the function definition to `def test_mpdist_custom_func(T_A, T_B):` (note the removal of `k`)

Right! I will revise it. Thank you for catching it.

> 2. The goal of this test is simply to see if `_mpdist` understands how to accept a custom function and to infer `k` from it. From what I can see, there is no default `percentage` value for `custom_func` but we are testing `custom_func` against the default `percentage` value of `_mpdist`, which is perfectly fine. Again, the goal is to simply check if the `custom_func`, which is a wrapped partial function, can produce the same outputs via `_mpdist` as the naive version when neither `k` nor `percentage` is provided as parameters to `_mpdist`. This is the most basic test but, I think, a reasonable one as-is

Got it now. Thank you for your explanation!

NimaSarajpoor: @seanlaw 

> 1. Since we don't actually use `k` anywhere in this test, we should remove `@pytest.mark.parametrize("k", k)` and update the function definition to `def test_mpdist_custom_func(T_A, T_B):` (note the removal of `k`)

I just noticed that I already removed it. Is there any other function that you wanted to bring my attention into?

seanlaw: No, I think that was it. I haven't gone through `naive.py` and the unit tests yet

seanlaw:
@NimaSarajpoor It looks like everything is good. Are we ready to merge?

NimaSarajpoor:
@seanlaw 
I checked the changes again. I think it is all good.

seanlaw:

seanlaw:
@NimaSarajpoor It looks like a test failed, would you mind taking a look? 
https://github.com/TDAmeritrade/stumpy/actions/runs/4647330152/jobs/8224276965

NimaSarajpoor:
@seanlaw 
> Perhaps, replacing line 211 with np.clip would resolve it?

Yes, I think that should resolve the coverage issue. Can I push to this PR using the same branch?

seanlaw:
> Yes, I think that should resolve the coverage issue. Can/ Should I push to this PR using the same branch?

This seems unrelated to `mpdist` and this branch/PR is already closed. I'll go ahead and make the change in `main`

seanlaw:
> Come to think of it, maybe min(1.0, pearson) is better. I will handle it :)

Fixed in 575a9a49efa09dfe049a2c1fc55f35642d5d308b

PR #823, opened 2023-04-05
Fixed bug in module `snippet`

NimaSarajpoor:
Previously, we tried to address issue #677 in PR #678. However, the aforementioned PR: (1) did not correctly address the issue and (2) faced some "loss of precision" issue and we decided to handle it in a separate PR. 

This PR is to just fix the bug in `snippet`.


NimaSarajpoor:
@seanlaw 
locally, I tested `test_snippet` and all of its test functions were passing, However, it seems that we now have a very small "loss of precision" issue.

stumpy/snippets.py
```diff
@@ -110,11 +110,11 @@ def _get_all_profiles(
             S_i,
             T,
             s,
-            M_T[start : start + s],
-            Σ_T[start : start + s],
+            M_T[start : stop - s + 1],
```
seanlaw, line 113: Why did we need to change this? How does this compare to the original Matlab code?

seanlaw:
> @seanlaw locally, I tested `test_snippet` and all of its test functions were passing, However, it seems that we now have a very small "loss of precision" issue.

What is the source of the loss in precision?

stumpy/snippets.py
NimaSarajpoor: > Why did we need to change this? 

`M_T` should contain the sliding mean of `S_i` (the first argument of this function, i.e. `_mpdist_vect `). Note that `S_i` is set to `S_i = T[start:stop]`. Also, note that `M_T` is the sliding mean of subsequences (with **length `s`**) of `T` (see the line before the for-loop). 

Therefore, the sliding mean of `S_i` will be `M_T[start : stop - s + 1]`. The same is true for `Σ_T ` and `T_subseq_isconstant`. Now, you may wonder why this issue was not exposed before.  The reason is that the slice `start : start + s` covers the slice `start : stop - s + 1` when `s > m/2`. Note that I added new value for s in the test functions. The new values reflect `s < m / 2` scenario and that is why we got error in the first commit. In the second commit, there should have been no error according to my local testing.  I guess the current error in the last push is related to precision. I will investigate it.

> How does this compare to the original Matlab code?

I will check the MATLAB code.

NimaSarajpoor: > I will check the MATLAB code.

The link provided in the paper is broken, and not working.

NimaSarajpoor:
> What bug are you trying to fix? Is it still regarding precision or something else?

This PR is to only address the bug 
`Quick review of the bug`: the number of subsequences of length `s` in `T[start:stop]` is: `stop - start - s + 1`, and their corresponding values in `M_T` is `M_T[start : stop - s + 1]` (Just to reiterate: **`M_T[start : start + s]` is not correct**. I think it only works when `s > m / 2`)


> What is the source of the loss in precision?

I investigated one of the cases. The root cause of the loss of precision (in performant version ) is the following line:

https://github.com/TDAmeritrade/stumpy/blob/814977f110403bcfd1629bcf1ba4f90938394bfe/stumpy/core.py#L998

Even if we use numpy to compute `μ_Q`, `M_T`, `QT` directly, the issue still exists. To reproduce the issue:

```
# computing the distance between a subsequence with itself!

seed = 13
np.random.seed(seed)
T = np.random.uniform(-1000, 1000, [64]).astype(np.float64)

idx = 44  # start index of subsequence
m = 4  # length of window

Q = T[idx : idx + m]
QT = np.dot(Q, Q)
μ_Q = np.mean(Q)
M_T = np.mean(Q)

σ_Q = np.std(Q)
Σ_T = np.std(Q)

denom = m * σ_Q * Σ_T
ρ = (QT - m * μ_Q * M_T) / denom 

>>> ρ
0.9999999999999999
```

seanlaw:
> This PR is to only address the bug

Okay, good. I will ignore precision issues here

> I think it only works when s > m / 2

I don't remember why but there is something about this statement that feels oddly familiar. That is, somehow `s` must be greater than `m/2`. However, it's purely a gut feeling and I don't have anything to back it up.

Can you confirm that this change doesn't screw up the snippets tutorial? If not, then I can merge it

NimaSarajpoor:
@seanlaw 
Please allow me to take a look at the paper and see if I can find anything there regarding `s`.  I will run the notebook and will update you.

NimaSarajpoor:
> Can you confirm that this change doesn't screw up the snippets tutorial?

The snippet tutorial notebook was run successfully.

> I don't remember why but there is something about this statement that feels oddly familiar. That is, somehow s must be greater than m/2. However, it's purely a gut feeling and I don't have anything to back it up.

I couldn't find any relevant information in the paper. However, I noticed that the snippet tutorial notebook contains an example that reproduces a figure of paper, and `s < m/2` in that example.

seanlaw:
> The snippet tutorial notebook was run successfully.

To be more precise, do the plots look the same/similar before and after the change? If the plots in the tutorial look substantially different then there is a problem.

> I couldn't find any relevant information in the paper. However, I noticed that the snippet tutorial notebook contains an example that reproduces a figure of paper, and s < m/2 in that example.

If I recall correctly, you were the one who helped add snippets to STUMPY and so I mostly defer to you on the correctness 😄

NimaSarajpoor:
> If I recall correctly, you were the one who helped add snippets to STUMPY and so I mostly defer to you on the correctness

Yes...I think I was involved in snippet (code / tutorial). And, I think I took a look at the matlab code to set `s` in the tutorial notebook. Not sure though.



> > The snippet tutorial notebook was run successfully.
> 
> To be more precise, do the plots look the same/similar before and after the change? If the plots in the tutorial look substantially different then there is a problem.

I forgot to re-install stumpy after switching branch. I realized that the tutorial snippet is already broken in branch `main`. Since the merge of that tutorial, the API of private functions were changed. The error raised in the notebook shows that the issue is related to the bug we are discussing here. According to my investigation, at the time of the merge of the snippet tutorial, this current API 

https://github.com/TDAmeritrade/stumpy/blob/2d003ff2c2e0212eba32b22211a743306bafa992/stumpy/core.py#L1609-L1620

was different and it only had `Q, T, and m` as the arguments. So, the sliding mean and sliding std will be computed and it will be correct! Now, according to existing version, we compute these values in the snippet module and then pass them all the way to the API provided above. That is why the  tutorial notebook is now broken in branch `main`

seanlaw:
> The tutorial notebook runs successfuly in this PR

Great!

seanlaw:
> I investigated one of the cases. The root cause of the loss of precision (in performant version ) is the following line:

I should've been more clear and asked, "where is the difference in precision coming from?". Because, at the end of the day, we are going to assert that two values are the same but they are not. And so why are the two different ways of computing some value diverging?

NimaSarajpoor:
@seanlaw 

> I should've been more clear and asked, "where is the difference in precision coming from?". Because, at the end of the day, we are going to assert that two values are the same but they are not. And so why are the two different ways of computing some value diverging?

> We shouldn't deal with it here if this bug does not address anything related to precision and move it to a different issue

I opened the issue #828

NimaSarajpoor:
@seanlaw 
I am doing some further investigation. I removed `percentage==0.4` from our unit testing. I will bring it back. This is to just  help me with my investigation.

NimaSarajpoor:
I just wanted to see if the trick 3 mentioned here https://github.com/TDAmeritrade/stumpy/pull/829#issuecomment-1509571481 can fix the loss of precision in this PR or not.

seanlaw:
> I just wanted to see if the trick 3 mentioned here [#829 (comment)](https://github.com/TDAmeritrade/stumpy/pull/829#issuecomment-1509571481) can fix the loss of precision in this PR or not.

This does not bother me.

NimaSarajpoor:
@seanlaw 
Should we merge this?

stumpy/core.py
```diff
@@ -1675,6 +1683,9 @@ def _mass_distance_matrix(
                 T_subseq_isconstant,
             )
 
+            if query_idx is not None:
```
seanlaw, line 1686: I think at the top of this function, we should do:

```
if query_idx is not None:
    query_idx = int(query_idx)

for i in range(distance_matrix.shape[0]):
```

NimaSarajpoor: Will add it to the code.

seanlaw:
I will merge this once the tests pass

tests/test_snippets.py
```diff
@@ -7,8 +7,8 @@
 from stumpy.snippets import _get_all_profiles
 
 test_data = [np.random.uniform(-1000, 1000, [64]).astype(np.float64)]
-s = [6, 7, 8]
-percentage = [0.7, 0.8, 0.9]
+s = [3, 6, 7, 8]
+percentage = [0.4, 0.7, 0.8, 0.9]
```
seanlaw, line 11: Is this the version that has precision issues before?

NimaSarajpoor: So, this version (i.e. `s==3` and `percentage=0.4`) exposed the bug and we fixed in this PR. However, we also faced the loss-of-precision as well in this version. And, we fixed that in this PR as well

PR #826, opened 2023-04-07
Testing snippet

NimaSarajpoor:
This is to investigate `snippet` (tutorial notebook and the code) of the branch `main`

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -9,7 +9,9 @@ site_pkgs=$(python -c 'import site; print(site.getsitepackages()[0])')
 # Parse command line arguments
 for var in "$@"
 do
-    if [[ $var == "unit" ]]; then
+    if [[ $var == "check_notebooks" ]]; then
+        test_mode="check_notebooks"
+    elif [[ $var == "unit" ]]; then
         test_mode="unit"
     elif [[ $var == "coverage" ]]; then
         test_mode="coverage"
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -9,7 +9,9 @@ site_pkgs=$(python -c 'import site; print(site.getsitepackages()[0])')
 # Parse command line arguments
 for var in "$@"
 do
-    if [[ $var == "unit" ]]; then
+    if [[ $var == "check_notebooks" ]]; then
```
seanlaw, line 12: Instead pf "check_notebooks" I would just call it "notebooks". So, from the command line it would be `./test.sh notebooks`

seanlaw: I would add it after `coverage`

test.sh
```diff
@@ -236,6 +238,15 @@ clean_up()
     fi
 }
 
+test_notebooks()
```
seanlaw, line 241: The function name should be `convert_notebooks`

test.sh
```diff
@@ -236,6 +238,15 @@ clean_up()
...
+{
+    echo "Testing notebooks"
+    for testfile in docs/Tutorial_*.ipynb
```
seanlaw, line 244:
```
for notebook in `grep ipynb docs/tutorials.rst | sed -e 's/^[ \t]*//'`
do
    jupyter nbconvert --to notebook --execute "$notebook"
    check_errs $?
done```

File: test.sh [hunk 4]
```diff
@@ -236,6 +238,15 @@ clean_up()
...
+    do
+        jupyter nbconvert --to notebook --execute "$testfile"
+        check_errs $?
+    done
+}
 ###########
```
seanlaw, line 250: Add a newline above

test.sh
NimaSarajpoor: So we are targetting the main tutorials. Cool.

PR #829, opened 2023-04-11
Add param isconstant to snippet

NimaSarajpoor:


stumpy/snippets.py
```diff
@@ -64,6 +65,16 @@ def _get_all_profiles(
         `MPdist` value. The `percentage` and `k` parameters are ignored when
         `mpdist_custom_func` is not None.
 
+    T_subseq_isconstant : numpy.ndarray or function, default None
+        A boolean array that indicates whether a subsequence (of length `s`) in `T`
```
NimaSarajpoor, line 69: I added `(of length s) `  to the description of this parameter.

stumpy/snippets.py
```diff
@@ -83,22 +94,25 @@ def _get_all_profiles(
             f"Please try `m <= len(T) // 2`."
         )
 
+    if s is not None:
+        s = min(int(s), m)
+    else:
+        percentage = np.clip(percentage, 0.0, 1.0)
+        s = min(math.ceil(percentage * m), m)
+
```
NimaSarajpoor, line 102: I moved this if-else block as we need to have `s` before computng `T_subseq_isconstant`.

stumpy/snippets.py
```diff
@@ -83,22 +94,25 @@ def _get_all_profiles(
...
     right_pad = 0
+    T_subseq_isconstant = core.rolling_isconstant(T, s, T_subseq_isconstant)
```
NimaSarajpoor, line 104: I think we need to compute `T_subseq_isconstant` before extending `T`. This is because the user might have provided the value of `T_subseq_isconstant`. If the provided value is a boolean array, then it has shape `len(T) - s + 1`. So, I think we need to compute it here, and then extends it in the if-block below by padding it with `False`.

stumpy/snippets.py
```diff
@@ -83,22 +94,25 @@ def _get_all_profiles(
...
     D = np.empty(((n_padded // m) - 1, n_padded - m + 1), dtype=np.float64)
 
-    if s is not None:
-        s = min(int(s), m)
-    else:
-        percentage = np.clip(percentage, 0.0, 1.0)
-        s = min(math.ceil(percentage * m), m)
-
-    T_subseq_isconstant = core.rolling_isconstant(T, s)
     M_T, Σ_T = core.compute_mean_std(T, s)
```
NimaSarajpoor, line 116: Since I moved the line `T_subseq_isconstant = core.rolling_isconstant(T, s)`, I was wondering if we should move this line above as well for the sake of consistency and flow in the code.

tests/naive.py
```diff
@@ -1449,35 +1448,42 @@ def get_all_mpdist_profiles(
     s=None,
     mpdist_percentage=0.05,
     mpdist_k=None,
-    mpdist_vect_func=mpdist_vect,
+    T_subseq_isconstant=None,
```
NimaSarajpoor, line 1451: Note that I changed the signature of this naive function so that we can pass `T_subseq_isconstant` from the naive function `mpdist_snippet` to this function, and then use it (in this function) to provide values for `T_A_subseq_isconstant` and `T_B_subseq_isconstant` for `mpdist_vect`.

Also, note that I added the non-normalized version of this function to `naive.py`. (See the function `get_all_aampdist_profiles`).

NimaSarajpoor (commented):
@seanlaw 
Would you mind taking a look the few comments I have provided?

Note1:
I have had a difficult time figurig out a way to wrap  param `T_subseq_isconstant` into a partial function `mpdist_vect_func` in the NAIVE function `get_all_mpdist_profiles`. So, I created the non-normalized version of this naive function, which is similar the way we handle things in the performant version (I mean we have the private function `_get_all_mpdist_profiles` in the two modules `snippets` and `aampdist_snippets`, and they have different signatures.)

Note2: I think we may have a loss-of-precision issue here as well! I will update you.

NimaSarajpoor:
@seanlaw 
This is regarding the loss-of-precision issue, and we may continue the dicussion somewhere else. We can see the loss-of-precision issue in comparing `snippets_fractions`

https://github.com/TDAmeritrade/stumpy/actions/runs/4705862963/jobs/8346703141?pr=829

```
            ref_fractions, cmp_fractions, decimal=config.STUMPY_TEST_PRECISION
        )
E       AssertionError: 
E       Arrays are not almost equal to 5 decimals
E       
E       Mismatched elements: 2 / 3 (66.7%)
E       Max absolute difference: 0.01785714
E       Max relative difference: 0.05
E        x: array([0.375  , 0.33929, 0.28571])
E        y: array([0.35714, 0.35714, 0.28571])
```

To reproduce a similar issue:
```
    )

m = 9 
k = 3
s = 6

seed = 62 # seeds 0 to 61 are okay
np.random.seed(seed)
T = np.random.uniform(-1000, 1000, [64]).astype(np.float64)
```

And then, at some point in the algorith, we need to compute the distance matrix between `T[36:36+m]` and `T[38:38+m]` for subsequences with length `s`. The result shows that the distance between `T[38:38+s]` and itself is not zero in performant version. The same issue exists for subsequence ``T[39:39+s]`. 

I can think of three ill solutions:

(1) Add `atol` to this line:
https://github.com/TDAmeritrade/stumpy/blob/b7185099502830dba5c12648a63ea2767c618b44/stumpy/snippets.py#L285
and change it to:
`mask = snippets_profiles[i] <= total_min + atol`

(2) Add the following if-block:
```
If ρ > 1-1e-10:
    ρ = 1.0
```


(3) In API for `_mpdist_vect`, we add param `i=None` :

https://github.com/TDAmeritrade/stumpy/blob/b7185099502830dba5c12648a63ea2767c618b44/stumpy/mpdist.py#L16

If `i` is not None, this means `Q` is coming from `T`, where `Q` is actually `T[i : i + len(Q)]`. Then, we can easily set the overlapping subseqs to `0`. Note that the distance between `T[i: i + s]` and `Q[:s]` is 0. The same is true for `T[i+1: i+1+s]` and `Q[1:1+s]`. So, if we reset those values to 0, we can fix this error. This may not necessarily fix all loss-of-precision issues, but it shoulf fix most of it, at least for the snippet module. I think all of the loss-of-precision issues I noticed recenrly is coming from the aforementioned issue above, where a distance between a subesq and itself is not zero in performant version.

NimaSarajpoor:
> (3) In API for _mpdist_vect, we add param i=None :

In the last commit, I have used this trick to resolve the slight loss of precision that may appear in the distance between a subseq and itself. The solution is not really "ill" as we are not introducting any arbitrary value. (I will apply this change to the snippet bug PR and see if this resolves the issue there or not.)

If we decide to not go with this solution, I will undo the changes and git push again.

stumpy/snippets.py
seanlaw: I think it's fine either way

stumpy/snippets.py
seanlaw: I don't know :) I trust your judgement here but please double check

stumpy/snippets.py
```diff
@@ -187,6 +209,16 @@ def snippets(
         The p-norm to apply for computing the Minkowski distance. This parameter is
         ignored when `normalize == True`.
 
+    T_subseq_isconstant : numpy.ndarray or function, default None
```
seanlaw, line 212: It feels like we should be calling it `mpdist_T_subseq_isconstant`?

stumpy/snippets.py
```diff
@@ -253,6 +285,7 @@ def snippets(
         s=s,
         mpdist_percentage=mpdist_percentage,
         mpdist_k=mpdist_k,
+        T_subseq_isconstant=T_subseq_isconstant,
```
seanlaw, line 288: See comment above

stumpy/snippets.py
seanlaw: I know it's a LOOOOONG name

tests/naive.py
seanlaw: I see

seanlaw:
> > (3) In API for _mpdist_vect, we add param i=None :
> 
> In the last commit, I have used this trick to resolve the slight loss of precision that may appear in the distance between a subseq and itself. The solution is not really "ill" as we are not introducting any arbitrary value. (I will apply this change to the snippet bug PR and see if this resolves the issue there as well or not.)
> 
> If we decide to not go with this solution, I will undo the changes and git push again.

I don't mind this as it is limiting the "correction" to a known, unambiguous case where the self-match SHOULD/MUST be zero. So, there are no nasty assumptions here. We should make sure that we add the test case to `tests/test_precision.py`

stumpy/snippets.py
NimaSarajpoor: It is long but it clearly separates it from other cases where window size is `m`. Also, we are just passing it to another function and it does not appear that much in the code. So, I think it is good :) I will revise the code.

stumpy/snippets.py
NimaSarajpoor: I will check it again.

NimaSarajpoor:
> So, there are no nasty assumptions here. We should make sure that we add the test case to tests/test_precision.py and document what each test is testing for and where imprecision would/could creep in.

So, this precision issue is exposed in #823 (after fixing the snippet bug) and in this PR (after adding the new param). So, we can merge snippet bug PR (#823) first (I added the `query_idx` param there as well so to pass the unit tests). Then, we can add a new unit test for precision and open the PR (but we keep it open for now and just update it). Then, we merge this PR and resume our work regarding adding param "isconstant" to modules. 


Sounds good @seanlaw ?

stumpy/snippets.py
```diff
@@ -254,6 +285,7 @@ def snippets(
         s=s,
         mpdist_percentage=mpdist_percentage,
         mpdist_k=mpdist_k,
+        T_subseq_isconstant=mpdist_T_subseq_isconstant,
```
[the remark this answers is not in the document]
seanlaw: May I ask why? At the end of the day, the `T_subseq_isconstant` is being passed into and used in an `mpdist` function and so shouldn't we prepend it with `mpdist`? It is a bit different if `T_subseq_isconstant` was being used directly in `snippets`. Just thinking out loud here.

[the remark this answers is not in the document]
NimaSarajpoor: > At the end of the day, the `T_subseq_isconstant` is being passed into and used in an `mpdist` function

You are absolutely correct here. The only reason I kept it as `T_subseq_isconstant` is be consistent with the naming of the two variables `M_T` and `Σ_T`. So, If we decide to go with `mpdist_T_subseq_isconstant`, then I think we should rename those two variables to `mpdist_M_T` and `mpdist_Σ_T`.

Alternatively, I can do:
```
# mpdist_T_subseq_isconstant is the argument
T_subseq_isconstant = core.rolling_isconstant(T, s, mpdist_T_subseq_isconstant)
```
And now I can keep the names "M_T" and "Σ_T". 

Are you okay with the latter approach?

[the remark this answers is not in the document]
seanlaw: sounds good

seanlaw:
@NimaSarajpoor thanks for addressing this!

PR #830, opened 2023-04-12
Add test function for `_mpdist_vect`

NimaSarajpoor:
I think we should add test function (in `tests/test_mpdist.py`) to test param "isconstant" of function `stumpy/mpdist._mpdist_vect`. To do so, we need to revise the naive function `mpdist_vect` so that it can accept param "isconstant". This can help us to add the param `isconstant` in the snippet module more easily.

### DIFF
--- a/tests/naive.py
+++ b/tests/naive.py
@@ -1308,7 +1308,15 @@ def aamp_ostinato(Ts, m, p=2.0):
     return radius, Ts_idx, subseq_idx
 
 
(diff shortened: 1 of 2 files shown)


NimaSarajpoor:
@seanlaw 
After merging this, I will add param "isconstant" to snippet module.

NimaSarajpoor:
Yeah...I think it is good to be merged.

PR #831, opened 2023-04-12
add code to test the execution of tutorial notebooks

NimaSarajpoor:
This PR adds code that allows a user to test the execution of tutorial notebooks through their command-line shell.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -247,7 +247,6 @@ convert_notebooks()
     do
         jupyter nbconvert --to notebook --execute "docs/$notebook"
         check_errs $?
-        rm docs/*.nbconvert.ipynb
     done
 }
 
--- a/test.sh
+++ b/test.sh
@@ -243,9 +243,9 @@ convert_notebooks()
     echo "testing notebooks"
     for notebook in `grep ipynb docs/tutorials.rst | sed -e 's/^[ \t]*//'`
     do
-    jupyter nbconvert --to notebook --execute "docs/$notebook"
-    check_errs $?
-    rm docs/*.nbconvert.ipynb
+        jupyter nbconvert --to notebook --execute "docs/$notebook"
+        check_errs $?
+        rm docs/*.nbconvert.ipynb
     done
 }
 
--- a/test.sh
+++ b/test.sh
@@ -245,6 +245,7 @@ convert_notebooks()
     do
     jupyter nbconvert --to notebook --execute "docs/$notebook"
     check_errs $?
+    rm docs/*.nbconvert.ipynb
     done
 }
 
--- a/test.sh
+++ b/test.sh
@@ -13,6 +13,8 @@ do
         test_mode="unit"
     elif [[ $var == "coverage" ]]; then
         test_mode="coverage"
+    elif [[ $var == "notebooks" ]]; then
+        test_mode="notebooks"
     elif [[ $var == "custom" ]]; then
         test_mode="custom"
     elif [[ $var == "silent" || $var == "print" ]]; then
@@ -236,6 +238,16 @@ clean_up()
     fi
 }
 
+convert_notebooks()
+{
+    echo "testing notebooks"
+    for notebook in `grep ipynb docs/tutorials.rst | sed -e 's/^[ \t]*//'`
+    do
+    jupyter nbconvert --to notebook --execute "$notebook"
+    check_errs $?
+    done
+}
+
 ###########
 #   Main  #
 ###########
@@ -248,7 +260,10 @@ check_docstrings
 check_print
 check_naive
 
-if [[ $test_mode == "unit" ]]; then
+if [[ $test_mode == "notebooks" ]]; then
+    echo "Executing Tutorial Notebooks Only"
+    convert_notebooks
+elif [[ $test_mode == "unit" ]]; then
     echo "Executing Unit Tests Only"
     test_unit
 elif [[ $test_mode == "coverage" ]]; then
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -236,6 +238,17 @@ clean_up()
     fi
 }
 
+convert_notebooks()
+{
+    echo "testing notebooks"
+    for notebook in `grep ipynb docs/tutorials.rst | sed -e 's/^[ \t]*//'`
+    do
+    jupyter nbconvert --to notebook --execute "docs/$notebook"
+    check_errs $?
```
NimaSarajpoor, line 247:
```suggestion
```
I think the line shown above should be removed here as the function `check_errs` is for pytest.

test.sh
```diff
@@ -236,6 +238,17 @@ clean_up()
     fi
 }
 
+convert_notebooks()
+{
+    echo "testing notebooks"
+    for notebook in `grep ipynb docs/tutorials.rst | sed -e 's/^[ \t]*//'`
+    do
+    jupyter nbconvert --to notebook --execute "docs/$notebook"
```
NimaSarajpoor, line 246: I tried to chain this command with `|| echo "Error!"` so that we just print `echo` rather than being interrupred. It did not work.

NimaSarajpoor (commented):
@seanlaw 
I left a couple of comments.

test.sh
seanlaw: Actually, I would keep it and change:

https://github.com/TDAmeritrade/stumpy/blob/575a9a49efa09dfe049a2c1fc55f35642d5d308b/test.sh#L39

to

```
echo "Error: Test execution encountered exit code $1"
```

The `check_errs` function isn't `pytest` specific and simply checks whether or not the last executed command terminated "successfully". If not, the script itself should also "die" without proceeding any further

test.sh
seanlaw: That seems strange. So, you did:

```
jupyter nbconvert --to notebook --execute "docs/$notebook" || echo "Error!"
```

What happened? The echo part only prints if the `jupyter` command fails. If it succeeds, then nothing is echoed

test.sh
NimaSarajpoor: Yeah...I actually meant the message (inside that function) was about pytest. I should have spoken more clearly. I can see your point. I will revise the line.

test.sh
NimaSarajpoor: Okay... I have no idea what happend before :smile:  I probably missed the message despite me running it several times. I tried it again and I found the message printed by `echo` when the first command fails.

test.sh
```diff
@@ -236,6 +238,17 @@ clean_up()
...
 }
 
+convert_notebooks()
+{
+    echo "testing notebooks"
+    for notebook in `grep ipynb docs/tutorials.rst | sed -e 's/^[ \t]*//'`
+    do
+        jupyter nbconvert --to notebook --execute "docs/$notebook"
+        check_errs $?
+        rm docs/*.nbconvert.ipynb
```
seanlaw, line 248: 1. When the nbconvert errors out, then the `rm` command doesn't get executed
2. Since nbconvert is generating files, we should add `*.nbconvert.ipynb` to the `.gitignore` file

NimaSarajpoor: > 1. When the nbconvert errors out, then the `rm` command doesn't get executed
> 
> 2. Since nbconvert is generating files, we should add `*.nbconvert.ipynb` to the `.gitignore` file 

Very good point! I noticed the rm command not being executed when an error occurs but I forgot to think and address it! 

Since rm command is not effective when there is an error, I think we should just remote it and handle it in gitignore. Is that correct?

seanlaw: It is still good to remove the files when possible but, in this case, I would move the `rm` command into the `clean_up` function. When there is failure, one would just need to clean things up manually

Separately, regardless, we should add the file type to `.gitignore`

test.sh
```diff
@@ -236,6 +238,17 @@ clean_up()
     fi
 }
 
+convert_notebooks()
+{
+    echo "testing notebooks"
+    for notebook in `grep ipynb docs/tutorials.rst | sed -e 's/^[ \t]*//'`
+    do
+        jupyter nbconvert --to notebook --execute "docs/$notebook"
```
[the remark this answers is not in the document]
seanlaw: What's the point of setting a limit?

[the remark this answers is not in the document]
seanlaw: That flag doesn't do anything for me when I do:

```
upyter nbconvert --to notebook --ExecutePreprocessor.timeout=10 --execute "docs/Tutorial_Matrix_Profiles_For_Streaming_Data.ipynb"
```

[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 

> What's the point of setting a limit?

Just to make sure the cells have enough time to be executed. We can set to `-1` (no limit). 

> That flag doesn't do anything for me when I do

Thank you for checking it. So, it may not be effecitve (?) I am going to ignore this flag and leave it as is.

NimaSarajpoor:
@seanlaw 
I tested it again and it works! (FYI, I locally removed the streaming data tutorial notebook as its execution takes a long time.) 

I think it is good to be merged!

seanlaw:

NimaSarajpoor:
@seanlaw 
And thank you for being a wonderful teacher!

NimaSarajpoor:
@seanlaw 
It seems that doube quotation makes the wildcard become ineffective.

https://github.com/TDAmeritrade/stumpy/blob/28932abf2aa819e1b2fc53a428ebad7136c49041/test.sh#L236

Would you mind fixing it? Sorry for the incovenience.

seanlaw:
Done!

PR #838, opened 2023-04-22
Add isconstant scrump

NimaSarajpoor:
param isconstant should be added to `prescrump` and `scrump`

- [ ] prescrump
- [ ] scrump


NimaSarajpoor:
@seanlaw 
I do not know why the first two commits that are related to precision are shown above. Maybe I mistakenly made changes in this branch regarding the test of precision but forgot to do `git reset`. I took a look at the changed files and they seem fine. If you think merging this branch may cause an issue, I can delete and submit another PR.

stumpy/scrump.py
```diff
@@ -415,10 +438,6 @@ def _prescrump(
     indices : numpy.ndarray
         The subsequence indices to compute `prescrump` for
 
-    idx_ranges : numpy.ndarray
-        The (inclusive) start indices and (exclusive) stop indices referenced
-        in the `indices` array
```
NimaSarajpoor, line 420: [irrelevant to this PR]
I tried to modify the docstring.py so that it can capture a case like this, where docstring han the description of a variable that is not in the signature. we can start by narrowing down the discovered variables in a docstring to the ones that are before Return. I tried to consider return with/without s and colon in regex. but I wasn't successful since the term return might appear more than once in a docstring. For example, a sentence in docstring might contain this word.

Maybe the following solution works:
We may try to search for different versions of r"Return\n---------" and find its position in the docstring, then read the docstring upto that position, and then use python set method symmetric_difference() to capture any variable that is in the signature or docstring but not both.

stumpy/scrump.py
```diff
@@ -244,10 +270,7 @@ def _compute_PI(
             continue
 
         j = nn_i
-        # Given the squared distance, work backwards and compute QT
-        QT_j = (m - P_squared[thread_idx, i, 0] / 2.0) * (Σ_T[j] * σ_Q[i]) + (
-            m * M_T[j] * μ_Q[i]
-        )
+        QT_j = QT[j]
```
NimaSarajpoor, line 273: Note: Using `P_squared` to compute `QT_j` is not an alternative option anymore since the matrix profile value is post-processed internally depending on whether `T_x_subseq_isconstant` is True or False.

NimaSarajpoor (commented):
@seanlaw 
I think this PR is ready for the review.

stumpy/scrump.py
seanlaw: So, is this a problem? I don't understand what the solution is? `QT_j = QT[j]` seems wrong?

stumpy/scrump.py
seanlaw: I don't think I understand why you are trying to do this. Please create a new issue if this is irrelevant to this PR

stumpy/scrump.py
```diff
@@ -579,14 +635,21 @@ def prescrump(T_A, m, T_B=None, s=None, normalize=True, p=2.0, k=1):
     )
 
     if k == 1:
-        return P.flatten().astype(np.float64), I.flatten().astype(np.int64)
+        return (P.flatten().astype(np.float64), I.flatten().astype(np.int64))
```
seanlaw, line 638: Why did you add extra parantheses?

seanlaw (commented):
Everything mostly looks good (I didn't look at the new tests too closely) but I left some comments.

stumpy/scrump.py
NimaSarajpoor: `QT_j = QT[j]` is okay, and there is no issue here.

I just wanted to bring your attention to this fact that `QT` of two subesquences cannot be computed based on their z-norm euclidean distance unless both of those subsequences are treated as non-constant (i.e. their corresponding value in `T_x_subseq_isconstant` is `False`). This is a minor side effect of adding param `T_x_subseq_isconstant` since it manipulates the value of distance, and hence QT cannot be computed in a backward manner. I wanted to share it with you so that if you want to use this approach (i.e. using distance to compute QT) in the future for some module, you can be careful regarding this matter.

stumpy/scrump.py
NimaSarajpoor: My bad. I was doing some investigation. I did undo changes but forgot to remove the parantheses added automatically by `black`. I will fix it.

stumpy/scrump.py
seanlaw: But how is it possible that we can set `QT_j = QT[j]`? If `QT` itself was an empty array to begin with then setting `QT_j = QT[j]` seems wrong as `QT` itself may be wrong (i.e., empty/random numbers).

NimaSarajpoor: I think `QT` cannot be empty here as we use it to compute distance profile of a subsequence (based on which we find its nearest neighbor later). In other words, this is how I see it:

```
subsequence Q --> Compute QT array --> Compute distance profile for Q --> find nn of Q --> QT[at nn index]
```
In the last step, we simply get the dot product of Q and its nearest neighbor using the aleady-computed QT and the discovered nn index.  

---

Note that before we reach this part, we already have done two things:
(1) QT array is computed
(2) nn of Q is computed

And that is all we need to get the dot product between Q and its nearest neighbor, which is QT[nn].


Does this help? If your question was about something else, please let me know.

seanlaw: I understand. It's just that the`QT_j = (m - P_squared[thread_idx, i, 0] / 2.0) * (Σ_T[j] * σ_Q[i]) + ( m * M_T[j] * μ_Q[i])` was taken from the original Matlab code and I can't help but feel like there was a reason for this. Simply setting `QT_j = QT[j]` feels like they they wouldn't have missed that.

Can you:

1. Create `T = np.random.rand(10000)`
2. Turn off `numba` (i.e., use pure `python` mode)
3. Keep the original code but perform a check for `if QT_j != QT[j]`
4. See if we ever enter that `if` condition
5. Repeat 10,000 times or some large number
6. Repeat for AB-joins

We need to convince ourselves that we aren't reading the code wrong

NimaSarajpoor: Sure. I will update you after getting the result.

NimaSarajpoor: @seanlaw 
`QT_j != QT[j]` becomes False; however, I think that is just related to loss-of-precision. The maximum absolute difference is in the order of `1e-15 `, and the test function `npt.assert_almost_equal` is passing,

seanlaw: Can you please share the full test function that you used?

NimaSarajpoor: As a follow up to my previous email, I checked it for AB-join, the condition `QT_j == QT[j]` is, again, not satisfied but the `npt.assert_almost_equal` is passing.

---

I also generated random time series using `np.random.uniform(-1000, 1000, n)` instead of `np.random.uniform(n)`.  Again, `QT_j` and `QT[j]` are not exactly equal (the max abs diff is in the order of `1e-09`) but they are almost equal according to `npt.assert_almost_equal`.

seanlaw: Thank you. Can you double check scraamp as well to see if we need to do something similar there?

seanlaw:
@NimaSarajpoor Do you agree that this is ready to be merged now?

stumpy/scrump.py
```diff
@@ -835,7 +945,9 @@ def __init__(
                     indices,
                     s,
                     excl_zone,
-                ) = _preprocess_prescrump(T_A, m, s=s)
+                ) = _preprocess_prescrump(
+                    T_A, m, s=s, T_A_subseq_isconstant=T_A_subseq_isconstant
```
NimaSarajpoor, line 949: In class `scrump`, when `pre_scrump==True`, we compute the prescrump.  

@seanlaw 
I think there is a code smell here but I am not sure. I would like to bring your attention to the following items:
* The input `T_A` must be original and not the preprocessed version `self._T_A`; because otherwise, the output `M_T` will not be correct and it will not have `np.inf` where it should have. One idea is to add the argument `T_subseq_isfinite` to the function `_preprocess_prescrump`.
* We pass `T_A_subseq_isconstant` and `T_B_subseq_isconstant`, and then we get `Q_subseq_isconstant` and `T_subseq_isconstant` in output. However, they are basically the same.

NimaSarajpoor (commented):
I think it is ready to be merged. I just left one comment. If you think we should ignore that comment, then we can merge it as I think it is ready.

stumpy/scrump.py
seanlaw: > The input T_A must be original and not the preprocessed version self._T_A; because otherwise, the output M_T will not be correct and it will not have np.inf where it should have.

I'm not seeing the problem. Can you please elaborate? Currently, the code is correct, right?

> We pass T_A_subseq_isconstant and T_B_subseq_isconstant, and then we get Q_subseq_isconstant and T_subseq_isconstant in output. However, they are basically the same.

So, you think that `_preprocess_prescrump` should not pass these back? Or should we have changed `Q_subseq_isconstant` and `T_subseq_isconstant` to `T_A_subseq_isconstant` and `T_B_subseq_isconstant`?  What do you propose?

NimaSarajpoor: > Currently, the code is correct, right?

Yes

> So, you think that `_preprocess_prescrump` should not pass these back? Or should we have changed `Q_subseq_isconstant` and `T_subseq_isconstant` to `T_A_subseq_isconstant` and `T_B_subseq_isconstant`? What do you propose?

I feel we are re-computing variables. So, I was thinking that we may do something like this (see code below) instead of calling the function `_preprocess_prescrump` and compute ALL required variables from scratch. 

```
if pre_scrump:
     _, σ_Q = core.compute_mean_
     _, Σ_T = core.compute_mean_std(T_B, self._m)
     indices = np.random.permutation(range(0, self._l, s)).astype(np.int64)
     P, I = _prescrump(...)
    
```
We just compute three more values and that's it. However, after implementing it locally, I got the following error:
```
SystemError: CPUDispatcher(<function _prescrump at 0x7fde740065f0>) returned a result with an exception set
```
Furthermore, I now think it is not as readable as the original version. 

So, maybe we should just leave it as it.

NimaSarajpoor: > `Q_subseq_isconstant` and `T_subseq_isconstant` to `T_A_subseq_isconstant` and `T_B_subseq_isconstant`

I am going to push a commit to improve consistency in the code.

NimaSarajpoor:
@seanlaw 

stumpy/scrump.py
```diff
@@ -209,9 +236,8 @@ def _compute_PI(
     for i in indices[start:stop]:
         Q = T_A[i : i + m]
         QT[:] = core._sliding_dot_product(Q, T_B)
-        squared_distance_profile[:] = core._mass(
-            Q,
-            T_B,
+        squared_distance_profile[:] = core._calculate_squared_distance_profile(
```
seanlaw, line 239: Nice!

seanlaw:

PR #840, opened 2023-04-28
[WIP] Fixed #839

### LINKED ISSUE #839: Enhance `docstring.py`
Currently, `dosctring.py` checks whether an argument in a function (or  class) has  a description in its docsting or not.

https://github.com/TDAmeritrade/stumpy/blob/89db087914084a5e9e5915567c5f1a3bc9ec895d/docstring.py#L25-L30

I was wondering if we could add function to check the opposite, i.e. to see if there is a description for an argument that is not in the signature of a function. This is not a big deal as adding such function to `docstring.py` may not help users that much. It can help with keeping the docstrings clean.

---

Example:
The function `_prescrump` does not have the argument `idx_ranges` but we can see its description in its docstring (see below).

https://github.com/TDAmeritrade/stumpy/blob/89db087914084a5e9e5915567c5f1a3bc9ec895d/stumpy/scrump.py#L367-L420

NimaSarajpoor:


docstring.py
```diff
@@ -10,7 +10,9 @@ def get_docstring_args(fd):
     Extract docstring parameters from function definition
     """
     docstring = ast.get_docstring(fd)
-    args = re.findall(r"(\w+)\s+\:", docstring)
+    params_section = re.findall(r"(?<=Parameter(.*)(?=Return", docstring, re.DOTALL)[0]
```
seanlaw, line 13: Looks like you are missing two closing parantheses:

```
re.findall(r"(?<=Parameters)(.*)(?=Returns)"
```

Note that `(?<=Parameters)` and `(?=Returns)` are unique for regex as they are considered "look ahead"/"look behind" methods and both need their own parantheses

NimaSarajpoor: Right! I missed that. Will fix it.

NimaSarajpoor:
This is getting very interesting! The error shows:
```
Traceback (most recent call last):
  File "./docstring.py", line 74, in <module>
    check_args(docstring_args, signature_args, filepath.name, fd.name)
  File "./docstring.py", line 53, in check_args
    raise RuntimeError(msg)
RuntimeError: Found one or more unsupported arguments/parameters with docstring in 
file: aampdist.py
function/method: _aampdist_vect
parameter(s): {'i'}
```

However, `i` is not a parameter and just appear in the docstring as `T[i : i + ...]`. One idea is to just capture the ones that are preceded by a new line. But, I am not sure if that guarantees us in future. Now that I think more about it, I feel that adding such feature to `docstrings.py` (i.e. finding unsupported args but with description) may open a can of worm. 

@seanlaw 
What do you think?

seanlaw:
Actually, it looks like the `_aampdist_vect` is missing a:

```
    Returns
    --------
```

It's not clear why it's saying that `i` is a parameter because it shouldn't/wouldn't have matched at all. When I do this:

```
docstring = """
    Compute the non-normalized (i.e., without z-normalization) matrix profile distance
    measure vector between `Q` and each subsequence, `T[i : i + len(Q)]`, within `T`.

    Parameters
    ----------
    Q : numpy.ndarray
        Query array

    T : numpy.ndarray
        Time series or sequence

    m : int
        Window size
    
    percentage : float, 0.05
        The percentage of distances that will be used to report `mpdist`. The value
        is between 0.0 and 1.0. This parameter is ignored when `k` is not `None` or when
        `custom_func` is not None.

    k : int, default None
        Specify the `k`th value in the concatenated matrix profiles to return. When `k`
        is not `None`, then the `percentage` parameter is ignored. This parameter is
        ignored when `custom_func` is not None.

    custom_func : function, default None
        A custom user defined function for selecting the desired value from the
        unsorted `P_ABBA` array. This function may need to leverage `functools.partial`
        and should take `P_ABBA` as its only input parameter and return a single
        `MPdist` value. The `percentage` and `k` parameters are ignored when
        `custom_func` is not None.

    p : float, default 2.0
    
    Returns
    -------
"""

params_section = re.findall(r"(?<=Parameters)(.*)(?=Returns)", docstring, re.DOTALL)[0]
args = re.findall(r"(\w+)\s+\:", params_section)
print(args)
```

I get `['Q', 'T', 'm', 'percentage', 'k', 'custom_func', 'p']`, which is doing what I expect.

docstring.py
```diff
@@ -10,7 +10,13 @@ def get_docstring_args(fd):
     Extract docstring parameters from function definition
     """
     docstring = ast.get_docstring(fd)
-    args = re.findall(r"(\w+)\s+\:", docstring)
+    params_section = re.findall(r"(?<=Parameter)(.*)(?=Return)", docstring, re.DOTALL)
```
seanlaw, line 13: It should be `Parameters` (plural) and not `Parameter`

NimaSarajpoor:
> It's not clear why it's saying that i is a parameter because it shouldn't/wouldn't have matched at all

Because I changed the code and it now consider the full doctoring if it finds no `Return`.

docstring.py
NimaSarajpoor: I used singular for both `Parameter` and `Return` just to be on the safe side. Otherwise, if by mistake, we have Parameter instead of Parameters, we cannot find the `params_section`.

seanlaw: I see

seanlaw:
> And, when you add the "Return", it is working as expected.

Yeah, with regex, we shouldn't try to get too cute and catch everything. Instead, we should catch the simplest/obvious case. Maybe we'd want to check if `Parameter` and `Return` are present first and then, separately, process the parameters. If both are not present then we should raise an error. If `params_section` has nothing in it then we should raise an error. We should try to be specific in our assumptions of the structure and not try to force it to do too much within a single regex.

NimaSarajpoor:
> Maybe we'd want to check if `Parameter` and `Return` are present first and then, separately, process the parameters.

I totally agree with you. I think  we should check both "Parameter" and "Return" first. For instance, `cache.py::_get_cache` does not have `Parameter` section. Do you think we should add just an empty section for "Parameter". If yes, then we can go ahead and check this first. Fix them. Then, our code should work then.

NimaSarajpoor:
> If `params_section` has nothing in it (i.e., length zero) then we should raise an error. We should try to be specific in our assumptions of the structure and not try to force it to do too much within a single regex.

Correct. It is better to expose the error and catch them. This can also help us become more consistent. Otherwise, we are fixing inconsistency with just adding flexibility to the script `docstring.py`

seanlaw:
> For instance, cache.py::_get_cache does not have Parameter section.

I just pushed a commit with `Parameters` section added

seanlaw:
@NimaSarajpoor I believe this is fixed now. Would you mind reviewing my changes? If it looks fine to you then I can merge it

NimaSarajpoor:
> @NimaSarajpoor I believe this is fixed now. Would you mind reviewing my changes? If it looks fine to you then I can merge it

Thanks for your help :) I will get back to you by end of today.

stumpy/gpu_stump.py
```diff
@@ -50,8 +50,8 @@ def _compute_and_update_PI_kernel(
 
     Parameters
     ----------
-    i : int
-        Sliding window `i`
+    idx : int
+        The index for sliding window `i`
```
NimaSarajpoor, line 54: I think the description should say `idx`

```suggestion
        The index for sliding window `idx`
```

stumpy/gpu_aamp.py
```diff
@@ -47,8 +47,8 @@ def _compute_and_update_PI_kernel(
 
     Parameters
     ----------
-    i : int
-        sliding window `i`
+    idx : int
+        The index for sliding window `i`
```
NimaSarajpoor, line 51: I think the description should say `idx`

```suggestion
        The index for sliding window `idx`
```

stumpy/aamp_stimp.py
```diff
@@ -320,13 +324,21 @@ def PAN_(self):
...
         return self.pan().astype(np.float64)
 
     @property
     def M_(self):
         """
         Get all of the (breadth first searched (level) ordered) subsequence window sizes
+
+        Parameters
+        ----------
+        None
```
NimaSarajpoor, line 341: The section for `Returns`  is missing.

stumpy/aamp_stimp.py
```diff
@@ -320,13 +324,21 @@ def PAN_(self):
         """
         Get the transformed (i.e., normalized, contrasted, binarized, and repeated) pan
         matrix profile
+
+        Parameters
+        ----------
+        None
```
NimaSarajpoor, line 330: The section for `Returns` is missing

stumpy/aamp_stimp.py
```diff
@@ -218,6 +218,10 @@ def update(self):
         Update the pan matrix profile by computing a single matrix profile using the
         next available subsequence window size
 
+        Parameters
+        ----------
+        None
+
```
NimaSarajpoor, line 224: The section for `Returns` is missing.

NimaSarajpoor (commented):
@seanlaw 
I can see that many docstrings have been fixed. Cool :) You made a great effort in enhancing them.

I left a few comments.

NimaSarajpoor:
Also, it seems that some of the functions have the purpose of raising error, and they have section called `Raises` in their docstrings. For instance, see `core.check_nan`. 

So, if you think it matters, you may want to add this section to other functions that have the same purpose, i.e. raising error. Some of the changed docstrings have only `Parameters` and `Returns`, but not the section `Raises`

stumpy/gpu_stump.py
seanlaw: I am leaving it as `i` on purpose

seanlaw:
> Also, it seems that some of the functions have the purpose of raising error, and they have section called Raises in their docstrings in addtion to Parameters and Returns. For instance, see core.check_nan.

The `Raises` section is not required. It is only there when the only real result is to raise an error

stumpy/aamp_stimp.py
seanlaw: Since it is a class method, we do not require a return

docstring.py
```diff
@@ -5,13 +5,36 @@
...
+        msg = "Missing required 'Parameters' section in docstring in \n"
+        msg += f"file: {file_name}\n"
+        if class_name is not None:
+            msg += f"class: {class_name}\n"
+        msg += f"function/method: {func_name}\n"
+        raise RuntimeError(msg)
+    if class_name is None and len(re.findall(r"Returns", docstring)) != 1:
+        msg = "Missing required 'Returns' section in docstring in \n"
+        msg += f"file: {file_name}\n"
+        if class_name is not None:
```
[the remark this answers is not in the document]
seanlaw: Good point. I copy/pasted from other sections but you are right!

seanlaw:
@NimaSarajpoor I think only one thing was needed. The rest should be fine

stumpy/aamp_stimp.py
NimaSarajpoor: I see. Good to know :) 
Btw, its sibling, `stimp`, has the section `Returns`. e.g.: function `M_`.

However, being inconsistent may not be a big deal here. I mean we can try to be perfect but the docstring is good and clear.

NimaSarajpoor:
> @NimaSarajpoor I think only one thing was needed. The rest should be fine

Thank you for taking care of this PR! I will stat adding `isconstant` param (to remaining modules) after you merge this PR.

PR #843, opened 2023-05-01
Add isconstant motifs

NimaSarajpoor:
This PR is to add isconstant param to module `motifs`

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -2403,7 +2403,7 @@ def rolling_isconstant(a, w, a_subseq_isconstant=None):
 
     elif isinstance(a_subseq_isconstant, np.ndarray):
         isconstant_func = None
-        if a.ndim != a_subseq_isconstant.ndim:
+        if a.ndim != a_subseq_isconstant.ndim:  # pragma: no cover
             msg = (
                 "The arrays `a` and `a_subseq_isconstant` must have same "
                 + "number of dimensions, {a.ndim} != {a_subseq_isconstant.ndim}"
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -2400,16 +2400,16 @@ def rolling_isconstant(a, w, a_subseq_isconstant=None):
             raise ValueError(msg)
 
         isconstant_func = a_subseq_isconstant
+
+    elif isinstance(a_subseq_isconstant, np.ndarray):
+        isconstant_func = None
         if a.ndim != a_subseq_isconstant.ndim:
             msg = (
                 "The arrays `a` and `a_subseq_isconstant` must have same "
-                +"number of dimensions, {a.ndim} != {a_subseq_isconstant.ndim}"
+                + "number of dimensions, {a.ndim} != {a_subseq_isconstant.ndim}"
             )
             raise ValueError(msg)
-    
-    elif isinstance(a_subseq_isconstant, np.ndarray):
-        isconstant_func = None
-        
+
     else:  # pragma: no cover
         msg = (
             "`T_subseq_isconstant` must be of type `np.ndarray` or a callable "
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -2400,9 +2400,16 @@ def rolling_isconstant(a, w, a_subseq_isconstant=None):
             raise ValueError(msg)
 
         isconstant_func = a_subseq_isconstant
-
+        if a.ndim != a_subseq_isconstant.ndim:
+            msg = (
+                "The arrays `a` and `a_subseq_isconstant` must have same "
+                +"number of dimensions, {a.ndim} != {a_subseq_isconstant.ndim}"
+            )
+            raise ValueError(msg)
+    
     elif isinstance(a_subseq_isconstant, np.ndarray):
         isconstant_func = None
+        
     else:  # pragma: no cover
         msg = (
             "`T_subseq_isconstant` must be of type `np.ndarray` or a callable "
--- a/stumpy/motifs.py
+++ b/stumpy/motifs.py
@@ -151,7 +151,13 @@ def _motifs(
     return motif_distances, motif_indices
 
 
-@core.non_normalized(aamp_motifs)
+@core.non_normalized(
+    aamp_motifs,
+    exclude=[
+        "normalize",
+        "T_subseq_isconstant",
+    ],
+)
 def motifs(
     T,
     P,
@@ -366,6 +372,7 @@ def motifs(
         "Σ_T",
         "T_subseq_isfinite",
         "T_subseq_isconstant",
+        "Q_subseq_isconstant",
         "p",
     ],
     replace={"M_T": "T_subseq_isfinite", "Σ_T": None},
--- a/stumpy/motifs.py
+++ b/stumpy/motifs.py
@@ -521,7 +521,6 @@ def match(
             T_subseq_isconstant=T_subseq_isconstant[i],
             Q_subseq_isconstant=Q_subseq_isconstant[i],
             query_idx=query_idx,
-
         )
     D = np.mean(D, axis=0)
 
--- a/stumpy/motifs.py
+++ b/stumpy/motifs.py
@@ -484,17 +484,23 @@ def match(
     if np.any(np.isnan(Q)) or np.any(np.isinf(Q)):  # pragma: no cover
         raise ValueError("Q contains illegal values (NaN or inf)")
 
-    if len(Q.shape) == 1:
-        Q = Q[np.newaxis, :]
-    if len(T.shape) == 1:
-        T = T[np.newaxis, :]
-
-    d, n = T.shape
-    m = Q.shape[1]
+    m = Q.shape[-1]
     excl_zone = int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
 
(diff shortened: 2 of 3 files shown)


NimaSarajpoor (commented):
@seanlaw 
I left a few notes/comments.

stumpy/motifs.py
```diff
@@ -466,18 +484,23 @@ def match(
...
+    excl_zone = int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
+
+    Q_subseq_isconstant = core.rolling_isconstant(Q, m, Q_subseq_isconstant)
+    T_subseq_isconstant = core.rolling_isconstant(T, m, T_subseq_isconstant)
+
+    if Q.ndim == 1:
         Q = Q[np.newaxis, :]
-    if len(T.shape) == 1:
-        T = T[np.newaxis, :]
+    if Q_subseq_isconstant.ndim == 1:
```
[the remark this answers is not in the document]
seanlaw: This could work?

```
if Q.ndim == 1 and Q.ndim == Q_subseq_isconstant.ndim:
    Q = Q[np.newais, :] Q_subseq_isconstant = 
    Q_subseq_isconstant[np.newaxis, :]
```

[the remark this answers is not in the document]
seanlaw: But the multiple if statements isn't horrible

tests/test_motifs.py
```diff
@@ -6,9 +8,64 @@
 from stumpy import core, match, motifs
 
 
-def naive_match(Q, T, excl_zone, max_distance, max_matches=None):
+def naive_match_multi(
```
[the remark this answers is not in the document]
seanlaw: maybe? this function looks reasonable but should be called `naive_multi_match`

stumpy/motifs.py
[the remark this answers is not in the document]
NimaSarajpoor: I think the check `Q.ndim == Q_subseq_isconstant.ndim` may not be effecitve here. Note that we have
```
Q_subseq_isconstant = core.rolling_isconstant(Q, m, Q_subseq_isconstant)
```
before this if-block. Hence, I assume Q_subseq_isconstant and Q must be of same ndim, and if they are not, an error (probably?) will be raised in `core.rolling_isconstant`.

tests/test_motifs.py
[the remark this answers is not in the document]
NimaSarajpoor: I will revise the name of function.

stumpy/motifs.py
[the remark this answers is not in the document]
seanlaw: Sounds good.

stumpy/core.py
```diff
@@ -2403,6 +2403,13 @@ def rolling_isconstant(a, w, a_subseq_isconstant=None):
 
     elif isinstance(a_subseq_isconstant, np.ndarray):
         isconstant_func = None
+        if a.ndim != a_subseq_isconstant.ndim:  # pragma: no cover
+            msg = (
+                "The arrays `a` and `a_subseq_isconstant` must have same "
```
seanlaw, line 2408: must have THE same

seanlaw (changes requested):

seanlaw:

stumpy/core.py
NimaSarajpoor: I really need to start paying more attention to the article THE. Thanks a lot for pointing it out. I will fix it.

NimaSarajpoor:
@seanlaw 
> Is this ready to be merged?

Not yet. I think I need to add param `T_x_subseq_isconstant` to function `motifs`. I am trying to fix the bug in `motifs` first before moving forward.

NimaSarajpoor:
@seanlaw 
If you are happy with the latest changes, then I think it is ready to be merged.

stumpy/motifs.py
```diff
@@ -126,6 +126,7 @@ def _motifs(
             atol=atol,
             query_idx=candidate_idx,
             T_subseq_isconstant=T_subseq_isconstant,
+            Q_subseq_isconstant=T_subseq_isconstant[:, [candidate_idx]],
```
seanlaw, line 129: Why not just `T_subseq_isconstant[:, candidate_idx],`?

Is the extra square brackets necessary?

NimaSarajpoor: I think it is necessary as other wise we are passing 1D boolean array that corresponds to a multi-dim time series, and hence each 1D Q (of multi-dim Q) corresponds to a single boolean value. However, this does not work as we need to provide boolean array (of shape `(1,)`) and not just boolean value. 

It might be a good idea to add support for boolean value in function `core.rolling_isconstant`. So, when we pass a boolean value to this function, it will be converted to 1d array (of length 1).

seanlaw: From what I can tell, this:

```
T_subseq_isconstant[:, candidate_idx]
```

returns a 1D array with a single element. When this gets passed into `match`, the `match` function will check the dimensions of `Q_subseq_isconstant` and makes sure it is appropriately 2-dimensional. Maybe I'm missing something?

NimaSarajpoor: I got confused by the dimension and I explained it in the wrong way. 

---

In `motifs.py`, we add a new axis to `T` and pass it to `_motifs`. (So, the input `T` in function `_motifs` is 2D. We may add it to the description of `T` in the docstring of `_motifs` that says it has the shape of `(1, len(T))`.

In function `_motifs`, we have:  
```
Q = T[:, candidate_idx : candidate_idx + m] 
```
which means `Q` is 2D. 

Note that `Q` and `Q_subseq_isconstant` that are being pass to function `match` must have the same dimension. Otherwise, we would  get an error when we do::

```
Q_subseq_isconstant = core.rolling_isconstant(Q, m, Q_subseq_isconstant)
```

Please note that the following if-block appears later in the code.
```
if Q_subseq_isconstant.ndim == 1:
    # add new dim
```

I hope that I did not make any mistage this time :smile:

stumpy/motifs.py
```diff
@@ -240,6 +248,18 @@ def motifs(
...
+        A boolean array that indicates whether a subsequence in `T` is constant
+        (True). Alternatively, a custom, user-defined function that returns a
+        boolean array that indicates whether a subsequence in `T` is constant
+        (True). The function must only take two arguments, `a`, a 1-D array,
+        and `w`, the window size, while additional arguments may be specified
+        by currying the user-defined function using `functools.partial`. Any
+        subsequence with at least one np.nan/np.inf will automatically have its
+        corresponding value set to False in this boolean array. Note that it is
+        the responsibility of user to ensure that the provided matrix profile `P`
+        is computed while considering `T_subseq_isconstant`.
```
NimaSarajpoor, line 261: Not sure if the last sentence above is a good addition to the description of param `T_subseq_isconstant`

stumpy/motifs.py
seanlaw: Okay. I think my biggest issue is that it's not obvious to me what the output dimensions would be with that syntax. What about:

```
Q_subseq_isconstant=np.atleast_2d(T_subseq_isconstant[:, candidate_idx])
```

stumpy/motifs.py
seanlaw: I don't understand what that sentence is trying to convey :(

stumpy/motifs.py
NimaSarajpoor: I trust your judgement as you have better insight. I think `np.atleast_2d` should be clear enough,

stumpy/motifs.py
NimaSarajpoor: `mateix profile (P)` is one of the parameters of function `motifs`. So, the user needs to compute it and pass  it to `motifs`. However, the user needs to be careful and conside the `T_subseq_isconstant` in the computation of matrix profile if they would like to pass  `T_subseq_isconstant` to `motifs` . 

In other words, if `T_subseq_isconstant` is not None, then the following code is not correct::
```
# T_subseq_isconstant is not None

mp = stumpy.stump(T, m)
P = mp[:, 0].astype(np.float64)
out = motifs(T, P, ..., T_subseq_isconstant=T_subseq_isconstant)
```

The correct way is:
```
# T_subseq_isconstant is not None

mp = stumpy.stump(T, m, T_A_subseq_isconstant=T_subseq_isconstants)
P = mp[:, 0].astype(np.float64)
out = motifs(T, P, ..., T_subseq_isconstant=T_subseq_isconstant)
```

Note that I am passing `T_subseq_isconstants` to function `stump` as well.

seanlaw: > In other words, if T_subseq_isconstant is not None, then the following code is not correct:

Is it incorrect only if `T_subseq_isconstant` is some custom function or custom array (that is different from what is produced by `core.rolling_isconstant`)? If so, I don't feel like the last sentence is necessary. To some extent, it is similar to when an exclusion zone is specified and I think that's an advance feature.

NimaSarajpoor: > To some extent, it is similar to when an exclusion zone is specified and I think that's an advance feature.

I see your point! I will remove the last sentence.

seanlaw:
@NimaSarajpoor Thank you for this contribution

PR #846, opened 2023-05-02
Fixed #844

### LINKED ISSUE #844: bug in motifs.py
In function `_motifs`  (in module `motifs.py`), we see the following lines of code:

https://github.com/TDAmeritrade/stumpy/blob/43a1cea710c004ed08fb1de0e712e3335917f61d/stumpy/motifs.py#L119-L125

which is not correct as we should have passed `max_matches` to this function, but we are passing `None`. I think that is the main issue behind https://github.com/TDAmeritrade/stumpy/discussions/794. Note that if we set `max_motifs=np.inf` and `cutoff=np.inf`, the existing code results in ONE motif only, which is what mentioned in the discussion #794:

> Specifically, the indices and distances returned by the call are of size (1, x) in Stumpy 1.11 and size (3, x) in Stumpy 1.9.1 (x depends on max_matches).

NimaSarajpoor:
Fixed #844


NimaSarajpoor:
@seanlaw 
Have you seen this?
https://github.com/TDAmeritrade/stumpy/actions/runs/4887949074/jobs/8725124869

NimaSarajpoor:
Messed up something! Sorry! Going to close this one as well!

Issue #708, opened 2022-11-07
Check fastmath=True


seanlaw:
In many `@njit` cases, we use `fastmath=True`. However, this might be bad when the input arguments and/or output results contain `np.nan` or `np.inf` (see [Numba fastmath](https://numba.pydata.org/numba-doc/latest/user/performance-tips.html#fastmath)). In those cases, we should use more specific but slower `fastmath` [flags](https://llvm.org/docs/LangRef.html#fast-math-flags) like `@njit(fastmath={"nsz", "arcp", "contract", "afn", "reassoc"})`

We should spend some time to reassessing whether all of the `fastmath` specifications are accurate or if they need to be updated

seanlaw:
There is a nice article [here](https://simonbyrne.github.io/notes/fastmath/) on some of the differences in the fastmath flags. I wonder if it is possible/might be worth to go down a path where we assume that all `njit` functions must contain finite inputs/outputs. This will require some further preprocessing of the inputs AND post-processing of the outputs but it might be well worth the effort.

seanlaw:
@NimaSarajpoor Did you actually spend time checking this in PR #1048 or does this work still need to happen?

NimaSarajpoor:
@seanlaw 
I did not. So,  I think we should keep this issue open. 

IIRC, in #1025, we had some discussion around a code that finds caller(s) of njit function. We can add a function in `./fastmath.py` that raises an error if it finds a njit function whose fastmath flag is not a subset of the fastmath flag of its caller njit function(s). In addition, any njit function with `fastmath=True` needs to be checked [manually] to make sure it does not take an input that contains np.inf / np.nan.


What are your thought?

seanlaw:
@NimaSarajpoor That sounds good.

> In addition, any njit function with fastmath=True needs to be checked [manually] to make sure it does not take an input that contains np.inf / np.nan.

Yes, this is probably the first step

> We can add a function in ./fastmath.py that raises an error if it finds a njit function whose fastmath flag is not a subset of the fastmath flag of its caller njit function(s).

Yes, adding a function to `fastmath.py` is fine. Is there anything I can assist/help you with? I think the current PR associated with this issue is too complicated and we should try to keep it simple/flat.

NimaSarajpoor:
> I think the current PR associated with this issue is too complicated and we should try to keep it simple/flat.

I do agree. 

>  Is there anything I can assist/help you with?

[...]

PR #847, opened 2023-05-02
fix dependencies

NimaSarajpoor:


seanlaw:
I'm wondering if we should/could write some code to detect this during testing?

NimaSarajpoor:
> I'm wondering if we should/could write some code to detect this during testing?

To detect if an error is raised bacause of a feature being depreciated / discontinued? That sounds like a cool idea.

seanlaw:
Not quite. I meant to simply detect if somebody is using `np.int` instead of `np.int64`

NimaSarajpoor:
Related: #708

NimaSarajpoor:
> I meant to simply detect if somebody is using `np.int` instead of `np.int64`

Maybe we can add such check to the following, existing function?
https://github.com/TDAmeritrade/stumpy/blob/43a1cea710c004ed08fb1de0e712e3335917f61d/stumpy/core.py#L311

------

**To further improve the code, we may take care of warnings:**


```
tests/test_core.py ..................................................... [ 22%]
........................................................................ [ 52%]
........................................................................ [ 82%]
............................s..............                              [100%]

=============================== warnings summary ===============================
tests/test_core.py::test_calculate_squared_distance_profile[Q0-T0]
  /home/runner/work/stumpy/stumpy/stumpy/core.py:2355: NumbaExperimentalFeatureWarning: Use of isinstance() detected. This is an experimental feature.
    return np.where(out == 0.0, True, False)
```

The function [numpy.where](https://github.com/numpy/numpy/blob/main/numpy/array_api/_searching_functions.py) calls `_normalize_two_args`, which calls `_new`, in which `isinstance` is used. To resolve this issue, we have two solutions: 
(1) we can just return `out == 0` , 
or 
(2) we can exclude the return statement from the numba-jit function, and instead return the array `out`. Then, we can do `np.where(out==0, True, False)` outside of the numba-jit function

---

**I also found another warning:**

tests/test_core.py::test_apply_exclusion_zone_bool
  /home/runner/work/stumpy/stumpy/stumpy/core.py:332: FutureWarning: Conversion of the second argument of issubdtype from `bool` to `np.generic` is deprecated. In future, it will be treated as `np.bool_ == np.dtype(bool).type`.
    if not np.issubdtype(a.dtype, dtype):

seanlaw:
I'm a little bit confused. Can you link to where you are seeing these warnings? Should the title be "fix deprecations" rather than "dependencies"?

NimaSarajpoor:
@seanlaw 
>  Can you link to where you are seeing these warnings?

See the `Run Unit Tests` step in the following testings:

https://github.com/TDAmeritrade/stumpy/actions/runs/4868285457/jobs/8681637411

https://github.com/TDAmeritrade/stumpy/actions/runs/4868285457/jobs/8681637702

---

Should the title be "fix deprecations" rather than "dependencies"?

I've been meaning to ask you to provide me with a bette name for the title! Thank you for the input. I think yours is bette. I will edit it.

---


> Can you provide the code that are triggering the need for these changes?

The code that triggers the need for the applied changes are some of the `test_mstump` that replace some finite values with nan/inf in input array `T`. I checked out the issue #850 and it seems this has been discovered there as well.  I tracked it and noticed the `np.inf` value does not affect result as expected when being passed to the njit function.

What version of Python and numba are you using?
Python 3.9, numpy 1.24, and numba 0.57. I had to upgrade numpy and numba to get the same error locally.

seanlaw:
> tests/test_core.py::test_apply_exclusion_zone_bool
/home/runner/work/stumpy/stumpy/stumpy/core.py:332: FutureWarning: Conversion of the second argument of issubdtype from bool to np.generic is deprecated. In future, it will be treated as np.bool_ == np.dtype(bool).type.
if not np.issubdtype(a.dtype, dtype):

I think we should update `apply_exclusion_zone` with:

```
def apply_exclusion_zone(a, idx, excl_zone, val):
    """
    """
    check_dtype(a, dtype=np.dtype(type(val)).type)
    _apply_exclusion_zone(a, idx, excl_zone, val)
```

NimaSarajpoor:
> @seanlaw

> Python 3.9, numpy 1.24, and numba 0.57. I had to upgrade numpy and numba to get the same error locally.

I checked again. My python is 3.10.

tests/test_core.py
```diff
@@ -792,7 +792,7 @@ def test_apply_exclusion_zone_int():
 
 
 def test_apply_exclusion_zone_bool():
-    T = np.ones(10, dtype=bool)
+    T = np.ones(10, dtype=np.dtype(bool).type)
```
seanlaw, line 795: This doesn't make sense to change

NimaSarajpoor: Correct! My bad!

NimaSarajpoor:
@seanlaw 
I need your help here :) I am trying to understand the error raised here: https://github.com/TDAmeritrade/stumpy/actions/runs/4878569344/jobs/8704334878?pr=847

I do not know where `np.int32` comes from. 

Would you mind take a look?

seanlaw:
> I do not know where np.int32 comes from.

I wonder if this is a Windows problem as all of the tests seem to be failing on Windows and I cannot reproduce the error on macos. However, I'm guessing that in `core.apply_exclusion_zone(comp, i, exclusion_zone, -1)` the last argument is `-1`, which is somehow being evaluated as an `np.int32` by `np.dtype(type(val)).type` on Windows. 

I think `np.dtype(type(val)).type` may or may not be the best idea. Let's revert this and deal with it in a separate PR

NimaSarajpoor:
> I wonder if this is a Windows problem as all of the tests seem to be failing on Windows and I cannot reproduce the error on macos

Thank you for taking a look at this! Got it now! I changed it back to the original version.

seanlaw:
> The code that triggers the need for the applied changes are some of the test_mstump that replace some finite values with nan/inf in input array T. I checked out the issue https://github.com/TDAmeritrade/stumpy/issues/850 and it seems this has been discovered there as well. I tracked it and noticed the np.inf value does not affect result as expected when being passed to the njit function.

Do you think your new `fastmath` change actually fixes the issues in #850? I guess we'll find out after we merge this PR.

Assuming that all tests pass, is this ready to be merged @NimaSarajpoor?

NimaSarajpoor:
@seanlaw 
You may want to change the title of PR again to something like "Fix deprecations and fastmath".  If it is not necessary, please ignore this comment.

seanlaw:
> tests/test_core.py::test_apply_exclusion_zone_bool
/home/runner/work/stumpy/stumpy/stumpy/core.py:332: FutureWarning: Conversion of the second argument of issubdtype from bool to np.generic is deprecated. In future, it will be treated as np.bool_ == np.dtype(bool).type.
if not np.issubdtype(a.dtype, dtype):

Btw, we shouldn't worry about this as this warning is only present in a very old version of `numpy` (only shows up in minimum version tests). I'm seeing if I can fix it but, if not, it's really a non-issue and can be completely ignored.

PR #856, opened 2023-05-04
Fix #844 Fix bug in module `motifs`

### LINKED ISSUE #844: bug in motifs.py

NimaSarajpoor:
@seanlaw 
Sorry for the noise! I am not getting revenge btw :laughing: I am just newbie. So, the nature of our noises are different :)

### DIFF
--- a/tests/test_motifs.py
+++ b/tests/test_motifs.py
@@ -6,6 +6,68 @@ import pytest
 from stumpy import core, match, motifs
 
 
+def naive_motifs(T, m, max_motifs, max_matches):
+    # To avoid complexity, this naive function is written
+    # such that each array in the ouput has shape
+    # (max_motif, max_matches).
+
+    # To this end, the following items are considered:
+    # 1. `max_distance` and `cutoff` are both hardcoded and
+    # set to np.inf
+    # 2. If the number of subsequence, i.e. `len(T)-m+1`, is
+    # not less than `m * max_motifs * max_matches`, then the
+    # output definitely has the shape (max_motif, max_matches).
+
+    l = len(T) - m + 1
+    excl_zone = int(np.ceil(m / 4))
+
+    output_shape = (max_motifs, max_matches)
+    motif_distances = np.full(output_shape, np.NINF, dtype=np.float64)
+    motif_indices = np.full(output_shape, -1, dtype=np.int64)
+
+    D = naive.distance_matrix(T, T, m)
+    for i in range(D.shape[0]):
+        naive.apply_exclusion_zone(D[i], i, excl_zone, np.inf)
+
+    P = np.min(D, axis=1)
+    for i in range(max_motifs):
+        distances = []
+        indices = []
+
+        idx = np.argmin(P)
+
+        # self match
+        distances.append(0)
+        indices.append(idx)
+        naive.apply_exclusion_zone(P, idx, excl_zone, np.inf)
+
+        # Explore distance profile D[idx] till `max_matches` are found.
+        naive.apply_exclusion_zone(D[idx], idx, excl_zone, np.inf)
+        for _ in range(l):
+            if len(distances) >= max_matches:
+                break
+
+            nn = np.argmin(D[idx])
+            distances.append(D[idx, nn])
+            indices.append(nn)
+
+            # Update D[idx] to avoid finding matches that are trivial to
+            # each other.
+            naive.apply_exclusion_zone(D[idx], nn, excl_zone, np.inf)
+
+            # Update P after the discovery of each match so that the
+            # match cannot be selected as the motif next time.
+            naive.apply_exclusion_zone(P, nn, excl_zone, np.inf)
+
+            # Note that a discovered match cannot be selected as motif but
+            # it can still be selected again as a match for another motif.
+
+        motif_distances[i] = distances
+        motif_indices[i] = indices
+
+    return motif_distances, motif_indices
+
+
 def naive_match(Q, T, excl_zone, max_distance, max_matches=None):
     m = Q.shape[0]
     D = naive.distance_profile(Q, T, m)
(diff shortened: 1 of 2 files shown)


NimaSarajpoor:
**For future reference:**
Fixing the bug resolves the exposed error and exposes an issue in one of the existing test functions. We are going to fix it in the next commit.

tests/test_motifs.py
```diff
@@ -112,8 +112,8 @@ def test_motifs_max_matches():
     # This test covers the following:
 
     # A time series contains motif A at four locations and motif B at two.
-    # If `max_motifs=2` the result should contain only the top two matches of motif A
-    # and the top two matches of motif B as two separate motifs.
+    # If `max_moitf=2` and `max_matches=2`, the result should contain two
```
seanlaw, line 115: Fixing a few spelling mistakes and changing the numbers, I think we should change the language to say:

"If `max_motifs=2` and `max_matches=3`, the result should contain (at most) two sets of motifs and each motif set should contain (at most) the top three matches"

I like to make the numbers different so as not to imply that there is any significance in the numbers both being the same (i.e., 2)

seanlaw: Of course, you'll need to change `max_matches` to `3` in the actual code

tests/test_motifs.py
```diff
@@ -170,6 +170,72 @@ def test_motifs_max_matches():
     npt.assert_almost_equal(left_profile_values, right_distance_values, decimal=4)
 
 
+def test_motifs_max_matches_max_distances_inf():
+    # This test covers the following:
+
+    # A time series contains motif A at two locations and motif B at two.
+    # If `max_moitf=2` and `max_matches=2`, and `max_distance=inf`, the
```
seanlaw, line 177: See above

tests/test_motifs.py
NimaSarajpoor: @seanlaw 
I see your point. I will revise it. 

Just as a heads up, it is possible that we get three matches instead of two as the number of returned matches depends on the `max_distance` as well. If I notice such thing, I will motidy the input time series `T`.

tests/test_motifs.py
```diff
@@ -301,3 +368,67 @@ def test_match_mean_stddev_isconstant(Q, T):
     )
 
     npt.assert_almost_equal(left, right)
+
+
+def test_motif_random_input():
```
NimaSarajpoor, line 373: I think the name of this func should be changed. However, I do not know  have any suggestion.

tests/test_motifs.py
```diff
@@ -301,3 +368,67 @@ def test_match_mean_stddev_isconstant(Q, T):
...
+    # since len(T) is greater than m * max_motifs * max_matches,
+    # the shape of output is defiitely `(max_motifs, max_matches)`
+    output_shape = (max_motifs, max_matches)
+
+    # naive approach
+    ref_distances = np.full(output_shape, np.NINF, dtype=np.float64)
+    ref_indices = np.full(output_shape, -1, dtype=np.int64)
+
+    k = 10  # set k that is the greater than m * (max_matches - 1)
+    mp = naive.stump(T, m, row_wise=True, k=k)
```
NimaSarajpoor, line 392: Instead of using top-k matrix profile in naive approach, we may just compute the full distance matrix and find the motifs (and their matches) by using the distance matrix. I think it is better to be independent of top-k feature

NimaSarajpoor (commented):
@seanlaw 
I just left two comments. Please add your comments as well if you have any.

tests/test_motifs.py
seanlaw: I agree

tests/test_motifs.py
seanlaw: What is the point of testing random input? If there is none, then why not simpy `test_motifs` (unless it's already taken)?

NimaSarajpoor: I wanted to reflect the differences between test functions in their names. You are right..."why not just `test_motids`?" :) 

I am going to use `test_motifs`

seanlaw:
I never noticed the `decimal=4` in some of the `npt.assert_almost_equal` and I wonder if we even need to specify it? There are multiple instances throughout this file

NimaSarajpoor:
> I never noticed the `decimal=4` in some of the `npt.assert_almost_equal` and I wonder if we even need to specify it? There are multiple instances throughout this file.

I will remove it. If I notice an error raised by a specific test(s) as a result of such change, I will create an issue so that we can handle that specifc test(s) in another PR.


> Also, maybe we should move the naive portion out into its own def naive_motifs function above def naive_match function

Correct. Later, we can re-use it for testing `T_subseq_isconstant`. 

Note that the `max_distance` and `cutoff` parameters are hard-coded and they are both set to `np.inf`. We can add them to naive function as arguments (so that they can match the signature of performant func more closely) at the cost of making the naive func more complicated.

seanlaw:

NimaSarajpoor:
@seanlaw 

---

FYI: In the function `naive_motifs`, we can add the following if-check to make sure that `T` is  long enough for our test.

```
k = len(T) - m + 1
if k < m * max_matches * max_motifs:
      raise ValueError(...)
```

I think `m * max_matches * max_motifs` is a good, weak lower bound. However, since it is a naive function, I decided to avoid raising error, and instead, I just added a comment regarding this matter.

seanlaw:
Thanks @NimaSarajpoor

PR #860, opened 2023-05-06
Fix loss of precision for distance between a subseq and itself

NimaSarajpoor:
According to [this error](https://github.com/TDAmeritrade/stumpy/actions/runs/4902125133/jobs/8753783558?pr=856), we can realize that there is a loss of precision in self match in core.mass. This can result in loss-of-precision in the output of `motifs`. 

We first add a test function to expose this error.

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -1245,6 +1245,9 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0, query_idx=None):
         raise ValueError(f"`Q` is {Q.ndim}-dimensional and must be 1-dimensional. ")
 
     check_window_size(m, max_size=Q.shape[-1])
+    if query_idx is not None:
+        query_idx = int(query_idx)
+        QT_query_idx_allclose = np.allclose(Q, T[query_idx : query_idx + m])
 
     T = _preprocess(T)
     n = T.shape[0]
@@ -1265,20 +1268,17 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0, query_idx=None):
     if np.any(~np.isfinite(Q)):
         distance_profile[:] = np.inf
     else:
+        if T_subseq_isfinite is None:
+            T, T_subseq_isfinite = preprocess_non_normalized(T, m)
+        distance_profile[:] = _mass_absolute(Q, T, p)
         if query_idx is not None:  # pragma: no cover
-            query_idx = int(query_idx)
-            if not np.allclose(Q, T[query_idx : query_idx + m]):
+            if not QT_query_idx_allclose:
                 msg = (
                     "Subsequences `Q` and `T[query_idx:query_idx+m]` are "
                     + "different but were expected to be identical. Please "
                     + "verify that `query_idx` is correct."
                 )
                 warnings.warn(msg)
-
-        if T_subseq_isfinite is None:
-            T, T_subseq_isfinite = preprocess_non_normalized(T, m)
-        distance_profile[:] = _mass_absolute(Q, T, p)
-        if query_idx is not None:  # pragma: no cover
             distance_profile[query_idx] = 0.0
 
         distance_profile[~T_subseq_isfinite] = np.inf
@@ -1584,6 +1584,9 @@ def mass(
         raise ValueError(f"Q is {Q.ndim}-dimensional and must be 1-dimensional. ")
 
     check_window_size(m, max_size=Q.shape[-1])
+    if query_idx is not None:
+        query_idx = int(query_idx)
+        QT_query_idx_allclose = np.allclose(Q, T[query_idx : query_idx + m])
 
     T = _preprocess(T)
     n = T.shape[0]
@@ -1604,16 +1607,6 @@ def mass(
     if np.any(~np.isfinite(Q)):
         distance_profile[:] = np.inf
     else:
-        if query_idx is not None:
-            query_idx = int(query_idx)
-            if not np.allclose(Q, T[query_idx : query_idx + m]):  # pragma: no cover
-                msg = (
-                    "Subsequences `Q` and `T[query_idx:query_idx+m]` are "
-                    + "different but were expected to be identical. Please "
-                    + "verify that `query_idx` is correct."
-                )
-                warnings.warn(msg)
-
         T, M_T, Σ_T, T_subseq_isconstant = preprocess(
             T,
             m,
@@ -1644,6 +1637,13 @@ def mass(
         )
 
         if query_idx is not None:
+            if not QT_query_idx_allclose:  # pragma: no cover
+                msg = (
+                    "Subsequences `Q` and `T[query_idx:query_idx+m]` are "
+                    + "different but were expected to be identical. Please "
+                    + "verify that `query_idx` is correct."
+                )
+                warnings.warn(msg)
             distance_profile[query_idx] = 0
 
     return distance_profile
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -1265,7 +1265,7 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0, query_idx=None):
     if np.any(~np.isfinite(Q)):
         distance_profile[:] = np.inf
     else:
-        if query_idx is not None:
+        if query_idx is not None:  # pragma: no cover
             query_idx = int(query_idx)
             if not np.allclose(Q, T[query_idx : query_idx + m]):
                 msg = (
@@ -1606,7 +1606,7 @@ def mass(
     else:
         if query_idx is not None:
             query_idx = int(query_idx)
-            if not np.allclose(Q, T[query_idx : query_idx + m]):
+            if not np.allclose(Q, T[query_idx : query_idx + m]):  # pragma: no cover
                 msg = (
                     "Subsequences `Q` and `T[query_idx:query_idx+m]` are "
                     + "different but were expected to be identical. Please "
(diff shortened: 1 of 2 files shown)


NimaSarajpoor:
@seanlaw 
Please feel free to review (and merge if there is notthing to change). After the merge, I will resume my work on fixing-the-motif PR.

stumpy/core.py
```diff
@@ -1261,6 +1268,10 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0):
         if T_subseq_isfinite is None:
             T, T_subseq_isfinite = preprocess_non_normalized(T, m)
         distance_profile[:] = _mass_absolute(Q, T, p)
+        if query_idx is not None:  # pragma: no cover
```
seanlaw, line 1271: What do you think about doing:

```
if query_idx is not None:
    query_idx = int(query_idx)
    if not np.allclose(Q, T[query_idx:query_idx+m]):
        msg = "Subsequences `Q` and `T[query_idx:query_idx+m]` are different but were expected to be identical. Please verify that `query_idx` is correct."
        warnings.warn(msg)
    distance_profile[query_idx] = 0.0
```

stumpy/core.py
```diff
@@ -1605,6 +1624,11 @@ def mass(
             T_subseq_isconstant,
         )
 
+        if query_idx is not None:
+            query_idx = int(query_idx)
+            if np.isfinite(M_T[query_idx]) and np.isfinite(μ_Q[0]):
```
seanlaw, line 1629: Wait, why is this dependent on the mean(s) being finite? 

Also, please add warning from above

stumpy/core.py
NimaSarajpoor: Yeah..I think it is a good idea! I will add it.

stumpy/core.py
NimaSarajpoor: > Wait, why is this dependent on the mean(s) being finite?

Becuase I thought the inf value in the distance profile (caused by np.nan/np.inf value in original `T` or `Q`) may be reset to 0.0 otherwise. Now that I think more about it, I realize that such case cannot happen here. This is because we first check if `Q` has any `np.inf / np.nan` value, and if not, then we start computing its distance profile. Therefore, it should be okay to remove the if check you pointed out.

stumpy/core.py
```diff
@@ -1576,6 +1604,16 @@ def mass(
     if np.any(~np.isfinite(Q)):
         distance_profile[:] = np.inf
     else:
+        if query_idx is not None:
+            query_idx = int(query_idx)
+            if not np.allclose(Q, T[query_idx : query_idx + m]):
+                msg = (
```
NimaSarajpoor, line 1610: Note 1) This warning cannot be outside of the if/else block (i.e. above  the line`if np.any(~np.isfinite(Q))`). This is because in such case, the following inputs will result in warning:
```
Q = [np.nan, np.nan, np.nan]
T = [np.nan, np.nan, np.nan, 0, 1, 2]
query_idx = 0
```

Note 2) This warning message must be before calling `core.preprocess(...)` since we need to have the original `T` here. If we put this warning at the end (i.e. after preprocesssing `T`), then the following input does not result in warning.

```
Q = [0, 0, 0, 1]
T = [np.nan, np.nan, np.nan, 1]
query_idx = 0
```

NimaSarajpoor: As a follow up to my previous comment, what if the user provide preprocess `T` and `M_T` in the input. Then, the last example in previous example still results in wrong warning.

stumpy/core.py
```diff
@@ -1258,9 +1265,22 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0):
...
```
NimaSarajpoor, line 1271: Same as my other comment, we will get wrong warning if the provided input `T` is the preprocessed version of `T` and `T_subseq_isfinite` is provided by the user.

stumpy/core.py
seanlaw: Technically, the comparison can be done very early on:

```
check_window_size(m, max_size=Q.shape[-1])
if query_idx is not None:
    query_idx = int(query_idx)
    QT_query_idx_allclose = np.allclose(Q, T[query_idx : query_idx + m])
T = _preprocess(T)
```

Then, later:

```
    if np.any(~np.isfinite(Q)):
        distance_profile[:] = np.inf
    else:
        if query_idx is not None and not QT_query_idx_allclose:  # pragma: no cover
                msg = (
```

stumpy/core.py
seanlaw: See comment below

stumpy/core.py
NimaSarajpoor: nice! Thank you for the suggestion!!

stumpy/core.py
```diff
@@ -1238,6 +1245,9 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0):
         raise ValueError(f"`Q` is {Q.ndim}-dimensional and must be 1-dimensional. ")
 
     check_window_size(m, max_size=Q.shape[-1])
+    if query_idx is not None:  # pragma: no cover
+        query_idx = int(query_idx)
+        QT_query_idx_allclose = np.allclose(Q, T[query_idx : query_idx + m])
```
seanlaw, line 1250: Now that I think about it further, there's no reason why we can't warn ASAP:

```
    check_window_size(m, max_size=Q.shape[-1])
    if query_idx is not None:  # pragma: no cover
        query_idx = int(query_idx)
        if not np.allclose(Q, T[query_idx : query_idx + m]):
            msg = (
```

and then later:

```
if query_idx is not None: # pragma: no cover
    distance_profile[query_idx] = 0.0
```

I think this is much cleaner

NimaSarajpoor: @seanlaw 
I think this does not work when there is `np.nan`. In fact, I tried it and we get warning when there is nan. 

For instance, if you try the following  code, you will get two warnings. However, I think users should not se any warning. 

```
T = np.random.rand(10)
m = 3

T[:m] = np.nan
T[-m:] = np.nan
for i in range(len(T) - m + 1):
    Q = T[i:i+m]
    D = core.mass(Q, T, query_idx=i)
```

So, I think your previous proposal, i.e. saving the output of `np.allclose(...)` in variable `QT_query_idx_allclose` and then use it later, is good. I can add a short comment to explain
why we do not add the warning part in the beginning.

stumpy/core.py
```diff
@@ -1217,6 +1217,13 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0):
     p : float, default 2.0
         The p-norm to apply for computing the Minkowski distance.
 
+    query_idx : int, default None
+        This is the index position along the time series, `T`, where the query
+        subsequence, `Q`, is located. `query_idx` should be set to None if `Q`
+        is not a subsequence of `T`. If `Q` is a subsequence of `T`, provding
+        this argument is optional. If provided, the precision of computation
+        can be slightly improved.
```
NimaSarajpoor, line 1225: @seanlaw 
How do you feel about the following description?

```suggestion
        can be slightly improved by setting the distance between a subsequence to self to 0.0 when all the values of the subsequence are finite.
```

tests/test_precision.py
```diff
@@ -38,3 +38,20 @@ def test_mpdist_snippets_s():
     npt.assert_almost_equal(
         ref_fractions, cmp_fractions, decimal=config.STUMPY_TEST_PRECISION
     )
+
+
+def test_distace_profile():
+    # This test function raises an error when the distance profile between
+    # the query `Q = T[i: i+m]`  and `T` becomes non-zero at index `i`.
```
NimaSarajpoor, line 45: According to this comment, we may change the test function to something like this:

```
npt.assert_almost_equal(0.0, D_comp[i])
```

@seanlaw 
What do you think?

seanlaw: Isn't `npt.assert_almost_equal(D_ref, D_comp)` already a superset of `npt.assert_almost_equal(0.0, D_comp[i])`? If not, maybe you can assert both?

stumpy/core.py
seanlaw: I don't even think we should say "slight improved". We should state concrete facts:

> If provided, the distance between `Q` and `T[query_idx : query_idx + m]` will automatically be set to zero.

stumpy/core.py
NimaSarajpoor: I think if we want to consider comparing just the finite values of a subsequence, then we should do:

```
Q_idx = np.isfinite(Q) 
T_idx = np.isfinite(T[query_idx : query_idx + m])

if not np.all(Q_idx==T_idx) or not np.allcose(Q[Q_idx], T[T_idx]):
     warning.warn(...)
```

Otherwise, we do not get warning for the following case:

```
Q = [3, np.inf, 1]
T_at_query_idx = [3, 2, np.inf]

```

stumpy/core.py
NimaSarajpoor: Correct! I will revise the description.

tests/test_precision.py
NimaSarajpoor: Yeah...but I had two reasons in mind:
(1) we are just focusing on testing wether the distance to self is 0 or not. However, now that I think more about it, testing all the values of the array is better as it ensures us that we are not messing with the output when `query_idx` is provided.

(2) Fortunately, we can trust naive function for our case as naive version gives 0 for self-match distances. But using 0.0 is more direct and does not need us to know whether naive gives `0.0` for the self-match distances or not.

stumpy/core.py
seanlaw: I am cool with that

seanlaw:
@NimaSarajpoor It looks like this is ready to be merged. Can you please confirm?

NimaSarajpoor:
I think it is ready.

(FYI: I did not change the added test function. So, the ref value is coming from Naive func as usual.)

seanlaw:
@NimaSarajpoor Thank you for addressing this!

Issue #707, opened 2022-11-06
Add Ability to Reproduce Failed Unit Tests


seanlaw:
Currently, when there are precision-related issues, it is not obvious what random seed produced those failed unit/coverage tests. It would be nice to figure out a way to log this information in PyTest so that we can easily reproduce the failure. It'll require so trial-and-error and may not be easy/possible.

seanlaw:
Based on some brief research, I've learned that, we should stop setting the `np.random.seed` since (according to [this](https://builtin.com/data-science/numpy-random-seed)) 

> The problem comes in larger projects or projects with imports that could also set the seed. Using np.random.seed(number) sets what NumPy calls the global random seed, which affects all uses to the np.random.* module. Some imported packages or other scripts could reset the global random seed to another random seed with np.random.seed(another_number), which may lead to undesirable changes to your output and your results becoming unreproducible. For the most part, you will only need to ensure you use the same random numbers for specific parts of your code (like tests or functions).

So, if we have multi-threading when running our tests, each thread could potentially be setting the random global random seed at the same time and therefore it can change the arrays that are produced. Instead, within each test file, we should 

1. Generate a random integer, `test_state` (essentially a seed)
2. Create a (local) pesudo random number generator, `rng`, based on the `test_state`
3. Use the random number generator, `rng`, everywhere within this test file where a random number or array is needed
4. Pass the `test_state` into every test function so that it will be written/recorded when a test fails


# Untested Example

```
# In some test file
import numpy as np
test_state = np.random.randint(1_000_000)
rng = np.random.default_rng(test_state)

test_data = [
    (
        np.array([9, 8100, -60, 7], dtype=np.float64),
        np.array([584, -11, 23, 79, 1001, 0, -19], dtype=np.float64),
    ),
    (
        rng.random.uniform(-1000, 1000, [8]).astype(np.float64),
        rng.random.uniform(-1000, 1000, [64]).astype(np.float64),
    ),
]


@pytest.mark.parametrize("state", test_state)
@pytest.mark.parametrize("Q, T", test_data)
def test_compute_mean_std_multidimensional(state, Q, T):
    m = Q.shape[0]

    Q = np.array([Q, rng.random.uniform(-1000, 1000, [Q.shape[0]])])
    T = np.array([T, T, rng.random.uniform(-1000, 1000, [T.shape[0]])])

    ref_μ_Q, ref_σ_Q = naive_compute_mean_std_multidimensional(Q, m)
[...]

PR #864, opened 2023-05-22
add param isconstant to ostinato

NimaSarajpoor:


tests/test_ostinato.py
```diff
@@ -78,3 +80,60 @@ def test_deterministic_ostinatoed(seed, dask_cluster):
...
+        Ts = [np.random.rand(n) for n in [64, 128, 256]]
+
+        l = 64 - m + 1
+        subseq_isconsant = np.full(l, 0, dtype=bool)
+        subseq_isconsant[np.random.randint(0, l)] = True
+        Ts_subseq_isconstant = [
+            subseq_isconsant,
+            None,
+            isconstant_custom_func,
+        ]
```
NimaSarajpoor, line 128: Please see the lines above. Is it okay that the list `Ts_subseq_isconstant` contains different types in the unit testing? My goal is:
(1) to consider different types for T_subseq_isconstant
(2) avoid having constant subseq in one of the time series (in `Ts`) as I think the radius beocmes 0 otherwise. That is why one of element in `Ts_subseq_isconstant` is `None`.

NimaSarajpoor (commented):
@seanlaw 
I just left one comment. Other than that, I think this PR is ready.

tests/test_ostinato.py
seanlaw: I think this is okay.

seanlaw:
@NimaSarajpoor Very clean. Well done! Thank you

seanlaw:
Hmmm. @NimaSarajpoor Would you mind taking a look?

https://github.com/TDAmeritrade/stumpy/actions/runs/5051590328/jobs/9063569551

NimaSarajpoor:
@seanlaw 
Hard to reproduce the error without having `T`. I tried it for 300 randomly-generated `T` (simply for seed==0 to 300) but couldn't get the error. 

Do you think it would be a good idea to add `seed = np.ranomd.randint(10000)` in this test function? So, if we get an error at some point in the future, we know that the input `T` can be reproduced by a seed that is in range(10000).

seanlaw:
I will need to think about this more. We should track this in issue #707

PR #870, opened 2023-05-28
Fix #869 fixing docstrings

### LINKED ISSUE #869: mstump: description of idx in private functions
### Discussed in https://github.com/TDAmeritrade/stumpy/discussions/867

<div type='discussions-op-text'>

<sup>Originally posted by **NimaSarajpoor** May 28, 2023</sup>
**In module mstump.py:**

The function `_mstump` calls the function `_compute_multi_D`  which has the argument `idx` with the following description:

```
idx : int
        The row index in `T`
```

The function `_mstump` also calls the function `core._compute_multi_PI` that has the same argument `idx` with the same description.

However, I think this description might be incorrect. If `T` is multi-dimensional, i.e. 2D numpy array, where each row corresponds to one time series, then the description of `idx` should be: 
```
the column index of `T`
```

This is based on what I can see, regarding `idx`, in the following lines of code:

https://github.com/TDAmeritrade/stumpy/blob/76cb9806dec2d32f80ff20a5a3c484c34b7a54bf/stumpy/mstump.py#L800-L814</div>

NimaSarajpoor:


seanlaw:
Thanks @NimaSarajpoor

PR #871, opened 2023-05-29
add isconstant param to mstump

NimaSarajpoor:


tests/naive.py
```diff
@@ -559,17 +602,25 @@ def multi_distance_profile(query_idx, T, m, include=None, discords=False):
...
-def mstump(T, m, excl_zone, include=None, discords=False):
+def mstump(T, m, excl_zone, include=None, discords=False, Ts_subseq_isconstant=None):
+    # Ts_subseq_isconstant is None, or list
     T = T.copy()
 
     d, n = T.shape
     k = n - m + 1
 
+    if Ts_subseq_isconstant is None:
+        Ts_subseq_isconstant = [None] * d
```
NimaSarajpoor, line 613: In ostinato, input is `Ts` (a list of time series) and hence it is reasonable to choose the name `Ts_susbeq_isconstant`. 

Here, in `mstump`, the input `T` is a 2D numpy.ndarray. However, we still need to consider type `list` for the isconstant param. This is because a user may provide a list  of callable objects. Hence, I was wondering if it should be okay to choose the name `Ts_subseq_isconstant` here.

NimaSarajpoor (commented):
@seanlaw 
Could you please check out the only comment I left for now? Afterwards, I start with adding test function and then adding the param isconstant to the performant module.

tests/naive.py
seanlaw: No, I would keep the name as `T_subseq_isconstant`. In the case of `ostinato`, the time series are unrelated. In this case, it is clear that the time series, `T`, is a 2-dimensional time series, which is related to each other. So, either `T_subseq_isconstant` is a 2D bool array, a list of callable functions, a list of 1D bool and callable objects, or a single callable function. Does that answer your question?

NimaSarajpoor: Yeah! It does! Thank you for the suggestion. I will revise the code accordingly.

stumpy/mstump.py
```diff
@@ -459,6 +492,8 @@ def _multi_distance_profile(
         μ_Q[:, query_idx],
         σ_Q[:, query_idx],
         T_subseq_isconstant,
+        Q_subseq_isconstant[:, query_idx].reshape(-1, 1),
```
NimaSarajpoor, line 495: @seanlaw 
Is line above (i.e. line 495) readable? (Note that the output of `np.atleast_2d` is different. That is why I did not use it here)

stumpy/mstump.py
```diff
@@ -459,6 +492,8 @@ def _multi_distance_profile(
...
+        query_idx=query_idx,
```
NimaSarajpoor, line 496: @seanlaw 
I added this param `query_idx` so that we can pass it to `core.mass` and automatically set the self-match distances to 0. (If you think this small change needs its own PR, please let me know).

stumpy/mstump.py
```diff
@@ -525,13 +569,32 @@ def multi_distance_profile(
         and the Euclidean distance, respectively. This parameter is ignored when
         `normalize == True`.
 
+    T_subseq_isconstant : numpy.ndarray or callable or list, default None
```
NimaSarajpoor, line 572: @seanlaw 
Are you happy with the description of this parameter?

tests/naive.py
```diff
@@ -559,17 +597,27 @@ def multi_distance_profile(query_idx, T, m, include=None, discords=False):
...
-def mstump(T, m, excl_zone, include=None, discords=False):
+def mstump(T, m, excl_zone, include=None, discords=False, T_subseq_isconstant=None):
     T = T.copy()
 
     d, n = T.shape
     k = n - m + 1
 
+    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
+        T_subseq_isconstant = [T_subseq_isconstant] * d
+    # else means T_subseq_isconstant is list or a numpy 2D array
```
NimaSarajpoor, line 608: @seanlaw 
This is naive function. So, I assume you are okay with having the comment above in the code. If not, please let me know.

NimaSarajpoor (commented):
@seanlaw 
I have a few comments. Can you please go through them and let me know if you have any feedback?

NimaSarajpoor:
There is still one function that needs to support the parameter `T_subseq_isconstant`, and that is the function `subspace`.

In this function, we can see the following lines:

https://github.com/TDAmeritrade/stumpy/blob/e91d813f553f14a7207c310ccaef1887fea02ab5/stumpy/mstump.py#L185-L193

where the variables `subseqs ` and `neighbors` seem to have the shape `(d, m)`. In other words, they are just a multi-dim subseq (or 2D subseq). 

Hence, the D in line 193, i.e. `D = np.linalg.norm(disc_subseqs - disc_neighbors, axis=1) `, is an array of length `d` representing the z-norm euclidean distance between `subseqs` and `neighbors` (AFTER being discretized). Hence, `D[idx]` is basically the z-norm euclidean distance between `subseqs[idx]`  and `neighbors[idx]` (AFTER being discretized). 

Now the question is: How should we apply the "isconstant" concept here? One straightforward approach is to just say if both `subseqs[idx]`  and `neighbors[idx]` are constant, their distance is 0, i.e. `D[idx]=0`.  

@seanlaw 
To me, if a user says this subseq is constant, then it is constant (I mean...it should be treated like an all-zero subseq). So, if we discretize this all-zero subseq, it should still be constant. Hence, it makes sense to just set `D[idx]=0` when both subseqs[idx] and neighbors[idx] are constant before the descritization. 

What do you think? Is there anything that I am missing here?

seanlaw (changes requested):

stumpy/mstump.py
```diff
@@ -1170,8 +1180,27 @@ def mstump(T, m, include=None, discords=False, normalize=True, p=2.0):
     T_A = T
     T_B = T_A
 
-    T_A, M_T, Σ_T, T_subseq_isconstant = core.preprocess(T_A, m)
-    T_B, μ_Q, σ_Q, Q_subseq_isconstant = core.preprocess(T_B, m)
+    T_A = core._preprocess(T_A)
+    T_B = core._preprocess(T_B)
+
+    T_A_subseq_isconstant = T_subseq_isconstant
+    if T_A_subseq_isconstant is None or callable(T_A_subseq_isconstant):
```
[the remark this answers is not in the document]
seanlaw: It's probably fine leaving it out.

seanlaw:
> One straightforward approach is to just say if both subseqs[idx] and neighbors[idx] are constant, their distance is 0, i.e. D[idx]=0.

That sounds reasonable to me.

> To me, if a user says a subseq is constant, then it is constant (I mean...it should be treated like an all-zero subseq). So, if we discretize this all-zero subseq, it should still be constant. Hence, it makes sense to just set D[idx]=0 when both subseqs[idx] and neighbors[idx] are constant before the descritization. What do you think? Is there anything that I am missing here?

Yeah, that sounds fine. At the end of the day, the user providing `T_subseq_isconstant` is likely rare and it is important to ensure that the result they get by applying the default (i.e., doing nothing) is identical to when they provide a function that is identical to the default `rolling_isconstant` function. Prior to this PR, what would happen to the subspace calculation with just default values? We need to make sure that we don't change the behavior of the outputs of the PREVIOUS defaults with this PR. Does that make sense?

stumpy/mstump.py
NimaSarajpoor: Cool! That is more readable for sure. I will revise the code.

stumpy/mstump.py
```diff
@@ -525,13 +569,32 @@ def multi_distance_profile(
         and the Euclidean distance, respectively. This parameter is ignored when
         `normalize == True`.
 
+    T_subseq_isconstant : numpy.ndarray or callable or list, default None
+        A parameter that is used to show whether a subsequence of a time series in `T`
+        is constant(True) or not. T_subseq_isconstant can be a 2D boolean numpy.ndarry
+        or it be a callable function, which will be applied to each time series in `T`.
+        It can also be list of length `T.shape[0]`, i.e. dimension size of `T`. In this
```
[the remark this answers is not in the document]
NimaSarajpoor: Thanks for the suggestion! I was wondering if we should avoid `T.shape[0]`. Because, I think `T` can also be a dataframe in which each time series is provided as one column. So, in that case, the number of time series is `T.shape[1]`.

stumpy/mstump.py
```diff
@@ -525,13 +569,32 @@ def multi_distance_profile(
...
+
     Returns
     -------
     D : numpy.ndarray
         Multi-dimensional distance profile for the window with index equal to
         `query_idx`
     """
-    T, M_T, Σ_T, T_subseq_isconstant = core.preprocess(T, m)
+    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
+        T_subseq_isconstant = [T_subseq_isconstant] * T.shape[0]
```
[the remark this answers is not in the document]
NimaSarajpoor: Yeah. In fact, I think it SHOULD BE POSSIBLE to just pass `None` or function. However, since I wanted to handle list (containing a combination of 1D boolean array, function, `None`), I decided to just repeat them and make a list first. And, finally just use the list.

NimaSarajpoor:
>  and it is important to ensure that the result they get by applying the default (i.e., doing nothing) is identical to when they provide a function that is identical to the default `rolling_isconstant` function. 

Right! I should check that out in the main branch and just get some sense about the output and see what happens in default.

> Prior to this PR, what would happen to the subspace calculation with just default values? We need to make sure that we don't change the behavior of the outputs of the PREVIOUS defaults with this PR. Does that make sense?

Yep! I will go through it more carefully to make sure that we are not messing with the default mode, i.e. output of existing version.


> In case it matters, you would likely apply the same logic to `mdl` as well.

Right! I missed that while I was going through the script. Thanks for pointing that out.

NimaSarajpoor:
I have been thinking about the discretizing process and if we should be worried about how it handles (user-defined) constant subseqs. I now think it should be straightforward. This is what I noticed:

Before discretization, we first z-normalize the subseqs. Now, something very cool is happening here: the mean of the z-normalized subseq is zero. This means that the z-normalized subseq is either ALL zero or it has BOTH positive and negative values. (Because otherwise, the mean cannot be zero). 

Therefore, when we discretize a subseq S, I think we cannot have identical values in the discretized version unless the non-discretized version, i.e. `S`, is all zero. For instance, let's assume `a` is the original subsequence which is NOT an all-zero array. Hence, its z-normalized version, `az`, has BOTH positive and negative values . Also, let's assume the value of param `n_bit`  in the function `mstump._inverse_norm` is 1. This result in bin with edge 0, which makes sense since `0` is exactly the middle of normal distribution. Hence, negative values in `az` are replaced with 0 and positive values are replaced with `1`. In other words, a non-constant subsequence has a normalized version that has both neg and pos values, and therefore the discretized version has at least one occurance of 1  and  at least one occurance of 0. Hence, it cannot be constant. 

Now, what if user defines constant subseqs by passing `isconstant` param? I think it should be fine. The only thing we need to do is to replace those subseq with all-zero subseq, and that's it. It will be handled correctly via discretization, and the distance will be computed according to what is implemented already in the code.

seanlaw:
> Now, what if user defines constant subseqs by passing isconstant param? I think it should be fine. The only thing we need to do is to replace those subseq with all-zero subseq, and that's it. It will be handled correctly via discretization, and the distance will be computed according to what is implemented already in the code. Also, it is important to note that we do not z-normalize subsequence after discretization. Hence, the constant-related issue does not matter at that point, i.e. after discretization, as the discretization process reflects the constant subsequence as explained earlier.

Ahh, I see what you mean. Since we z-normalize prior to discretization then we should be mostly safe (at least in the case where the subsequence is "truly" constant). In the case where the user provides `isconstant` then we "fake" the z-normalization and simply set the z-normalized subsequence to all-zero and, to the discretization, this subsequence is indistinguishable from the first case ("truly" constant). Is that correct?

If so, then it sounds like this greatly simplifies things! Thanks for thinking through it @NimaSarajpoor. It's been so long that I've forgotten all of the inner workings. Hopefully, the code wasn't too painful to understand?

NimaSarajpoor:
> "fake" the z-normalization and simply set the z-normalized subsequence to all-zero and, to the discretization, this subsequence is indistinguishable from the first case ("truly" constant). Is that correct?

That is correct! 


> Hopefully, the code/logic wasn't too painful to understand?

I think it was good overal :)

NimaSarajpoor (commented):
@seanlaw 
I think we are close to finishing this PR. I left a few comments (some of them are for me. So, you can ignore those as I will take care of them for sure unless you have a suggestion / opposite opinion regarding them)

seanlaw:
> I think we are close to finishing this PR. I left a few comments (some of them are for me. So, you can ignore those as I will take care of them for sure unless you have a suggestion / opposite opinion regarding them)

Looks good to me. I will do a more thorough pass at the end

> Right! I should check that out in the main branch and just get some sense about the output and see what happens in default.

Please confirm that you've checked this already?

NimaSarajpoor:
> > Right! I should check that out in the main branch and just get some sense about the output and see what happens in default.
> 
> Please confirm that you've checked this already?

This is what I did:

```
# data and code
d, n = 3, 64
m = 8

seed = 1000
np.random.seed(seed)

T = np.random.rand(d, n)
P, I = stumpy.mstump(T, m)
IDX = np.argmin(P, axis=1)
nn_IDX = I[np.arange(len(IDX)), IDX]
bit_sizes, subspaces = mdl(T, m, IDX, nn_IDX)
```

I considered the following cases:
* case 1: compare the output of branch "main" and "this PR"
* case 2: set `T[:, 5 : 5 + m] = 2.0` and `T[:, 15 : 15 + m] = -2.0` and compare "main" vs "this PR"
* case 3: set `T[:, 5 : 5 + m] = T[:, 5 : 5 + m] * 1e-6` and `T[:, 15 : 15 + m] = T[:, 15 : 15 + m] * 1e-6` in main and in this PR (with/without passing `T_subseq_isconstant`)
* case 4: just pass `T_subseq_isconstant` where `T_subseq_isconstant[:, 5]` and `T_subseq_isconstant[:,15]` are True. 

Does this suffice?

seanlaw:
> Does this suffice?

Sounds good to me! Thanks

NimaSarajpoor:
@seanlaw 
The comments are addressed. The support for param "isconstant" is also added to the "mstumped" module.

seanlaw:
> @seanlaw 
> The comments are addressed. The support for param "isconstant" is also added to the "mstumped" module.
> 
> 

Thank you. I will find some time to review it

stumpy/mstump.py
```diff
@@ -316,6 +367,15 @@ def mdl(
         and the Euclidean distance, respectively. This parameter is ignored when
         `normalize == True`.
 
+    T_subseq_isconstant : numpy.ndarray or callable or list, default None
```
seanlaw, line 370: "T_subseq_isconstant : numpy.ndarray, function, or list, default None"

stumpy/mstump.py
```diff
@@ -316,6 +367,15 @@ def mdl(
...
+        A parameter that is used to show whether a subsequence of a time series in `T`
+        is constant (True) or not. T_subseq_isconstant can be a 2D boolean numpy.ndarry
+        or a function that can be applied to each time series in `T`. Alternatively, for
+        maximum flexibility, a list (with length equal to the total number of time
+        series) may also be used. In this case, T_subseq_isconstant[i] corresponds to
+        the i-th time series T[i] and each element in the list can either be 1D boolean
```
seanlaw, line 376: "be A 1D boolean..."

stumpy/mstump.py
```diff
@@ -178,14 +215,27 @@ def subspace(
     T = core._preprocess(T)
     core.check_window_size(m, max_size=T.shape[-1])
 
+    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
```
seanlaw, line 218: If this region of code is identical and repeated in multiple functions then we should refactor and move it into `core.py`

```
def process_multi_isconstant(T, m, T_subseq_isconstant):
    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
        T_subseq_isconstant = [T_subseq_isconstant] * T.shape[0]

    T_subseq_isconstant = np.array(
        [
            core.rolling_isconstant(T[i], m, T_subseq_isconstant[i])
            for i in range(T.shape[0])
        ]
    )

    return T_subseq_isconstant
```

seanlaw: This will make it easier to maintain in the future as we'd only need to fix/change it in one place

stumpy/mstump.py
```diff
@@ -353,6 +413,16 @@ def mdl(
...
```
seanlaw, line 416: See refactor above

stumpy/mstump.py
```diff
@@ -481,9 +562,18 @@ def _multi_distance_profile(
     return D
 
 
-@core.non_normalized(maamp_multi_distance_profile)
+@core.non_normalized(
+    maamp_multi_distance_profile, exclude=["T_subseq_isconstant", "normalize"]
```
seanlaw, line 566: Please flip the order to be `["normalize", "T_subseq_isconstant"]` to reflect the order in which they appear in the function signature

stumpy/mstump.py
```diff
@@ -525,13 +615,34 @@ def multi_distance_profile(
         and the Euclidean distance, respectively. This parameter is ignored when
         `normalize == True`.
 
+    T_subseq_isconstant : numpy.ndarray or callable or list, default None
```
seanlaw, line 618: Please replace the first line with "T_subseq_isconstant : numpy.ndarray, function, or list, default None"

stumpy/mstump.py
```diff
@@ -525,13 +615,34 @@ def multi_distance_profile(
...
+        A parameter that is used to show whether a subsequence of a time series in `T`
+        is constant (True) or not. T_subseq_isconstant can be a 2D boolean numpy.ndarry
+        or a function that can be applied to each time series in `T`. Alternatively, for
+        maximum flexibility, a list (with length equal to the total number of time
+        series) may also be used. In this case, T_subseq_isconstant[i] corresponds to
+        the i-th time series T[i] and each element in the list can either be 1D boolean
```
seanlaw, line 624: "be A 1D boolean"

stumpy/mstump.py
```diff
@@ -525,13 +615,34 @@ def multi_distance_profile(
...
+
     Returns
     -------
     D : numpy.ndarray
         Multi-dimensional distance profile for the window with index equal to
         `query_idx`
     """
-    T, M_T, Σ_T, T_subseq_isconstant = core.preprocess(T, m)
+    T = core._preprocess(T)
+    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
```
seanlaw, line 634: See refactor above

stumpy/mstump.py
```diff
@@ -1059,6 +1179,15 @@ def mstump(T, m, include=None, discords=False, normalize=True, p=2.0):
         and the Euclidean distance, respectively. This parameter is ignored when
         `normalize == True`.
 
+    T_subseq_isconstant : numpy.ndarray or callable or list, default None
```
seanlaw, line 1182: This first line should be "T_subseq_isconstant : numpy.ndarray, function, or list, default None"

stumpy/mstump.py
```diff
@@ -1059,6 +1179,15 @@ def mstump(T, m, include=None, discords=False, normalize=True, p=2.0):
...
+        A parameter that is used to show whether a subsequence of a time series in `T`
+        is constant (True) or not. T_subseq_isconstant can be a 2D boolean numpy.ndarry
+        or a function that can be applied to each time series in `T`. Alternatively, for
+        maximum flexibility, a list (with length equal to the total number of time
+        series) may also be used. In this case, T_subseq_isconstant[i] corresponds to
+        the i-th time series T[i] and each element in the list can either be 1D boolean
```
seanlaw, line 1188: "be A 1D boolean"

stumpy/mstump.py
```diff
@@ -1100,8 +1229,27 @@ def mstump(T, m, include=None, discords=False, normalize=True, p=2.0):
     T_A = T
     T_B = T_A
 
-    T_A, M_T, Σ_T, T_subseq_isconstant = core.preprocess(T_A, m)
-    T_B, μ_Q, σ_Q, Q_subseq_isconstant = core.preprocess(T_B, m)
+    T_A = core._preprocess(T_A)
+    T_B = core._preprocess(T_B)
+
+    T_A_subseq_isconstant = T_subseq_isconstant
+    if T_A_subseq_isconstant is None or callable(T_A_subseq_isconstant):
```
seanlaw, line 1236: See refactor above

stumpy/mstump.py
seanlaw: Also, do we need to handle `fix_isconstant_isfinite_conflicts` too?

stumpy/mstump.py
```diff
@@ -1100,8 +1229,27 @@ def mstump(T, m, include=None, discords=False, normalize=True, p=2.0):
...
+
+    T_A_subseq_isconstant = np.array(
+        [
+            core.rolling_isconstant(T_A[i], m, T_A_subseq_isconstant[i])
+            for i in range(T_A.shape[0])
+        ]
+    )
+    T_B_subseq_isconstant = T_A_subseq_isconstant
+
+    T_A, M_T, Σ_T, T_subseq_isconstant = core.preprocess(
```
seanlaw, line 1247: A little bit of code smell here. Feels a tiny bit redundant to do `_preprocess` at the top and then `preprocess` (which has the `_preprocess` step). I understand why it needs to be done this way

stumpy/mstumped.py
```diff
@@ -184,8 +185,20 @@ def _dask_mstumped(
...
+    maamped,
+    exclude=["normalize", "T_subseq_isconstant"],
+)
+def mstumped(
+    client,
+    T,
+    m,
+    include=None,
+    discords=False,
+    p=2.0,
```
seanlaw, line 198: Thanks for catching that

stumpy/mstumped.py
```diff
@@ -225,11 +238,25 @@ def mstumped(client, T, m, include=None, discords=False, normalize=True):
...
+        The p-norm to apply for computing the Minkowski distance. Minkowski distance is
+        typically used with `p` being 1 or 2, which correspond to the Manhattan distance
+        and the Euclidean distance, respectively.
+
     normalize : bool, default True
         When set to `True`, this z-normalizes subsequences prior to computing distances.
         Otherwise, this function gets re-routed to its complementary non-normalized
         equivalent set in the `@core.non_normalized` function decorator.
 
+    T_subseq_isconstant : numpy.ndarray or callable or list, default None
```
seanlaw, line 251: First line: "T_subseq_isconstant : numpy.ndarray, function, or list, default None"

stumpy/mstumped.py
```diff
@@ -225,11 +238,25 @@ def mstumped(client, T, m, include=None, discords=False, normalize=True):
...
+        A parameter that is used to show whether a subsequence of a time series in `T`
+        is constant (True) or not. T_subseq_isconstant can be a 2D boolean numpy.ndarry
+        or a function that can be applied to each time series in `T`. Alternatively, for
+        maximum flexibility, a list (with length equal to the total number of time
+        series) may also be used. In this case, T_subseq_isconstant[i] corresponds to
+        the i-th time series T[i] and each element in the list can either be 1D boolean
```
seanlaw, line 257: "be A 1D boolean"

stumpy/mstumped.py
```diff
@@ -275,8 +302,27 @@ def mstumped(client, T, m, include=None, discords=False, normalize=True):
     T_A = T
     T_B = T_A
 
-    T_A, M_T, Σ_T, T_subseq_isconstant = core.preprocess(T_A, m)
-    T_B, μ_Q, σ_Q, Q_subseq_isconstant = core.preprocess(T_B, m)
+    T_A = core._preprocess(T_A)
+    T_B = core._preprocess(T_B)
+
+    T_A_subseq_isconstant = T_subseq_isconstant
```
seanlaw, line 308: See factor above

stumpy/mstumped.py
```diff
@@ -275,8 +302,27 @@ def mstumped(client, T, m, include=None, discords=False, normalize=True):
...
+
+    T_A_subseq_isconstant = np.array(
+        [
+            core.rolling_isconstant(T_A[i], m, T_A_subseq_isconstant[i])
+            for i in range(T_A.shape[0])
+        ]
+    )
+    T_B_subseq_isconstant = T_A_subseq_isconstant
+
+    T_A, M_T, Σ_T, T_subseq_isconstant = core.preprocess(
```
seanlaw, line 320: See code smell comment above

seanlaw (changes requested):
@NimaSarajpoor Let's start with these comments

stumpy/mstump.py
```diff
@@ -43,6 +54,16 @@ def _multi_mass(Q, T, m, M_T, Σ_T, μ_Q, σ_Q, T_subseq_isconstant):
     T_subseq_isconstant : numpy.ndarray
         A boolean array that indicates whether a subsequence in `T` is constant (True)
 
+    Q_subseq_isconstant : numpy.ndarray
```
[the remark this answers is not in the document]
NimaSarajpoor: Right! How about "that indicates whether the subsequence Q is constant"

stumpy/mstump.py
```diff
@@ -141,6 +169,15 @@ def subspace(
         and the Euclidean distance, respectively. This parameter is ignored when
         `normalize == True`.
 
+    T_subseq_isconstant : numpy.ndarray or callable or list, default None
```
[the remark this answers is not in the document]
NimaSarajpoor: I thought I fixed the "callable" thing. Sorry about that.

stumpy/mstump.py
NimaSarajpoor: > Also, do we need to handle fix_isconstant_isfinite_conflicts too?

I think so. If a user provides a 2D numpy boolean array, it should be checked.  I may go back and check all modules that use `T_x_subseq_isconstant` and see if they need to handle it or not.

stumpy/mstump.py
NimaSarajpoor: > A little bit of code smell here. Feels a tiny bit redundant to do `_preprocess` at the top and then `preprocess` (which has the `_preprocess` step). I understand why it needs to be done this way

One way to resolve this code smell is to put this code inside the `core.preprocess`. Note that this function, i.e. `core.preprocess`, already supports multi-dim time series. So, we can add support for multi-isconstant "T_subseq_isconstant", when provided as type `list` by the user.

What do you think @seanlaw ?

seanlaw: Sure. As long as it is still readable

seanlaw:
@NimaSarajpoor Do I need to review anything yet?

NimaSarajpoor:
> @NimaSarajpoor Do I need to review anything yet?

No, I haven't made the changes yet. Hopefully, I will push them in upcoming days.

stumpy/mstump.py
NimaSarajpoor: @seanlaw 
I was wondering if it would be a good idea to add the function `process_isconstant` that supports BOTH 1D and 2D (multi-) time series data, and (as the name may suggest) apply `fix_isconstant_isfinite_conflicts` too.

So, I have something like this in mind:

```
def process_isconstant(T, m, T_subseq_isconstant):
    """docstring"""
    ndim = T.ndim

    T = np.atleast_2d(T)
    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
        T_subseq_isconstant = [T_subseq_isconstant] * T.shape[0]

    T_subseq_isconstant = np.array(
        [
            rolling_isconstant(T[i], m, T_subseq_isconstant[i])
            for i in range(T.shape[0])
        ]
    )

    T_subseq_isconstant = fix_isconstant_isfinite_conflicts(T, m, T_subseq_isconstant)

    if ndim == 1:
        T_subseq_isconstant = T_subseq_isconstant.flatten()
    
    return T_subseq_isconstant

```

So, this function supports 1D and 2D time series `T` and the dimention of its output depends on the dimention of the input `T`. I remember you mentioned that you prefer to keep  "the computation of isconstant" and "the check for having conflict with isfinite" separated. I think we still have that separation. We just provide a wrapper around them so that we do not forget to check the isfinite whenever we use `core.rolling_isconstant`.

---

For instance, in `core.preprocess`, we can do:

```
def preprocess(...):
    """docstring"""
    T = _preprocess(T, copy)
    check_window_size(m, max_size=T.shape[-1])

    T[np.isinf(T)] = np.nan

    T_subseq_isconstant = process_isconstant(T, m, T_subseq_isconstant)
    if M_T is None or Σ_T is None:
    T[np.isnan(T)] = 0
```

What do you think?

seanlaw: Yep, that makes sense. Can you please:
1. Make this change (i.e., adding a `process_isconstant` function in separate issue/PR
2. Fix all PREVIOUS `isconstant` functions and submit a separate PR (if applicable)
3. Merge the above changes into this PR
4. Update this PR to leverage this change

NimaSarajpoor: Thanks for the feedback! I will create a new PR to take care of this first before resuming our work here.

stumpy/mstump.py
```diff
@@ -426,6 +482,10 @@ def _multi_distance_profile(
     T_subseq_isconstant : numpy.ndarray
         A boolean array that indicates whether a subsequence in `T_A` is constant (True)
 
+    Q_subseq_isconstant : numpy.ndarray
+        A boolean array that indicates whether the query subsequence `T_B` is
```
NimaSarajpoor, line 486:
```suggestion
        A boolean array that indicates whether a subsequence in `T_B` is
```

NimaSarajpoor: side note:
The description of the argument `query_idx` is not complete.

```
query_idx : int
    The start index of the (multi-dimensional) query subsequence in `T_B`
```

NimaSarajpoor:
@seanlaw 
I think this PR is good. Plz let me know any comment you may have.

seanlaw:
> @seanlaw I think this PR is good. Plz let me know any comment you may have.

Thank you. I will find some time to review it

tests/naive.py
```diff
@@ -532,11 +550,30 @@ def apply_include(D, include):
...
+    if T_subseq_isconstant is None or callable(T_subseq_isconstant):
+        T_subseq_isconstant = [T_subseq_isconstant] * d
+
+    T_subseq_isconstant = np.array(
+        [rolling_isconstant(T[i], m, T_subseq_isconstant[i]) for i in range(d)]
+    )
+
     Q = T[:, query_idx : query_idx + m]
-    D = multi_mass(Q, T, m, include, discords)
+    Q_subseq_isconstant = T_subseq_isconstant[:, query_idx].reshape(-1, 1)
```
seanlaw, line 567: Should we use `expand_dims` or `atleast_2d` here instead of `reshape`?

seanlaw (changes requested):
Please see comment

tests/naive.py
NimaSarajpoor: Correct! I will fix this one (and other similar cases) in naive.

seanlaw: Please do a quick grep for `reshape(-1, 1)` in the codebase and replace where appropriate

NimaSarajpoor: I fixed the cases in `.py` scripts (in both `./tests/` and `./stumpy/` directories). 

Also: 
I found a few cases in the `.ipynb` notebooks in `docs/` but did not change them. If needed, please let me know and I will fix those too.

seanlaw:
@NimaSarajpoor this was a more challenging one. Thank you for your contribution and persistence!

Issue #213, opened 2020-07-10
Add Left and Right Matrix Profile Indices to Scrump


seanlaw:
Current only global matrix profile values are being tracked but it would be useful to have the left and right matrix profile indices as well

Issue #212, opened 2020-07-03
Finding Anomalies Using STUMPI


seanlaw:
@hmcoservit originally [wrote](https://github.com/TDAmeritrade/stumpy/issues/207#issuecomment-653441736): 

> @seanlaw According to my tests so far on anomaly detection, I can see the effect of Z-normalization more and more. It really makes it difficult to detect the peaks in time series (using STUMPI left MP). Also when sub sequences are flat (but noisy) the matrix profile would not give ideal results for anomaly discovery specially when the sub sequence value is small. Therefore, impacting the precision heavily. What do you think about this? I am not sure if it is possible to adapt it to suit anomaly detection more.

seanlaw:
@hmcoservit Moving the discussion here as the other issue is focused on implementing non-normalized distances. 


So, since matrix profiles mainly look at subsequences, if you are looking for (roughly) single point peaks in a time series then matrix profiles are not a good approach. Have you ever looked at:

[Twitter’s Anomaly Detection ](https://blog.twitter.com/engineering/en_us/a/2015/introducing-practical-and-robust-anomaly-detection-in-a-time-series.html)

Or

[Netflix Outlier Detection](https://link.medium.com/dtpwLGE3O7)

Perhaps, you can provide some examples that we can explore together? I am sure that this will help others as well

hmcoservit:
@seanlaw excluding contextual anomalies in time series, the target is to detect collective and point anomalies. 

The effect of z-normalization might be more visible when either the sub sequence value is too small or there are flat regions in the time series as  [discussed in this paper](https://www.researchgate.net/publication/338811628_Implications_of_Z-Normalization_in_the_Matrix_Profile). They propose a new solution to overcome this in the case of anomaly detection. Another thing that can be done to overcome this is to increase the sub sequence length to a long period, but doing so would eventually add delays to detection as new points coming in do not update the MP values right away. 

Regarding Netflix's solution, is is mainly for processing data in batches, and unsuitable for real-time streaming applications [according to this article ](https://www.sciencedirect.com/science/article/pii/S0925231217309864). And for Twitter's one, it doesn't seem to be maintained and has been known to [perform worse than RRCF ](https://github.com/numenta/NAB) according to NAB. 

I will try to share some visual examples here as soon as possible ;)

seanlaw:
@hmcoservit I'm going to close this for now but please feel free to re-open when you have some examples to share

hmcoservit:
[...]

Issue #195, opened 2020-06-02
Remove Reduced Precision in tests/test_mstump.py Unit Tests


seanlaw:
Currently, the `tests/test_mstump.py::test_multi_mass_seeded` and `tests/test_mstump.py::test_multi_mass` unit tests have an assertion that is accurate to only 6 decimal places (rather than 7):

```
npt.assert_almost_equal(left, right, decimal=6)
```

Hopefully, this condition can be removed once the `centered sum-of-products` is implemented in issue #196 for computing the distance

Here is a [recent paper describing better precision and parallelization for mSTAMP](https://link.springer.com/chapter/10.1007/978-3-030-50743-5_6)

Issue #186, opened 2020-05-26
Add Minimum Description Length MDL Algorithm


seanlaw:
According to the [mSTAMP paper](https://www.cs.ucr.edu/~eamonn/Motif_Discovery_ICDM.pdf) one could use the "minimum description length" (MDL) algorithm to decide how many dimensions to select. It might be helpful to add this to STUMPY.

See Figure 6 and starting in the paragraph right above Figure 6. Additionally, see this [Matrix Profile III paper](https://www.cs.ucr.edu/~eamonn/PID4481999_Matrix%20Profile_III.pdf) which discusses the MDL framework in more detail.

There is also a [background paper by Tanaka et al](https://www.researchgate.net/publication/243134007_Discovery_of_Time-Series_Motif_from_MultiDimensional_Data_Based_on_MDL_Principle) and a [subsequent paper by Hu et al](https://www.researchgate.net/publication/220766245_Discovering_the_Intrinsic_Cardinality_and_Dimensionality_of_Time_Series_Using_MDL)

seanlaw:
@shfa5275 @mexxexx Do you have any ideas how MDL works? I've been looking at some of the papers but it's not clear to me how to implement MDL relative to the mSTAMP paper. Do you have any experience with this?

seanlaw:
@mcyeh We've already implemented mSTOMP but the details on how to apply MDL to the results is vague. Would you mind giving me a high level overview of the MDL steps that we need to take after the multi-dimensional matrix profile (and indices) are computed?

shfa5275:
Hej @seanlaw!

I did not pursue the MDL myself. But the following paper uses the same MDL principle as the 
"Yeh, Chin-Chia Michael, et al. (2017) Matrix Profile VI: Meaningful Multidimensional Motif Discovery. ICDM:565-574"

The paper is entitled:
E. Galbrun, P. Cellier, N. Tatti, A. Termier, and B. Crémilleux. Mining periodic patterns with a MDL criterion. In the proceedings of the European Conference on Machine Learning and Principles and Practice of Knowledge Discovery in Databases (ECML-PKDD'18), September 2018

See the references to the MDL in both the papers, i.e.,:
J. Rissanen. Modeling by shortest data description. Automatica, 14(5):465–471, 1978.
(I was unable to get hold of this paper myself)

The work of E. Galbrun et al., is also implemented in Python. The code is available at:
https://github.com/nurblageij/periodic-patterns-mdl


[...]

PR #872, opened 2023-06-11
Fixed #606: DAMP Notebook Reproducer / Tutorial

### LINKED ISSUE #606: DAMP
[Here](https://sites.google.com/view/discord-aware-matrix-profile) is the supporting website

NimaSarajpoor:
# Pull Request Checklist

- [ ] Create a new branch


docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,408 @@
+{
```
NimaSarajpoor, line 1: <p>Add link for twin-freak</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

NimaSarajpoor, line 1: <p>The terms "train" and "test" are used here so to remain faithful to the paper. However, personally speaking, the term "train"  may not make sense in offline setting. (It may make sense online setting but we have to wait and read the corresponding part in the paper later)</p><p> </p><p>The paper introduces the term "split index"  which is the index that divides the time series <code>T</code> into two parts: train part <code>T[: split_index] </code> and test part <code>T[split_index:]</code> Then, to detect whether the subsequence <code>S_i = T[i:i+m]</code> is discord or not, all of its left neighbors upto index <code>i </code> are considered. In other words, BOTH train and test parts are considered in the decision making process for the subsequence  <code>S_i</code>.  So, it feels a little bit strange to use the terms "train" and "test"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.        Compute the top-1 discord in <code>T</code>, where the subsequence discord resides in T[split_index:]
</pre><p>We can easily improve this function to return top-k discords. However, since the implementation of DAMP in this notebook support top-1 discord for now, I decided to avoid add param <code>k</code> to the naive function.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #21.        stumpy.config.STUMPY_EXCL_ZONE_DENOM = 1
</pre><p>According to paper, the exclusion zone is set to <code>m</code>. However, I decided to not hardcode it. In <code>STUMPY</code>,we may use the default setting and see how it affects the result.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #31.                PL[i] = np.linalg.norm(core.z_norm(T[i : i + m]) - core.z_norm(T[nn_i : nn_i + m]))
</pre><p>To get profile value from profile index, we can  probably use the function <code>core._idx_to_mp</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #31.        """
</pre><p> </p><p>Add the "Returns" section</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #91.        prefix = np.power(2, int(np.ceil(np.log(m) / np.log(2))))
</pre><p> </p><p>According to the paper, MASS works better if the length of time series is a power of two. Two notes here:</p><p>(1) Make sure the lenght of time series that will be used is EXACTLY a power of two, and not just close it.</p><p>(2) Is <code>MASS</code> "really" faster when the length of time series is a power of two? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #93.        while left_nn_distance &gt;= discord_score:
</pre><p>We should be able to avoid while loop here. Note that the condition is used again later in the body of while loop to break out of it. We can use a for-loop and consider a very generous upper bound for the number of iterations, and that is the total number of subsequences to the left of <code>query_idx</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #98.                    T[start_idx : query_idx + m],
</pre><p>In line #94, we update <code>start_idx</code> as follows:</p><pre class="ql-syntax" spellcheck="false">start_idx = max(0, query_idx + m - prefix) 
</pre><p> </p><p>where prefix is doubled in each iteration. Note that the function MASS computes the distance between the query</p><pre class="ql-syntax" spellcheck="false">Q=T[query_idx : query_idx + m] 
</pre><p>and <code>T[start_idx : query_idx + m]</code> . Therefore, in the next iteration, when <code>start_idx</code> gets lower, we need to compute distance profile for a longer time series, i.e. <code>T[start_idx_lower : query_idx + m]</code></p><p> </p><p>However, this means that we are unnecessary re-computing the distance between <code>Q</code> and the subsequences in <code>T[start_idx : query_idx + m]</code> .</p><p> </p><p>Fig. 8 of the <a href="https://www.cs.ucr.edu/~eamonn/DAMP_long_version.pdf" target="_blank">paper</a> shows the correct way. However, Table 2 seems to not update the stop index.</p><p> </p><p> </p><p> </p><p> </p><p>Side Note:</p><p>The DAMP paper  compute the distance between the query <code>Q</code> and <code>T[start_idx : query_idx]</code>   as they consider one full window size for exlcusion zone. However, in this implementation, I considered to compute the distance upto <code>query_idx + m</code>  but apply <code>core.apply_excl	usion_zone</code>  instead.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #110.                dist_profile = core.mass(
</pre><p>This <code>else</code> part is very similar to the <code>if  </code> part. Hence, it seems that we can avoid repeating ourselves and avoid if/else here. Needs some improvement. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #127.    def DAMP(T, m, split_index):
</pre><p>Later, we may add the param <code>T_subseq_isconstant</code> to this function </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #145.        PL : np.ndarray
</pre><p>Please fix the "Returns" in the docstring</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #160.                PL[i] = PL[i-1]
</pre><p>If <code>i</code> is pruned, we may just set the value of <code>PL[i] </code> to <code>np.inf </code> so to make it more intuitive and readable. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #165.        PL = np.where(PL==np.inf, np.NINF, PL)
</pre><p> </p><p>Replace it with:</p><pre class="ql-syntax" spellcheck="false">mask = PL == np.inf
PL[mask] = np.NINF
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor: No need. The paper DAMP mentioned it.

NimaSarajpoor: The code is chagned to return the original PL as well. Hence, we keep `PL` and assign its modified version to a new variable.

```
PL_modified = np.where(PL==np.inf, np.NINF, PL)
```

NimaSarajpoor: Since `PL==np.inf` has a special meaning (i.e. the subsequence has at least one non-finite value or the subsequence has no left neighbor), it is better to keep it as is.

NimaSarajpoor: It is important to be aware of (close to) constant subsequences and handle them before computing the discord. 

Also, see the note in the MATLAB code shareed by the authors of DAMP: https://drive.google.com/file/d/1EPDhFXQ2goTJ5m_x1S84-KNpQFsRNVmf/view

NimaSarajpoor: > However, Table 2 seems to not update the stop index.

Also, it seems that the authors of `DAMP` in the MATLAB code did consider updating the `stop_index` which is reasonable.

NimaSarajpoor: (1) is done. (2) can be investigated later if needed.

NimaSarajpoor: For now, let's keep it as is since this is a naive function.

NimaSarajpoor:
[update]
* I made some minor improvements to the code such as enhancing docstrings, texts, etc.

* Fig. 3 and Fig. 4 of the paper are reproduced.

* To reproduce the figures and check the answers, we can use [link1](https://sites.google.com/view/discord-aware-matrix-profile/reproducibility?authuser=0) or [link2](https://docs.google.com/presentation/d/1tQQfKuKrOa3j5-WJ9peoYfZGOnJbBcl9/edit#slide=id.p34)

* DAMP code (in particular, the "BackProcessing" part) is a bit different compared to what presented in the paper. As mentioned in one of the previous comments in this PR, the "BackProcessing" algorithm provided in the table is not optimized as it recompute some of previously-analyzed subsequences. The MATLAB code, however, considers this: [see this MATLAB code](https://drive.google.com/file/d/1EPDhFXQ2goTJ5m_x1S84-KNpQFsRNVmf/view)

* DAMP's performance needs to be improved. We can check if the MATLAB code considers any other logic to improve performance as it seems it is different than what presented in the paper. The current notebook in this PR is based on the algorithm provided in the paper.

NimaSarajpoor:
[update]

I improved performance by:
* using `core._mass` intead of `core.mass`
* avoiding recomputing distances for already-analyzed subsequences (this is different than the paper but somewhat similar to the code provided in MATLAB)

Also, I simplified the code by removing if/else and applying necessary changes to the code

I was able to find top-1 discord in a time series with length ~250k, in less than three minutes.

NimaSarajpoor:
<blockquote>I am purposely trying to avoid using "anomaly" where possible and to try and be precise</blockquote><p>I see! Thanks for the input! </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505753'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<blockquote>My definition for a "discord" is a "potential anomaly" but one needs to verify whether it is an actual anomaly</blockquote><p>Cool! </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505786'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>Noted. Will revise it accordingly.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505818'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>Again, the missing article THE :) Shame on me :D Thanks for catching that. </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505840'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>I will revise the code accordinly. </p><p> </p><p>Side note: </p><p>I think I designed the code in a way that should work for any exclusion zone. I need to test it. (The algorithms in the paper are written in  way that only consider <code>excl_zone == m</code>.  For now, I am trying to be faithful to the paper as much as possible. That is why I am setting the excl_zone denom to 1 to reflect the paper's implementation) </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505870'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>I think the only way one can get a discord with infinte value (score) is to have at least one non-finite value in the time series. Wouldn't it be better to not add the proposed argument to the API and instead just let user know regarding this matter?</p><p> </p><p>I mean they can just do: <code>core.rolling_finite(T[split_index:], m)</code> .</p><p> </p><p>What do you think? </p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505896'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>Will do so.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505919'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

seanlaw:
<blockquote>just one question here: Did you intentionally remove "catching twin-freak cases" ? If yes, should we remove "i.e." as well?</blockquote><p>Oops! How about:</p><p>"left nearest neighbors so AS NOT TO miss the case where you have two rare subsequences that happen to also be similar to each other (i.e., a "twin freak")"</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505753'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

seanlaw:
<p>I agree. Let's just put a clear note in the description for PL (i.e., all infinite distances are ignored)</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505896'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>Sounds good. I will revise the code accordingly.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505753'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>You are right! In <code>_foreward_processing</code> they are always the same value.  I will remove the intermediate variables <code>start</code>.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587505992'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

NimaSarajpoor:
<p>Will revise the code accordingly.</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587506112'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

seanlaw:
<p>I think that sounds reasonable</p>

---

 <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/#comment-1587506053'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,847 @@
...
```
seanlaw, line 1: <p>The comma in the first sentence can be omitted</p><p> </p><p>For a tutorial, it would be important to explain the difference between a left MP and full MP. Then explain why a left MP is used to find discords instead of a full MP. Afterward, I would omit the parentheses and just write out "left nearest neighbor" and "left matrix profile" </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

seanlaw, line 1: <p>"discovering POTENTIAL anomalies for the remaining points"</p><p> </p><p>"offline/batch setting"</p><p> </p><p>Change to <code>split_idx</code> (see below)</p><p> </p><p>"and the POTENTIAL anomalies should be coming from..."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

seanlaw, line 1: <p>tqdm isn't really doing anything here and the reader might think that they need to install it. Please remove</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

seanlaw, line 1: <p>"To find the top discord"</p><p> </p><p>"one can compute the left matrix profile and..." (i.e., omit "array")</p><p> </p><p>"approximate" (do not truncate)</p><p> </p><p>"Let's say we have a time series T and we begin checking each subsequence starting from left to right and updating our best-so-far discord as a rarer, more unique subsequence is encountered. At the i-th subsequence, the distance for the best-so-far discord is simply the distance between the rarest/most unique subsequence discovered somewhere amongst <code>T[:i]</code> and its nearest neighbor."</p><p> </p><p>I think we should just call it "discord distance" or "just distance" rather than "discord score". At the end of the day, it's just a Euclidean distance, right?</p><p> </p><p>"Naively, one can compare all of the subsequences, <code>T[:i]</code>,that exist to the left of i-th subsequence, <code>S_i = T[i : i + m],</code> and then find its one-nearest-neighbor. However, this is very time consuming and DAMP employs an early abandoning approach that helps to dramatically reduce this computational time." </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #177.    def DAMP(T, m, split_index):  # Add param <code>T_subseq_isconstant</code>  
</pre><p>Change to <code>split_idx</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor: <p>Thanks for your several suggestions and feedback! I will revise the notebook accordingly.</p>

NimaSarajpoor:
@seanlaw 
I have added a few notes to the initial post of this PR at the top of this page. These note are based on the MATLAB code and the slides that are avaialble on the supporting webpage. While there are more notes to add regarding DAMP, I tried to avoid putting them all here.

docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,778 @@
...
```
seanlaw, line 1: <p>"left (right)" is very confusing to read.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

seanlaw, line 1: <p>Why is this considered "offline/batch"? </p><p> </p><p>Separately, it feels too advanced to start off with splitting the data. You haven't really motivated why this splitting is necessary. Why not just take a time series, say that you are applying DAMP to all of it (without splitting) and then find the discord(s)?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

seanlaw, line 1: <p>Instead of saying "start index of discord", why not say something like "the discord is the subsequence located at index i". I find "start index" to be hard to grasp</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #18.    def _foreward_processing(
</pre><p>Let's shorten this to <code>_foreward_process</code> and <code>_backward_process</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #79.                T[start : start + m],
</pre><p>Shouldn't this be <code>T[query_idx, query_idx + m]</code> ?</p><p> </p><p>Otherwise, the foreward code looks very neat and tidy. Though, I would put the backward function first and then the foreward function (right now foreward is first)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #93.        return pruned
</pre><p>There's no point in returning anything since <code>pruned</code> is a reusable array that is updated inplace</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #96.    def _backward_processing(
</pre><p>The flow of this code feels weird and non-obvious to me what is happening. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #148.        excl_zone_st = query_idx - excl_zone
</pre><p>It's not clear what <code>st</code> stands for (I think it's <code>start</code> but it's not obvious)  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #155.            stop_idx = start_idx + m
</pre><p>Would it make more sense to initialize both <code>start</code> and <code>stop</code>   OUTSIDE the for loop and then update both at the very end of the for loop? Would that simplify things?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #160.                T[start_idx : stop_idx],
</pre><p>Inconsistent naming. In <code>_foreward_process</code> you use <code>start</code> and <code>stop</code> but here you use <code>start_idx</code> and <code>stop_idx</code>. Just use <code>start</code> and <code>stop</code> to make things less confusing        </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #174.            if stop_idx &gt;= excl_zone_st:
</pre><p>I actually think it's better to write <code>if stop &gt;= query_idx - excl_zone</code> because it's very clear that you want to stop before you get to the <code>query_idx</code> and, if the exclusion zone is greater than zero then make sure <code>stop</code> does not go past that barrier either.   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #175.                idx = max(0, excl_zone_st - start_idx)
</pre><p>This is really hard to follow</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor: <p>Agreed. I will provide separate senteneces.</p>

NimaSarajpoor: <p>[I will read the paper to provide the answers to your questions. For now, I am going to provide some explanations that come to my mind as I do not want to forget them later. I will keep this series of comments open]</p><p> </p><p>I think the only reason for the term "offline" here is that we use the look-ahead (<code>_forward_process</code> ) to prune some of the upcoming subsequences (so we know that those subsequences cannot be a discord and hence we do not need to do <code>_backward_process</code>   for them). As mentioned in the MATLAB code (and I think in the paper too), when look-ahead is set to 0, i.e. no pruning, then the algorithm becomes online. </p><p> </p><p> </p><blockquote>You haven't really motivated why this splitting is necessary. Why not just take a time series, say that you are applying DAMP to all of it (without splitting) and then find the discord(s)?</blockquote><p> </p><p>[Right! I need to go and check the paper to see what justification they have provided regarding this matter.]</p><p> </p><p>I think one of the reason is that we are looking LEFT discord! So, some of the first subsequences have no non-trivial LEFT neighbour. The very first subsequence that has one non-trivial neighbour cannot be good candidate as we haven't got enough information to whether choose that subsequence as discord or not. </p><p> </p><p>[I will elaborate more in the upcoming days]</p>

NimaSarajpoor: <p><strong>Regarding "why </strong><code><strong>offline</strong></code><strong> ":</strong></p><p>As a follow up to my previous comment, below I have provided three bullet points mentioned in the DAMP paper regarding <code>lookahead</code> </p><p> </p><ul><li>If lookahead is zero, DAMP is a fast true online algorithm.</li><li>If lookahead is allowed to be arbitrarily large, DAMP is an ultrafast batch algorithm</li><li>Even if lookahead is a small (but non-zero) number, DAMP is effectively online algorithm,</li></ul><p> </p><p>The paper also says:</p><p> </p><blockquote>The algorithm as explained in Table 1 is a batch algorithm. To make it an online algorithm, we simply must reduce the size of the lookahead (Table 3, line 1) to the largest delay we are willing to accept (including possibly zero delay)</blockquote><p> </p><p>The current implementation in the notebook follows the algo provided in the Tables 1,2,3 of the paper.  The current implementation uses next-power-of-two of <code>m</code>  as the value of <code>lookahead</code> and thus, it is not small. Hence, it is offline algorithm. </p><p> </p><p> </p><p><strong>Regarding "why splitting is neccary":</strong></p><p>Personally I think the term train/test split, as used in the paper, may not make sense. The reason behind my opposition is that the train set (i.e. <code>T[:split_idx]</code> ) is not fixed  and it will change as we move towars to the end of time series. Note that for the query <code>T[query_idx : query_idx+m]</code> , we compare it with all of its left neighbors, i.e subsequences in <code>T[:query_idx]</code> , and not just the ones in <code>T[:split_idx]</code>   </p><p> </p><p>The paper also provides discussion on the <code>split_idx </code> :</p><blockquote>Recall that Classic DAMP has a parameter called spIndex, which sets the location of the split point between the training and test data in the initial state. When Classic DAMP processes a time series, it assumes that the data before spIndex, T[1:spIndex-1] are normal, which may lead to three issues. First, this causes the algorithm to ignore the potential anomalous behavior present in T[1:spIndex-1], resulting in certain false-negative results. Second, this approach may have the algorithm wasting time searching redundant data. It is possible that the patterns in T[1:spIndex-1] are highly redundant, such as 1,000 heartbeats that are essentially identical. If the heartbeats all have the same pattern, it would suffice for the algorithm to take just one of them to learn3 ; there is no need to consider the same pattern 1,000 times, which will waste a lot of time. Further, it may be difficult for T[1:spIndex-1] to contain every normal pattern, which can cause the algorithm to incorrectly identify normal behavior that does not appear in&nbsp; T[1:spIndex-1] as an anomaly. For example, if T[1:spIndex-1] only contains data on the solar zenith angle during the day, the algorithm may incorrectly identify normal solar zenith angles at night as anomalies. These potential problems can undermine the accuracy and efficiency of the algorithm</blockquote><p> </p><blockquote>Golden DAMP is our proposed solution to the above three problems</blockquote><p> </p><p>Two notes:</p><p>(1) As discussed above, <code>split_idx</code> is a timestamp before which ALL regular behaviors of the system appear with No anomaly. </p><p> </p><p>(2) As discussed above, the paper will handle the <code>split_idx</code> with the algorithm called Golden DAMP. In this algorithm, we have a vector (a time series) called Golden batch that contains ALL regular behaviors (without any redundant patters) . We can start from the very first index of the main input <code>T</code> . Although it has no left neighbors, That is not an issue sinec we do NOT look at the left neighbors anymore! We just compare a query with a Golden Batch. </p>

NimaSarajpoor: <p>Please let know if further discussion is needed.</p>

NimaSarajpoor: <p>How about this:</p><p> </p><p>To find the top discord, one can compute the left matrix profile, PL, and locate the index <code>i</code>  where PL  has the maximum finite value. The discord is the subsequence located at index i, i.e. the subsequence <code>T[i:i+m</code> .</p><p> </p><p> </p><p>?</p>

NimaSarajpoor: <p>Will revise the name of these two functions.</p>

NimaSarajpoor: <p>You are right! I will fix Line #79. Thanks for catching it!</p><p> </p><blockquote>Though, I would put the backward function first and then the foreward function (right now foreward is first)</blockquote><p>One thing that I would like to point out here is that <code>_backward_process</code> is a mandatory step and <code>_forward_process </code> is an optional step that can speed up the process by pruning. So, I was wondering if we should keep the optional function at top and the two other functions, i.e. the main function and the <code>_backward_process </code> , next to each other. </p><p> </p><p>What do you think? </p><p> </p>

NimaSarajpoor: <p>Right! I will fix it! I will just do bare <code>return</code>  <span class="ql-cursor">﻿</span></p>

NimaSarajpoor: <p>I will change the code back to <code>while</code> (original format) and see if I can find other ways to simplity the code. We can then start from there and enhance it if needed. </p>

NimaSarajpoor: <p>Yes! I should have used the full word "start". I will fix it.</p>

NimaSarajpoor: <p>Thanks for input! I will fix it.</p>

NimaSarajpoor: <blockquote>&nbsp;actually think it's better to write</blockquote><p>Just to make sure that I am not missing anything here:</p><p>Are you suggesting that I should avoid using the intermediate variable <code>excl_zone_st</code>and just write <code>query_idx - excl_zone</code>  in the inequality ?</p><p> </p><p> </p><p> </p><p> </p><blockquote>&nbsp;if the exclusion zone is greater than zero then make sure&nbsp;<code>stop</code>&nbsp;does not go past that barrier either.</blockquote><p> </p><p>I think it all depends on how we want to definte "stop". Is "stop" the stop index of a time series OR is it the stop index before which the last subsequence of the time series starts? </p><p> </p><p>e.g. <code>T = [0, 1, 2, 3, 4, 5]; m=3</code>  </p><p> </p><p>The "stop" is <code>6</code> or <code>4</code>? </p><p> </p><p>A non-trivial neighbor is a subsequence whose start index is outside of the exclusion zone. So, If I define "stop" as the stop index of the time series (e.g. <code>6</code>  in example above) then I am okay if <code>stop</code> goes past the barrier as long as the start index of that subsequence is before the barries. </p><p> </p><p>Please allow me to revise the code. Your comment, this long explanation (a red flag) as well as your next comment regarding line 175 all indicate that my code is not clear enough. I will try to clean it.</p>

NimaSarajpoor: <p>Agreed! I will revise the function.</p>

seanlaw: <p>I think we need to build up the intuition step-by-step. To me <code>split_idx</code> seems like an advanced topic and not a parameter that is automatically intuitive to the average user. How can we build up to using <code>split_idx</code>? </p>

seanlaw: <p>I am good with that</p>

seanlaw: <p>I get your point and maybe I'm overthinking things. As I'm reading from top to bottom, seeing <code>_foreward_process</code> first makes me think that it's more important than <code>_backward_processing</code> and I don't think that's the case, right?  </p>

seanlaw: <blockquote>Are you suggesting that I should avoid using the intermediate variable&nbsp;<code>excl_zone_st</code>and just write&nbsp;<code>query_idx - excl_zone</code>&nbsp;in the inequality ?</blockquote><p>Yes, I find <code>excl_zone_st</code> to be ugly :(</p><p> </p><blockquote>Please allow me to revise the code. Your comment, this long explanation (a red flag) as well as your next comment regarding line 175 all indicate that my code is not clear enough. I will try to clean it.</blockquote><p>Thank you! I'm really glad that you got my point. It will be very hard to maintain and reason with in the future if we don't get this right</p><p> </p>

NimaSarajpoor: Correct. Since we are dealing with left neighbors (in classic `DAMP`), `_backward_process` is more important. You can even remove the `_forward_process` as the function is just for speeding up the computation. 

I will put `_backward_process` at the top. Later if you change your mind, I will change it back to current format.

NimaSarajpoor: > It will be very hard to maintain and reason with in the future if we don't get this right. I don't doubt that the code "works" but it needs to be more obvious and also consistent (regarding what `start` and `stop` refer to)

I totally understand. I checked the MERLIN notebook a few days ago and it was not easy for me to follow the code despite the fact that I wrote the code myself. So, I do not want to upset my future self this time :)

NimaSarajpoor:
@seanlaw 
I need to read the notebook one more time. I will then answer some of the remaining comments according to new notebook and see what you feel.

seanlaw:
@NimaSarajpoor Can you please merge the latest commits to this PR? I believe that it should fix the failed tests

docs/Tutorial_DAMP.ipynb
NimaSarajpoor: I changed the design a little bit and initialize them outside of the for-loop. I think it should be more readable now.

NimaSarajpoor: I kept the for-loop but revised the code to make it more readable. Let's see how you feel about it.

NimaSarajpoor: According to the paper, `split_idx` is problem dependent and user needs to have some insight about the system and its usual / regular patterns. So, they can say "upto the time stamp `split_idx`, the data shows [all kind of] regular behaviors and there is no anomaly". 

I added some explanation regarding this matter to the notebook.

docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,885 @@
...
```
NimaSarajpoor, line 1: <p>I added this example to explain the motivation behind  the split index. I also explained that this value is problem dependent and should be set by the user who has insight about the system and its regular / usual behaviors.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #37.        IL[:split_idx] = -1  # <code>split_idx</code> or <code>split_idx-m</code>?
</pre><p> </p><p>I added this comment for our future reference. In the paper, a subsequence is compared with all subsequences in <code>T[:split_idx]</code> , hence, the latest subsequence in the aforementioned chunk is a subsequence that starts at location <code>split_idx -  m</code>. Although <code>excl_zone</code> is set to <code>m</code> in the paper,  the value <code>split_idx - m</code> should still remain the same even with different excl_zone  [IMO].</p><p> </p><p>Should we think about how data arrives? [e.g. In batch of size of <code>m</code> , etc]   </p><p>@NimaSarajpoor: Read the paper regarding this matter. [note to self]</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    T = np.random.rand(100000)
</pre><p>I increased the length of <code>T</code> to  better show the difference between naive version and DAMP algo. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #8.    naive_DAMP(T, m, split_index)  # dummy run
</pre><p>I added a dummy run to make sure the running time does not contain the compilation time of njit funcs.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #60.        for _ in range(n_chunks_upperbound):    
</pre><p>Originally, there was a while loop. I replace it with a for-loop and add a couple of <code>break</code> statements to terminate the loop. Note that the number of iterations shoud be, at least, equal to the number of chunks in <code>T[:query_idx]</code>  So, one easy answer  is <code>query_idx</code> since it is definitely larger than the number of chunks. But, just to add some sense to the code and make it a bit more readable, I computed a tighter upperbound for total number of chunks. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #90.        return nn_distance, bsf
</pre><p>We may combine the two <code>if / break</code> blocks  and we can just return <code>nn_distance</code> which is either exact [if it is greater than bfs] or approx [if otherwise]. We may even return the <code>is_exact</code> flag to make it more explicit. Then, we can update <code>bfs</code> outside of this function. </p><p> </p><p>[I am trying to see if I can find a way to return one object rather than two objects] </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #212.                _foreward_process(T, m, M_T, Σ_T, T_subseq_isconstant, excl_zone, i, bsf, pruned)
</pre><p>Instead of <code>_forward_process</code>, we may say: <code>_update_pruned</code>   </p><p> </p><p>And, similarly, instead of <code>_backward_process</code> , we may say: <code>_calculate_distance</code> . As a follow up to my previous comment, we can just return <code>nn_distance</code> . In line #213 (i.e. next line after the line above), we can have:</p><p> </p><pre class="ql-syntax" spellcheck="false">if nn_distance &gt; bfs:  
    bfs = nn_distance  # update discord score

# alternatively: bfs = max(bfs, nn_distance)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #11.    print('running time [sec]: ', t_end - t_start)
</pre><p>A few notes regarding running time:</p><p> </p><p>1) IMO, the running time depends on the location of the discord. If it is located to the end of the time series <code>T</code>,then that means the best-so-far value we are looking for will be updated to be the discord distance at the end. Therefore, there is probably a less chance of early-abondoning  in the process of analyzing subsequences from <code>split_idx</code> to that discord index.</p><p> </p><p>2) I think ranomly-generated time series may not be a good choice for comparing the running time between a naive versiona and DAMP version. This is because DAMP looks for discord by considering left neighbors only. As we get closer to the end of time series, a subsequence will have more and more left neighbors. This can increase its chance of having a neighbor to which its distance is less than <code>bsf</code> [and hence not a discord]. That is why we usually see the discord index in the randomly-generated time series is usually close to the <code>split_idx </code>. So, in naive version, it naively computes exact left matrix profile, and in the DAMP, it looks for the discord with early-abondoning. And since th discord index in this case is very close to the beginning, i.e. close to <code>split_idx</code>, we may get higher chance for early-abondoning and hence the running time of DAMP will be very low. [So, a better way might be applying DAMP to a real-world data and see how it performs.]</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

seanlaw: <p>Returning 2 items is fine I think?</p>

seanlaw:
@NimaSarajpoor How does it feel to return to this code after 2 months? I think this is a good example where you get to critique your own code for readability.

NimaSarajpoor:
Correct! I am going to put a time gap between my visits to the PR :) Next time, I may provide some other comments with fresh eye.

docs/Tutorial_DAMP.ipynb
seanlaw: > I think ranomly-generated time series may not be a good choice for comparing the running time between a naive version and DAMP version

In general, it is always best to report the average/majority case running time rather than the "absolute best case scenario" or the "absolute worst case scenario".

NimaSarajpoor: Exactly! 

I think randomly-generated time series for discord discovery may give **[almost] absolute best case scenario for DAMP**. Since the top discord is probably located close to the beginning of the time series, the best-so-far discord distance will be updated to the value of true discord distance in the beginning of the process and this results in early abandoning in all of the backward processes of the subsequences that appear later in the time series. I will try to do some experiments.

seanlaw: For testing, could we do:

1. Create a sine/cosine wave
2. Add a tiny bit of random noise
3. Insert a discord at any desired location that is composed of random values and see how it changes the computation for different length time series

NimaSarajpoor: smart! Will do it!

NimaSarajpoor: @seanlaw 

image

```
n = 100001
m = 200 # length of subsequence

seed = 0
np.random.seed(seed)

noise = 0.1 * np.random.rand(n)
T = np.linspace(-500 * np.pi, 500 * np.pi, n)
T[:] = np.sin(T) * (1.0 + noise)

split_idx = 10000
# dummy run
naive_DAMP(T, m, split_idx)
DAMP(T, m, split_idx)

l = n - m + 1
step = 5000
discord_loc = np.arange(split_idx, l, step)

discord = np.random.rand(m)
naive_runnig_time = []
performant_runnig_time = []

for idx in discord_loc:
    T_with_discord = T.copy()
    T_with_discord[idx : idx + m] = discord
    
    ts = time.time()
    naive_DAMP(T_with_discord, m, split_idx)
    te = time.time()
    naive_runnig_time.append(te - ts)
    
    ts = time.time()
    DAMP(T_with_discord, m, split_idx)
    te = time.time()
    performant_runnig_time.append(te - ts)
```

NimaSarajpoor: The jump in the beginning of the plot of the naive version can be because of njit compilation time. I noticed similar behavior before. So, it is not just that the first call has higher running time because of compilation time. It seems that the higher running time might exist for the first few calls.

The running time of the DAMP increases as the location of the discord gets closer to the end of time series

NimaSarajpoor: Going to "resolve" this conversation and add the figure to the notebook.

NimaSarajpoor: Going to leave it as is. [Also: Now that I think more about it, it is better to remain faithful to the algorithm as much as possible in the first stage of our work. And then, if needed, we can revise it to improve it]

NimaSarajpoor: <p>To avoid this in <code>naive_DAMP</code>  and to improve readability, the manipulation of IL is avoided. Instead, we just start computing PL from <code>split_idx</code>  <span class="ql-cursor">﻿</span></p>

NimaSarajpoor: <p>It is now changed to make the code closer to the proposed algorithm</p>

docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,900 @@
...
```
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #79.            chunk_stop = chunk_start + m
</pre><p>This is different than the algorithm in the paper, but "similar" to what provided in the MATLAB code.</p><p> </p><p>In the line 10 of the Backward processing algorithm proposed in the paper, the start index of chunk gets updated to consider more subsequences; however, the chunk's stop index is fixed at i . In the MATLAB code, they have the optimized version where the algorithm avoids computing the distance between the query Q and the subsequences in the previously-considered chunk. To do this, the MATLAB code updates the stop index of chunk too. There are still some differences between the code here and the MATLAB code. </p><p> </p><pre class="ql-syntax" spellcheck="false">X_start = i-X+1+(expansion_num * SubsequenceLength);
X_end = i-(X/2)+(expansion_num * SubsequenceLength);
approximate_distance = min( real(MASS_V2(T(X_start:X_end), query))); 
       

# To update X and expansion_num at the end of each iteration in while-loop 
X = 2*X;
expansion_num = expansion_num+1;
</pre><p> </p><p>[Note to self]</p><p>Further analysis is needed to understand the reason behind using <code>expansion_num</code>  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #133.        chunk_start = query_idx + m
</pre><p>According to paper, this is to avoid trivial match. As discussed before, the paper considers<code>excl_zone=m</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #186.        excl_zone = int(np.ceil(m / stumpy.config.STUMPY_EXCL_ZONE_DENOM))
</pre><p> </p><p>This line can be deleted as <code>excl_zone</code> is not being used anymore.  The <code>excl_zone</code> is now set to <code>m</code>  in the performant algorithm so as to remain faithful to the paper. </p><p> </p><p> </p><p><strong>Additional discussion: </strong></p><p>If we consider <code>T[:split_idx]</code> as the training part, and we want to make a decision about the new subsequence <code>T[split_idx : split_idx + m]</code> , what should we do? Should we compare it to the subsequences in <code>T[:split_idx]</code>   Or should we compare it to the subsequences in <code>T[:split_idx + m]</code> while ignoring the trivial zone? I think the former is the paper's approach. If we want to go with the latter, then we should bring back the <code>excl_zone</code> . </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #195.                PL[i] = PL[i-1]
</pre><p> </p><p>If <code>pruned[i] == True</code> then the approximate value for <code>PL[i]</code> will be set to <code>PL[i-1]</code> . This results in having more than one global maxima. Although this is the approach of the paper, the MATLAB code shows a different approach:</p><p> </p><pre class="ql-syntax" spellcheck="false">% We subtract a very small number here to avoid the pruned
% subsequence having the same discord score as the real discord
Left_MP(i) = Left_MP(i-1)-0.00001;
</pre><p> </p><p>However, I think this is not an ideal approach as the value <code>0.00001</code> is hardcoded. Alternatively, we may do:</p><p> </p><pre class="ql-syntax" spellcheck="false"># see the end of the function _forward_process
mask = np.argwhere(D &lt; bsf) + chunk_start
pruned[mask] = True

# NEW
PL[mask] = D[D &lt; bsf] 
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

seanlaw: <p>In the future, we should make sure to use the same exclusion zone as motifs (i.e., so that the discord discovered via DAMP will be the same as the classical <code>stumpy.stump</code> approach) </p>

seanlaw: <p>As mentioned above, the output from DAMP needs to be the same as the discord discovered via <code>stumpy.stump</code> .  Otherwise, we will waste a lot of time explaining it to users</p>

seanlaw: <blockquote>However, I think this is not an ideal approach as the value&nbsp;<code>0.00001</code>&nbsp;is hardcoded.</blockquote><p> </p><p>I agree. This hack feels like it's a problem waiting to happen</p>

seanlaw: <p>Could we please split up each function and place it in its own cell? This way, the discussion/comments can be more focused on each section of code. Is it worth submitting a new PR?</p>

seanlaw, line 1: <p>I have a stupid question. Does DAMP only produce the top-1 discord? Is there an easy way to get the top-k discords?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/872/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

NimaSarajpoor: Right. later, I will bring back the `excl_zone` to make it behave like the discord discovery via `stumpy.stump`

NimaSarajpoor: It turns out one can use the following statement to get more than one discord.

see [DAMP_2.0 How To Use.pptx](https://docs.google.com/presentation/d/1_-LGilUJpYRbRZpitw05EgkiOZX52kRd/edit?usp=sharing&ouid=115719825860849832698&rtpof=true&sd=true)
where the authors say:
> The highest peak before the top-1 discord, has a correct value

However, IIUC, this approach does not necessarily give the top-k discord (from left matrix profile). E.g. let's say I have the following approximate left matrix profile vector:

[1, 2, 3, 4, 3.9, 3.8, 3.9, 3.8, 5, 3.7, 4.1]

In this case, the top-1 discord has the discord distance `5 ` (at index `idx`).  Then, we can say that the next top discord in `T[:idx]` (NOT in `T`) is `4`. However, what if I want to consider the whole `T`? 

**Solution:**
I have not thought this through but we might be able to use a sorted vector of length `k` to keep track of top-k discord and do prunning according to the k-th discord distance. I need to:
(1) Check paper 
(2) and/or Check code 
(3) and/or Think about this logic
to see if this can work!

NimaSarajpoor: >  Is it worth submitting a new PR?

Now that you mentioned it, I think I may submit a new PR to clean things up.

NimaSarajpoor: I get the point. Later, I will bring back the `excl_zone` to make it behave like the discord discovery approach when using `stumpy.stump`

seanlaw: > However, what if I want to consider the whole T?

We probably want to be able to do both (i.e., get top-k discords from left MP and get top-k from global MP). What do you think?

NimaSarajpoor: I am not sure if the current, proposed DAMP can find top-k from global MP. However, we may think later and see how to add such support. (I have some rough idea but haven't implemented it yet).

**Note that my previous comment was regarding left MP.** I THINK the proposed approach for top-k in DAMP does not give the top-k from left matrix profile. Instead, it finds the second-top discord that is located before the top-discords, and find the third-top discord that is located before the second-top discord, and so on.  I THINK THIS is different than finding the top-k from left matrix profile.

NimaSarajpoor: Related: 
https://sites.google.com/view/discord-aware-matrix-profile/documentation#h.3jzxuzat4jfj

NimaSarajpoor: Thanks for the suggestion! Will consider it in the new PR#920.

Issue #874, opened 2023-06-16
Error in minimum version testing (caused by python 3.7 in macOS)


NimaSarajpoor:
**Python 3.7 (to be more precise:  3.7.17) issue in macOS**

See: https://github.com/TDAmeritrade/stumpy/actions/runs/5285945141/jobs/9564904199

Related: 
* https://github.com/actions/setup-python/issues/682
* https://github.com/web-platform-tests/wpt/issues/40525

seanlaw:
@NimaSarajpoor Thanks for reporting it. It looks like an issue unrelated to STUMPY that will require others to fix the base Python image. Let's wait a couple of days to see if it resolves itself

seanlaw:
I believe this is resolved now

PR #875, opened 2023-06-25
Add function `process_isconstant`

NimaSarajpoor:
The new function, `process_isconstant`, is added to the `stumpy.core` module to handle:
* 2D time series when `T_subseq_isconstant` is provided as a list
* To fix the conflict between `T_subseq_isconstant` and `T_subseq_isfinite`

Two notes:
(1) The function `core.rolling_isconstant` can handle 2D time series, and the parameter `T_subseq_isconstant` can still be passed as `numpy.ndarray`, `function`, or `None`. However, it does not suppoert the `list` type.

(2) I think `core.preprocess` supports 2D array. So, we may change its docstring to reflect that the param `T_subseq_isconstant` can be a list. (since this function now calls the function `process_isconstant` which can support `list` type for `T_subseq_isconstant`)

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -4248,8 +4248,8 @@ def process_isconstant(T, m, T_subseq_isconstant, T_subseq_isfinite=None):
     if isinstance(T_subseq_isconstant, list):
         if T.ndim != 2:  # pragma: no cover
             msg = (
-                "When `T_subseq_isconstant` is provided as a list, the `T` "
-                + f"must be a 2D array. Got {T.ndim} dimension instead."
+                "When `T_subseq_isconstant` is provided as a list, `T` "
+                + f"must be a 2D array. Found {T.ndim} dimension instead."
             )
             raise ValueError(msg)
 
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -4246,8 +4246,12 @@ def process_isconstant(T, m, T_subseq_isconstant, T_subseq_isfinite=None):
         Rolling window isconstant
     """
     if isinstance(T_subseq_isconstant, list):
-        ndim = T.ndim
-        T = np.atleast_2d(T)
+        if T.ndim != 2:
+            msg = (
+                "When `T_subseq_isconstant` is provided as a list, the `T` "
+                + f"must be a 2D array. Got {T.ndim} dimension instead."
+            )
+            raise ValueError(msg)
 
         if len(T_subseq_isconstant) != T.shape[0]:
             msg = (
@@ -4262,9 +4266,6 @@ def process_isconstant(T, m, T_subseq_isconstant, T_subseq_isfinite=None):
                 for i in range(T.shape[0])
             ]
         )
-
-        if ndim == 1:
-            T_subseq_isconstant = T_subseq_isconstant.flatten()
 
     T_subseq_isconstant = rolling_isconstant(T, m, T_subseq_isconstant)
     T_subseq_isconstant[...] = fix_isconstant_isfinite_conflicts(
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -4210,3 +4210,63 @@ def _mpdist(
     MPdist = _select_P_ABBA_value(P_ABBA, k, custom_func)
 
     return MPdist
+
+
+def process_isconstant(T, m, T_subseq_isconstant):
+    """
+    Compute the rolling isconstant for 1-D and 2-D arrays.
+    This is accomplished by comparing the min and max within each window and
+    assigning `True` when the min and max are equal and `False` otherwise. If
+    a subsequence contains at least one NaN, then the subsequence is not constant.
+
+    If `T_subseq_isconstant` is provided as boolean array, its element will be set
+    to False if their corresponding value in `T_subseq_isfinite` is False.
+
+    Parameters
+    ----------
+    T : numpy.ndarray
+        The input 1D or 2D array
+
+    m : numpy.ndarray
+        The rolling window size
+
+    T_subseq_isconstant : np.ndarray, function, or list, default None
+        A parameter that is used to show whether a subsequence of a time series in `T`
+        is constant (True) or not. T_subseq_isconstant can be a 1D or 2D boolean
+        numpy.ndarry (depending on the dimension of `T`) or a function that can be
+        applied to each time series in `T`. Alternatively, for  maximum flexibility, a
+        list (with length equal to the total number of time series) may also be used.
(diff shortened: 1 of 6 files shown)


tests/test_core.py
```diff
@@ -1669,3 +1669,80 @@ def test_compute_P_ABBA_with_isconstant(T_A, T_B):
...
+
+    T_subseq_isconstant_comp = core.process_isconstant(T, m, T_subseq_isconstant)
+
+    npt.assert_array_equal(T_subseq_isconstant_ref, T_subseq_isconstant_comp)
+
+    # case 2: with nan
+    T = np.random.rand(d, n)
+    row = np.random.randint(d)
+    col = np.random.randint(n)
+    T[row, col] = np.nan
```
NimaSarajpoor, line 1733: Since the last item in `T_subseq_isconstant` (see below) is a boolean array, it makes sense to do `T[-1, col] = np.nan` AND also `T_subseq_isconstant[-1][col]=True`. This way the impact of `fix_isconstant_isfinite_conflicts` will be tested.

stumpy/stumpi.py
```diff
@@ -324,7 +324,7 @@ def _update_egress(self, t):
             σ_Q = np.nan
             Q_subseq_isconstant = False
         else:
-            Q_subseq_isconstant = core.rolling_isconstant(
+            Q_subseq_isconstant = core.process_isconstant(
```
NimaSarajpoor, line 327: There is no need to replace `rolling_isconstant` with `process_isconstant` here since the param `T_subseq_isconstant` that is being passed here always has the "function" type. Having said that, I think it is better to use a higher-level API so to be consistent with other parts of the code.

stumpy/core.py
```diff
@@ -4210,3 +4206,70 @@ def _mpdist(
...
+        A parameter that is used to show whether a subsequence of a time series in `T`
+        is constant (True) or not. T_subseq_isconstant can be a 1D or 2D boolean
+        numpy.ndarry (depending on the dimension of `T`) or a function that can be
+        applied to each time series in `T`. Alternatively, for  maximum flexibility, a
+        list (with length equal to the total number of time series) may also be used.
+        In this case, T_subseq_isconstant[i] corresponds to the i-th time series T[i]
+        and each element in the list can either be 1D boolean np.ndarray, a function,
+        or None.
+
+    T_subseq_isfinite : numpy.ndarray, default None
```
NimaSarajpoor, line 4239: I added this param since `fix_isconstant_isfinite_conflicts` , which is used in this function, has the param `T_subseq_isfinite`. However, I now think it should bettter be removed as this may lead to a strange result if a user provides their own `T_subseq_isfinite` that is different than the outcome of `core.rolling_isfinite(T, m)`.

stumpy/core.py
```diff
@@ -4210,3 +4206,70 @@ def _mpdist(
...
+            raise ValueError(msg)
+
+        T_subseq_isconstant = np.array(
+            [
+                rolling_isconstant(T[i], m, T_subseq_isconstant[i])
+                for i in range(T.shape[0])
+            ]
+        )
+
+    T_subseq_isconstant = rolling_isconstant(T, m, T_subseq_isconstant)
```
NimaSarajpoor, line 4270: I could have added `else` here, i.e.
```
else:
     T_subseq_isconstant = rolling_isconstant(T, m, T_subseq_isconstant)
```

However, I think it is cleaner to avoid using `else`.

NimaSarajpoor (commented):
@seanlaw 
The [current error](https://github.com/TDAmeritrade/stumpy/actions/runs/5368164129/jobs/9738837681) is mentioned in #874. As you said, we can wait and see how it goes.

For now, I have provided a few comments. It would be geat if you could go through them and let me know what you think. After merging this PR, I will resume my work in PR #871

stumpy/core.py
```diff
@@ -4210,3 +4206,70 @@ def _mpdist(
...
+
+    Returns
+    -------
+    T_subseq_isconstant : numpy.ndarray
+        Rolling window isconstant
+    """
+    if isinstance(T_subseq_isconstant, list):
+        if T.ndim != 2:  # pragma: no cover
+            msg = (
+                "When `T_subseq_isconstant` is provided as a list, the `T` "
```
seanlaw, line 4251: Remove "the"

stumpy/core.py
seanlaw: I think `else` is fine in this case. Otherwise, if a `list` is provided, you end up calling `rolling_isconstant` again which is unnecessary

stumpy/stumpi.py
seanlaw: sounds good.

seanlaw (changes requested):
@NimaSarajpoor Everything looks good. On your end, can you please confirm that the tests all pass AND we have 100% coverage?

NimaSarajpoor:
> @NimaSarajpoor Everything looks good. On your end, can you please confirm that the tests all pass AND we have 100% coverage?

I will apply your comments and run the unit test (+ coverage) on colab.

seanlaw:
@NimaSarajpoor Can you please merge the latest commits to this PR? I believe that it should fix the failed tests

stumpy/core.py
NimaSarajpoor: okay. So, let's keep the parameter `T_subseq_isfinite` for now. If we assume this is going to be used by advanced users, they probably won't mess it up.

NimaSarajpoor:
@seanlaw 
Note sute if you still need it, but just in case: I ran the tests on colab (with GPU). All tests pass and coverage is 100%.

stumpy/core.py
```diff
@@ -4210,3 +4206,71 @@ def _mpdist(
     MPdist = _select_P_ABBA_value(P_ABBA, k, custom_func)
 
     return MPdist
+
+
+def process_isconstant(T, m, T_subseq_isconstant, T_subseq_isfinite=None):
+    """
+    Compute the rolling isconstant for 1-D and 2-D arrays.
```
seanlaw, line 4213: I think this should start off by saying:

"This is a convenience wrapper around the `rolling_isconstant` and `fix_isconstant_isfinite_conflicts`

seanlaw (commented):
@NimaSarajpoor Just one minor suggestion

Do you think this is ready?

stumpy/core.py
NimaSarajpoor: Good point! I added this sentence at the top of the docstring.

NimaSarajpoor:
> Do you think this is ready?

I think it is ready! Please let me know if there is anything else that I should take care of.

seanlaw:

PR #878, opened 2023-07-11
Add param isconstant to stimp

NimaSarajpoor:


seanlaw:
Looks good so far!

NimaSarajpoor:
@seanlaw 
Tests are passing now. We may investigate the error in the previous run of GitHub actions in separate PR. I can open an issue for it so that we do not forget. What do you think? 

Also:
Is there anything else needs to be done for this PR?

seanlaw:
> The tests are passing now. We may investigate the error in the previous run of GitHub actions in a separate PR. I can open an issue for it so that we do not forget. What do you think?

Yes, please do.

> Also:
Is there anything else that needs to be done for this PR?

Thanks again for completing this PR, @NimaSarajpoor !

PR #886, opened 2023-07-22
Fix #842 Fix mass_PI

### LINKED ISSUE #842: wrong post-processing on distance matrix in naive mass_PI
In `naive.py::mass_PI`, we can see the following code:

https://github.com/TDAmeritrade/stumpy/blob/8cc59b2bc5ec7222002ae68234303b09d007bc4a/tests/naive.py#L147-L160

However, the last line `D[np.isnan(D)] = np.inf ` cannot be always effective as the value might be already set to a finite value because of the elif in line 151. To sovle this, we have two solutions:

(1) `D[np.isnan(D)] = np.inf ` right after computing D in line 147, and then add a if-check before applying isconstant-related post-processing as follows:

```
for i in range(len(T) - m + 1):
     if np.isfinite(D[i]):
         if Q_subseq_isconstant and T_subseq_isconstant[i]: 
              ...
         elif .... 
```  

Or alternatively: 

(2) we can compute D_isfinite = np.isfinite(D) right after computing `D`. Then, at the end, we can do:
```
D[~D_isfinite] = np.inf
```

NimaSarajpoor:


NimaSarajpoor:
@seanlaw 
Ideally we should have a test function that raises an error, and then it will be fixed with the changes applied to the function. However, there was no test function for `mass_PI`. Also, the function `mass_PI` is used in the functions that are no longer supported. So, I just revised the code to address the issue mentioned in #842. 

If it is okay with you, plz feel free to close this PR once the tests pass.

seanlaw:
Thank you @NimaSarajpoor !

PR #887, opened 2023-07-22
Fix #750 Clean `scrump` and `scraamp`

### LINKED ISSUE #750: if-continue block in `scrump` and `scraamp`
It seems that this

https://github.com/TDAmeritrade/stumpy/blob/8da71ecec45e947cf448eb3dba8a75c98f5dbfe2/stumpy/scrump.py#L209-L211

is never executed, and even if it is executed, there will be no change! 

The reason is that: 
(1) all elements of `P_squared` are initially set to `np.inf` and their corresponding indices in `I` are initially set to -1. 
(2) `P_squared` is sorted ascendingly. Thus, after finding at least one finite distance, the value `P_squared[thread_idx, i, 0]` becomes finite. This is because the smallest distance will be at index 0.

Therefore, after finding the first finite distance between subseq `i` and one of its neighbor, the value at `P_squared[thread_idx, i, 0]` becomes finite. Note that if we never find a finite distance for subseq `i`, then all values in `P_squared[thread_idx, i, :]` remain `np.inf` and their corresponding indices remain `-1`. Hence, if-block results in no change.

We have a similar situation in `scraamp`. see [Lines 194-197](https://github.com/TDAmeritrade/stumpy/blob/main/stumpy/scraamp.py#L194-L197)

---

The only drawback is that `P_squared` is a variable that is passed to the function `_prescrump`. Hence, we may need to improve the docstring to let user know that the default value in `P_squared` is `np.inf`, and its corresponding index in `I` is `-1`.

NimaSarajpoor:


NimaSarajpoor:
@seanlaw 
I think this PR is ready [to be merged].

seanlaw:
Thank you @NimaSarajpoor for this PR!

PR #888, opened 2023-07-23
Fix #868 

### LINKED ISSUE #868: numba JIT-compiled func `core._compute_multi_PI` is not parallel?
The funciton `core._compute_multi_PI` seems to be parallel according to its decorator:

https://github.com/TDAmeritrade/stumpy/blob/76cb9806dec2d32f80ff20a5a3c484c34b7a54bf/stumpy/core.py#L3955-L3960

However, it uses `range` instead of `prange`:
https://github.com/TDAmeritrade/stumpy/blob/76cb9806dec2d32f80ff20a5a3c484c34b7a54bf/stumpy/core.py#L3997-L3998

NimaSarajpoor:


NimaSarajpoor:
Surprisingly, removing the `parallel` from the njit decorator resulted in a significant decrease in the running time.

```
import time

import numpy as np
import stumpy

seed = 1000
np.random.seed(seed)

T = np.random.rand(2, 10000)
m = 50

stumpy.mstump(T, m)

lst = []
for _ in range(10):
    ts = time.time()
    stumpy.mstump(T, m)
    te = time.time()
    lst.append(te - ts)

print(np.mean(lst))
```

* branch `main` --> ~16sec
* this branch --> ~9sec

NimaSarajpoor:
@seanlaw 
I think this PR is ready.

stumpy/core.py
```diff
@@ -3956,7 +3956,6 @@ def _mdl(disc_subseqs, disc_neighbors, S, n_bit=8):
 
 @njit(
     # "(i8, i8, f8[:, :], f8[:], i8, f8[:, :], i8[:, :], f8)",
-    parallel=True,
```
seanlaw, line 3959: So, where's the a parallelization happening? I looked at the code but do not see `prange` anywhere and it should be somewhere. Otherwise, `mstump` would be slow

seanlaw: I guess `compute_multi_PI` cannot be parallel because the for-loop relies on the previous iteration (i.e., each iteration of the for loop is NOT independent of the previous step). Also, the part that needs to be "fast" and parallelized is in `compute_multi_D`

seanlaw: In `compute_multi_PI`:

```
D_prime[:] = 0.0
    for i in range(d):
        D_prime = D_prime + np.power(D[i], 1.0 / p)

        min_index = np.argmin(D_prime)
        pos = idx - range_start
        I[i, pos] = min_index
        P[i, pos] = D_prime[min_index] / (i + 1)
        if np.isinf(P[i, pos]):  # pragma nocover
            I[i, pos] = -1
```

technically, each column of `D_prime` can be updated in parallel like:

```
D_prime[:] = 0.0
    for i in range(d):
        for j in prange(len(D_prime):
            D_prime[j] = D_prime[j] + np.power(D[i, j], 1.0 / p)

        min_index = np.argmin(D_prime)
        pos = idx - range_start
        I[i, pos] = min_index
        P[i, pos] = D_prime[min_index] / (i + 1)
        if np.isinf(P[i, pos]):  # pragma nocover
            I[i, pos] = -1
```

although, I'm not sure if it's worth it or if it's even "faster" than the vectorized version

seanlaw: @NimaSarajpoor Would you mind testing this (`prange` version) and comparing it against the speed of `parallel=False`? Also, do the tests pass for `mstump` and `maamp` for this `prange` version?

NimaSarajpoor: > So, where's the a parallelization happening? I looked at the code but do not see `prange` anywhere and it should be somewhere

see the inner for-loop: https://github.com/TDAmeritrade/stumpy/blob/main/stumpy/mstump.py#L896-L899



> I guess `compute_multi_PI` cannot be parallel because the for-loop relies on the previous iteration (i.e., each iteration of the for loop is NOT independent of the previous step)

Correct!

> technically, each column of D_prime can be updated in parallel like:
...
although, I'm not sure if it's worth it or if it's even "faster" than the vectorized version

Correct! Maybe we  compare the performance [for different length of `T`]

> @NimaSarajpoor Would you mind testing this (prange version) and comparing it against the speed of parallel=False? 

So, just to confirm: 
You meant `prange` version [over columns] in `core._compute_multi_PI`. Correct? And, then I should compare it with the vectorized approach [with `parallel==False`].

seanlaw: > You meant prange version [over columns] in core._compute_multi_PI. Correct? And, then I should compare it with the vectorized operation [with parallel==False, which is basically the existing approach].

Yes, exactly! 

> Maybe we compare the performance [for different length of T]

Maybe test different length T and a range of dimensions (i.e., number of time series) `range(1, 11)`

NimaSarajpoor:
For our future reference:
```
# a simple script to compare the vecotrized operation vs parallel operation
import time

import numpy as np
import stumpy

from numba import prange, njit


@njit(fastmath=True)
def func1(x, y):
    x[:] = x + np.power(y, 1.0/3)


@njit(parallel=True, fastmath=True)
def func2(x, y):
    for i in prange(len(x)):
        x[i] = x[i] + np.power(y[i], 1.0/3)    


if __name__ == "__main__":
    seed = 1000
    np.random.seed(seed)
   
    for n in [1000, 2000, 5000, 10000, 20000, 50000, 100000, 500000, 1000000]:
        x = np.random.rand(n)
        y = np.random.rand(n)

        x_ref = x.copy()
        func1(x_ref, y)

        x_ref = x.copy()
        ts = time.time()
        func1(x_ref, y)
        te = time.time()
        t_func1 = te - ts


        x_cmp = x.copy()
        func2(x_cmp, y)

        x_cmp = x.copy()
        ts = time.time()
        func2(x_cmp, y)
        te = time.time()
        t_func2 = te - ts
        
        print(f'n --> {n}')
        print('t_func1: ', t_func1)
        print('t_func2: ', t_func2)
```

The parallel operation is faster when `n`, the length of `T`, is 500k or 1M. 

---

Next task: compare the running time of `mstump` between "vectorized operation" and "parallel operation" in `core._compute_multi_PI`

NimaSarajpoor:
Due to a high running time, I just tested it for `d = [1, 3, 5]` and `n = [10-000, 20-000, 50-000]` on my pc, which has two cores, and each has one thread! (Yes! It is old :smile: ). The y-value of each point is the difference between the running time with prange (over columns) and the vectorized operation (this branch). So, a positive y-value means that the computing time with `prange` (over columns) is higher than the existing, vectoried operation.


running_time_fig

Colab has two cores, each with two threads. Two questions:
* Do you think we should check the performance `d=1 and n=500-000` on Colab?
* As we increase the number of threads, we should probably get better performance for `prange-over-columns` approach. So, I was wondering how one should compare the two approaches in a "fairly" manner.

seanlaw:
No, I think it is clear that we should just go with the vectorized version.

NimaSarajpoor:
@seanlaw 
I think this PR is ready [to be merged].

seanlaw:
Thank you @NimaSarajpoor !

PR #889, opened 2023-07-25
Fix #769 fix name and order of variables 

### LINKED ISSUE #769: Enhance consistency of the code regarding `T_A` and `T_B`
It seems there are a few inconsistencies in the naming of variables and the way they are being passed to functions.

(i) In stump, we have this:
https://github.com/TDAmeritrade/stumpy/blob/f078424350ab97d02a517da9e29b254309b27574/stumpy/stump.py#L634-L650

So, `T_A` corresponds to Q-related variables, and `T_B` corresponds to T-related variables. However, we see something different in `gpu_stump`:
https://github.com/TDAmeritrade/stumpy/blob/f078424350ab97d02a517da9e29b254309b27574/stumpy/gpu_stump.py#L577-L578

(ii) Also, the module `stump` itself shows some sort of inconsistency:
https://github.com/TDAmeritrade/stumpy/blob/f078424350ab97d02a517da9e29b254309b27574/stumpy/stump.py#L678-L695

Since we pass the arguments as `_stump(T_A, T_B, ...)`, we should expect to see the same order in the rest of the arguments. However, this is not true here. As you may have noticed already, `M_T` (which corresponds to `T_B`) is located before `μ_Q` (which corresponds to `T_A`).

---

In my opinion, these inconsistencies occur because we, unintentionally, confuse ourselves by using the `T_A`, `T_B` naming and `Q` and `T` naming. I think we should usually consider `T_A` as the sequence that contains queries (hence `Q`), and `T_B` is `T`. 

So, if we do:
```
Q, μ_Q, σ_Q = core.preprocess(T_A, m)
T, M_T, Σ_T = core.preprocess(T_B, m) 
```

We will be consistent and we will not be confused when we try to be consistent regarding the order of arguments in a function.
(Btw, I just provided two examples. I haven't checked all modules yet)

NimaSarajpoor:


NimaSarajpoor:
In #769, @seanlaw pointed out that:

>  T_A should always correspond to Q and, vice verse, with T_B and T. Although, I wonder if it is confusing when we have functions that only take T and m but no Q? In that case, T has no direct relation to T_B

I think it would be reanable to think about this as follows:
* sometimes we just have a time series `T` and we are interested in finding its [sliding] mean, std, etc.
* sometimes we have the time series `T`, but we also have another time series / sequence `Q` which is our query 

@seanlaw
What do you think? For now, I just revised `stumpy/stump.py` (and I had to do minor changes for two more modules). I will go through the other modules one by one, even the ones that were changed because of the changes in `stump.py`.  If you are happy with this, we can start applying similar changes to other modules where appropriate.

seanlaw:
> What do you think? 

I'm not sure I understand. 

> I think it would be reanable to think about this as follows:
sometimes we just have a time series T and we are interested in finding its [sliding] mean, std, etc.
sometimes we have the time series T, but we also have another time series / sequence Q which is our query

These sounds like true statements but I don't see how this is related to the code. Can you provide more context? At the end of the day, `Q` should really be a single subsequence (either 1D or multi-dimensional) but it should not contain more than one subsequence

NimaSarajpoor:
> > What do you think?
> 
> I'm not sure I understand.
> 
> >

I want to know if it is okay [from your perspective] to use the name `Q` instead of `T_A`. But, I think I get my answer when you said:

> At the end of the day, Q should really be a single subsequence (either 1D or multi-dimensional) but it should not contain more than one subsequence

Regarding:
These sounds like true statements but I don't see how this is related to the code. Can you provide more context?

My goal is to get rid of `T_A` in the following line:
https://github.com/TDAmeritrade/stumpy/blob/286cb178d201ea4378d8dde70f6d24c1b8caa5a2/stumpy/stump.py#L667-L674

And, I wanted to replace it with `Q`. Then, when I want to pass variables to `_stump`, their names would contain either `Q` and `T` [as opposed to the existing version where the variables are `T_A`, `T_B`, `μ_Q`, `M_T`, and so on]

---

If you think we should keep the name `T_A` [and avoid using `Q`], then I can keep the names and  just fix the order of arguments in the [private] functions. For instance, see the changes in the signature of `_stump` in this PR.

seanlaw:
> I can keep the names and just fix the order of arguments in the [private] functions. 

Yes, let's fix this

stumpy/gpu_stump.py
```diff
@@ -170,16 +170,16 @@ def _compute_and_update_PI_kernel(
             )
 
             QT_out[0] = QT_first[i]
-        if math.isinf(M_T[j]) or math.isinf(μ_Q[i]):
+        if math.isinf(μ_Q[j]) or math.isinf(M_T[i]):
```
NimaSarajpoor, line 173: It might be a good idea to swap the name of variables `i` and `j` so as to be consistent with `stumpy/stump.py` where `i` is used for `μ_Q` (which is associated with `T_A`) and `j` for `M_T` (which is associated with `T_B`).

seanlaw: Sure... if it works

NimaSarajpoor:
Quick review:
(1) `stump`, `stumped`, and `scrumped` are modified to reflect a better order of arguments in the private functions

(2) The non-normalized version of the modules mentioned in (1) are checked and they are good.

(3) Also, `gpu_stump` is modified so that the order of arguments in its private functions becomes consistent with the module `stump`. The unit test `tests/test_gpu_stump.py` is checked with Colab. Its tests are passing.

(4) @seanlaw: When you get a chance, please take a look at `gpu_stump`. If the changes are okay with you, I will apply similar changes to `gpu_aamp`

(5) I have avoided touching the module `mstump` since it is complicated, and, also, at some point I realized that it might be a good idea to leave it as is. For instance, see the following lines [from `mstump` module]:

https://github.com/TDAmeritrade/stumpy/blob/286cb178d201ea4378d8dde70f6d24c1b8caa5a2/stumpy/mstump.py#L514-L525

Here, the **query** is `T_B[:, query_idx : query_idx + m]`, so I think the name `μ_Q` that is associated with `T_B` here may not be  bad idea. So, that is why I did not change anything in `mstump`.

seanlaw (changes requested):
@NimaSarajpoor Looks good. Please see the one requested change

NimaSarajpoor:
[update] 
I revised `gpu_aamp.py` according to the changes applied to `gpu_stump.py`. Its tests are passing (in Colab).

@seanlaw 
I did a quick scan over the modules and I think we are good.

stumpy/gpu_aamp.py
```diff
@@ -439,8 +440,6 @@ def _gpu_aamp(
 
 
 def gpu_aamp(T_A, m, T_B=None, ignore_trivial=True, device_id=0, p=2.0, k=1):
-    # function needs to be revised to return (top-k) matrix profile and
```
seanlaw, line 442: @NimaSarajpoor Can you also remove this from `aamped.py` as well (assuming it's already been fixed there)

NimaSarajpoor: I fixed that and two other modules that have the same out-dated comment.

seanlaw:
@NimaSarajpoor Thanks again for doing this! My head is feeling much clearer now :)

NimaSarajpoor:
Same here :)

PR #892, opened 2023-08-01
Fix #881 Snippets Unit Test Assertion Failure

### LINKED ISSUE #881: Snippets Unit Test Assertion Failure
It appears that snippets unit tests are failing [here](https://github.com/TDAmeritrade/stumpy/actions/runs/5569517430/jobs/10172918123) and it's not clear if it may be related to the snippet comment in #828

NimaSarajpoor:
Fix #881

### DIFF
--- a/stumpy/snippets.py
+++ b/stumpy/snippets.py
@@ -136,10 +136,6 @@ def _get_all_profiles(
             query_idx=start,
         )
 
-        # post-processing `D` to avoid loss of precision
-        for j in range(0, i * m, m):
-            D[i, j] = D[j // m, i * m]
-
     stop_idx = n_padded - m + 1 - right_pad
     D = D[:, :stop_idx]
 
--- a/stumpy/snippets.py
+++ b/stumpy/snippets.py
@@ -136,6 +136,10 @@ def _get_all_profiles(
             query_idx=start,
         )
 
+        # post-processing `D` to avoid loss of precision
+        for j in range(0, i * m, m):
+            D[i, j] = D[j // m, i * m]
+
     stop_idx = n_padded - m + 1 - right_pad
     D = D[:, :stop_idx]
 
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -103,3 +103,49 @@ def test_snippets():
     )
     npt.assert_almost_equal(ref_areas, cmp_areas, decimal=config.STUMPY_TEST_PRECISION)
     npt.assert_almost_equal(ref_regimes, cmp_regimes)
+
+
+def test_calculate_squared_distance():
+    # This test function raises an error if the distance between a subsequence
+    # and another does not satisfy the symmetry property
+    seed = 332
+    np.random.seed(seed)
+    T = np.random.uniform(-1000.0, 1000.0, [64])
+    m = 3
+
+    T_subseq_isconstant = core.rolling_isconstant(T, m)
+    M_T, Σ_T = core.compute_mean_std(T, m)
+
+    i, j = 2, 10
+    # testing the distance between query `Q_i = T[i : i + m]`, and the
+    # subsequence T[j : j + m] should be the same as the distance between
+    # the query `Q_j = T[j : j + m]` and the subsequence T[i : i + m]
+
+    QT_i = core.sliding_dot_product(T[i : i + m], T)
+    dist_ij = core._calculate_squared_distance(
+        m,
+        QT_i[j],
+        M_T[i],
+        Σ_T[i],
+        M_T[j],
+        Σ_T[j],
+        T_subseq_isconstant[i],
+        T_subseq_isconstant[j],
+    )
+
+    QT_j = core.sliding_dot_product(T[j : j + m], T)
+    dist_ji = core._calculate_squared_distance(
+        m,
+        QT_j[i],
+        M_T[j],
+        Σ_T[j],
+        M_T[i],
+        Σ_T[i],
+        T_subseq_isconstant[j],
+        T_subseq_isconstant[i],
+    )
+
+    comp = dist_ij - dist_ji
+    ref = 0.0
+
+    npt.assert_almost_equal(ref, comp)
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -101,7 +101,8 @@ def test_calculate_squared_distance():
             comp = dist_ij - dist_ji
             ref = 0.0
 
-            npt.assert_almost_equal(ref, comp, decimal=15)
+            assert ref == comp
+            # npt.assert_almost_equal(ref, comp, decimal=15)
 
 
 def test_snippets():
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -111,8 +111,7 @@ def test_calculate_squared_distance():
             comp = dist_ij - dist_ji
             ref = 0.0
 
-            assert ref == comp
-            # npt.assert_almost_equal(ref, comp, decimal=15)
+            npt.assert_almost_equal(ref, comp, decimal=15)
 
 
 def test_snippets():
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -59,52 +59,6 @@ def test_distace_profile():
         npt.assert_almost_equal(D_ref, D_comp)
 
 
-def test_snippets():
-    m = 10
-    k = 3
-    s = 3
-    seed = 332
-    np.random.seed(seed)
-    T = np.random.uniform(-1000.0, 1000.0, [64])
-
-    isconstant_custom_func = functools.partial(
-        naive.isconstant_func_stddev_threshold, quantile_threshold=0.05
-    )
-    (
-        ref_snippets,
-        ref_indices,
-        ref_profiles,
-        ref_fractions,
-        ref_areas,
-        ref_regimes,
-    ) = naive.mpdist_snippets(
-        T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func
-    )
-    (
-        cmp_snippets,
-        cmp_indices,
-        cmp_profiles,
-        cmp_fractions,
-        cmp_areas,
-        cmp_regimes,
-    ) = stumpy.snippets(T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func)
-
-    npt.assert_almost_equal(
-        ref_snippets, cmp_snippets, decimal=config.STUMPY_TEST_PRECISION
-    )
-    npt.assert_almost_equal(
-        ref_indices, cmp_indices, decimal=config.STUMPY_TEST_PRECISION
-    )
-    npt.assert_almost_equal(
-        ref_profiles, cmp_profiles, decimal=config.STUMPY_TEST_PRECISION
-    )
-    npt.assert_almost_equal(
-        ref_fractions, cmp_fractions, decimal=config.STUMPY_TEST_PRECISION
-    )
-    npt.assert_almost_equal(ref_areas, cmp_areas, decimal=config.STUMPY_TEST_PRECISION)
-    npt.assert_almost_equal(ref_regimes, cmp_regimes)
-
-
 def test_calculate_squared_distance():
     # This test function raises an error if the distance between a subsequence
     # and another does not satisfy the symmetry property.
@@ -149,3 +103,49 @@ def test_calculate_squared_distance():
     ref = 0.0
 
     npt.assert_almost_equal(ref, comp)
+
+
+def test_snippets():
+    m = 10
+    k = 3
+    s = 3
+    seed = 332
+    np.random.seed(seed)
+    T = np.random.uniform(-1000.0, 1000.0, [64])
+
+    isconstant_custom_func = functools.partial(
+        naive.isconstant_func_stddev_threshold, quantile_threshold=0.05
+    )
+    (
+        ref_snippets,
+        ref_indices,
+        ref_profiles,
+        ref_fractions,
+        ref_areas,
+        ref_regimes,
+    ) = naive.mpdist_snippets(
+        T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func
+    )
+    (
+        cmp_snippets,
+        cmp_indices,
+        cmp_profiles,
+        cmp_fractions,
+        cmp_areas,
+        cmp_regimes,
+    ) = stumpy.snippets(T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func)
+
+    npt.assert_almost_equal(
+        ref_snippets, cmp_snippets, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(
+        ref_indices, cmp_indices, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(
+        ref_profiles, cmp_profiles, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(
+        ref_fractions, cmp_fractions, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(ref_areas, cmp_areas, decimal=config.STUMPY_TEST_PRECISION)
+    npt.assert_almost_equal(ref_regimes, cmp_regimes)
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -1,3 +1,5 @@
+import functools
+
 import naive
 import numpy as np
 import numpy.testing as npt
@@ -55,3 +57,49 @@ def test_distace_profile():
         )
 
         npt.assert_almost_equal(D_ref, D_comp)
+
+
+def test_snippets():
+    m = 10
+    k = 3
+    s = 3
+    seed = 332
+    np.random.seed(seed)
+    T = np.random.uniform(-1000.0, 1000.0, [64])
+
+    isconstant_custom_func = functools.partial(
+        naive.isconstant_func_stddev_threshold, quantile_threshold=0.05
+    )
+    (
+        ref_snippets,
+        ref_indices,
+        ref_profiles,
+        ref_fractions,
+        ref_areas,
+        ref_regimes,
+    ) = naive.mpdist_snippets(
+        T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func
+    )
+    (
+        cmp_snippets,
+        cmp_indices,
+        cmp_profiles,
+        cmp_fractions,
+        cmp_areas,
+        cmp_regimes,
+    ) = stumpy.snippets(T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func)
+
+    npt.assert_almost_equal(
+        ref_snippets, cmp_snippets, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(
+        ref_indices, cmp_indices, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(
+        ref_profiles, cmp_profiles, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(
+        ref_fractions, cmp_fractions, decimal=config.STUMPY_TEST_PRECISION
+    )
+    npt.assert_almost_equal(ref_areas, cmp_areas, decimal=config.STUMPY_TEST_PRECISION)
+    npt.assert_almost_equal(ref_regimes, cmp_regimes)
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -1,11 +1,21 @@
 import functools
+from unittest.mock import patch
 
 import naive
 import numpy as np
 import numpy.testing as npt
+import pytest
+from numba import cuda
 
 import stumpy
 from stumpy import config, core
+
+try:
+    from numba.errors import NumbaPerformanceWarning
+except ModuleNotFoundError:
+    from numba.core.errors import NumbaPerformanceWarning
+
+TEST_THREADS_PER_BLOCK = 10
 
 
 def test_mpdist_snippets_s():
@@ -151,10 +161,14 @@ def test_snippets():
     npt.assert_almost_equal(ref_regimes, cmp_regimes)
 
 
+@pytest.mark.filterwarnings("ignore", category=NumbaPerformanceWarning)
+@patch("stumpy.config.STUMPY_THREADS_PER_BLOCK", TEST_THREADS_PER_BLOCK)
 def test_distance_symmetry_property_in_gpu():
+    if not cuda.is_available():  # pragma: no cover
+        pytest.skip("Skipping Tests No GPUs Available")
+
     # This test function raises an error if the distance between a subsequence
     # and another does not satisfy the symmetry property.
-
     seed = 332
     np.random.seed(seed)
     T = np.random.uniform(-1000.0, 1000.0, [64])
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -74,7 +74,7 @@ def test_calculate_squared_distance():
     k = n - m + 1
     for i in range(k):
         for j in range(k):
-            QT_i = core.sliding_dot_product(T[i : i + m], T)
+            QT_i = core._sliding_dot_product(T[i : i + m], T)
             dist_ij = core._calculate_squared_distance(
                 m,
                 QT_i[j],
@@ -86,7 +86,7 @@ def test_calculate_squared_distance():
                 T_subseq_isconstant[j],
             )
 
-            QT_j = core.sliding_dot_product(T[j : j + m], T)
+            QT_j = core._sliding_dot_product(T[j : j + m], T)
             dist_ji = core._calculate_squared_distance(
                 m,
                 QT_j[i],
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -107,7 +107,7 @@ def test_snippets():
 
 def test_calculate_squared_distance():
     # This test function raises an error if the distance between a subsequence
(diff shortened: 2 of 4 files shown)


NimaSarajpoor:
@seanlaw 
After merging #893, I will update this branch and run the tests again.

NimaSarajpoor (commented):
@seanlaw 
Do you think we can come up with a more elegant solution?

seanlaw:
> @seanlaw Do you think we can come up with a more elegant solution?

It feels like a hack and has a code smell. In this instance, how do we know which direction (AB vs BA) is suffering from a loss of precision?

NimaSarajpoor:
> It feels like a hack and has a code smell. 

I agree!

> In this instance, how do we know which direction (AB vs BA) is suffering from a loss of precision?

Need to dig deeper :) Will provide an update.

NimaSarajpoor:
[update]

> how do we know which direction (AB vs BA) is suffering from a loss of precision?

For the case added to the test function, the answer is Both! However, the loss of imprecision in one resulted in increasing the ref by +1e-14, and in the other, resulted in increasing the ref by -1e-14. since the delta value are not identical, it resulted in a change in the final outcome!


I followed the code all the way and noticed this particular loss of precision is coming from the function 
`core._calculate_squared_distance` in which the distance between `T[i : i + s]` and `T[j : j + s]` is not the same as  the distance between `T[j : j + s]` and `T[i : i + s]`, where `i==2`, `j==10`. 

Note 1: I set `fastmath` to `False` but that did not help
Note 2: I tried something else locally. Going to push it here and see how it goes.

stumpy/snippets.py
```diff
@@ -136,6 +136,10 @@ def _get_all_profiles(
             query_idx=start,
         )
 
+        # post-processing `D` to avoid loss of precision
```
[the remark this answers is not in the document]
NimaSarajpoor: [update]
cancelled due to a recent change in the code

NimaSarajpoor:
@seanlaw 
I think I got to the bottom of it! 

```
# A failure case

m = 10
k = 3
s = 3

seed = 332
np.random.seed(seed)
T = np.random.uniform(-1000.0, 1000.0, [64])

M_T, Σ_T = stumpy.core.compute_mean_std(T, s)

i, j = 2, 10

# denom for computing the rho (see func `core._calculate_squared_distance` in the branch `main`)
denom_v1 = print(s * Σ_T[i] * Σ_T[j])
denom_v2 = print(s * Σ_T[j] * Σ_T[i])
# denom_v1 and denom_v2 have a very small difference

# denom for computing  m * rho (see the last commit in this PR) 
denom_v1 = print(Σ_T[i] * Σ_T[j])
denom_v2 = print(Σ_T[j] * Σ_T[i])
# denom_v1 and denom_v2 are exactly the same
```

seanlaw:
> I think I got to the bottom of it!

Can you please add this test case to `test_precision.py`?

NimaSarajpoor:
@seanlaw 
> I'd like to post this in the numba forum

I think this is related to just python (I ran my simple experiment outside of numba!)

> Can you tell me a set of values for σ_Q, Σ_T, and m that would cause denom to be different if we changed the order?

How about this:
```
lst = []
for seed in list(range(1000)):
    np.random.seed(seed)
    vals = np.random.uniform(-1000, 1000, size=2)
    x = vals[0]
    y = vals[1]
    m = 3

    A = m * x * y 
    B = m * y * x
    
    if A == B:
        continue
    else:
        lst.append(seed)
```

Then, run the same thing but change the order of multiplication. Check `lst` in both cases.
(Spoiler alert: the `lst` stays empty if you change the order of multiplication.)

> With those same values, are you able to check if gpu_stump might also suffer from the same issue:

Do you think I should still check this given that the issue is coming from python?

seanlaw:
Thanks @NimaSarajpoor. It looks like this would make things more precise:

```
A = decimal.Decimal(3) * decimal.Decimal(650.912209452633) * decimal.Decimal(722.0717285148526)
B = decimal.Decimal(3) * decimal.Decimal(722.0717285148526) * decimal.Decimal(650.912209452633)
```

However, note that you'll have to cast it back to `float(A)` and `float(B)` before that value can be used to combine with other float values

NimaSarajpoor:
Just to make sure we are on the same page: are we talking about comparing the performance of stump? (And maybe gpu_stump too?)

NimaSarajpoor:
(New lesson for me: avoid hack as much as possible! Now that I think more, and after that I read your "Decimal" approach, I can see that swapping the variables is like a hack! These hacks should be our last resort! Thank you for the lesson !!)

seanlaw:
> Just to make sure we are on the same page: are we talking about comparing the performance of stump? (And maybe gpu_stump too?)

So, since the problem is really in `_calculate_squared_distance`, I would call the parent function `_calculate_squared_distance_profile` and pass in sets of `QT, μ_Q, σ_Q, M_T, Σ_T` that are of varying lengths (which would then execute `_calculate_squared_distance` repeatedly) to see if using `decimal` has any affect on the computation time.

Does that make sense?

> (New lesson for me: avoid hack as much as possible! Now that I think more, and after that I read your "Decimal" approach, I can see that swapping the variables is still like a hack! These hacks should be our last resort! Thank you for the lesson !!)

I'm glad that you understood. It is certainly easier to hack together a solution but it is another thing to:

1. Isolate the problem (what you did) and then
2. Understand what causes the problem (i.e., general issues with floating point representation)
3. The best/typical options that others have used to address it cleanly (what I did)
4. Try those solutions

NimaSarajpoor:
@seanlaw 
I just noticed that we have another similar case when computing the distance. Kindly see the line below:

https://github.com/TDAmeritrade/stumpy/blob/6c367accc523708f6e4ec6701da76447cc5b1b0b/stumpy/core.py#L1103

`m * μ_Q * M_T` is the similar part where changing the order can change the precision. For now, I made a small change to address this and the one in denom. I know that is a hack but it is a cleaner one compared to previous hacks. I will consider it as our last solution.

I will go and apply your proposal to the code for both `m * μ_Q * M_T`  and the denom `m * σ_Q * Σ_T`. I will provide an update.

seanlaw:
> I will go and apply your proposal to the code for both m * μ_Q * M_T and the denom m * σ_Q * Σ_T. I will provide an update.

Sounds good!

NimaSarajpoor:
[update]

Apparently, numba does not support decimal types. See: [this Github issue](https://github.com/numba/numba/issues/8769) and [this post in numba Discourse](https://numba.discourse.group/t/numba-with-decimal-package-for-increasing-the-accuracy/944).

---

**[just for our future reference if needed]**
I also compared the running time of the original version with this PR. They are the same.

hack


```
# code
n_values = 100000 * np.arange(11)
n_values[0] = 5000

print('n_values:', n_values)


m = 100


running_time=[]
for i in range(len(n_values)):
    np.random.seed(i)
    n = n_values[i]
    T = np.random.uniform(-1000, 1000, size=n)
    j = np.random.randint(n - m + 1)
    Q = T[j : j + m]

    QT = stumpy.core._sliding_dot_product(Q, T)
    

    Q_subseq_isconstant = False
    T_subseq_isconstant = np.zeros(n - m + 1, dtype=bool)
    
    M_T, Σ_T = stumpy.core.compute_mean_std(T, m)
    
    
    lst = []
    for _ in range(5):
        ts = time.time()
        _calculate_squared_distance_profile(
            m,
            QT,
            M_T[j],
            Σ_T[j],
            M_T,
            Σ_T,
            Q_subseq_isconstant,
            T_subseq_isconstant,
        )
        te = time.time()
        lst.append(te - ts)
    
    running_time.append(np.mean(lst))
```

seanlaw:
Thank you. I find it rather surprising that adding parentheses did not solve the problem.

NimaSarajpoor:
> Thank you. I find it rather surprising that adding parentheses did not solve the problem.

Yes, it is surprising indeed!

---

Btw, I tried `np.float128` as follows:

```
denom = m * np.float128(σ_Q * Σ_T) 
```

In an isolated test, it seems that it can solve the precision issue. But, when I want to use it in njit func, I get numba error:
```
 464         msg = (f"{str(e).rstrip()} \n\nThis error may have been caused "
    465                f"by the following argument(s):\n{args_str}\n")
    466         e.patch_message(msg)
--> 468     error_rewrite(e, 'typing')
    469 except errors.UnsupportedError as e:
    470     # Something unsupported is present in the user code, add help info
    471     error_rewrite(e, 'unsupported_error')

File ~/miniconda3/envs/stumpypy39/lib/python3.10/site-packages/numba/core/dispatcher.py:409, in _DispatcherBase._compile_for_args.<locals>.error_rewrite(e, issue_type)
    407     raise e
    408 else:
--> 409     raise e.with_traceback(None)

TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Failed in nopython mode pipeline (step: nopython frontend)
float128
During: typing of get attribute at /tmp/ipykernel_13752/2182441483.py (58)

File "../../../../../tmp/ipykernel_13752/2182441483.py", line 58:
<source missing, REPL/exec in use?>

During: resolving callee type: type(CPUDispatcher(<function _calculate_squared_distance at 0x7fc6a21fc8b0>))
During: typing of call at /tmp/ipykernel_13752/2182441483.py (121)

During: resolving callee type: type(CPUDispatcher(<function _calculate_squared_distance at 0x7fc6a21fc8b0>))
During: typing of call at /tmp/ipykernel_13752/2182441483.py (121)


File "../../../../../tmp/ipykernel_13752/2182441483.py", line 121:
<source missing, REPL/exec in use?>
```

NimaSarajpoor:
> Also, can you please add the precision unit test above for _calaculate_squared_distance?

will do.

seanlaw:
> I find it rather surprising that adding parentheses did not solve the problem.

So, when I do:

```
print(3 * 650.912209452633 * 722.0717285148526)
print(3 * 722.0717285148526 * 650.912209452633)
```
then the output is different:

```
1410015.9125726535
1410015.9125726533
```

However, if I add parentheses:
```
print(3 * (650.912209452633 * 722.0717285148526))
print(3 * (722.0717285148526 * 650.912209452633))
```

the issue appears to be resolved:
```
1410015.9125726535
1410015.9125726535
```

Similarly, adding parentheses here:

```
import numpy as np

np.random.seed(0)
vals = np.random.uniform(-1000, 1000, size=2)
x = vals[0]
y = vals[1]
m = 3

A = m * (x * y)
B = m * (y * x)
```

Also produces identical results
```
print(A)
print(B)

126049.76376646347
126049.76376646347
```

NimaSarajpoor:
> However, if I add parentheses:
> 
> ```
> print(3 * (650.912209452633 * 722.0717285148526))
> print(3 * (722.0717285148526 * 650.912209452633))
> ```
> 
> the issue appears to be resolved:

Maybe I made a mistake when I was investigating the impact of parentheses. Because I just tried it with/without numba and it worked !! Thanks for checking that on your end @seanlaw. That was my bad. 

I am going to use parentheses and see if the newly-added tests pass or not.

NimaSarajpoor:
> Maybe I made a mistake when I was investigating the impact of parentheses.

okay...I think I may know what happened. When I use parentheses, it passes the test function `tests/test_precision.py::test_calculate_squared_distance` (the test function I just added recenlty) but it does NOT pass the test function `tests/test_precision.py::test_snippets`, which is the test function I added in the beginning of this PR! So, I think I checked "parentheses" using the latter test function at that time and I thought it does not resolve the precision in `3 * 650.912209452633 * 722.0717285148526`.

So, the error in snippet needs to be investigated again. I am going to push the commit with parentheses so you can see the error too in the test function `test_snippet`. I will then investigate the snippet again!

seanlaw:
> So, the error in snippet needs to be investigated again. I pushed the commit with parentheses so you can see the error too in the test function test_snippet. I will then investigate the snippet again!

All tests are passing 🤷‍♂️

NimaSarajpoor:
@seanlaw 
What is happening here? :laughing: 

So, I ran it locally, and I got error. And, then I pushed it here and I noticed an error on the github webpage. Then, after an hour or so, I checked my github app and noticed the tests were still running. 

**Did you trigger the tests again?**

I am not sure if something is messed up on my end or not. So, this time, I ran it in google colab and I got assertion failuer!

```
tests/test_precision.py ...F                                             [100%]

=================================== FAILURES ===================================
________________________________ test_snippets _________________________________

    def test_snippets():
        m = 10
        k = 3
        s = 3
        seed = 332
        np.random.seed(seed)
        T = np.random.uniform(-1000.0, 1000.0, [64])
    
        )
        (
            ref_snippets,
            ref_indices,
            ref_profiles,
            ref_fractions,
            ref_areas,
            ref_regimes,
        ) = naive.mpdist_snippets(
            T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func
        )
        (
            cmp_snippets,
            cmp_indices,
            cmp_profiles,
            cmp_fractions,
            cmp_areas,
            cmp_regimes,
        ) = stumpy.snippets(T, m, k, s=s, mpdist_T_subseq_isconstant=isconstant_custom_func)
    
            ref_snippets, cmp_snippets, decimal=config.STUMPY_TEST_PRECISION
        )
E       AssertionError: 
E       Arrays are not almost equal to 5 decimals
E       
E       Mismatched elements: 10 / 30 (33.3%)
E       Max absolute difference: 1546.55831117
E       Max relative difference: 20.34471955
E        x: array([[  61.66845,  600.71676,   93.43324,  380.43982, -341.5808 ,
E                455.06518,  -92.4037 , -321.6817 ,  731.41598, -694.49969],
E              [ 743.18048, -887.71494, -286.93457, -562.81888,  306.91077,...
E        y: array([[  61.66845,  600.71676,   93.43324,  380.43982, -341.5808 ,
E                455.06518,  -92.4037 , -321.6817 ,  731.41598, -694.49969],
E              [ 743.18048, -887.71494, -286.93457, -562.81888,  306.91077,...

tests/test_precision.py:138: AssertionError
========================= 1 failed, 3 passed in 9.72s ==========================
```

NimaSarajpoor:
@seanlaw 
Can you please run the tests again here? Tbh, I have no idea what is going on :)

seanlaw:
> Did you trigger the tests again?

Nope, it was the one and only run. I did not trigger the tests again

seanlaw:
> Can you please run the tests again here? Tbh, I have no idea what is going on :)

Yep, I will re-run the tests again now.

NimaSarajpoor:
@seanlaw 
I have been checking the logs for the ongoing run of tests and it seems that I cannot find `test_precision`. Would you mind checking the log and see if you can find it there?

seanlaw:
> So, I ran it locally, and I got error. And, then I pushed it here and I noticed an error on the github webpage. Then, after an hour or so, I checked my github app and noticed the tests were still running.

Since it was never executed on Github, doesn't this imply that there is something wrong with the parentheses + `numba`? There is no error on the coverage testing where `numba` is turned off.

NimaSarajpoor:
> > So, I ran it locally, and I got error. And, then I pushed it here and I noticed an error on the github webpage. Then, after an hour or so, I checked my github app and noticed the tests were still running.
> 
> Since it was never executed on Github, doesn't this imply that there is something wrong with the parentheses + `numba`? There is no error on the coverage testing where `numba` is turned off.

Ahh! Right! Going to play with fastmath flags, the one(s) related to the precision of floating point in computation.

NimaSarajpoor:
@seanlaw 
After merging #896, I will update this PR. If tests pass, we are good!

seanlaw:
What if we allow `reassoc` but change the order as you had proposed:

```
denom = (σ_Q * Σ_T) * m
denom = max(denom, config.STUMPY_DENOM_THRESHOLD)

ρ = (QT - (σ_Q * Σ_T) * m) / denom
```

> I just noticed that we have another similar case when computing the distance. Kindly see the line below:

Can you also check/test/fix this case as well as `gpu_stump`?

NimaSarajpoor:
> What if we allow `reassoc` but change the order as you had proposed:

Will re-order the operations and check `gpu_stump`. But, before this, I would like to investigate this case further. The added test function to check the symmetry property of distance (i.e, `dist_ij  - dist_ji == 0`) still pass even with numba. But, the `test_snippet` function (in `tests/test_precision.py`) does not pass. I am curious to know 
 why the problem is fixed after changing the flag. Because, as mentioned earlier, the symmetry property seems to work fine even with `fastmath=True`. I will provide an update

seanlaw:
@NimaSarajpoor In case it matters, in all previous cases (see `core.py`), we've specified `fastmath={"nsz", "arcp", "contract", "afn", "reassoc"}` (so we allow `nan` and `inf` to be present for those functions).

NimaSarajpoor:
[please see the test function `tests/test_precision.py::test_calculate_squared_distance` in this PR. This test function raises an error if the symmetry property of distance between two subsequence is violated.]

After some investigations, I noticed the following things:

Note 1: `npt.assert_almost_equal(ref, comp, decimal=7)` does not raise error even with `fastmath=True`

Note 2: `npt.assert_almost_equal(ref, comp, decimal=15)` raises an error unless **at least** one of the flags `nnan` or `ninf` is removed from `fastmath`.

Note 3: Although we may think `decimal=7` is okay, we can see that there is a difference if we use `decimal=15`. Note that this level of precision is important since, in the snippet, the values  of an array are being added up, and then the `np.argmin` operation will be applied. `np.argmin` considers all decimal in a value when it wants to find the minimum value and not just till 7-th decimal position. [As mentioned before, this simply can result in a change in the index of snippet, which can result in assertion failure]

Note 4: We may remove the last test function in `test_precision.py` (i.e. `test_snippet.py`) since we are tackling the problem at its core. But, we may keep it just to be on the safe side.

Note 5: I need to find a way to see how I can test it in `gpu_stump`. So, I first need to answer this question: "what does conidering `gpu_stump` mean here regarding this PR. Maybe we just call `core._calculate_squared_distance` in a device function and see how it performs when it is used [in GPU]. 

What do you think @seanlaw ?

NimaSarajpoor:
> `npt.assert_almost_equal(ref, comp, decimal=15)` raises an error unless **at least** one of the flags `nnan` or `ninf` is removed from `fastmath`.

NimaSarajpoor:
> What if we allow `reassoc`...

I allowed it and the tests pass. The problem was with the flags `nnan` and `ninf` when they are BOTH allowed in fastmath. I removed them both. 

> ... but change the order as you had proposed:

I just pushed a commit to take care of it as well just to increase the safety/stability of the code regarding precision.

> I think we should just do assert ref==comp

I now use `assert`. I can revert it back to `npt.assert_almost_equal` if needed.

seanlaw:
> I think we should just do `assert ref==comp` in this particular test to make sure all decimals are considered.

I find `npt.assert_almost_equal(ref, comp, decimal=15)` to be more explicit. `assert ref==comp` feels like a typo or inconsistent. The fact that `decimal=15` is called out seems very deliberate and is therefore important

seanlaw:
> Note 4: We may remove the last test function in test_precision.py (i.e. test_snippet.py) since we are tackling the problem at its core. But, we may keep it just to be on the safe side.

I would just keep it. At the minimum, it's there to remind us that we've had precision-related issues in the past

> Note 5: I need to find a way to see how I can test it in gpu_stump. So, I first need to answer this question: "what does conidering gpu_stump mean here regarding this PR?
> Maybe we just call core._calculate_squared_distance in a device function and see how it performs when it is used [in GPU].

I think what I'm interested in knowing is whether:

```
A = 3 * 650.912209452633 * 722.0717285148526
B = 3 * 722.0717285148526 * 650.912209452633
```

produces the same results in a GPU. If "yes", then there is nothing to do. If order matters, then we should change the code accordingly

NimaSarajpoor:
@seanlaw 
Do you like the new test function added to check the precision in gpu? 
I tested in colab. It fails using `gpu_stump` of the main branch. 

Adding parentheses [for multiplication in GPU] did not resolve the assertion failure. But, when I re-order the multiplication, it resolves the issue. Can you please try it out on your end as well?

tests/test_precision.py
```diff
@@ -55,3 +67,133 @@ def test_distace_profile():
...
+                Σ_T[i],
+                T_subseq_isconstant[j],
+                T_subseq_isconstant[i],
+            )
+
+            comp = dist_ij - dist_ji
+            ref = 0.0
+
+            assert ref == comp
+            # npt.assert_almost_equal(ref, comp, decimal=15)
```
seanlaw, line 115: Please use the `decimal=15` version

seanlaw:
> Adding parentheses [for multiplication in GPU] did not resolve the assertion failure. But, when I re-order the multiplication, it resolves the loss-of-precision issue. Can you please try it out on your end as well?

Excellent. I think the GPU test looks great! Testing on Colab now

tests/test_precision.py
seanlaw: Otherwise, the unit test is still failing:

https://github.com/TDAmeritrade/stumpy/actions/runs/5773239904/job/15648976680?pr=892

seanlaw (commented):
Can you please the latest commit into this repo? This way the `test.sh` is the the most up-to-date for testing the GPU on colab

NimaSarajpoor:
> GPU on colab (I don't think `test_precision` is executed for unit testing in this repo when I clone it on colab)

Sure. [In the meantime, you may just use the flag `custom` to run the test for `test_precsion.py` only]

tests/test_precision.py
NimaSarajpoor: Doesn't unit test may still fail since difference is in the order of `1e-15`? It fails in **minimum** version testing.

NimaSarajpoor: I updated the branch and applied the changes, and pushed the commit. Going to have my eye on the tests to see how they go.

NimaSarajpoor: [Before I forget]
As one of our last resorts, we may remove the flag `reassoc` from fastmath and see if it solves the issue and [if yes] how much it affects the running time.

seanlaw:
> Excellent. I think the GPU test looks great! Testing on Colab now

`test_precision.py` is passing for me on colab. After the tests pass here, I think we are good to merge, right @NimaSarajpoor?

tests/test_precision.py
seanlaw: Sounds good!

NimaSarajpoor:
> After the tests pass here (and you check the performance), I think we are good to merge, right @NimaSarajpoor?

Right..I think we should be good.

---

I noticed the assertion failure in the minimum version testing is for:

```
# python 3.7 
# numpy 1.17
# note: no numba

import numpy as np
seed = 332
np.random.seed(seed)
T = np.random.uniform(-1000.0, 1000.0, [64])
m = 3

i, j = 0, 51

ij_dot = np.dot(T[i:i+m], T[j:j+m])
ji_dot = np.dot(T[j:j+m], T[i:i+m])

npt.assert_almost_equal(ij_dot, ji_dot, decimal=15)
```

And, it fails. I tried to change the version of python and/or numpy to find what version preserves the symmetry property of dot product but couldn't find it as I still got failure! [which is strange since when I checked it in colab with python `3.10.12` and numpy `1.22.4`, I get no error!] I will look into it.

seanlaw:
> Is it possible that a specific versio of numpy works with a specific version of python considering their patches?

It's certainly possible. However, you're only calling `np.dot`, which I doubt has been changed in a long, long time and, my guess, it shouldn't have issues with backwards compatibility. 

I think this is worth posting as an issue on the `numba` Github page

seanlaw:
@NimaSarajpoor In case it might make our lives easier, I was thinking about bumping the minimum Python version since Python 3.7 reached end of life about a month ago. Of course, it'll require some work to figure out what other corresponding minimum `numpy`, `numba`, and `scipy` versions should accompany Python 3.8

NimaSarajpoor:
> I think this is worth posting as an issue on the `numba` Github page

I created a new conda env with just python and numpy (so no numba), and got the assetion failure. 

> 

> I noticed the assertion failure in the minimum version testing is for:
> 
> ```
> # os: linux
> # python 3.7 
> # numpy 1.17
> # note: no numba
> 
> import numpy as np
> seed = 332
> np.random.seed(seed)
> T = np.random.uniform(-1000.0, 1000.0, [64])
> m = 3
> 
> i, j = 0, 51
> 
> ij_dot = np.dot(T[i:i+m], T[j:j+m])
> ji_dot = np.dot(T[j:j+m], T[i:i+m])
> 
> npt.assert_almost_equal(ij_dot, ji_dot, decimal=15)
> ```

The assertion will not fail if I do `T=np.random.uniform(-1000.0, 1000.0, [64]).astype(np.float128)`

---

I tried to do:
```
conda create -n py38 python=3.8
conda activate py38
conda install numpy
```

And, I tried the case above and got assetion failure.

---

[not sure if it is related or not] 
I noticed that I get no asserion failure in two of my conda env. In both, I have `numpy-base` package as well.

NimaSarajpoor:
> n case it might make our lives easier, I was thinking about bumping the minimum Python version since Python 3.7 reached end of life about a month ago. Of course, it'll require some work to figure out what other corresponding minimum `numpy`, `numba`, and `scipy` versions should accompany Python 3.8

Dropping support for 3.7 may not be a bad idea as I noticed several packages did the same. I am still curious to understand why setting up my conda env with python 3.8 and installing a numpy package does not resolve the assertion failure.

seanlaw:
> Is there anything I can do? Should I start with changing the minimum version in requirement.txt and see if the tests pass?

See issue #897

seanlaw:
@NimaSarajpoor What do you think? Is this ready to be merged?

NimaSarajpoor:
I think it is ready 👍

NimaSarajpoor:
Btw, one of the new test func has no comment. Is that okay?

NimaSarajpoor:
[update]
> Btw, one of the new test func has no comment. Is that okay?

I took care of it. 

Ready to review and merge [if nothing else needs to be done here]

seanlaw:
Thanks for working on this @NimaSarajpoor !

PR #896, opened 2023-08-05
Fix 895 update unit test

### LINKED ISSUE #895: `test_precision` is missing in uni test `test.sh`

NimaSarajpoor:
This PR fixes #895

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -200,6 +200,8 @@ test_unit()
     check_errs $?
     pytest -rsx -W ignore::RuntimeWarning -W ignore::DeprecationWarning -W ignore::UserWarning tests/test_non_normalized_decorator.py
     check_errs $?
+    pytest -rsx -W ignore::RuntimeWarning -W ignore::DeprecationWarning -W ignore::UserWarning tests/test_precision.py
+    check_errs $?
 }
 
 test_coverage()
--- a/test.sh
+++ b/test.sh
@@ -137,71 +137,11 @@ test_custom()
 test_unit()
 {
     echo "Testing Numba JIT Compiled Functions"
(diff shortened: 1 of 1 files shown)


seanlaw:
I can't remember why but for unit tests, they must be listed out as shown rather than the way that you had simplified.

seanlaw:
I noticed that it took 2 h and 37 mins to run...

Re-running again

NimaSarajpoor:
> It may have been because of `dask` or `gpu`-related functions

I was trying to see if I could find any difference in the commands in the shell script when it needs to run on GPU but couldn't find anything. 

> Can you confirm that all tests are being executed in test_unit and that there are no side effects on colab?

(1) I will check the logs
(2) I will test it on colab

seanlaw:
> I tried to see if I could find any difference in the commands in the shell script when a test needs to run on GPU but couldn't find anything. [so, I thought a for-loop should be safe]

What I would do is check that the number of tests being executed for each file is the same with/without the for-loop. For example, `test_aamp.py` says `27 passed in 21.45s`. This needs to be the same for all `test_x.py` files in both non-for-loop and for-loop scenarios.

NimaSarajpoor:
> . This needs to be the same for all `test_x.py` files in both non-for-loop and for-loop scenarios.

They are the same (I compared one unit testing check in for-loop scenario with one unit testing check in non-for-loop scenario, and I checked the counts of tests being passed/skipped in each `test_*.py` file)

---

Also: these new tests will be covered after the merge:
* test_chain.py
* test_floss.py
* test_stamp.py
* test_stomp.py (discontinued?)

NimaSarajpoor:
> Next task: check colab

Done! All tests are passing and they were compared with the outcome of running `./test.sh` in branch main. All good.

seanlaw:
Thanks @NimaSarajpoor for the refactor. We may be able to simplify this down further

PR #74, opened 2019-10-10
MPdist documentation

Deepakgthomas:

Issue #67, opened 2019-08-25
Using DataFrame Inputs with MSTUMP/MSTUMPED


seanlaw:
@marcusau asked:

I have tried multi-dimensional array to  mstump with stock prices and its technical indicators 

    #### step 2 : Feature creation
    df1=df.copy()

    df1['Close_pct']=np.log(df1['Close'] / df1['Close'].shift(1)).dropna()
    df1['STDDEV']= ta.STDDEV(df1['Close'], timeperiod=5, nbdev=1)
    ### Volume Indicator Functions
    df1['OBV']=ta.OBV(df1['Close'], df1['Volume'])
    df1['Chaikin AD'] = ta.AD(df1['High'], df1['Low'], df1['Close'], df1['Volume'])

    ### momentum Indicator

    macd, macdsignal, macdhist = ta.MACD(df1['Close'], fastperiod=12, slowperiod=26, signalperiod=9)
    df1['macdhist']=macdhist

    df1['RSI']= ta.RSI(df1['Close'], timeperiod=14)

    # Volatility Indicator Functions
    df1['NATR'] = ta.NATR(df1['High'], df1['Low'], df1['Close'], timeperiod=14)
    df1['TRANGE'] = ta.TRANGE(df1['High'], df1['Low'], df1['Close'])
    print(df1.tail())

    print(list(set(df.columns) ^ set(df1.columns)))

    feature_cols=list(set(df.columns) ^ set(df1.columns))
    df1=df1.loc[:,feature_cols].dropna()
    print(df1.head())

    #Store these values in the NumPy array for using in our models later:
    f_x=()
    for f in feature_cols:
      f_x += (df1[f].values.reshape(-1,1),)
    X = np.concatenate(f_x,axis=1)
    X=X.T
    print(X.shape)
    >>>> (8, 2429)
<img width="839" alt="Screenshot 2019-08-25 at 6 55 35 PM" src="https://user-images.githubusercontent.com/31444195/63648956-fc4e6380-c769-11e9-81c2-88d71f7a558c.png">

    window_size = 10  # Approximately, how many data points might be found in a pattern

    matrix_profile, matrix_profile_indices = stumpy.mstump(X, m=window_size)

    left_matrix_profile_index = matrix_profile[:, 2]
    right_matrix_profile_index = matrix_profile[:, 3]

as you said, mstumpy only support 1-D data, however, what is the explanation of the results in mstumpy?

seanlaw:
[...]

PR #920, opened 2023-10-03
Tutorial notebook DAMP

NimaSarajpoor:
This PR is a replacement for PR #872. 

This PR is to address #606.


docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,276 @@
+{
```
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #52.        chunk_stop = query_idx
</pre><p>This is based on the original algorithm proposed in the paper, where exclusion zone is set to <code>m</code>. This will be changed later to reflect the default <code>excl zone</code> in STUMPY.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #85.                    chunk_start = max(0, chunk_stop - chunksize)
</pre><p>Similar to the original paper, we have fixed the <code>chunk_stop</code> . However, this approach is not optimal as each new chunk will contain the subsequences of the previous chunk. According to the MATLAB code, the <code>chunk_stop</code> should be updated too.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

NimaSarajpoor:
@seanlaw 
For now, I have only implemented the backward processing algorithm as proposed in the [DAMP paper](https://www.cs.ucr.edu/~eamonn/DAMP_long_version.pdf). I have commented on the parts that can be changed to enhance the code.

docs/Tutorial_DAMP.ipynb
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #51.        chunksize = next_pow2(m) 
</pre><p>In contrast to the algorithm proposed in the paper, the MATLAB code uses <code>next_pow2(8 * m)</code></p><p> </p><p>Why <code>8</code> ? It has a code smell. Also, in another part of the code, we see:</p><pre class="ql-syntax" spellcheck="false">#from MATLAB
addParameter(p,'lookahead',2^nextpow2(16*SubsequenceLength));
</pre><p>which seems to be for <strong>forward</strong> processing. But, why <code>16</code>?  </p><p> </p><p>MAYBE the authors wanted to make sure that the algorithm can still consider a good number of subsequences in the first round even when the subsequence length m is [close to] a power of two. For instance, if <code>m == 32</code> ,  <code>next_pow2(m)</code> returns 32.  In this case, the first chunksize will be 32, meaning that it contains only one subsequence if exclusion zone is set to <code>m</code> </p><p> </p><p>I prefer to avoid using 8 or 16 for now. [We can later test and see how much it improves the performance in edge cases.]</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #85.                    chunk_start = max(0, chunk_stop - chunksize)
</pre><p>According to the MATLAB code <code>DAMP_2_0.m</code> , the authors use the following approach to update the chunk (in backward processing).</p><p> </p><p>In the first iteration, they do: </p><pre class="ql-syntax" spellcheck="false">#MATLAB
approximate_distance = min( real(MASS_V2(T(i-X+1:i), query))); 
</pre><p>where X is the initial chunksize.</p><p> </p><p>Then, while <code>approximate_distance &gt; best-so-far-discord-distance</code> , they do: </p><pre class="ql-syntax" spellcheck="false">#MATLAB

# update chunksize
X = 2 * X  # doubling the chunksize
expansion_num = expansion_num + 1   # initially, expansion_num is set to 0 

# update chunk start/end points
X_start = i-X+1+(expansion_num * SubsequenceLength);
X_end = i-(X/2)+(expansion_num * SubsequenceLength);

# compute distance
approximate_distance = min( real(MASS_V2(T(X_start:X_end), query))); 
</pre><p> </p><p>Here, the authors go backward in the time series. They double the chunksize in each iteration. IIUC, the reason for having <code>expansion_num</code> in the code is nothing but just because of the way they update the chunk start, end points using the query index <code>i</code> . </p><p>This is not a clean way! Instead, one can just use the chunk end point and find the chunk start point from it by going backward. This results in a cleaner code compared to the current approach where the authors use the query index to compute the chunk start, end point in each iteration.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

seanlaw: <p>Agreed (re: code smell). </p>

seanlaw: <p>If the chunk size is known ahead of time, one should pre-compute the start/stop indices for each chunk using the <code>stumpy.core._get_ranges</code> function</p>

NimaSarajpoor: @seanlaw 
IIUC, in the original proposed algorithm, the chunksize is getting doubled when the current chunk is checked and it does not result in early-abandoning. So, the next chunk will have the length `2 * len(current_chunk)`.

NimaSarajpoor: Cool! For now, I am going to leave it as is and resolve this conversation. I will add a note to the comment  at the top of this PR so that we do not forget about it.

NimaSarajpoor: <p>As explained by <a href="https://github.com/seanlaw" target="_blank">@seanlaw</a> in PR#872, we should consider <code>excl_zone</code>s so that the output from DAMP to be the same as the discord discovered via <code>stumpy.stump</code>  </p><p> </p><p>Since the naive DAMP uses <code>STUMPY.config.STUMPY_EXCL_ZONE_DENOM</code> , I think it is better to add the <code>excl_zone</code> now rather than later.  </p>

seanlaw: <p>Sure</p>

docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,378 @@
...
```
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #24.        For the given index i, segmentize the array np.arange(i) into 
</pre><p>Is there a better way to describe the task of this function?</p><p> </p><p>"Given the index <code>i</code>, divide the array <code>np.arange(i)</code> into chunks. Ensure the last chunk's size is <code>chunksize_init</code>, with the preceding chunks doubling in size. The output consists of (start, stop) indices of each chunk."</p><p> </p><p><strong>Also, we may add: </strong></p><p>The left-most chunk always start at 0.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #42.        out : np.ndarray
</pre><p>Are you okay with the description of <code>out</code>?   </p><p> </p><p>How about this:</p><p>"A 2D numpy array where each row represents a (start, stop) index pair. <code>out[0]</code> corresponds to the chunk immediately preceding index <code>i</code>, and <code>out[-1]</code> corresponds to the left-most chunk in the time series, starting at index 0."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #74.        else:
</pre><p> </p><p>To reduce the cognitive load of developer, we may remove the <code>else</code> part, and simply do:</p><pre class="ql-syntax" spellcheck="false">bfs = max(bfs, nn_distance)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

NimaSarajpoor:
@seanlaw 
I have provided a few comments. Please let me know what you think.

docs/Tutorial_DAMP.ipynb
seanlaw: <p>Wait, are you doing a <code>for-else</code>? If so, let's avoid that as it will confuse people  </p>

seanlaw: <p>This function seems weird. It's not clear why you need to end with the last element being <code>chunksize_init</code> </p><p> </p><p>Something feels backwards here. </p>

seanlaw: <p>Also, do we actually need to call the function multiple times? Or is it possible to call the function once and then shift the indices somehow?</p>

NimaSarajpoor: > Something feels backwards here

Well, it is backward (I mean... the process itself is backward). but, I trust your lens! Please allow me to think more and see if I can come up with something that is more elegant.


> Also, do we actually need to call the function multiple times? Or is it possible to call the function once and then shift the indices somehow?

I wanted to do that but I decided to start simple. What we can do is to keep shifting the start, stop indices of all chunks by one in each iteration. So, for instance, if I get the chunks for the subsequences in `T[:idx]`. Then, I can find the chunks for `T[:(idx+1)]` by just adding start/stop indices of chunks by one to the right. We just need to keep track of the left-most chunk though as it can become the power two of a number at some point. In such case, the number of chunks will be increased by one.

NimaSarajpoor: Please ignore it for now as I may revise the function.

NimaSarajpoor: Yes. It is the for-else approach... and as you said... it is not easy to read. Going to change it.

NimaSarajpoor: [FYR...before I forget]
Another approach is to just offset from the original stop index `idx` by `s`, `2s`, `4s`, `8s`, and so on, where `s` is power of two.  The good thing is that the difference between any two consecutive offset is STILL a power of two. This may result in cleaner code.  IIRC, I think I saw something similar in the MATLAB code before. Going to check that as well.

NimaSarajpoor: The MATLAB code uses the following lines for updating the start / stop of each chunk:

```
# Query is at index `i`, i.e. `query = T(i:i+SubsequenceLength-1)`
# X is the chunk size, initially set to `2 ^ nextpow2(8 * SubsequenceLength)`

X_start = i - X + 1 + (expansion_num * SubsequenceLength)
X_end = i - (X / 2) + (expansion_num * SubsequenceLength)

# and then
approximate_distance = min( real(MASS_V2(T(X_start:X_end), query))); 
```

The term `(expansion_num * SubsequenceLength)` is to take into account the elements of the last subsequence in a chunk that were not considered in the previous chunk. To keep the length of chunk untouched, the `X_start` will have the same shift. 

---

Note 1:
According to the paper / MATLAB code, the length of chunk, `T`, in `core.mass(Q, T)` should be a power-of-two, and NOT the number of subsequences in the chunk.

Note 2:
The reason behind considering the length `power-of-two` for chunk is to speed up the MASS algorithm (according to the authors)

So, I did a quick check by considering different sizes for chunk in range `(2 ** 20 - 1000 - 1, 2 ** 20 + 1000 + 1)`

```
seed = 0
np.random.seed(seed)

T = np.random.rand(10_000_000)
m = 50

T, M_T, Σ_T, T_subseq_isconstant = stumpy.core.preprocess(T, m)

query_idx = 600000
t_lst = []
for stop in range(2 ** 20 - 1000 - 1, 2 ** 20 + 1000 + 1):
    time_start = time.time()

    QT = core.sliding_dot_product(
        T[query_idx : query_idx + m], 
        T[start : stop],
    )
    
    D = core._mass(
        T[query_idx : query_idx + m],
        T[start : stop],
        QT=QT,
        μ_Q=M_T[query_idx],
        σ_Q=Σ_T[query_idx],
        M_T=M_T[start : stop - m + 1],
        Σ_T=Σ_T[start : stop - m + 1],
        Q_subseq_isconstant=T_subseq_isconstant[query_idx],
        T_subseq_isconstant=T_subseq_isconstant[start : stop - m + 1],
    )

    time_end = time.time()
    t_lst.append(time_end - time_start)
```

And, the I plot it:


```
indices = np.arange(2 ** 20 - 1000 - 1, 2 ** 20 + 1000 + 1)
indices = indices[2:]
t_lst = t_lst[2:]

idx = np.flatnonzero(indices == 2 ** 20)[0]

plt.figure(figsize=(20, 5))
plt.scatter(indices[idx], t_lst[idx], color='r', label='chunksize = 2 ** 20')
plt.plot(indices[idx-200 : idx+200], t_lst[idx-200:idx+200]) 
plt.ylabel('running time')
plt.legend()
plt.show()
```

image

Well, it seems the running time for the chunk size `2 ** 20` is low (not necessarily the lowest) but it should be okay. To remain faithful to the paper, I am going to consider the length power-of-two for the chunk size.

docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,569 @@
...
```
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #28.        while start &lt; chunk_stop:
</pre><p>Is this code readable? I tried to vectorize it but couldn't figure out "how". If the number of subsequences in each chunk were supposed to be a power of two, I think I would be able to vectorize it. However, according to the paper, the size of the chunk itself should be a power of two. In other words, the number of subsequences is like... <code>2 ** num - m + 1</code> </p><p> </p><p>Since I couldn't find out a cleaner solution, I am going with <code>naive_get_range_damp</code> for now. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #59.            chunks_range = naive_get_range_damp(idx - excl_zone + m, m, start=0)
</pre><p>@NimaSarajpoor</p><p>Add comment to why we do <code>idx - excl_zone + m</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #63.        # compared with.
</pre><p>@NimaSarajpoor</p><p>remove <code>with</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #67.            PL[start : stop - m + 1] = np.minimum(PL[start : stop - m + 1], D)
</pre><p>Previously, we had the variable <code>pruned</code> , which was a boolean vector, where <code>pruned[i]</code> is pruned if the i-th subsequence is pruned. And, instead of line above (i.e. line #67), we had:</p><p> </p><pre class="ql-syntax" spellcheck="false">mask = np.argwhere(D &lt; bsf) + start
pruned[mask] = True
</pre><p> </p><p>Recall that the paper's original algorithm does not compute <code>PL[i]</code> if <code>pruned[i]</code> is True. It just skips it. In such case, the original algorithm set <code>PL[i]</code> as follows:   </p><p> </p><pre class="ql-syntax" spellcheck="false">PL[i] = PL[i-1]
</pre><p> </p><p>which makes it difficult to find the correct index of discord. The MATALB code does a hack instead as follows:</p><pre class="ql-syntax" spellcheck="false">PL[i] = PL[i-1] - 0.000001 
</pre><p> </p><p>and this does not seem to be a clean solution.</p><p> </p><p>So, instead, I am updating the (approximate) matrix profile <code>PL</code> by using the computed distance profile D . This helps us to avoid the hack, and I think It should not be computationally more expensive compared to <code>np.argwhere(D &lt; bfs)</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <p>Lines 48-55 updates the <code>chunks_range</code> by shifting it by one to the right in each iteration, and reset the first index to 0.</p><p> </p><p>I think the codes in line 48-55 are not clean! Still trying to figure out if there is a better way for updating the chunks range.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

NimaSarajpoor:
@seanlaw 

Would you please check out the comments and let me know what you think. I need your support to figure out:
(1) If there is a better / cleaner way to get the chunks range (see function `naive_get_range_damp`)
(2) If there is a better to update the chunks range in each iteration in the main function `DAMP` 

**Note [Regarding exclusion zone]:**
Also, as discussed, I am considering `excl_zone` (instead of the hardcoded value `m`) to make the result align with what one obtains using the `stumpy.stump`

docs/Tutorial_DAMP.ipynb
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #51.        distance : float
</pre><p>@NimaSarajpoor</p><p>change the name to <code>nn_distance</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #51.        chunksize = next_pow2(m)
</pre><p> </p><p>@NimaSarajpoor</p><p>Add the parameter <code>lookahead</code> to the function and pass the value <code>next_pow2(m)</code> as the argument. No need to re-compute it in every call of  <code>_forward_process</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #39.        next_expected_chunk_size =  next_pow2(m) * (2 ** (n_chunks - 1)) 
</pre><p>@NimaSarajpoor</p><p>Add comment to explain <code>n_chunks - 1</code>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

NimaSarajpoor: <p>Done.</p>

NimaSarajpoor: <p>Done.</p>

NimaSarajpoor: <p>Done</p>

NimaSarajpoor: <p>Going to remove comment as it seems to be unnecessary.</p>

NimaSarajpoor: <p>Will take care of it.</p>

docs/Tutorial_DAMP.ipynb
```diff
@@ -0,0 +1,592 @@
...
```
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #45.        if discord_dist == np.NINF:
</pre><p>This should be <code>if discord_dist == np.inf</code> </p><p>Also: no need for the next line <code>discord_dist = np.inf</code>   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #13.        Update the (approximate) left matrix profile <code>PL</code> in place by checking the right neighbors of
</pre><p> </p><p>revise the definition as the current description can be misleading. We may say:</p><pre class="ql-syntax" spellcheck="false">Update the (approximate) left matrix profile <code>PL</code> for a chunk of forthcoming subsequences by computing their distance to their left neighbor <code>T[query_idx : query_idx+m]</code>

</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #36.        chunks_range = naive_get_range_damp(split_idx - excl_zone + m, m, start=0)
</pre><p><strong>[suggestion]</strong></p><p>We might use the name <code>chunks_geometric_range</code> as the range of chunks are following the geometric series with the initial value of next_pow2(m) (the coefficient) and the ratio of 2 . It is better to set this initial value and pass it to the function <code>naive_get_range_dump</code> because the same initial value will be used later to compute <code>expected_last_chunk_size</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #33.        k = len(T) - m + 1  # len(IL)
</pre><p>We should avoid using <code>k</code> to represent the number of subsequences. Instead we should use <code>l</code>    </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #35.        for i in range(split_idx, k):
</pre><p>Instead of this <code>for</code> loop, can we simplify and do:</p><pre class="ql-syntax" spellcheck="false">PL = core._idx_to_mp(IL, T, m, check_neg=False) 
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #41.        discord_idx = np.argmax(
</pre><p>maybe slightly cleaner to do:</p><pre class="ql-syntax" spellcheck="false">discord_idx = np.where(np.isinf(PL), np.NINF, PL).argmax()
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    def next_pow2(v):
</pre><p>Replace <code>v</code> with <code>val</code> instead  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.        For the given range <code>(start, stop)</code>, segmentize it into 
</pre><p>"segment" instead of "segmentize"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.        chunks, each with a length of power of two. Each chunk is
&nbsp; &nbsp; preceded by one that is twice in size
</pre><p>So, for <code>naive_get_range_damp(stop=100, m=5)</code> what should the output be? I am seeing:</p><pre class="ql-syntax" spellcheck="false">array([[ 92, 100],
&nbsp; &nbsp; &nbsp; &nbsp;[ 81,&nbsp; 97],
&nbsp; &nbsp; &nbsp; &nbsp;[ 54,&nbsp; 86],
&nbsp; &nbsp; &nbsp; &nbsp;[&nbsp; 0,&nbsp; 59]])
</pre><p>but the difference in chunk size is:</p><pre class="ql-syntax" spellcheck="false">array([ 8, 16, 32, 59])
</pre><p>Which is inconsistent with your description. Why is the final one <code>59</code> and then why is each one preceded by HALF in size? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #23.            range (start, stop) of the i-th chunk. out[0][1]
</pre><p>Should this be <code>out[0, 1]</code> ? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/920/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

seanlaw:
@NimaSarajpoor Let's start with these comments first before we dive into DAMP.

docs/Tutorial_DAMP.ipynb
NimaSarajpoor: That's right! Will revise it 👍

NimaSarajpoor: @seanlaw 
We have two things here:
(1) The direction we traverse the time series data (which is backward here, i.e. from `stop` to `start=0`)
(2) The order we represent the chunks range in the numpy array

> Which is inconsistent with your description
> why is each one preceded by HALF in size?

I think (1) and (2) got mixed. To better explain this, let's reverse the order of the chunks returned by the function `naive_get_range_damp`:

```
out = naive_get_range_damp(stop=100, m=5)
out = out[::-1]
```

which gives:
```
>>> out
array([
       [  0,  59], 
       [ 54,  86],
       [ 81,  97],
       [ 92, 100]
])
```

Now, we can see that as we move in the time series from right (`stop=100`) to left (`start=0`), AND starts from the last row of `out`, then each chunk is PRECEDED BY TWICE in size (except the first one!)

Note that the length of the chunk `out[1]` (with the new order above) is 32. So, the most-left chunk, i.e. `out[0]`, should have a length of 64. But, it is just truncated at index 0.

NimaSarajpoor: Correct. Will revise it.

seanlaw: # The diff between stop-start should always be an incrementing power of 2

The problem is how do we know the set of "incrementing powers of 2" that will fit into the `range(start, stop)`? First, we must make the simpler observation that, in the worst case scenario, the cumulative sum of the "incrementing powers of 2" must be (just) larger than or equal to `stop-start`. For example, if `start=0` and `stop=100`, then we need to look for some set of `1 + 2 + 4 + ... + n` that will be just larger than or equal to `stop - start = 100` so that it can act as a bound for our `for` loop. However, after staring at the cumulative sum for different values `n`, I noticed a strange pattern:

| n | cumsum | pattern |
|---|--------|---------|
| 0 | 0      | $2^0-1$ |
| 1 | 1      | $2^1-1$ |
| 2 | 3      | $2^2-1$ |
| 3 | 7      | $2^3-1$ |
| 4 | 15     | $2^4-1$ |
| 5 | 31     | $2^5-1$ |
| 6 | 63     | $2^6-1$ |

And so, for any `n`, the pattern for the cumulative sum of the "incrementing powers of 2" is simply $2^n-1$! Again, we must solve for $n_{upper}$ (upper bound) under the condition that the cumulative sum is (just) larger than or equal to `stop - start = 100`:

$$
\begin{align}
2^{n_{upper}}-1 &\ge 100
\newline 2^{n_{upper}} &\ge 101
\newline \log_22^{n_{upper}} &\ge log_2101
\newline {n_{upper}} &\ge log_2101
\newline {n_{upper}} &= \lceil{log_2101}\rceil
\newline {n_{upper}} &= \lceil{6.658211482751795}\rceil
\newline {n_{upper}} &= 7
\end{align}
$$

# Example

```
start, stop = 0, 100
n_upper = math.ceil(math.log2(stop-start+1))
n_upper

# 7
```

Of course, you may have noticed that we should NOT have started our cumulative sum at `1` and, instead, we should start our cumulative sum at the next power of 2 that is closest to the window size, `m`. So, if `m=5`, then we should start the cumulative sum at $n_{lower}=3$. This is easily remedied by subtracting the cumulative sum leading up to `8`:

$$
\begin{align}
\left( 2^{n_{upper}}-1 \right) - \left( 2^{n_{lower}}-1 \right) &\ge 100
\newline  2^{n_{upper}}-1 &\ge 100 + \left( 2^{n_{lower}}-1 \right)
\newline  2^{n_{upper}}-1 &\ge 100 + \left( 2^3-1 \right)
\newline  2^{n_{upper}}-1 &\ge 100 + \left( 8-1 \right)
\newline  2^{n_{upper}}-1 &\ge 100 + 7
\newline  2^{n_{upper}} &\ge 107 + 1
\newline  \log_22^{n_{upper}} &\ge log_2108
\newline  {n_{upper}} &\ge log_2108
\newline  {n_{upper}} &= \lceil{log_2108}\rceil
\newline  {n_{upper}} &= \lceil{6.754887502163468}\rceil
\newline  {n_{upper}} &= 7
\end{align}
$$

```
m = 5
chunk_start = next_pow2(m)
n_lower = math.ceil(math.log2(chunk_start))
n_upper = math.ceil(math.log2(stop + 2**n_lower - 1 + 1))
n_upper

# 7
```

Applying this, we get a new function:

```
def get_damp_range(stop, m, start=0):
    out = []
    n_lower = math.ceil(math.log2(next_pow2(m)))
    n_upper = math.ceil(math.log2(stop + 2**n_lower)) + 1
    chunk_stop = stop
    
    for chunk_size in np.power(2, range(n_lower, n_upper)):
        chunk_start = chunk_stop - chunk_size
        out.append([max(chunk_start, start), chunk_stop])
        chunk_stop = chunk_start + m

        if chunk_start <= 0:
                break
        
    return np.array(out)
```

And we can verify that the results are the same:

```
import numpy.testing as npt

for stop in range(1, 1000):
    for m in range(1, 25):
        ref = naive_get_range_damp(stop, m)
        cmp = get_damp_range(stop, m)
        npt.assert_almost_equal(ref, cmp)
```

All of the tests pass. Hopefully, I didn't make too many mistakes.

NimaSarajpoor: **Note 1:**
I noticed that two adjacent chunks range have one subsequence in common.
```
# stop=100, 
# m =5
>>> out
array([[ 92, 100],
       [ 81,  97],
       [ 54,  86],
       [  0,  59]])
```

The subsequence `T[92:97]` is in both chunks `[92, 100]` and `[81, 97]`.  This can be resolved by changing the line `chunk_stop = chunk_start + m` to `chunk_stop = chunk_start + m - 1`


**Note 2:**
@seanlaw 
I think I get what you did. But, still... there is something that I cannot get my head around. 

> in the worst case scenario, the cumulative sum of the "incrementing powers of 2" must be (just) larger than or equal to stop-start

I am not sure about this. Let's say `start` is set to `0`, and $n_{lower}$ is the minimum integer value that satisfies $2^{n_{lower}} \ge m$. Note that the total number of subsequences is: `stop - m + 1`. The size of each chunk is power of two, i.e. $2^{num}$. Hence, the NUMBER of SUBSEQUENCES within this chunk is: $2^{num} - m + 1$. So, we have:

$(2^{n_{lower}}  - m + 1) +  (2 * 2^{n_{lower}}  - m + 1) + ... + (2^{n-1} * 2^{n_{lower}} - m + 1) \ge stop -m + 1$
where `n` is the number of chunks. The left side is the cumulative sum of number of subsequences in chunks, and the right side is the total number of subsequences.

Hence:
$2^{n_{lower}} \left(1 + 2 + 4 + ... + 2^{n-1} \right) - m * n +  n   \ge stop - m + 1$
$2^{n_{lower}} (2^{n} - 1) - mn + n \ge stop - m + 1$
$2^{n_{lower}} (2^{n} - 1)  \ge stop + (n - 1)(m - 1)$


We can convert the inequation above to the following format: $Ae^{x} + Bx + C = 0$, and solve it. I do not want to go to details. I just wanted to mention that this is different than what you obtained, which is: 
$2^{n_{lower}} (2^{n} - 1)  \ge stop$
$2^{n_{lower} + n}  \ge stop + 2 ^ {n_{lower}}$
where, $n_{lower} + n$ is $n_{upper}$.

NimaSarajpoor: > I'm not sure what the problem is. Both naive_get_range_damp and naive_get_range_damp produce identical results. So, I'm just following what you are algorithm is doing. Feel free to adjust accordingly.

That was a general note (like a self note so that I do not forget).  I think I should have mentioned in a separate comment section to avoid creating confusion.

 
> So, my inequality just an upper bound for the for loop

Yes...I think I was trying to confirm that. I think I can now see what you did!...
$2^{n_{lower}} (2^{n} - 1)  \ge 2^{n_{lower}} (2^{n} - 1)  - (n - 1)(m - 1) \ge stop $

In fact, when you said " upper bound", it helps me to understand what you meant by 
> First, we must make the simpler observation that, in the worst case scenario, the cumulative sum of the "incrementing powers of 2" must be (just) larger than or equal to stop-start

Instead of $2^{num} - m + 1$, you just considered the upper bound $2 ^ {num}$! 

> Maybe you disagree but I think my proposal is pretty straightforward to comprehend.

I think the lock is unlocked in my brain 😄 I think your solution (with that upper bound trick) is waaaay more cleaner.

seanlaw: Right. I haven't thought this through but since we are constantly shifting the stop index by `m`, the upper bound might be even worse if `len(T)` is super long and `m` is super small since there will be a ton of overlap between ranges. Maybe we should test this (as I've only used a small `len(T)` and small `m`. Any issues may reveal itself when `len(T)` is large and `m` is small (or large?!?).

NimaSarajpoor: Going to ignore this for now.

NimaSarajpoor: Done!

NimaSarajpoor:
**A few updates:**

(1) The proposed `get_damp_range` function is added to the notebook.

(2) The running time of DAMP algorithm increases drastically when `m` becomes larger. See the new example 3 in the notebook. The running time of naive DAMP (stumpy-based DAMP) is ~5 sec, and the running time of DAMP algorithm is ~55 sec. **I think it is mostly related to the computation of sliding dot product using `core. sliding_dot_product`**. However, if we use the NUMBA JIT-compiled function `core._ sliding_dot_product` instead, the running time becomes ~41 sec. And, if parallelize this function `core._ sliding_dot_product` (by passing `parallel=True` to the njit decorator  and using `prange`), the running time becomes ~12 sec. Still, this is longer than the `stumpy-based` naive DAMP. 

* Since I do not want to do any premature optimization, I am avoiding making any changes.
* It might be worth it to check the paper and see if the authors compare DAMP with parallelized STOMP algorithm or not. Is DAMP really better than `stumpy.stump` in a batch case? 

(3) An  [experimental top-k DAMP code](https://drive.google.com/file/d/1SRpPXDB2SHFeXsTP2dLfFHVWyOVpFmP_/view?usp=sharing) is provided on the [supporting webpage](https://sites.google.com/view/discord-aware-matrix-profile/documentation). I tried to make some changes on our current DAMP to reflect top-k code. I did a few tests and I think it works (i.e. the results are the same as the top-k discords from naive DAMP). 

**How it works:** 
In short, we just set bsf (best-so-far score) to the `k`-th discord score, and not the top-1. In contrast to the stumpy-based naive DAMP, we need to set `k` in the beginning of the algorithm. 


(4)  Is only ONE discord interesting?
I am not sure... but I FEEL that it should be useful in an online case (i.e. streaming data) when a user wants to see if the current subsequence (containing the latest data) is an anomaly or not when it is compared with a set of usual patterns (e.g. see GOLDEN DAMP in the paper)

seanlaw:
> I think it is mostly related to the computation of sliding dot product using core.sliding_dot_product

I think it is worth profiling the code to see which function most of the time is being spent. Does the paper offer any clues on what timings should be? How does the Matlab code perform?

NimaSarajpoor:
> I think it is worth profiling the code to see which function most of the time is being spent.


```
# example 

import numpy as np

seed = 100
np.random.seed(seed)

T = np.random.rand(100000)
m = 50
split_index = 5000
```

**I profiled the computing time for:** 

* backward proces, forward process, and chunks update block.
* Also, in backward prices and forward process each, I profiled the computing time of `QT` (the dot product part) and `core._mass`


```
# computing time [in seconds]

# in DAMP
backward_time:  49.28337216377258
forward_time:  8.625611543655396
chunkupdate_time:  0.16403627395629883

-----

# considering total calls of `_backward_process`
# QT is regarding the dot product part

backward_time_QT:  47.81996726989746
backward_time_mass:  0.7662274837493896

-----

# considering total calls of `_foreward_process`
# QT is regarding the dot product part

forward_time_QT:  8.326124429702759
forward_time_mass:  0.1750011444091797
```

> Does the paper offer any clues on what timings should be? How does the Matlab code perform?

On it... will provide an update.

seanlaw:
> The running time of naive DAMP (stumpy-based DAMP) is ~5 sec

I didn't realize that STUMPY was that performant even for `n = 100_000`! Out of curiosity, i tried it myself on my M2 Macbook Air and also saw ~4.46s. According to our old performance chart (in the README), it use to take ~19s on a much older 16-core Intel machine. This is exciting to see!

> I ran DAMP version 2.0 [provided on the supporting website](https://sites.google.com/view/discord-aware-matrix-profile/documentation) using the same input of Example 3 (of our notebook in this PR). The output matches (see below) and the running time is ~56 sec (which is close to our current implementation of DAMP)

Thank you. At least this confirms that we're not doing anything horribly wrong compared with the MATLAB code? But is this really scalable? After you've investigated, it might be good to check with Eamonn and his team.

> In contrast to the original algorithm proposed in the paper (and our implementation), the MATLAB code sets the initial chunk size to next_pow2(8 * m)

Presumably, this shouldn't change the performance much?

NimaSarajpoor:
>   didn't realize that STUMPY was that performant even for n = 100_000! 
> This is exciting to see!

Indeed exciting!

> For a [CPU that was released in Q2 of 2018 has only 6 cores/threads](https://www.intel.com/content/www/us/en/products/sku/129939/intel-core-i58500-processor-9m-cache-up-to-4-10-ghz/specifications.html), 22.3 seconds seems really fast for 8 million data points. I'd be curious if we could reproduce these results using their Matlab code or our DAMP code.

According to [this video](https://sites.google.com/view/discord-aware-matrix-profile/home), the `split_index` seems to be 2500, and the window size is `80`. However, according to [this](https://sites.google.com/view/discord-aware-matrix-profile/reproducibility), the window size seems to be set to `94`. The data itself is not a 1D array. It is a 2D array, with two rows, and each row has ~10M data points (not 8 million) 


> But is this really scalable? After you've investigated, it might be good to check with Eamonn and his team.
 
I will get to this after the investigation.

---

Before I forget, I am going to provide a couple of observations. We may discuss/ignore them later after trying out the case you suggested above.

(1) The trillion-experiment data is a text file, containing values, each in one line. The value returned by `wc -l` command is `118-540-512`, which is around 100M (so not trillion?)

Yeah! I will try to do that! The data and the code is provided at the bottom of this page:
https://sites.google.com/view/discord-aware-matrix-profile/dataset#h.dwws21dk9mej

The data is about ~3GB. The maximum size that can be uploaded to MATLAB online server is `256 MB`. So, I think I need to split it to several files, each with maximum size of 256MB, and then reconstruct the full time series on MATLAB after uploading the files there.

Also:  The code associated with this data has the following lines:

```
# MATLAB
T = load('MG.txt');

my_string_name_DOB = 'eamonnjohnkeogh04041968_office_machine';
my_seed = sum(double(my_string_name_DOB));

success_num = 0;
total_time = 0;

n = 100000000;
k = 1500;
for i = 1:k
    MG_anomaly = T(1:100000);
    MG = T(100001:n);
    randn('seed',my_seed*i);
    MG = MG + (randn(size(MG))/1000);
    T = [MG_anomaly;MG];
```

And `n` is 100M here?



(2) I think the performance can be affected by the location of discord. So, it is worth it to report the location of discord in my upcoming investigation (in addition to the running time)

NimaSarajpoor:
> According to [this video](https://sites.google.com/view/discord-aware-matrix-profile/home), the split_index seems to be 2500, and the window size is 80. However, according to [this](https://sites.google.com/view/discord-aware-matrix-profile/reproducibility), the window size seems to be set to 94. The data itself is not a 1D array. It is a 2D array (a matrix), with two rows, and each row has ~10M data points (not 8 million)

I couldn't find the time series with 8M data points. The `mit_long_term_ecg14046.mat` data provided for Fig. 12 on the [supporting webpage]() has two rows, each with 10M data points. In the following cases, I considered  the first row of this data, and I just considered the first 8M data points. The `split_idx` is set to 2500.  Both Python and MATLAB codes were run on MATLAB online server (Thank you @seanlaw  for setting this up!). 

### Case 1-1:
| m=80  |      Python      |  MATLAB |
|----------|:-------------:|------:|
| discord loc |  973932 | 973933 |
| discord dist |     10.433   |   10.4333 |
| running time [sec] | 282.23 |    1044.47 |


### Case 1-2:
| m=94   |      Python      |  MATLAB |
|----------|:-------------:|------:|
| discord loc | 3711  | 3712 |
| discord dist |    11.254    |   11.254 |
| running time [sec] | 428 |    1162  |


Note 1: Indexing in MATLAB starts at 1.
Note 2: The MATLAB code uses `DAMP 2.0`, which  is slightly different:
* In the beginning of the algorithm,  the program computes the full distance profile for the first `~ 16 * m` subsequences located AFTER `split_idx` (with a mix of forward_process).  And then, start using the backward-process / forward-process for the remaining test points (i.e. the points that are located AFTER  `split_idx + 16 * m`)
* Also, the initial chunksize in the backward process of DAMP 2.0 is  the minimum power of two that is $\ge 8m$

So, if I have a data that has some receptive patterns every, say, 2 * m period, then approx. 6 * m subsequences are being processed in the first chunk. However, it is possible to just early abandon the neighbours of a subsequences after considering just the first `2 * m`. So, I am trying to understand the logic behind `8m` vs `m`. I think it all comes to this point:

Is the running time of `MASS(Q, T1 U T2)` (roughly) equal to `MASS(Q, T1) U MASS(Q, T2)`? 


**Next step**
I can remove that first part of DAMP 2.0 that considers full distance profile for `~ 16 * m`, and then change `8m` to `m` for the initial chunksize, or change m to 8m in our python DAMP, and compare again...

seanlaw:
@NimaSarajpoor Thanks for running the comparison. From what I can tell, the Python implementation seems to be around the same performance as Matlab (maybe a little faster)? What should be the next steps? Do we have questions that we should pose to Eamonn?

NimaSarajpoor:
@seanlaw 
I tried to use `stumpi`-like concept (i.e. shifting `QT` and updating it as we consider new data points in `T`)

https://github.com/TDAmeritrade/stumpy/blob/21840d773470effd8bc3cd46c2ba1499f7f50bc1/stumpy/stumpi.py#L426

However, I think: 
(1) it can take more time to implement (as one needs to carefully track the slice of `T` that is processed and bypass QT computation without re-computing chunks range) , 
and 
(2) I have been doing pre-mature optimization! 

So, for now, I just added `@nji` decorator to the notebook function of DAMP.

----

> From what I can tell, the Python implementation seems to be around the same performance as Matlab (maybe a little faster)?

Correct. I just added `@njit` decorator with fast math flag (and `no parallelism`) and its performance becomes faster than MATLAB DAMP and MATLAB DAMP 2.0. 


> What should be the next steps?

(1) I think we can start reproducing a couple of figures of the paper.
(2) We can think about when it makes sense to use DAMP instead of stump (because `stump` uses parallelism but that is not the case in DAMP). 
(3) Add param `k` (for `top-k`), and enhance the code accordingly.


> Do we have questions that we should pose to Eamonn (especially with regards to reproducibility)?

I can think of three questions:
* Where is that 8M data? Because, the provided dataset has 10M data points.
* Has DAMP been compared with other options like `stumpi`, or `stump` (which uses parallel computing) ?
* Most examples provided on [the supporting webpage](https://sites.google.com/view/discord-aware-matrix-profile/reproducibility) except two have small `m`. I was wondering if the performance is tested for large `m` particularly when there is less periodic behaviour in the data.

seanlaw:
> (1) I think we can start reproducing a couple of figures of the paper.
(2) We can think about when it makes sense to use DAMP instead of stump (because stump uses parallel computing).
(3) Add param k (for top-k), and enhance the code accordingly.

Cool! I think reproducing the figures/performance found in the paper is key.

> Where is that 8M data? Because, the provided dataset has two time series each with 10M data points.

We should definitely get some clarification for this.

> Has DAMP been compared with other options like stumpi, or stump (which uses parallel computing) ?

Note that Eamonn likely has no idea what `stumpi` or `stump` is since we've renamed the original functions. Instead, refer to `stompi` and `stomp/mpx`

> Most examples provided on [the supporting webpage](https://sites.google.com/view/discord-aware-matrix-profile/reproducibility) except two have small m. I was wondering if the performance is tested for large m particularly when there is less periodic behaviour in the data.

Based on your test of this, DAMP seems pretty bad as window size increases. Can you confirm that this is also true when using the MATLAB code?

NimaSarajpoor:
For sliding dot product between a query, Q, and a time series T, I tried MATLAB (fft) and Python (fft / njit). The computing time for these three cases are shown below.  

image

Notice that Python with njit seems to be dependent on the window size. However, that is not the case for Python fft (with SciPy). However, to decorate the backward / forward process with njit decorator (to speed up computation), the sliding dot product cannot take advantage of scipy fft as Numba does not support that!


> **Regarding the last item (i.e. performance of DAMP when window size is large)** To do a quick comparison, I considered the first `100k` datapoints provided [here](https://drive.google.com/file/d/1EI_voQS4lsBKitUe6Rq6uKbH-Lrfn_Km/view).
> 
> Then:
> 
> ```
> # python
> data = np.array(loadmat('./DAMP_data/mit_long_term_ecg14046.mat')['mit_long_term_ecg_14046'][0]).astype(np.float64)
> 
> l = 100000 # length of data
> split_idx = 2500
> m in range(50, 1001, 50)
> ```
> 
> image

This figure that I provided previously shows the performance of Python DAMP and Stump. I am now thinking of checking out MATLAB DAMP performance considering different window size and see whether it is independent of `m` or not (According to the performance of sliding dot product in MATLAB, it should be. But, I think it would be good to check that).

@seanlaw 
What do you think? Please let me know if you have any suggestion.

seanlaw:
> Notice that Python with njit seems to be dependent on the window size. However, that is not the case for Python fft (with SciPy). However, to decorate the backward / forward functions with njit decorator (to speed up computation), the sliding dot product cannot take advantage of scipy fft as Numba does not support that!

Yikes, our `njit` version is pretty bad. When you say "scipy fft", are you referring to `scipy.signal.convolve`? Note that `convolve` requires executing an `FFT` 3 times so please make sure that you are doing the same thing everywhere and not simply a single "FFT". Additionally, I vaguely remember reading `scipy.signal.convolve` will attempt to use heuristics to choose the fastest algorithm (based on the length of the array). I believe that a different algo is used when the length is over 500 in length and there may be an even better algorithm for longer time series. I think it would be worthwhile testing longer time series.

Can you please share the code that you used?

I think it might be worthwhile posting any comparisons to both the `numba` discourse channel to see if `numba` can get closer (either by `dot` or by `fft`/`convolve`) and also post to the `scipy` Github issues.

NimaSarajpoor:
> are you referring to scipy.signal.convolve? Note that convolve requires executing an FFT 3 times so please make sure that you are doing the same thing everywhere and not simply a single "FFT"

Below, I am showing what I meant by each case :

* MATLAB fft --> The part of MATLAB MASS function that computes the sliding dot product with help of fft
* Python scipy fft --> `core.sliding_dot_product`, which uses `scipy.signal.convolve`
* Python njit --> `core._sliding_dot_product`, which uses `njit` on top of `np.dot`

Btw, I think it makes sense that the computing time in the last case depends on the window size `m` as it simply uses `np.dot`. So, this computation has $\mathcal{O}({nm})$ time complexity. However, computing the sliding dot product via `fft` trick is said to have $\mathcal{O}({n\log{n}})$ time complexity.

> Can you please share the code that you used?

All the following three cases were run on the MATLAB online server.

Case: [MATLAB] `sliding dot product` from MASS algorithm
```
# MATLAB 

t_lst = [];

idx = 1_000_000;
T = load("mit_long_term_ecg14046.mat");
T = T.mit_long_term_ecg_14046(1, 1:idx);
n = length(T);

m_values = 99:1:1000;
for m = m_values
    Q = T(1:m);
   
    t = tic; 
    Q = Q(end:-1:1); %Reverse the query
    Q(m+1:n) = 0;  %aappend zeros

    X = fft(T);
    Y = fft(Q);
    Z = X.*Y;
    z = ifft(Z);
    
    t_lst(end+1) = toc(t);
end

# later,  the slice `z(m:n)` is considered to compute the distance profile
```

Case: [Python] `core.sliding_dot_product(Q, T)`, which uses `scipy.signal.convolve`
```
from stumpy import core

idx = 1_000_000
T = np.array(loadmat('./mit_long_term_ecg14046.mat')['mit_long_term_ecg_14046'][0]).astype(np.float64)[:idx]

t_lst = []
for m in range(99, 1000):
    Q = T[:m]
    t1 = time.time()
    core.sliding_dot_product(Q, T)
    t2 = time.time()
    t_lst.append(t2 - t1)
```

Case: [Python] `core._sliding_dot_product(Q, T)`, which uses njit decorator on top of `np.dot`
```
from stumpy import core

idx = 1_000_000
T = np.array(loadmat('./mit_long_term_ecg14046.mat')['mit_long_term_ecg_14046'][0]).astype(np.float64)[:idx]

t_lst = []
for m in range(99, 1000):
    Q = T[:m]
    t1 = time.time()
    core._sliding_dot_product(Q, T)
    t2 = time.time()
    t_lst.append(t2 - t1)
```


>  I vaguely remember reading scipy.signal.convolve will attempt to use heuristics to choose the fastest algorithm (based on the length of the array). 

Right. I read the doc. It has a parameter called `method` that is set to `auto` by default, which automatically chooses between `fft` and another method.

> I think it would be worthwhile testing longer time series.

Will try it out. Note that this will affect the case that uses `scipy.signal.convolve`. Still, we cannot decorate it with `njit` (unless we use `np.convolve` with Numba as you suggested), which seems to have a considerable impact on reducing the running time of DAMP. I will try to check the performance of DAMP without njit decorator and using `core.sliding_dot_product`. I GUESS that its running time should not depend on `m` but it cannot outperform the case where DAMP's private functions are decorated with njit

> It appears that numba now supports np.convolve according to a [quick search](https://github.com/search?q=repo%3Anumba%2Fnumba%20convolve&type=code)? This might be relevant to compare

Interesting! I will see if I can easily/safely replace scipy convolve with numpy convolve, and then compare again.

> Note, at some point, we may want to implement our own parallel convolve (FFT) function from scratch without relying on numpy or scipy, which is likely using a much slower np.dot underneath the hood.

I understand. In fact, I already tried [rocket-fft](https://pypi.org/project/rocket-fft/), which supports Numba for `fft`, and then I tried to do what MATLAB code does in python using `np.fft`. However, I did not notice any improvement in the performance. I did not put much time into it though.

> I think it might be worthwhile posting any comparisons to both the numba discourse channel to see if numba can get closer (either by dot or by fft/convolve) and also post to the scipy Github issues.

@seanlaw  
... get closer to the MATLAB's performance? Will do so 👍

seanlaw:
I was curious as to why Matlab's FFT might be faster and came across [this post](https://www.mathworks.com/company/newsletters/articles/faster-finite-fourier-transforms-matlab.html#:~:text=With%20MATLAB%205.3%20and%20a,Matteo%20Frigo%20and%20Steven%20G.).

NimaSarajpoor:
> I was curious as to why Matlab's FFT might be faster and came across [this post](https://www.mathworks.com/company/newsletters/articles/faster-finite-fourier-transforms-matlab.html#:~:text=With%20MATLAB%205.3%20and%20a,Matteo%20Frigo%20and%20Steven%20G.).

Also: https://www.reddit.com/r/Julia/comments/5wel0j/fft_speed_vs_matlab/

seanlaw:
> Also: https://www.reddit.com/r/Julia/comments/5wel0j/fft_speed_vs_matlab/

In 2017, I experimented with FFTW and PyFFTW but hadn't noticed any significant performance difference while noting that they were much harder to use (due to its need to make "plans"). I understand that FFTW uses a ton of tricks and optimizations underneath the hood that could help make DAMP faster but, after a quick search, it looks like FFTW's license (GPL) is incompatible with STUMPY's (BSD3) license. At some point, we may just need to accept our performance-dependency trade off. 

Maybe there's a nice way to detect whether `pyfftw`+`FFTW3` are installed and, if so, hand off the calculations. Otherwise, default to `scipy.signal.convolve`. Or simply set a `config.FFTW = True` flag to force STUMPY to monkey patch itself or something like that. Regardless, we should create a new Github issue.

NimaSarajpoor:
>  At some point, we may just need to accept our performance-dependency trade off.

I understand. The best option is to see if we can improve the performance with the current dependencies. If not, then...

> Maybe there's a nice way to detect whether pyfftw+FFTW3 are installed and, if so, hand off the calculations. Otherwise, default to scipy.signal.convolve. Or simply set a config.FFTW = True flag to force STUMPY to monkey patch itself or something like that.

---

> Regardless, we should create a new Github issue.

It makes sense. I created the issue #938 to continue our conversation regarding the performance of sliding dot product.

NimaSarajpoor:
We recently noticed that Python DAMP's running time increases as window size, `m`, gets larger and larger. To investigate MATLAB DAMP regarding same matter, we first started with comparing MATLAB MASS (with fft trick) and PYTHON _MASS. We noticed that the performance of `stumpy.core.sliding_dot_product`, which uses `scipy.signal.convolve` (fft trick), seems to be independent of window size. This is what we want as MATLAB MASS function shows similar behaviour. However, this function cannot be nit-decorated. An alternative function, `stumpy.core.sliding_dot_product` is nit-decorated; however, it uses `np.dot` for the sliding dot product. This can results in having high computing time when `m` is large (see the following experiments).

(1) Comparing the "sliding dot product" functions 
https://github.com/TDAmeritrade/stumpy/pull/920#issuecomment-1854883771

(2) Comparing Python Naive (i.e. with STUMP using multiple threads) vs Python DAMP (using one thread but nit-decorated with fast math. We used `core._sliding_dot_product`)
https://github.com/TDAmeritrade/stumpy/pull/920#issuecomment-1846618845


> Based on your test of this, DAMP seems pretty bad as window size increases. Can you confirm that this is also true when using the MATLAB code? If "yes", then we should definitely ask Eamonn and his students.


Recall that the current version of MATLAB DAMP provided on the supporting webpage has some differences compared to the original algo proposed in the paper. To this end, I am going to show the result for both versions. To reproduce the following result, here is the code:
```
# MATLAB

running_time_lst_comp = [];
discord_lst_comp = [];

m_values = 50:50:1000;

idx = 100000;
T = load("mit_long_term_ecg14046.mat");
T = T.mit_long_term_ecg_14046(1, 1:idx);
location_to_start_processing = 2500;
for m = m_values
    t = tic;
    
    [discord_score,position] = DAMP(T, m, location_to_start_processing,'lookahead', m);
    
    running_time_lst_comp(end+1) = toc(t);
    discord_lst_comp(end+1) = position;
end

```

(1) DAMP Original 
where, the initial chunksize is `next_pow2(m)`.

The following figure shows the running time of DAMP original for different window sizes. Group of adjacent datapoints with the same color has the same `next_pow2(m)` value, which is the initial chunksize (in moving backward) and the lookahead size (for forward process).  

image

One reason for having ascending trend for group of window sizes with similar `next_pow2(m)` can be the fact that the initial chunk will have less and less subsequences. For instance, for `m=500`, the value of `next_pow2(500)` is `512`. So, this means we only have 13 subsequences in our first iteration of our backward process.


(2) DAMP v2
where, the initial chunksize is `next_pow2(16*m)`.

image

Compared to the previous case, we can see a slightly larger SLOPE in the running time for a group of adjacent data points. One reason can be: "For the same window size between the two cases DAMP original and DAMP v2, DAMP v2 needs to compare the query subsequence agains more subsequences as each chunk is larger in this version of DAMP"


Now, let's get back to the question... 

> Can you confirm that this is also true when using the MATLAB code?

According to my experiments, I think it is true.


---

Let's compare MATLAB DAMP with Python  DAMP (njit version), for different window sizes, by running  the latter on the MATLAB online services.

image


---

Now, let's compare again but with larger window size, `[3500, 4000, 4500]`

image


We can now see the impact of higher time complexity of sliding dot product in Python.

---

Finally, let's just focus on MATLAB DAMP and consider a wide range of values `np.arange(250, 4500+250, 250)` for window size `m` to better see the impact of window size on the performance MATLAB DAMP. The split index is set to `20_000` 

Group of **adjacent** data points with **similar** color are associated with window sizes that have the same `next_pow2(m)` value.


image

seanlaw:
Thanks @NimaSarajpoor

Kotrix:
Thanks for the good work @NimaSarajpoor 

Do you plan to add anything else or the DAMP is ready to roll?

One comment:
> The data is a txt file, containing values, each in one line. The value returned by `wc -l` command is `118-540-512`, which is around 100M (so not trillion?)

From slides here: https://docs.google.com/presentation/d/1tQQfKuKrOa3j5-WJ9peoYfZGOnJbBcl9/edit#slide=id.p31

> we count a trial successful if the top-1 discord is found in the first 100,000 datapoint (created by [e]), rather than from the appended ninety-nine million nine hundred thousand datapoints. **We conducted ten thousand such trials, and found no false positives.**
100 million times 10 thousand is 1 trillion

NimaSarajpoor:
@Kotrix 

> Do you plan to add anything else or the DAMP is ready to roll?

Not ready yet. We are currently focusing on implementing numba-friendly Fast Fourier Transform (FFT), which is an important component of DAMP (see #938  and #939 to track the progress for FFT).  

> One comment:
> 
> > The data is a txt file, containing values, each in one line. The value returned by `wc -l` command is `118-540-512`, which is around 100M (so not trillion?)
> 
> From slides here: https://docs.google.com/presentation/d/1tQQfKuKrOa3j5-WJ9peoYfZGOnJbBcl9/edit#slide=id.p31
> 
> > we count a trial successful if the top-1 discord is found in the first 100,000 datapoint (created by [e]), rather than from the appended ninety-nine million nine hundred thousand datapoints. **We conducted ten thousand such trials, and found no false positives.**
> > 100 million times 10 thousand is 1 trillion

Ahhh.... good catch! Thanks! 😄

seanlaw:
@NimaSarajpoor I've been wondering if I've overthought this and that maybe we could/should attempt to reproduce `DAMP` using pyFFTW and see how fast it is. Then, if it is indeed fast (with pyFFTW), then we make it an optional dependency (i.e., give the user a warning that they aren't using pyFFTW and they should consider installing it for better performance).

NimaSarajpoor:
>  but I think the "planning" and "wisdom" is specific to a given Q and T pair. So, if you repeatedly perform FFT on the same length Q and T, then you can benefit from the existing "plan". However, if your lengths vary then you will likely need a plan for all of the different lengths.

I did some research, and this is correct.  Let's focus on one array for now. For the array `arr`, the following code should build a plan:

```
import pyfftw
obj = pyfftw.builders.rfft(np.empty(len(arr)), ...)
```

and then, to calculate RFFT:
```
RFFT_arr = obj(arr)
```

IIUC, as long as I keep that `obj`,  I should be able to re-use the plan for faster computation of FFT for any array with length `len(arr)`. And if the length of array changes, then my `obj` is going to be new, and cannot access that previous plan anymore. I added the following code as "warm up" to DAMP. It creates RFFT / IRFFT plans for arrays with lengths `2^1, 2^2, ..., 2^20`, and save them in dictionaries. 

```
rfft_objects = {}  # to save RFFT plans
irfft_objects = {}  # to save IRFFT plans

for p in range(1, 21):
    n  = 2 ** p
    rfft_objects[n] = pyfftw.builders.rfft(
        np.empty(n), overwrite_input=True, n=n, threads=1
    )
    irfft_objects[n] = pyfftw.builders.irfft(
        rfft_objects[n].output_array, overwrite_input=True, n=n, threads=1
    )
```

For a given query `Q` in DAMP, we need to perform `sliding_dot_product(Q, T)` on many vary-in-length `T`s (chunks of data). The length of first chunk is `next_pow2(len(Q))`, and the second chunk is `2 * next_pow2(len(Q))`, and so on. The length of the last chunk may not be a power of two. For now, I am finding the `next_pow2` of that last chunk so that I can leverage the existing plans for that chunk too. So, I've added the following code to DAMP:


```
# pre-compute plans
rfft_objects = {}
irfft_objects = {}
for p in range(1, 21):
    n  = 2 ** p
    rfft_objects[n] = pyfftw.builders.rfft(
        np.empty(n), overwrite_input=True, n=n, threads=1
    )
    irfft_objects[n] = pyfftw.builders.irfft(
        rfft_objects[n].output_array, overwrite_input=True, n=n, threads=1
    )

# re-usable arrays 
out = np.empty(2 ** 20, dtype=np.complex128)  # re-usable array to store output of sliding dot product

t1 = time.time()
_DAMP(..., rfft_objects, irfft_objects, out)
t2 = time.time()
```

And then, _DAMP calls the sliding-dot-product function:

```
def next_pow2(val):
    return int(math.pow(2, math.ceil(math.log2(val))))


def pyfftw_sliding_dot_product(Q, T, rfft_objects, irfft_objects, out):
    m = len(Q)
    n = len(T)
    
    s = next_pow2(len(T))
    n_out = s // 2 + 1
    
    Q_pad = np.empty(s, dtype=np.float64)
    Q_pad[:m] = Q[::-1]
    Q_pad[m:] = 0.0
    
    T_pad = np.empty(s, dtype=np.float64)
    T_pad[:n] = T
    T_pad[n:] = 0.0
    
    rfft_plan = rfft_objects[s]
    out[:n_out] = rfft_plan(Q_pad)
    out[:n_out] = np.multiply(out[:n_out], rfft_plan(T_pad))
     
    irfft_plan =  irfft_objects[s]
    return irfft_plan(out[:n_out]).real[m-1:n]


def sliding_dot_product(Q, T, rfft_objects, irfft_objects, out):
    if len(Q) <= 64 or len(Q) == len(T):
        return njit_sliding_dot_product(Q, T)
    else:
        return pyfftw_sliding_dot_product(Q, T, rfft_objects, irfft_objects, out)
```

For example 3 of the notebook, I can see a performance but it is close to the performance of the naive DAMP.  Am I missing something here?

NimaSarajpoor:
@seanlaw 

> 1. Directly compare the FFT sliding dot product performance between Matlab FFTW and pyFFTW. Focusing on that will help us understand whether it is an pyFFTW performance issue
> 2. If performance from step 1 is the same then we should compare the Matlab DAMP implementation to your new pyFFTW DAMP implementation to identify which functions are slow

Good point! Will try to limit MATLAB FFTW to single thread so that I can do apple-to-apple comparison with pyFFTW with one thread. Then, I will remove that limit in MATLAB, and in pyFFTW, I will play with the number of threads.

> 3. Also, if the performance is close to naive DAMP then that sounds like a 5x improvement compared with your previous implementation (looking at the notebook results)

Correct! that was the good part of the story that I did not share. That boost in performance indicates that for larger cases, FFT-based sliding-dot-product performs better. But...I was hoping to see a much better performance than `naive DAMP`.  Maybe it is because of what you mentioned....

> Maybe Example 3 is too small/short of an example?

...that the example is not large enough, and `naive DAMP` still performs better thanks to the parallel mechanism of `stumpy.stump`.

Issue #38, opened 2019-06-18
Multidimensional Time Series Analysis with Continuous and Categorical Data


Sandy4321:
great code , thanks
can it be used for multidimensional time series with both continues and categorical data

seanlaw:
It can only be used for multi-dimensional time series with floating point numbers. It is not obvious how one would use this for categorical time series data since the meaning behind the statistical calculations would not make any sense. Please see issue #23 for more discussions about this.

seanlaw:
@Sandy4321 closing this for now but feel free to reopen if you have further questions.

Issue #33, opened 2019-06-12
Host Tutorial Notebooks on Binder


seanlaw:
[https://gke.mybinder.org/](https://gke.mybinder.org/)

PR #939, opened 2023-12-25
Fixed#938: Speed up njit-decorated function for sliding dot product [WIP]

NimaSarajpoor:
This PR addresses the issue raised in #938.  We want to create a new function for `sliding_dot_product` that:
(1) has a performance that is close to the performance of `core.sliding_dot_product` 
(2) can be called by a njit-decorated function

Currently, the existing alternative to `core.sliding_dot_product`  is `core._sliding_dot_product`; however, in contrast to `core.sliding_dot_product`, its performance is NOT independent of `m`, the length of query.

As suggested by @seanlaw, we are going to try [OTFFT](http://wwwa.pikara.ne.jp/okojisan/otfft-en/stockham2.html) to implement fft and ifft. Then, we use them to create a new function for `sliding_dot_product`.


docs/dev_sliding_dot_product.ipynb
```diff
@@ -0,0 +1,981 @@
+{
```
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.        As a bonus, it tests their performance too when <code>test_output</code> is set to True.
</pre><p>should be replaced with:</p><pre class="ql-syntax" spellcheck="false">As a bonus, it tests their outputs too when <code>test_output</code> is set to True.
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/939/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #20.    # I swapped the params of <code>fft</code> function to make its signature similar to <code>scipy.ftt.ftt</code>
</pre><p>This is an old comment. Remove it. Checked the other versions and fix them as well if needed.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/939/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #38.        _list7_ifft_v1(n, x, y)
</pre><p> </p><p>Need to create <code>y</code> in the line before line#38.</p><pre class="ql-syntax" spellcheck="false">y&nbsp;= np.empty(n, dtype=complex)   
</pre><p> </p><p>Check the other versions and fix them as well if needed.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/939/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

seanlaw, line 1: <p>2^15 seems short. Though, I don't know what is a "good" test. Maybe we should look for other FFT benchmarks to compare against</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/939/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

seanlaw, line 1: <pre class="ql-syntax" spellcheck="false">Line #33.        y = np.empty(n, dtype=np.complex128)
</pre><p>I don't think this needs to be <code>complex128</code> , right? Simply <code>dtype=complex</code> should suffice. <code>complex128</code> might possibly make the computation take longer. Also, I have been wondering if we could make <code>rfft</code> and <code>rifft</code> using the stockham algorithm? I've read that "real" only algos might be twice as fast as their complex counterpart and, similarly, single precision computation is much faster than double precision computation and so I'd expect 128-bit computation to be much slower than 64-bit (double precision) </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/939/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

NimaSarajpoor, line 1: <p>Add the following lines to the code:</p><pre class="ql-syntax" spellcheck="false">n = 2 ** 15
m_values = range(10, 1000 + 10, 10)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/939/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

NimaSarajpoor: <p>There was no particular rationale behind <code>2^15</code>. I just wanted a time series that can reflect the differences from one version to another to some extent. If I choose a short time series, I cannot see those differences easily.</p><p> </p><p>After selecting a version, we can then compare it with the reference for time series with different lengths, e.g <code>np.power(2, np.arange(10, 21)).</code></p><p>This comparison is provided at the bottom of the notebook.</p><p> </p><p>We may also get some idea from the OTFFT benchmark provided here: <a href="http://wwwa.pikara.ne.jp/okojisan/otfft-en/index.html#singlethread" target="_blank">http://wwwa.pikara.ne.jp/okojisan/otfft-en/index.html#singlethread</a>  </p>

NimaSarajpoor: <blockquote>I don't think this needs to be&nbsp;<code>complex128</code>&nbsp;, right? Simply&nbsp;<code>dtype=complex</code>&nbsp;should suffice.&nbsp;<code>complex128</code>&nbsp;might possibly make the computation take longer.&nbsp;</blockquote><p> </p><p>Cool! I am going to replace it with <code>complex</code> . I tested the following and it works.</p><pre class="ql-syntax" spellcheck="false">y = np.empty(n, dtype=np.complex_)
</pre><p>Also: I will replace other instances of <code>np.complex128</code> with <code>np.complex_</code> </p><p> </p><blockquote>Also, I have been wondering if we could make&nbsp;<code>rfft</code>&nbsp;and&nbsp;<code>rifft</code>&nbsp;using the stockham algorithm?</blockquote><p> </p><p>This should potentially boost the performance! I will try to do some quick review of FFT. I will go through the code of this notebook to see how we can use it to compute <code>rfft</code> only. </p><p> </p>

seanlaw: >  Also, I have been wondering if we could make rfft and rifft using the stockham algorithm? I've read that "real" only algos might be twice as fast as their complex counterpart and, similarly, single precision computation is much faster than double precision computation and so I'd expect 128-bit computation to be much slower than 64-bit (double precision)

@NimaSarajpoor There are some clues in section [2.6.2](https://www.researchgate.net/profile/Christos-Bechlioulis/publication/341270520_FFT_algorithms_are_not_mine_However_I_am_going_to_convince_you_soon_regarding_the_visit_of_RMS_to_our_university_Believe_it_or_not_this_is_me_This_is_us_Univeristy_of_Patras_you_have_chosen_a_quite_wr/links/5fa53ce7299bf10f7328c33b/FFT-algorithms-are-not-mine-However-I-am-going-to-convince-you-soon-regarding-the-visit-of-RMS-to-our-university-Believe-it-or-not-this-is-me-This-is-us-Univeristy-of-Patras-you-have-chosen-a-quite.pdf) (page 21). From what I can (barely) understand, it sounds like when we compute the FFT for something that has complex data type, we are computing the FFT for both the real space as well as the complex space simultaneously. However, if our input is only real valued then spend half of our time computing the complex FFT and then throwing the results away. Instead, there seems to be a way to "trick" it and push half of the real computation into the complex space (i.e., "n-Point Real FFT Using n/2-Point Complex FFT"). I hope that helps...

NimaSarajpoor: @seanlaw 
Thanks for pointing me to that section! I can see it has some algorithms to take advantage of this.

Btw, when the input array is real, both `fft` and `rfft` results in complex array still. 

```
T = [0, 1, 2, 3, 4]

>>> scipy.fft.fft(T)
[(10-0j),
 (-2.5+3.4409548011779334j),
 (-2.5+0.8122992405822659j),
 (-2.5-0.8122992405822659j),
 (-2.5-3.4409548011779334j)]

>>> scipy.fft.rfft(T)
[(10+0j), 
(-2.5+3.4409548011779334j), 
(-2.5+0.8122992405822659j)]
```

The second half of the fft result is just conjugate of the first half in reverse.  So, what `rfft` does is that it just computes the first half. So, we can construct fft based on that. 

Need to take a closer look at our 6-step / 8-step implementation and see how I can modify it to address this :) That should give us some boost in performance as you mentioned.

NimaSarajpoor: **[Update]**
> So, what rfft does is that it just computes the first half.
> there seems to be a way to "trick" it 

Okay...I think [2.6.2](https://www.researchgate.net/profile/Christos-Bechlioulis/publication/341270520_FFT_algorithms_are_not_mine_However_I_am_going_to_convince_you_soon_regarding_the_visit_of_RMS_to_our_university_Believe_it_or_not_this_is_me_This_is_us_Univeristy_of_Patras_you_have_chosen_a_quite_wr/links/5fa53ce7299bf10f7328c33b/FFT-algorithms-are-not-mine-However-I-am-going-to-convince-you-soon-regarding-the-visit-of-RMS-to-our-university-Believe-it-or-not-this-is-me-This-is-us-Univeristy-of-Patras-you-have-chosen-a-quite.pdf) was eye-opening! I now understand the trick! 

Thanks again  @seanlaw for bringing it to my attention!!!

Although I did not go through the `Listing 2.3` algo, I think I understand the proposed approach. I am going to provide a summary of it:

Step 1: Get input `T`, with length `n = 2 ** p`
Step 2: Compute `X`, with length `n // 2`, where `X[i]=T[2*i] + j T(2*i+1)`
Step 3: Compute  the fft of `X` using 6-step / 8-step algo, let's call it `Y`
Step 4: Compute just half of the fft of `T` using `2.20-2.24`
Step 5: Compute the other half by conjugating the first half

Will implement and test it to see if I understand it correctly.

seanlaw: Awesome! I haven't gone through it myself so I'm sure I'll be learning from you :)

NimaSarajpoor:
@seanlaw 
I think the previous notebook got too messy and there were different versions there. I did not remove it (I just kept it for our future reference). Instead, I just pushed a new notebook (see https://github.com/TDAmeritrade/stumpy/pull/939/commits/c2a9b439594c866a1a89e19731e6318488b690c6). My goal is to add functions step by step based on the discussions we have had so far on this PR and #938.  

For now I just `resolved` the previously- provided comments (by you and me) that were regarding the initial notebook. I will consider them in this new notebook though.

docs/OTFFT.ipynb
```diff
@@ -0,0 +1,178 @@
+{
```
NimaSarajpoor, line 1: <pre class="ql-syntax" spellcheck="false">Line #20.        x[:] = scipy.fft.fft(x)  # we will implement our fft shortly!
</pre><p>We will later work on implementing this <code>fft</code> function using 6-step / 8-step algorithm. We then change this function accordingly.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/TDAmeritrade/stumpy/pull/939/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

NimaSarajpoor:
I believe this PR is not needed anymore. See:

* https://github.com/stumpy-dev/stumpy/blob/main/stumpy/sdp.py
* https://github.com/stumpy-dev/sliding_dot_product
* https://github.com/stumpy-dev/stumpy/pull/1118#issuecomment-3854039523

PR #960, opened 2024-02-15
Fix typos

EitanHemed:
Fixed multiple occurrences of `Non` instead of `None` in doc-strings.


seanlaw:
Oops! Thanks for taking the time to help us improve the docs @EitanHemed! Contributions like these are super important and greatly appreciated.

PR #961, opened 2024-02-16
[WIP] Fix #958 Bump Supported Python Version to 3.12 

### LINKED ISSUE #958: Bump Supported Python Version to 3.12
[Numba now supports Python 3.12 now and STUMPY tests appear to be functioning properly](https://github.com/stumpy-dev/automate/actions/runs/7767351423/job/21300455571). We should consider bumping the maximum Python version to 3.12 in our Github Workflow.

NimaSarajpoor:


NimaSarajpoor:
**[Self note]**
When there is no GitHub Actions activity on this PR page, check the `Actions` tab. There is a failure saying:

```
[Invalid workflow file: .github/workflows/github-actions.yml#L9](https://github.com/TDAmeritrade/stumpy/actions/runs/7942992099/workflow)
error parsing called workflow
".github/workflows/github-actions.yml"
-> "./.github/workflows/reusable-workflow.yml"
: failed to fetch workflow: workflow was not found.
```

The first thing to do is to just check the path carefully, and make sure everything is correct.  As observed, the problem was with the name of the file.

.github/workflows/reusable-workflow.yml
```diff
@@ -0,0 +1,46 @@
...
+             if [ "$RUNNER_OS" == "macOS" ]; then
+                  brew link --force libomp
+             fi
+        shell: bash
+      - name: Show Full Numba Environment
+        run: python -m numba -s
+        shell: bash
+      - name: Run Tests
+        run: ./test.sh ${{ inputs.test-type }}
+        shell: bash
```
NimaSarajpoor, line 46: The two following steps are missing here:

```
- name: Check Coverage Report
        run: coverage report -m --fail-under=100 --skip-covered --omit=setup.py,docstring.py,min.py,stumpy/cache.py
        shell: bash
- name: Upload Coverage Tests Results
  uses: codecov/codecov-action@v4
``` 

So, we can first check If ${{ inputs.test-type }} is `coverage`, and, if yes, then run the steps above.

seanlaw:
@NimaSarajpoor If I understand correctly, you are refactoring the original workflow so that the common parts can be "imported". I have a few questions/comments:

1. Does it need to be split out into a separate file? Or is it possible to give it a name in the original file and then "use" it? Essentially, it is acting like a callable function from what I can tell?
2. Are we doing too much in this PR? What I had in mind for this issue was only to allow 3.12 to be tested. This is much simpler. It feels like a refactor of the workflow should be its own issue.

NimaSarajpoor:
> Does it need to be split out into a separate file? Or is it possible to give it a name in the original file and then "use" it? Essentially, it is acting like a callable function from what I can tell?

Good question! As you said, it is currently like a callable function, and using it in the main workflow is basically like copying its content wherever it is being called.
see: https://docs.github.com/en/actions/using-workflows/reusing-workflows

> I'm not convinced that having two separate files is making it easier to manage as there is not another workflow that needs to use the reusable workflow (unless you have plans for something in the future?). Analogously, I was forced to create a core.py when I noticed that multiple different modules needed to access the same set of functions OR we were having circular (import) dependencies. If that were not the case, I would've not created a core.py.

Very good point! I will do some digging to see how I can avoid creating an extra `yml` file, and just use it within our original `yml` file.


> Are we doing too much in this PR? What I had in mind for this issue was only to allow 3.12 to be tested. This is much simpler. While I like/appreciate the idea behind refactoring, it feels like a refactor of the workflow should be tracked as its own issue/PR so that it will be easier to reference in the future.

I totally agree with you! I think it should have its own PR. I just noticed the label `refactor` in #958 and thought what kind of `refactoring` I can do here 😅

seanlaw:
Thanks @NimaSarajpoor I think everything is ready. Shall I merge?

NimaSarajpoor:
@seanlaw 
Sure. All good.

PR #983, opened 2024-06-08
[WIP] Fix #982: Ostinato Overwrites Original Time Series

### LINKED ISSUE #982: Ostinato Overwrites Original Time Series
There appears to be a bug in the `ostinato` (and variants) algorithm where the list of time series being passed into the function are also (possibly, in some rare cases) being incorrectly overwritten after they have been `core.preprocessed`:

https://github.com/TDAmeritrade/stumpy/blob/9edd7f50e3e59f7212ca2ce399be075c01872960/stumpy/ostinato.py#L372-L375

This isn't a problem when there are no `np.nan`/`np.inf` in the data since the time series that gets passed back will be identical to the input time series. However, in the rare case, when there are `np.nan`/`np.inf` in the time series, those parts of the time series get preprocessed into `0.0` and then this converted time series overwrites the original element within the input list of time series `Ts`. 

What we should do is:

1. make a copy of the input list, `Ts`, and also each time series, `Ts[i]`, so that the original `Ts`/`Ts[i]` is NOT overwritten
2. Add a unit test to check that the input is not overwritten in `Ts` (especially when there are `np.nan`/`np.inf` in the data

Something like:

```
Ts_copy = []
for T in Ts:
    Ts_copy.append(T.copy())
```
and then use `Ts_copy` everywhere and overwrite where necessary

NimaSarajpoor:
See #982


tests/test_ostinato.py
```diff
@@ -137,3 +137,33 @@ def test_deterministic_ostinatoed_with_isconstant(seed, dask_cluster):
...
+    # without nan values
+    Ts_ref = Ts.copy()
+    Ts_comp = Ts.copy()
+    stumpy.ostinato(Ts_comp, m)
+    for i in range(len(Ts)):
+        npt.assert_almost_equal(Ts_ref[i], Ts_comp[i])
+
+    # with nan values
+    nan_size = [np.random.choice(np.arange(1, len(T) + 1)) for T in Ts]
+    for i in range(len(Ts_ref)):
```
NimaSarajpoor, line 159:
```suggestion
    for i in range(len(Ts)):
```

seanlaw (commented):
Looks good. Please add unit tests where you:

1. Run `ostinato` on the list of `Ts`
2. Find the consensus motif
3. Use `mass` to find the location of the motif in each `Ts[i]`
4. Set that subsequence in `Ts[i]` to `np.nan`
5. Run `ostinato` again on this modified list of time series and repeat above.

A telltale sign of something going wrong is if you get a `divide by zero` warning when computing one of the inverse means like `M_Ts`.

seanlaw:
@NimaSarajpoor Do we need to do the same thing for `aamp_ostinato` functions too?

NimaSarajpoor:
@seanlaw 

> Looks good. Please add unit tests where you:
> 
> 1. Run `ostinato` on the list of `Ts`
> 2. Find the consensus motif
> 3. Use `mass` to find the location of the motif in each `Ts[i]`
> 4. Set that subsequence in `Ts[i]` to `np.nan`
> 5. Run `ostinato` again on this modified list of time series and repeat above.
> 
> A telltale sign of something going wrong is if you get a `divide by zero` warning when computing one of the inverse means like `M_Ts`.

Can you please take a look at the test function ` test_extract_second_consensus` and let me know if that is what you have in mind?



> @NimaSarajpoor Do we need to do the same thing for `aamp_ostinato` functions too?

Good point! I think so. I am going to work on it after finalizing our thoughts for the test functions that have been recently added to `ostinato`.

seanlaw:
> Can you please take a look at the test function  test_extract_second_consensus and let me know if that is what you have in mind?

Yes, I will try to look at it tonight

tests/test_ostinato.py
```diff
@@ -137,3 +137,83 @@ def test_deterministic_ostinatoed_with_isconstant(seed, dask_cluster):
...
+    Ts_ref[central_Ts_idx][
+        central_subseq_idx : central_subseq_idx + m
+    ] = np.nan  # apply same changes to Ts_ref
+    for i in range(len(Ts)):
+        if i != central_Ts_idx:
+            D = stumpy.core.mass(consensus_motif, Ts_comp[i])
+            idx = np.argmin(D)
+            Ts_comp[i][idx : idx + m] = np.nan
+            Ts_ref[i][idx : idx + m] = np.nan  # apply same changes to Ts_ref
+
```
seanlaw, line 205: So here, please assert that the original `Ts` is unchanged (i.e., it doesn't contain any `np.nan` or runs of `zero` that don't belong.

I'm not sure that the inserting of single (non-consecutive) `np.nan` above is necessary or what the goal is there?

seanlaw (commented):
Looks good for the most part

tests/test_ostinato.py
```diff
@@ -137,3 +137,83 @@ def test_deterministic_ostinatoed_with_isconstant(seed, dask_cluster):
...
+@pytest.mark.parametrize(
+    "seed", np.random.choice(np.arange(10000), size=25, replace=False)
+)
+def test_input_not_overwritten(seed):
+    m = 50
+    np.random.seed(seed)
+    Ts = [np.random.rand(n) for n in [64, 128, 256]]
+
+    # without nan values
```
NimaSarajpoor, line 150: @seanlaw 
Do we need to check this scenario, i.e. having no `nan` value in the original input ?

tests/test_ostinato.py
NimaSarajpoor: I was wondering if I should do the assertion right after passing `Ts_comp` to `ostinato`. I mean:

```
central_radius, central_Ts_idx, central_subseq_idx = stumpy.ostinato(Ts_comp, m)
for i in range(len(Ts)):
    ref = Ts[i]
    comp = Ts_comp[i]
    npt.assert_almost_equal(ref[np.isfinite(ref)], comp[np.isfinite(comp)])
```

Right? 
This way I can check if the original input `Ts_comp` remains untouched after calling `ostinato`. Particularly, to see if the fix works, I need to have at least one `np.nan` in the original data (so that it can be replaced with `0.0` in the local variable in the called function `ostinato` , and then we can check to see if that changes the original data or not ).  Btw, I checked this particular scenario in `test_input_not_overwritten`

---

To extract the second consensus, I DO CHANGE the  input `Ts_comp` here (by setting the consensus and its NN motifs to `np.nan`) before passing it to `ostinato` for the second time.

tests/test_ostinato.py
seanlaw: I had no intentions to. Is it necessary?

tests/test_ostinato.py
seanlaw: Okay, I think this is fine then but please document the different/various intentions a little more clearly if possible

If it helps, it is important that the reason for each unit test is clear and, if it fails, it should be clear WHY it failed. We should avoid testing multiple different/unrelated things within the same unit test

NimaSarajpoor: @seanlaw 
Right! I think things got mixed a bit there in my tests. Will add supportive comments and remove unnecessary tests.

tests/test_ostinato.py
NimaSarajpoor: Not really. Initially I thought `ostinato` may change some values of the input time series to `nan` as it searches through data to find the consensus. I took a look and noticed that does not happen in the code (and in fact, you already mentioned that in #982).  I am going to remove it.

NimaSarajpoor:
@seanlaw 
After you had pointed out that the intention behind a test function should be clearer, I started thinking about what exactly we would like to test here, and I came to the following conclusion: 
At the end of the day, I want to know if the original array is remained unchanged after calling `ostinato`. 


So, I was wondering if 

> A telltale sign of something going wrong is if you get a divide by zero warning when computing one of the inverse means like M_Ts.

is just a special case of the test function 

```
def test_input_not_overwritten(seed):
    # ostinato preprocesses its input, a list of time series,
    # by replacing nan value with 0 in each time series
    # This test ensures that the original input is not overwritten
    m = 50
    np.random.seed(seed)
    Ts = [np.random.rand(n) for n in [64, 128, 256]]

    # insert at least one nan value
    nan_size = [np.random.choice(np.arange(1, len(T) + 1)) for T in Ts]
    for i in range(len(Ts)):
        IDX = np.random.choice(np.arange(len(Ts[i])), size=nan_size[i], replace=False)
        Ts[i][IDX] = np.nan

    # raise error if ostinato overwrite its input
    Ts_input = [T.copy() for T in Ts]
    stumpy.ostinato(Ts_input, m)
    for i in range(len(Ts)):
        T_ref = Ts[i]
        T_comp = Ts_input[i]
        npt.assert_almost_equal(T_ref[np.isfinite(T_ref)], T_comp[np.isfinite(T_comp)])
```
?

seanlaw:
> is just a special case of the test function

So, I am told that in test-driven-development we should do "red, green, refactor", which means that we should:

1. Red — think about what you want to develop
2. Green — think about how to make your tests pass
3. Refactor — think about how to improve your existing implementation

In other words:

1. Write a test that fails (without changing our code base) - so, take a user example that causes a failure
2. Consider what might make the test pass (without causing other existing tests to fail AND without changing existing tests)
3. Implement the new code 

In our case, the example that causes failure when using the original code is:

```
import pandas as pd
import stumpy
import numpy as np

Ts = []
for i in [6, 7, 9, 10, 16, 24]:
    Ts.append(pd.read_csv(f'https://zenodo.org/record/4288978/files/EOG_001_01_{i:03d}.csv?download=1').iloc[:, 0].values)

m = 50
radius, Ts_idx, subseq_idx = stumpy.ostinato(Ts, m)

consensus_indices = []
for k in range(5):
    consensus_motif = Ts[Ts_idx][subseq_idx : subseq_idx + m].copy()
    radius, Ts_idx, subseq_idx = stumpy.ostinato(Ts, m)
    nn_indices = []
    for i, T in enumerate(Ts):
        nn_idx = np.argmin(stumpy.core.mass(consensus_motif, T))
        nn_indices.append(nn_idx)
        # Remove matching motifs from T
        T[nn_idx : nn_idx + m] = np.nan
    consensus_indices.append(nn_idx)
```

This is basically the Ostinato tutorial but ran multiple times after clearing out the consensus motif. Currently, this gives me the following warnings:

```
[miniconda3/lib/python3.12/site-packages/stumpy/core.py:2243](miniconda3/lib/python3.12/site-packages/stumpy/core.py#line=2242): RuntimeWarning: divide by zero encountered in divide
  Σ_T_inverse = 1.0 [/](http://localhost:8888/) Σ_T
[miniconda3/lib/python3.12/site-packages/stumpy/core.py:3453](miniconda3/lib/python3.12/site-packages/stumpy/core.py#line=3452): UserWarning: A large number of values in `P` are smaller than 1e-06.
For a self-join, try setting `ignore_trivial=True`.
  warnings.warn(msg)
```

and the `consensus_indices` output is absolutely WRONG. So, the goal is a to write one or more unit tests that causes this same behavior and then refactor our code to fix it.

NimaSarajpoor:
> So, the goal is a to write one or more unit tests that causes this same behavior and then refactor our code to fix it.

My first two commits did "Red-Green-Refactor" but for the root cause though. My goal was to expose the issue with the new test provided in the [first commit](https://github.com/TDAmeritrade/stumpy/pull/983/commits/c60d38ea5d7f9a675b55773d5513f86c81439bc9) (Note: I just removed the `without nan values` part in my latest commit and some minor changes). I successfully got [the error](https://github.com/TDAmeritrade/stumpy/actions/runs/9428082592/job/25972981280) after pushing my first commit. The fix in [the second commit](https://github.com/TDAmeritrade/stumpy/pull/983/commits/5e985619e40cc47ee1098e39c5f4a0ef7559bbeb) addressed the exposed issue.

So, If I understand correctly, what we are trying to do here is to be extra careful and consider its side-effect as well and add a test function just for that. The new test function does not raise an error though if the already-implemented fix works for that too. But, I can see the point: If the fix does not work, then I know I need to do something else as well.

seanlaw:
> So, If I understand correctly, what we are trying to do here is to be extra careful and consider the side-effect of the issue as well and add a test function just for that. The new test function does not raise an error though if the already-implemented fix works for that too. 

Right. Sometimes (like in this case), the side effect of:

> RuntimeWarning: divide by zero encountered in divide  Σ_T_inverse = 1.0 [/](http://localhost:8888/) Σ_T

leads us to the real root cause (i.e., we were overwriting our original data). I think it is fine if the user's original data actually contained a bunch of constant regions but it shouldn't be caused by our code and this is ultimately the reason why the above code fails so we should prevent it from breaking. In this case, it's tricky because it's not actually throwing any error, it's only warning us :(

tests/test_ostinato.py
```diff
@@ -137,3 +137,59 @@ def test_deterministic_ostinatoed_with_isconstant(seed, dask_cluster):
         npt.assert_almost_equal(ref_radius, comp_radius)
         npt.assert_almost_equal(ref_Ts_idx, comp_Ts_idx)
         npt.assert_almost_equal(ref_subseq_idx, comp_subseq_idx)
+
+
+@pytest.mark.parametrize(
+    "seed", np.random.choice(np.arange(10000), size=25, replace=False)
```
seanlaw, line 143: Is there a particular reason why we need to set a random seed?

tests/test_ostinato.py
```diff
@@ -137,3 +137,59 @@ def test_deterministic_ostinatoed_with_isconstant(seed, dask_cluster):
...
+    for i in range(len(Ts)):
+        T_ref = Ts[i]
+        T_comp = Ts_input[i]
+        npt.assert_almost_equal(T_ref[np.isfinite(T_ref)], T_comp[np.isfinite(T_comp)])
+
+
+@pytest.mark.parametrize(
+    "seed", np.random.choice(np.arange(10000), size=25, replace=False)
+)
+def test_extract_seversl_consensus(seed):
```
seanlaw, line 171: Typo: `several`

seanlaw (changes requested):
I think this looks reasonable. Do you think it's ready to merge after these changes?

tests/test_ostinato.py
NimaSarajpoor: Good point! 
I don't think so. In fact, I now think it might be better to set the seed to a certain value so that we can remove this complexity from the test function. We can better find the issue if it raises an error in the future.

tests/test_ostinato.py
```diff
@@ -186,9 +186,15 @@ def test_extract_seversl_consensus(seed):
         radius, Ts_idx, subseq_idx = stumpy.ostinato(Ts_comp, m)
         consensus_motif = Ts_comp[Ts_idx][subseq_idx : subseq_idx + m].copy()
         for i in range(len(Ts_comp)):
-            nn_idx = np.argmin(stumpy.core.mass(consensus_motif, Ts_comp[i]))
-            Ts_comp[i][nn_idx : nn_idx + m] = np.nan
-            Ts_ref[i][nn_idx : nn_idx + m] = np.nan
+            if i == Ts_idx:
+                # The consensus motif is in the same time series
+                idx = Ts_idx
```
NimaSarajpoor, line 191: @seanlaw 
I am thinking of a situation where we might have several 0 in the distance profile (the output of `mass`). In that case, the value returned by `np.argmin(...)` may not be `Ts_idx`. That's why I created this if-else block.

NimaSarajpoor:
@seanlaw 

> Do you think it's ready to merge after these changes?

Not yet. Need to add similar test functions to `aamp_ostinato`.

tests/test_ostinato.py
```diff
@@ -137,3 +137,65 @@ def test_deterministic_ostinatoed_with_isconstant(seed, dask_cluster):
...
+    Ts = [np.random.rand(n) for n in [256, 512, 1024]]
+    Ts_ref = [T.copy() for T in Ts]
+    Ts_comp = [T.copy() for T in Ts]
+
+    m = 20
+
+    k = 5  # Get the first `k` consensus motifs
+    for _ in range(k):
+        # Find consensus motif and its NN in each time series in Ts_comp
+        # Rmove them from Ts_comp as well as Ts_ref, and assert that the
```
NimaSarajpoor, line 184: Typo: `Remove`

tests/test_ostinato.py
seanlaw: I don't think we need to touch/set the random seed at all.

NimaSarajpoor: @seanlaw 
Is this because of the purpose of this test function? (that we do not need to test the outputs for different inputs but just to see if the input remains unchanged?

seanlaw: Yes, that's correct. It's fine to add an `np.nan` in the original input but what we are testing is if the input, `Ts` remains completely unchanged after it is passed through `ostinato`

stumpy/ostinato.py
```diff
@@ -485,6 +486,7 @@ def ostinatoed(client, Ts, m, normalize=True, p=2.0, Ts_subseq_isconstant=None):
     if not isinstance(Ts, list):  # pragma: no cover
         raise ValueError(f"`Ts` is of type `{type(Ts)}` but a `list` is expected")
 
+    Ts = [T.copy() for T in Ts]
```
NimaSarajpoor, line 489: Since we are copying arrays here, should we do `copy=False` in the for-loop below?

```
for i, T in enumerate(Ts):
        Ts[i], M_Ts[i], Σ_Ts[i], Ts_subseq_isconstant[i] = core.preprocess(
            T, m, copy=False, T_subseq_isconstant=Ts_subseq_isconstant[i]
        )
```

I tested it, and it works.

----

Btw, the default value for `copy` is `True` in the `core.preprocess`. I am curious to know why `Ts[i], ... = core.preprocess(T, ....)`  did not work, and resulted in overwriting the array

NimaSarajpoor (commented):
@seanlaw 
I just added one comment.

Also: 
Do you think we should write tests for `ostinatoed`, `gpu_ostinato`, and their corresponding non-normalized functions?

stumpy/ostinato.py
seanlaw: > Btw, the default value for copy is True in the core.preprocess. I am curious to know why Ts[i], ... = core.preprocess(T, ....) did not work, and resulted in overwriting the array

Hopefully, I am understanding your question correctly but I don't think it has to do with the `numpy` `array` level but instead with the `python` `list` level. Let's say we remove `numpy` out of the equation:

```
def some_func(Ts):
    for i in range(len(Ts)):
        Ts[i] = i + 10

if __name__ == '__main__':
    Ts = [1, 2, 3, 4]
    some_func(Ts)
    print(Ts)
    # 10 11 12 13
```

Since `python` is always "pass-by-reference", a `python` `list` is NOT a copy when passed into a function and, instead, you have direct access to the original list, which you are then overwriting the contents of. In the case of `core.preprocess`, we are filling/overwriting `Ts[i]` with a copy of a preprocessed array. So, to state it more clearly, I am guessing that the `core.preprocess` makes a copy of the array `T == Ts[i]`, processes it, and replaces the `i`th element in `Ts` with this new preprocessed data. I think we should leave `copy=True` (default) here.

seanlaw: > should we do copy=False in the for-loop below?

Why? Is it purely to avoid a double `copy`? I think we should leave copy=True (default) here.

NimaSarajpoor: > > Btw, the default value for copy is True in the core.preprocess. I am curious to know why Ts[i], ... = core.preprocess(T, ....) did not work, and resulted in overwriting the array
> 
> Hopefully, I am understanding your question correctly but I don't think it has to do with the `numpy` `array` level but instead with the `python` `list` level. Let's say we remove `numpy` out of the equation:
> 
> ```
> def some_func(Ts):
>     for i in range(len(Ts)):
>         Ts[i] = i + 10
> 
> if __name__ == '__main__':
>     Ts = [1, 2, 3, 4]
>     some_func(Ts)
>     print(Ts)
>     # 10 11 12 13
> ```
> 
> Since `python` is always "pass-by-reference", a `python` `list` is NOT a copy when passed into a function and, instead, you have direct access to the original list, which you are then overwriting the contents of.

You understood it correctly indeed. The example and its following explanation were very helpful! 👍  Thanks!

NimaSarajpoor: @seanlaw 

> > should we do copy=False in the for-loop below?
> 
> Why? Is it purely to avoid a double copy? 

Yes. 

> I think we should leave copy=True (default) here. Otherwise, I would make a note above it just to say something like:
> `# We've already made a copy above`

I might be wrong but I feel your concern is coming from our future selves. I think you want to be on the safer side and leave it as `copy=True` (particularly if its contribution to the running time is negligible compared to the other parts of the code). This makes sense! I can see that this is a little price we agree to pay to keep ourselves in a better position in the long term.   So, I will go with your initial suggestion: 
> I think we should leave copy=True (default) here.

seanlaw: If anything, you can add a comment to say "setting `copy=False` is also fine here if performance becomes an issue"

seanlaw: @NimaSarajpoor I think I still prefer this:

```
# Avoid overwriting `Ts` during preprocessing
Ts_copy = [None] * len(Ts)
.
.
.
for i, T in enumerate(Ts):
        Ts_copy[i], M_Ts[i], Σ_Ts[i], Ts_subseq_isconstant[i] = core.preprocess(
            T, m, copy=True, T_subseq_isconstant=Ts_subseq_isconstant[i]
        )
```

In fact, if you go back to the original issue #983, my initial instinct was to propose something like:

```
Ts_copy = []
for T in Ts:
    Ts_copy.append(T.copy())
``` 

I am liking this better because:
1. We are being explicit with the copy
2. We gain the (tiny) performance benefit of not doing a double copy while having zero doubt that `Ts_copy` is a "copy" and therefore `copy=False` is fine.

Can we go with using the `Ts_copy` approach, please?

NimaSarajpoor: @seanlaw 
I can understand the  "being the explicit" part. It makes sense. I will make the changes. Thanks for the feedback!  (The only reason I went with `Ts = [T.copy() for T in Ts]` was to avoid making changes in other parts of the code. But, I can see that was not a very good reason here)

stumpy/ostinato.py
```diff
@@ -485,18 +495,22 @@ def ostinatoed(client, Ts, m, normalize=True, p=2.0, Ts_subseq_isconstant=None):
...
 
     if Ts_subseq_isconstant is None:
         Ts_subseq_isconstant = [None] * len(Ts)
     for i, T in enumerate(Ts):
-        Ts[i], M_Ts[i], Σ_Ts[i], Ts_subseq_isconstant[i] = core.preprocess(
-            T, m, T_subseq_isconstant=Ts_subseq_isconstant[i]
+        Ts_copy[i], M_Ts[i], Σ_Ts[i], Ts_subseq_isconstant[i] = core.preprocess(
+            T, m, copy=True, T_subseq_isconstant=Ts_subseq_isconstant[i]
         )
+        # Setting `copy=False` is also fine here if performance becomes an issue
```
seanlaw, line 509: This comment should be removed since we never make a copy of the original data

seanlaw (changes requested):

tests/test_aamp_ostinato.py
```diff
@@ -85,3 +85,60 @@ def test_deterministic_ostinatoed(seed, dask_cluster):
             npt.assert_almost_equal(ref_radius, comp_radius)
             npt.assert_almost_equal(ref_Ts_idx, comp_Ts_idx)
             npt.assert_almost_equal(ref_subseq_idx, comp_subseq_idx)
+
+
+def test_input_not_overwritten():
+    # ostinato preprocesses its input, a list of time series,
```
NimaSarajpoor, line 91:
```suggestion
    # aamp_ostinato preprocesses its input, a list of time series,
```

tests/test_aamp_ostinato.py
```diff
@@ -85,3 +85,60 @@ def test_deterministic_ostinatoed(seed, dask_cluster):
...
+    m = 50
+    Ts = [np.random.rand(n) for n in [64, 128, 256]]
+
+    # insert at least one nan value
+    nan_size = [np.random.choice(np.arange(1, len(T) + 1)) for T in Ts]
+    for i in range(len(Ts)):
+        IDX = np.random.choice(np.arange(len(Ts[i])), size=nan_size[i], replace=False)
+        Ts[i][IDX] = np.nan
+
+    # raise error if ostinato overwrite its input
```
NimaSarajpoor, line 103:
```suggestion
    # raise error if aamp_ostinato overwrite its input
```

tests/test_aamp_ostinato.py
```diff
@@ -85,3 +85,60 @@ def test_deterministic_ostinatoed(seed, dask_cluster):
...
+    Ts_input = [T.copy() for T in Ts]
+    stumpy.aamp_ostinato(Ts_input, m)
+    for i in range(len(Ts)):
+        T_ref = Ts[i]
+        T_comp = Ts_input[i]
+        npt.assert_almost_equal(T_ref[np.isfinite(T_ref)], T_comp[np.isfinite(T_comp)])
+
+
+def test_extract_several_consensus():
+    # This test is to further ensure that the function `ostinato`
```
NimaSarajpoor, line 113:
```suggestion
    # This test is to further ensure that the function `aamp_ostinato`
```

tests/test_gpu_aamp_ostinato.py
```diff
@@ -53,3 +53,60 @@ def test_deterministic_gpu_aamp_ostinato(seed):
...
+    m = 50
+    Ts = [np.random.rand(n) for n in [64, 128, 256]]
+
+    # insert at least one nan value
+    nan_size = [np.random.choice(np.arange(1, len(T) + 1)) for T in Ts]
+    for i in range(len(Ts)):
+        IDX = np.random.choice(np.arange(len(Ts[i])), size=nan_size[i], replace=False)
+        Ts[i][IDX] = np.nan
+
+    # raise error if gpu_ostinato overwrite its input
```
NimaSarajpoor, line 73:
```suggestion
    # raise error if gpu_aamp_ostinato overwrite its input
```

tests/test_gpu_aamp_ostinato.py
```diff
@@ -53,3 +53,60 @@ def test_deterministic_gpu_aamp_ostinato(seed):
...
+        # two time series are the same
+        radius, Ts_idx, subseq_idx = gpu_aamp_ostinato(Ts_comp, m)
+        consensus_motif = Ts_comp[Ts_idx][subseq_idx : subseq_idx + m].copy()
+        for i in range(len(Ts_comp)):
+            if i == Ts_idx:
+                query_idx = subseq_idx
+            else:
+                query_idx = None
+
+            idx = np.argmin(core.mass(consensus_motif, Ts_comp[i], query_idx=query_idx))
```
NimaSarajpoor, line 106:
```suggestion
            idx = np.argmin(core.mass(consensus_motif, Ts_comp[i], normalize=False, query_idx=query_idx))
```

NimaSarajpoor:
@seanlaw 
I did some minor touches. Do you think this PR is in good shape if all tests pass?

stumpy/gpu_aamp_ostinato.py
```diff
@@ -69,15 +69,17 @@ def gpu_aamp_ostinato(Ts, m, device_id=0, p=2.0):
     if not isinstance(Ts, list):  # pragma: no cover
         raise ValueError(f"`Ts` is of type `{type(Ts)}` but a `list` is expected")
 
+    # Avoid overwriting `Ts` during preprocessing
```
seanlaw, line 72: I don't think this comment is necessary since `Ts_copy` is pretty self-explanatory

stumpy/gpu_ostinato.py
```diff
@@ -108,23 +108,38 @@ def gpu_ostinato(Ts, m, device_id=0, normalize=True, p=2.0, Ts_subseq_isconstant
     if Ts_subseq_isconstant is None:
         Ts_subseq_isconstant = [None] * len(Ts)
 
+    # Avoid overwriting `Ts` during preprocessing
```
seanlaw, line 111: Please remove comment

stumpy/ostinato.py
```diff
@@ -365,25 +365,35 @@ def ostinato(Ts, m, normalize=True, p=2.0, Ts_subseq_isconstant=None):
     if not isinstance(Ts, list):  # pragma: no cover
         raise ValueError(f"`Ts` is of type `{type(Ts)}` but a `list` is expected")
 
+    # Avoid overwriting `Ts` during preprocessing
```
seanlaw, line 368: Please remove comment

stumpy/ostinato.py
```diff
@@ -485,18 +495,20 @@ def ostinatoed(client, Ts, m, normalize=True, p=2.0, Ts_subseq_isconstant=None):
...
```
seanlaw, line 498: Please remove comment

tests/test_aamp_ostinato.py
```diff
@@ -85,3 +85,60 @@ def test_deterministic_ostinatoed(seed, dask_cluster):
...
+
+def test_input_not_overwritten():
+    # aamp_ostinato preprocesses its input, a list of time series,
+    # by replacing nan value with 0 in each time series.
+    # This test ensures that the original input is not overwritten
+    m = 50
+    Ts = [np.random.rand(n) for n in [64, 128, 256]]
+
+    # insert at least one nan value
+    nan_size = [np.random.choice(np.arange(1, len(T) + 1)) for T in Ts]
```
seanlaw, line 98: If you only need a single random integer for each time series then why not just do:

```
nan_size = [ np.random.randint(1, len(T) + 1) for T in Ts]
```

It doesn't make sense to generate a range of numbers just to extract a single value from it

seanlaw: In fact, why should we allow so many elements to be `np.nan`? I'd limit it to as many as 10% of `len(T)` only

seanlaw (changes requested):

seanlaw:
> I did some minor touches. Do you think this PR is in good shape if all tests pass?

Yes, I think everything looks reasonable. I've left some suggestions for you to consider

stumpy/gpu_aamp_ostinato.py
NimaSarajpoor: Will remove it

tests/test_aamp_ostinato.py
NimaSarajpoor: > It doesn't make sense to generate a range of numbers just to extract a single value from it

Right! I do not know why my mind went to that complicated direction! I will revise it.

> In fact, why should we allow so many elements to be np.nan? I'd limit it to as many as 10% of len(T) only

Well, I think just a single `np.nan` should suffice actually. But then I thought why I should limit it to one element. So, I made it completely random.  I am just trying to understand "how" I should decide whether I should go with one random element, 10% of `len(T)` elements, or just any number of elements?

seanlaw: Honestly, I would've gone with no np.nan but if you mist then maybe only do one. It's okay to hardcode the location too

NimaSarajpoor: > I would've gone with no np.nan

I think we need at least one `np.nan` (in at least one of the time series in `Ts`) for this test function, as explained in [this comment](https://github.com/TDAmeritrade/stumpy/issues/982#issue-2337118449).  If there is no `np.nan`, then the following assertion always pass as the preprocessed data and the original data are the same.

```
npt.assert_almost_equal(T_ref[np.isfinite(T_ref)], T_comp[np.isfinite(T_comp)])
```

I will go with one `np.nan` and, I will hardcode the location (For the sake of simplicity, I can put it as the first element of each time series)

---

Note that in the other test function `test_extract_several_consensus`, there is no `np.nan` in the original input because `np.nan` will be created as we extract several consensus motifs.

NimaSarajpoor:
@seanlaw 
I addressed the comments. I also added two more test functions to `tests/test_aamp_ostinato.py` for `aamp_ostinatoed`

seanlaw (approved):
@NimaSarajpoor Everything looks good. Let me know if I can merge

NimaSarajpoor:
@seanlaw 

> @NimaSarajpoor Everything looks good. Let me know if I can merge

All good from my side. It is ready to be merged.

seanlaw:
@NimaSarajpoor Thank you for this PR!

PR #988, opened 2024-06-12
Fixed #973 updated min.py script with additional checks to scan files in the Stumpy repo that list package versions and confirm those versions align with the minimum supported versions

### LINKED ISSUE #973: Add additional checks to `min.py`
Currently, `min.py` finds and returns the minimum `python`, `numba`, `numpy`, and `scipy` versions that are compatible with the minimum supported `python` (i.e., NOT end-of-life). This is super handy but we can/should take this a step further and also have this script check the necessary files (i.e., `pyproject.toml`, `requirements.txt`, `environment.yml`, `.github/worflows/github-actions.yml`, and `README.md`) to detect whether those files have mismatches or need to be updated as well. 

Note that we don't necessarily want to change these files by default but, instead, be able to first CHECK and notify the user of all of the mismatches. Then possibly add an `argparse` flag like `-force` or something that will automatically make the changes to those files.

joehiggi1758:
See #973 

# Pull Request Checklist

- [x] Create a new branch


joehiggi1758:
@seanlaw as a heads up I'm trying to figure out what went wrong here locally to where none of my tests are passing on the PR! I'll be pushing again once I figure this out! 

Apologies for any confusion!

pypi.sh
```diff
@@ -5,11 +5,12 @@
 # 3. Update README with new features/functions/tutorials
 # 4. Determine minimum versions and dependencies with ./min.py
 # 5. Bump minimum versions and dependencies
-#    a) setup.py
+#    a) pyproject.toml
 #    b) requirements.txt
 #    c) environment.yml
 #    d) .github/worflows/github-actions.yml
 #    e) recipes/meta.yaml in conda-feedstock
+#    f) README.md
```
seanlaw, line 13: So, the latest version of `pypi.sh` already contains the updated lines:

https://github.com/TDAmeritrade/stumpy/blob/887feb8b8b6800e2e09d8f00ea2023caadad0427/pypi.sh#L8-L13

It is rather curious that your code base is now adding this line back in. This leads me to believe that your current development branch does not contain the latest updates.

min.py
```diff
@@ -135,6 +136,73 @@ def get_min_scipy_version(min_python, min_numpy):
...
+
+    Returns:
+        dict: A dictionary containing package names
+              as keys and lists of tuples as values.
+              Each tuple contains file path, line number,
+              and the outdated version.
+
+    """
+
+    def parse_version(version):
```
seanlaw, line 158: In general, I would prefer not to use a "closure" and, instead move this out to its own standalone function (rather than trapped inside another function). However, given how short the function is, it doesn't make sense to make this a function

min.py
```diff
@@ -146,9 +214,32 @@ def get_min_scipy_version(min_python, min_numpy):
...
+
     print(
         f"python: {MIN_PYTHON}\n"
         f"numba: {MIN_NUMBA}\n"
         f"numpy: {MIN_NUMPY}\n"
         f"scipy: {MIN_SCIPY}"
     )
+
+    # Instantiating a list of files to scan for minimum version mismatches
+    files_to_scan = [
```
seanlaw, line 226: Change `files_to_scan` to `fnames` (i.e., "file names") and then you would eventually do something like `for fname in fnames` (note the singular vs plural)

min.py
```diff
@@ -135,6 +136,73 @@ def get_min_scipy_version(min_python, min_numpy):
...
+        version_parts = version.split(".")
+
+        return tuple(map(int, version_parts))
+
+    # Instantiating a dictionary of lists for storing
+    # any hits on minimum version mismatches while scanning
+    matching_files = {package: [] for package in package_versions}
+
+    # Iteratively scanning files input for minimum version mismatches
+    for file_path in files:
```
seanlaw, line 178: Having things so deeply nested is concerning and it likely means that you are trying to do too much inside of this function. I think it might be a bit cleaner to, say, have a function called:

```
def find_pkg_mismatches(fnames, pkg, version):
```

So this function only checks for ONE specific `pkg` and `version` pair amongst the set of `fnames`. Then, you would have a separate `for-loop` that looks something like:

```
for pkg, version in package_versions:
    find_pkg_mismatches(fnames, pkg, version)
```

This way, just by looking at this `for-loop`, you know immediately what is happening and the `find_pkg_mismatch` function has a single responsibility (i.e., it doesn't look for all packages and all versions at once). Compact, "single responsibility" functions like this are usually easier to maintain in the longer term

joehiggi1758: Ahhh yeah that makes sense - I'm on it!

I'll try and break this out further to handle single responsibilities!

seanlaw: Thank you for being open to feedback!

joehiggi1758: @seanlaw Of course! I truly appreciate the direction, I can't grow without it!

As an FYI I'm hoping to have the new PR by end of week!

joehiggi1758: @seanlaw hey Sean - hope you're having a great one! 

Two clarifying questions...

- Would a dataframe work for the data structure to return regarding all mismatches noted? (or would it be most optimal to use a list or dictionary)
- Are there any opinions from the Stumpy team about default variables with functions? (the way I re-wrote the above seems to require me to have some default variables - but I'd like to stay in line with the Stumpy preferred approach)

seanlaw: > Would a dataframe work for the data structure to return regarding all mismatches noted? (or would it be most optimal to use a list or dictionary)

If it's not too painful, I'd like to use `stdlib` if possible but feel free to show your use of `pandas` and we can discuss as `pandas` is already imported. Note that `pandas` is technically NOT a `stumpy` dependency and that it is only there for development purposes only.

> Are there any opinions from the Stumpy team about default variables with functions? (the way I re-wrote the above seems to require me to have some default variables - but I'd like to stay in line with the Stumpy preferred approach)

I'm not sure I understand. I don't see any files/code committed in the current/latest commit

seanlaw (changes requested):

seanlaw:
@joehiggi1758 Can you please tell me all of the difference flavors of versioning styles that you've encountered in the files? I think there are ways for us to simplify things

joehiggi1758:
@seanlaw 

Yes, more than happy to share! 
-  The first regex (f"{pkg}-version: [") is searching for the format "python-version: ['3.8']" for example, line 13 in github-actions.yml
- The second regex (rf"({pkg}|{pkg})\s*==?\s*['\"]?(\d+\.\d+(\.\d+)?)['\"]?") is searching for the format "numba>=0.55.2" for example, line 8 in environment.yml
- The third regex (rf"({pkg})\s*>=\s*['\"]?(\d+\.\d+(\.\d+)?)['\"]?") is searching for the format "python>=3.8" for example, line 5 in environment.yml
- The fourth regex (rf"::\s*{pkg}\s*::\s*['\"]?(\d+\.\d+(\.\d+)?)['\"]?") is searching for the format "Programming Language :: Python :: 3.8" for example, line 26 in pyproject.toml
- The fifth regex (rf"{pkg}\s+[a-z]+\s+(\d+\.\d+(\.\d+)?)") is searching for the format "Programming Language :: Python :: 3.8" for example, line 26 in pyproject.toml
- The sixth format is "STUMPY supports `Python...." for example, line 328 in README.rst

seanlaw:
@joehiggi1758 If we change the README to `python>=3.8`, then do you think (building off of your work) something like the following minimum-reproducible-example might suffice?

```
import re

fnames = [
    "pyproject.toml",
    "requirements.txt",
    "environment.yml",
    ".github/workflows/github-actions.yml",
    "README.rst",
]

MIN_PYTHON = "2.7"
MIN_NUMBA = "0.1"
MIN_NUMPY = "0.1"
MIN_SCIPY = "0.1"

pkgs = {
    "numpy": MIN_NUMPY,
    "scipy": MIN_SCIPY,
    "numba": MIN_NUMBA,
    "python": MIN_PYTHON,
    "python-version": MIN_PYTHON,
}

for pkg_name, pkg_version in pkgs.items():
    for fname in fnames:
        with open(fname, "r") as file:
            for line_num, line in enumerate(file, start=1):
                l = line.strip().replace(" ", "").lower()
                matches = re.search(
                    rf"""
                                {pkg_name}  # Package name
                                [=><:"\'\[\]]+  # Zero or more special characters
                                (\d+\.\d+[\.0-9]*)  # Capture "version" in `matches`
                                """,
                    l,
                    re.VERBOSE,  # Ignores all whitespace in pattern
                )
                if matches is not None:
                    version = matches.groups()[0]
                    if version != pkg_version:
                        print(
                            f'Package Mismatch Found: "{pkg_name}" "{version}" in {fname}:{line_num} '
                        )
```

joehiggi1758:
@seanlaw

I do believe that would suffice, I can start reworking my local min.py - thanks for throwing that together! However I believe that regex misses the following edge cases...
- Programming Language :: Python :: 3.8
- STUMPY supports `Python 3.8`
- python-version: ['3.8']
- requires-python = ">=3.8"
- numba>=0.55.2

I will work on trying to get a regex to match all edge cases!

seanlaw:
> However I believe that regex misses the following edge cases (I tested using regex101.com)...

Hmm, can you tell me which lines those correspond to? I am seeing this output when I run the code:

```
Package Mismatch Found: "numpy" "1.21" in pyproject.toml:35
Package Mismatch Found: "numpy" "1.21" in requirements.txt:1
Package Mismatch Found: "numpy" "1.21" in environment.yml:6
Package Mismatch Found: "scipy" "1.10" in pyproject.toml:36
Package Mismatch Found: "scipy" "1.10" in requirements.txt:2
Package Mismatch Found: "scipy" "1.10" in environment.yml:7
Package Mismatch Found: "numba" "0.57.1" in pyproject.toml:37
Package Mismatch Found: "numba" "0.57.1" in requirements.txt:3
Package Mismatch Found: "numba" "0.57.1" in environment.yml:8
Package Mismatch Found: "python" "3.8" in pyproject.toml:8
Package Mismatch Found: "python" "3.8" in pyproject.toml:26
Package Mismatch Found: "python" "3.8" in environment.yml:5
Package Mismatch Found: "python" "3.8" in README.rst:328
Package Mismatch Found: "python-version" "3.8" in .github/workflows/github-actions.yml:13
Package Mismatch Found: "python-version" "3.8" in .github/workflows/github-actions.yml:66
Package Mismatch Found: "python-version" "3.8" in .github/workflows/github-actions.yml:113
``` 

In case you missed it, in addition to the regex, I also did `l = line.strip().replace(" ", "").lower()` and it appears to be enough to cover all cases. Otherwise, please let me know which lines in which file are missed.

joehiggi1758:
@seanlaw

I see, yes you're correct I overlooked that strip and replace line! 

The only edge case left is "python-version: ['3.8']" for example, line 13 in ".github/workflows/github-actions.yml"

PR #991, opened 2024-06-18
Fix #990: Add support for param `copy` to preprocess-related functions

### LINKED ISSUE #990: Add the parameter `copy` to `core.preprocess_non_normalized`
Currently, `core.preprocess` supports the parameter `copy`. This function calls `core._preprocess` and pass the argument `copy` to it: `T = _preprocess(T, copy)`.

So, I think a similar support should be provided for `core.preprocess_non_normalized` (This function also calls `core._preprocess`)

NimaSarajpoor:
see #990

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -2181,7 +2181,7 @@ def preprocess_non_normalized(T, m, copy=True):
     return T, T_subseq_isfinite
 
 
-def preprocess_diagonal(T, m, copy=True, T_subseq_isconstant=None):
+def preprocess_diagonal(T, m, T_subseq_isconstant=None, copy=True,):
     """
     Preprocess a time series that is to be used when traversing the diagonals of a
     distance matrix.
@@ -2204,10 +2204,6 @@ def preprocess_diagonal(T, m, copy=True, T_subseq_isconstant=None):
     m : int
         Window size
 
-    copy : bool, default True
-        A boolean value that indicates whether the process should be done on
-        input `T` (False) or its copy (True).
-
     T_subseq_isconstant : numpy.ndarray or function, default None
         A boolean array that indicates whether a subsequence in `T` is constant
         (True). Alternatively, a custom, user-defined function that returns a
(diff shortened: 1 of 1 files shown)


NimaSarajpoor:
In addition, I think it is reasonable to add a similar support to `core.preprocess_diagonal`.

https://github.com/TDAmeritrade/stumpy/blob/93ba25a4d49a7865e47f2ef6af1434d8248511dc/stumpy/core.py#L2180

 What do you think @seanlaw ?

seanlaw:
> In addition, I think it is reasonable to add a similar support to core.preprocess_diagonal.

Sure!

NimaSarajpoor:
@seanlaw 
Can you please check the changes and merge if you are okay with them?

stumpy/core.py
```diff
@@ -2177,7 +2181,7 @@ def preprocess_non_normalized(T, m):
     return T, T_subseq_isfinite
 
 
-def preprocess_diagonal(T, m, T_subseq_isconstant=None):
+def preprocess_diagonal(T, m, copy=True, T_subseq_isconstant=None):
```
seanlaw, line 2184: Be very careful. When you insert a keyword argument in the middle of a function, it may/can cause conflicts. You must go and ensure that all calls to `preprocess_diagonal` have `T_subseq_isconstant` explicitly defined by keyword and NOT by position. I just verified that this was `True` but I wanted to call it out for the future. Otherwise, if there were some code that was currently doing:

```
core.preprocess_diagonal(T, m, None)
```

Then, we could end up with `copy=None`, which would lead to unexpected behavior. As ugly/inconsistent as it may look, it is best to append `copy=True` to the end of the list of parameters. In this case, it is probably "okay" but "power users" may be calling `preprocess_diagonal` themselves for other purposes and it would cause problems

seanlaw (commented):
@NimaSarajpoor I left one comment for you to read but, after your response, I will merge

stumpy/core.py
NimaSarajpoor: > Then, we could end up with copy=None, which would lead to unexpected behavior. As ugly/inconsistent as it may look, it is best to append copy=True to the end of the list of parameters

> In this case, it is probably "okay" but "power users" may be calling preprocess_diagonal themselves for other purposes and it would cause problems

Right! I thought if the tests are passing, then we should be okay. However, I did not think about that particular scenario. Got it! I will revise the code accordingly. 

> Be very careful.

> but I wanted to call it out for the future. 

Thanks for bring it to my attention! 👍

NimaSarajpoor:
@seanlaw 
I revised the order of named arguments in the function `core.preprocess_diagonal` to ensure backward compatibility.

seanlaw:
Thank you for this PR @NimaSarajpoor

PR #994, opened 2024-06-22
Fixed #973 updated min.py script with additional checks to scan files in …

### LINKED ISSUE #973: Add additional checks to `min.py`

joehiggi1758:
See #973 

…the Stumpy repo that list package versions and confirm those versions align with the minimum supported versions

# Pull Request Checklist

- [x] Create a new branch

### DIFF
--- a/min.py
+++ b/min.py
@@ -136,33 +136,32 @@ def get_min_scipy_version(min_python, min_numpy):
     return df.SciPy_version
 
 
-def find_pkg_mismatches(pkgs, fnames):
+def find_pkg_mismatches(pkg_name, pkg_version, fnames):
     """
-    Determine if any package version mismatches
+    Determine if any package version has mismatches
     """
     pkg_mismatches = []
-
-    for pkg_name, pkg_version in pkgs.items():
-        for fname in fnames:
-            with open(fname, "r") as file:
-                for line_num, line in enumerate(file, start=1):
-                    l = line.strip().replace(" ", "").lower()
-                    matches = re.search(
-                        rf"""
-                            {pkg_name}  # Package name
-                            [=><:"\'\[\]]+  # Zero or more special characters
-                            (\d+\.\d+[\.0-9]*)  # Capture "version" in `matches`
-                          """,
-                        l,
-                        re.VERBOSE,  # Ignores all whitespace in pattern
-                    )
-                    if matches is not None:
-                        version = matches.groups()[0]
-                        if version != pkg_version:
-                            pkg_mismatches.append(
-                                f'Package Mismatch Found: "{pkg_name}" "{version}" '
-                                f"in {fname}:{line_num}"
-                            )
+    
+    for fname in fnames:
+        with open(fname, "r") as file:
+            for line_num, line in enumerate(file, start=1):
+                l = line.strip().replace(" ", "").lower()
+                matches = re.search(
+                    rf"""
+                        {pkg_name}  # Package name
+                        [=><:"\'\[\]]+  # Zero or more special characters
+                        (\d+\.\d+[\.0-9]*)  # Capture "version" in `matches`
+                        """,
+                    l,
+                    re.VERBOSE,  # Ignores all whitespace in pattern
+                )
+                if matches is not None:
+                    version = matches.groups()[0]
+                    if version != pkg_version:
+                        pkg_mismatches.append(
+                            f'Package Mismatch Found: "{pkg_name}" "{version}" '
+                            f"in {fname}:{line_num}\n"
+                        )
 
     return pkg_mismatches
 
@@ -187,7 +186,7 @@ if __name__ == "__main__":
     )
 
     pkgs = {
-        "numpy": "1.19",
+        "numpy": MIN_NUMPY,
         "scipy": MIN_SCIPY,
         "numba": MIN_NUMBA,
         "python": MIN_PYTHON,
@@ -202,4 +201,5 @@ if __name__ == "__main__":
         "README.rst",
     ]
 
-    print(find_pkg_mismatches(pkgs, fnames))
+    for pkg_name, pkg_version in pkgs.items():
+        print(find_pkg_mismatches(pkg_name, pkg_version, fnames))
--- a/min.py
+++ b/min.py
@@ -203,6 +203,6 @@ if __name__ == "__main__":
             pkg_name, pkg_version, fnames
         ):
             print(
-                f'Package Mismatch Found: "{pkg_name}" "{version}"'
-                f'in {fname}:{line_num}\n'
+                f"Package Mismatch Found: "{pkg_name}" "{version}""
+                f"in {fname}:{line_num}\n"
             )
--- a/min.py
+++ b/min.py
@@ -203,6 +203,6 @@ if __name__ == "__main__":
             pkg_name, pkg_version, fnames
         ):
             print(
-                f"Package Mismatch Found: "{pkg_name}" "{version}""
+                f"Package Mismatch Found: {pkg_name} {version}"
                 f"in {fname}:{line_num}\n"
             )
--- a/min.py
+++ b/min.py
@@ -86,7 +86,7 @@ def get_min_scipy_version(min_python, min_numpy):
         .replace({".x": ""}, regex=True)
         .pipe(
             lambda df: df.assign(
-                SciPy_version=df.SciPy_version.str.replace(
+                SciPy_version=df.SciPy_version.astype("str").str.replace(
                     r"\d\/", "", regex=True  # noqa
                 )
             )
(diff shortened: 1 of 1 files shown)


min.py
```diff
@@ -135,6 +136,37 @@ def get_min_scipy_version(min_python, min_numpy):
     return df.SciPy_version
 
 
+def find_pkg_mismatches(pkgs, fnames):
+    """
+    Determine if any package version mismatches
```
seanlaw, line 141: Maybe "Determine if any package version HAS mismatches"?

min.py
```diff
@@ -135,6 +136,37 @@ def get_min_scipy_version(min_python, min_numpy):
     return df.SciPy_version
 
 
+def find_pkg_mismatches(pkgs, fnames):
```
seanlaw, line 139: So, as mentioned before, I think it is better to make this function:

```
def find_pkg_mismatches(pkg_name, pkg_version, fnames):
```

rather than doing ALL packages at the same time. The code that I shared yesterday was purely for demo purposes and was not intended to suggest that we should do all packages at the same time so apologies for that confusion.

min.py
```diff
@@ -135,6 +136,37 @@ def get_min_scipy_version(min_python, min_numpy):
...
+    """
+    Determine if any package version mismatches
+    """
+    pkg_mismatches = []
+
+    for pkg_name, pkg_version in pkgs.items():
```
seanlaw, line 145: As mentioned in the comment above, if we change the function signature then this `for` loop can be eliminated. In general, we try to avoid having deeply nested loops as it as adds too much burden to the reader to understand the code. One for-loop is good, two nested for-loops is fine, but three or more nested for-loops usually means that we should reconsider the design/abstraction of our function(s)

min.py
```diff
@@ -135,6 +136,37 @@ def get_min_scipy_version(min_python, min_numpy):
...
+        for fname in fnames:
+            with open(fname, "r") as file:
+                for line_num, line in enumerate(file, start=1):
```
seanlaw, line 148: Cool! I learned something new here. I never knew that `enumerate` accepted a `start` parameter

min.py
```diff
@@ -146,9 +178,28 @@ def get_min_scipy_version(min_python, min_numpy):
...
+
     print(
         f"python: {MIN_PYTHON}\n"
         f"numba: {MIN_NUMBA}\n"
         f"numpy: {MIN_NUMPY}\n"
         f"scipy: {MIN_SCIPY}"
     )
+
+    pkgs = {
+        "numpy": "1.19",
```
seanlaw, line 190: Should this be `MIN_NUMPY` instead of `1.19`?

min.py
```diff
@@ -146,9 +178,28 @@ def get_min_scipy_version(min_python, min_numpy):
...
+
+    fnames = [
+        "pyproject.toml",
+        "requirements.txt",
+        "environment.yml",
+        ".github/workflows/github-actions.yml",
+        "README.rst",
+    ]
+
+    print(find_pkg_mismatches(pkgs, fnames))
```
seanlaw, line 205: If we change the function signature above then we should put the `for` loop here:

```
for pkg_name, pkg_version in pkgs.items():
    print(find_pkg_mismatches(pkg_name, pkg_version, fnames))
```

min.py
```diff
@@ -135,6 +136,37 @@ def get_min_scipy_version(min_python, min_numpy):
...
+                          """,
+                        l,
+                        re.VERBOSE,  # Ignores all whitespace in pattern
+                    )
+                    if matches is not None:
+                        version = matches.groups()[0]
+                        if version != pkg_version:
+                            pkg_mismatches.append(
+                                f'Package Mismatch Found: "{pkg_name}" "{version}" '
+                                f"in {fname}:{line_num}"
```
seanlaw, line 164: Will we need to append a `\n` (newline) to the end of the string?

seanlaw (changes requested):
@joehiggi1758 I have left some comments for you to consider

joehiggi1758:
@seanlaw 

Thanks so much for the direction - I'll start fixing things up!

QQ - I noticed I failed two tests from codecov - how can I remedy this?

min.py
joehiggi1758: Yes! That's the result of staying up too late 🤦

min.py
joehiggi1758: If I remove this for loop how should I return a set of results for a full scan (i.e. across all packages/versions), would this be acceptable if I fix the signature of `find_pkg_mismatches`?

`pkg_mismatches = [find_pkg_mismatches(pkg, version, fnames) for pkg, version in pkgs.items()]
 print(pkg_mismatches)`

min.py
joehiggi1758: I don't believe so! The split between line 163 and 164 was to satisfy flake8's length concern! 

It seems to append correctly!

seanlaw:
> QQ - I noticed I failed two tests from codecov - how can I remedy this?

Don't worry about those two failures. I am in the process of removing those checks. What matters is our coverage checks

min.py
seanlaw: Oh, I meant at the end of the combined strings. This way, they can be printed on separate lines later below in `__main__`

joehiggi1758:
@seanlaw ah got it - I will keep that in mind! 

I updated `find_pkg_mismatches` to print newlines when __main__ is run! Also FYI on line 89 I kept getting an error from Pandas and the only way I could remedy it was by adding `.astype("str")` - is it okay that I've included it in this most recent commit?

min.py
```diff
@@ -135,6 +136,36 @@ def get_min_scipy_version(min_python, min_numpy):
...
+                )
+                if matches is not None:
+                    version = matches.groups()[0]
+                    if version != pkg_version:
+                        pkg_mismatches.append(
+                            f'Package Mismatch Found: "{pkg_name}" "{version}" '
+                            f"in {fname}:{line_num}"
+                        )
+
+    return "\n".join(pkg_mismatches)
```
seanlaw, line 166: Maybe it's just me but I don't like doing the join here and I'd rather return a list of strings rather than a single long string. Perhaps it's because it keeps this function flexible.

I know that I'm being picky and possibly annoying now but maybe a list of strings is too rigid and, instead, we should do:

```
pkg_mistaches.append((pkg_name, version, fname, line_num))
```

And then all of the print formatting happens in `__main__`. I think I like that better because it'll likely be easier to maintain/change in the future. And "FUTURE Sean/Joey will thank PAST Sean/Joey" for not painting ourselves into a corner

min.py
```diff
@@ -146,9 +177,29 @@ def get_min_scipy_version(min_python, min_numpy):
...
+    fnames = [
+        "pyproject.toml",
+        "requirements.txt",
+        "environment.yml",
+        ".github/workflows/github-actions.yml",
+        "README.rst",
+    ]
+
+    for pkg_name, pkg_version in pkgs.items():
+        print(find_pkg_mismatches(pkg_name, pkg_version, fnames))
```
seanlaw, line 205: Based on my change suggestion above, I think this will become:

```
for name, version, fname, line_num in find_pkg_mismatches(pkg_name, pkg_version, fnames):
    print(f'Package Mismatch Found: "{pkg_name}" "{version}" in {fname}:{line_num}\n'
```

seanlaw (changes requested):
@joehiggi1758 Just a minor suggestion and then I think we're good to merge

seanlaw:
> Also FYI on line 89 I kept getting an error from Pandas and the only way I could remedy it was by adding .astype("str") - is it okay that I've included it in this most recent commit?

Yes, I'm noticing it too! I'd bet that the version table in the SciPy docs changed or was updated recently and now only contains floating point numbers (whereas something like `1.20.1` would cause the entire column to type cast as a string). Thanks for fixing it!

min.py
joehiggi1758: @seanlaw

No annoyance here - it's extremely helpful for me to see and understand your design decisions, so again - I very much appreciate the feedback! I've fixed the best I can according to this approach!

Let me know if I need to update further!

seanlaw:
> Let me know if I need to update further!

Everything looks good!

@joehiggi1758 [Previously](https://github.com/TDAmeritrade/stumpy/issues/973#issuecomment-2159199209), you had asked:

> QQ - do I need to create a unit test for min.py when I make this contribution?

And, originally, I had said "No". I now regret that response 😆 

Considering how many different cases we're trying to cover with our `regex`, I'd like for us to have a way to quickly check whether or not our `regex` has stopped covering one of our identified/existing cases. While I don't think it should be a `unit test` per se, we should at least think about how to protect ourselves.

I think it is okay to:

1. Deal with this in a separate (new) issue and solve it "later"
2. Document the list of all of the cases that we've identified so far in the new issue
3. Then, we can go ahead and merge this PR ASAP

joehiggi1758:
@seanlaw 

Sounds good I'm on it! I'll open an issue!

seanlaw:
@joehiggi1758  Thank you for contributing this PR!

> Should I work on that and try to open a PR? Happy and eager to keep contributing!

Yes, if you have some ideas/thoughts, please go ahead.

PR #1001, opened 2024-07-08
Fix #998: Speed up stumpi and aampi

### LINKED ISSUE #998: Speed Up STUMPI Update
STUMPI `.update()` is reasonably fast but @NimaSarajpoor [was able to speed it even more](https://github.com/TDAmeritrade/stumpy/discussions/924#discussioncomment-9901868)!

We should take a closer look and see if there is further refactoring/enhancement that could be had.

NimaSarajpoor:
See #998  .

- [x] Speed up stumpi, `_update_egress` method
- [x] Speed up stumpi, `_update` method
- [x] Speed up aampi, `_update_egress` method
- [x] Speed up aampi, `_update` method


stumpy/stumpi.py
```diff
@@ -471,6 +453,61 @@ def _update(self, t):
         self._Σ_T = Σ_T_new
         self._T_subseq_isconstant = T_subseq_isconstant_new
 
+    @staticmethod
```
NimaSarajpoor, line 456: @seanlaw 
Is this a good idea? I know if I move it to `core` (and write test for it), I do not need to define it again in `aampi`. But I wanted to limit the usability of this function to these classes only for now.

stumpy/stumpi.py
```diff
@@ -471,6 +453,61 @@ def _update(self, t):
...
+
+        Parameters
+        ----------
+        D : numpy.ndarray
+            A 1D array (with dtype float) representing the distance profile of
+            the last subsequence of T
+
+        P : numpy.ndarray
+            A 2D array representing the matrix profile of T,
+            with shape (len(T) - m + 1, k)
```
NimaSarajpoor, line 472: Add: 
```
, where `m` is the window size.
```

stumpy/stumpi.py
```diff
@@ -471,6 +453,61 @@ def _update(self, t):
...
+
+        I : numpy.ndarray
+            A 2D array representing the matrix profile index of T,
+            with shape (len(T) - m + 1, k)
```
NimaSarajpoor, line 476: Add: 
```
, where `m` is the window size.
```

stumpy/stumpi.py
```diff
@@ -471,6 +453,61 @@ def _update(self, t):
...
+            (idx -  excl_zone, idx + excl_zone + 1).
+
+        n_appended : int
+            Number of times the timeseries start point is shifted one to the right
+
+        Returns
+        -------
+        None
+        """
+        core._apply_exclusion_zone(D, D.shape[0] - 1, excl_zone, np.inf)
```
NimaSarajpoor, line 490: **Note:** I decided to include this in the function. Otherwise, one needs to make sure the elements corresponding to the trivial subsequences of last subsequences in `D` are set to `np.inf`

stumpy/stumpi.py
seanlaw: To be consistent, I would move it to `core.py`

stumpy/stumpi.py
seanlaw: It's probably okay

stumpy/core.py
```diff
@@ -4371,7 +4371,7 @@ def get_ray_nworkers(ray_client):
 
 
 @njit
-def _update_egress_PI(D, P, I, excl_zone, n_appended=0):
+def _update_incremental_PI(D, P, I, excl_zone, n_appended=0):
```
seanlaw, line 4374: Why change the name of the function if the function itself has not changed??

NimaSarajpoor: > Why change the name of the function if the function itself has not changed??

I am calling this function from both methods `_update` and `_update_egress` of `stumpy.stumpi` / `stumpy.aampi`. So, I thought using the term `_egress` might be misleading here. But, `_update_PI` is not clear enough. So, that is why I decided to go with `_update_incremental_PI`.

seanlaw: Nevermind, I didn't realize it was a new function altogether. I only saw the diff between commits.

stumpy/aampi.py
```diff
@@ -322,30 +305,17 @@ def _update(self, t):
...
         I_new = np.full(self._k, -1, dtype=np.int64)
-        for i, d in enumerate(D):
-            if d < P_new[-1]:  # maximum value in sorted array P_new
-                idx = np.searchsorted(P_new, d, side="right")
-                core._shift_insert_at_index(P_new, idx, d)
-                core._shift_insert_at_index(I_new, idx, i)
+        self._P = np.append(self._P, P_new.reshape(1, -1), axis=0)
+        self._I = np.append(self._I, I_new.reshape(1, -1), axis=0)
+
+        core._update_incremental_PI(D, self._P, self._I, self._excl_zone, 0)
```
NimaSarajpoor, line 313: **[Note to self]**
Be more explicit:
```suggestion
        core._update_incremental_PI(D, self._P, self._I, self._excl_zone, n_appended=0)
```

stumpy/stumpi.py
```diff
@@ -440,30 +423,18 @@ def _update(self, t):
...
         I_new = np.full(self._k, -1, dtype=np.int64)
-        for i, d in enumerate(D):
-            if d < P_new[-1]:  # maximum value in sorted array P_new
-                idx = np.searchsorted(P_new, d, side="right")
-                core._shift_insert_at_index(P_new, idx, d)
-                core._shift_insert_at_index(I_new, idx, i)
+        self._P = np.append(self._P, P_new.reshape(1, -1), axis=0)
+        self._I = np.append(self._I, I_new.reshape(1, -1), axis=0)
+
+        core._update_incremental_PI(D, self._P, self._I, self._excl_zone, 0)
```
NimaSarajpoor, line 431: [Note to self]
Be more explicit:
```suggestion
        core._update_incremental_PI(D, self._P, self._I, self._excl_zone, n_appended=0)
```

tests/test_core.py
```diff
@@ -1753,3 +1753,31 @@ def test_process_isconstant_2d():
...
+    I_comp = np.full_like(I_ref, -1)
+
+    mp_comp = stump(T[:-1], m, k=k)
+    P_comp[:-1, :] = mp_comp[:, :k]
+    I_comp[:-1, :] = mp_comp[:, k : 2 * k]
+
+    D = core.mass(T[-m:], T)
+    excl_zone = int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
+
+    core._update_incremental_PI(D, P_comp, I_comp, excl_zone, 0)
```
NimaSarajpoor, line 1779: **[Note to self]**

Be more explicit
```suggestion
    core._update_incremental_PI(D, P_comp, I_comp, excl_zone, n_appended=0)
```

Also: add test for a case where `n_appended>.0`

NimaSarajpoor:
@seanlaw 
I think it is ready. What do you think?

seanlaw:
@NimaSarajpoor Please allow me some time to review it

seanlaw:
@NimaSarajpoor For completeness, are you able to provide some timings for the speedup (before and after your code changes) here in the comments? I trust that the code is indeed faster but at least we can document it here.

NimaSarajpoor:
@seanlaw 
I checked the performance for time series with length `1000`, `10_000`, and `100_000`. 

```
import time
import numpy as np
import stumpy

def get_running_time(n, m=50):
    seed = 0
    np.random.seed(seed)
    T = np.random.rand(n)

    n_iter = 100
    vals = np.random.rand(n_iter)

    obj = stumpy.stumpi(T, m, egress=True)  # try with stumpi / aampi, and egress: True / False
    t_lst = []
    for val in vals:
        start = time.time()
        obj.update(val)
        t_lst.append(time.time() - start)
    
    return np.mean(t_lst[1:]), np.std(t_lst[1:])


if __name__ == "__main__":
        n = 1000  # try with 1000 / 10_000 / 100_000
        out = get_running_time(n)
        print(f'mean: {out[0]}, std: {out[1]}' )
```


In the following table, `n` is the length of time series. The results are obtained by running the script above on Apple M1 with 8G Memory. Each value provided in the running time columns is the average running time of 100 `.updates()` excluding the first one. The speedup percentage is provided in the right-most column.

| n = 1000         | running time <br> (current version) | running time <br> (PR's version) | Speedup (%) |
| -------------------- | --------------- | ------------ | ------------------------------- |
| stumpi(egress=True)  | 0.00035         | 0.00014      | 60                              |
| stumpi(egress=False) | 0.00033         | 0.00015      | 54.54545455                     |
| aampi(egress=True)   | 0.00026         | 0.00004      | 84.61538462                     |
| aampi(egress=False)  | 0.00022         | 0.00006      | 72.72727273                     |



| n = 10_000       | running time <br> (current version) | running time <br> (PR's version) | Speedup (%) |
| -------------------- | --------------- | ------------ | ------------------------------- |
| stumpi(egress=True)  | 0.00138         | 0.00022      | 84.05797101                     |
| stumpi(egress=False) | 0.00114         | 0.00023      | 79.8245614                      |
| aampi(egress=True)   | 0.00147         | 0.00032      | 78.23129252                     |
| aampi(egress=False)  | 0.00121         | 0.00032      | 73.55371901                     |


| n =  100_000      |  running time <br> (current version) | running time <br> (PR's version)  | Speedup (%) |
| -------------------- | --------------- | ------------ | ------------------------------- |
| stumpi(egress=True)  | 0.01162         | 0.00079      | 93.20137694                     |
| stumpi(egress=False) | 0.00943         | 0.00088      | 90.66808059                     |
| aampi(egress=True)   | 0.01387         | 0.00282      | 79.66834895                     |
| aampi(egress=False)  | 0.01154         | 0.00285      | 75.30329289                     |

seanlaw:
@NimaSarajpoor Can you tell me how you are computing "Speedup %"? The numbers don't look right to me. I think perhaps the wording should be "Percent Reduction" (i.e., `100 * (new-old)/old`).

I think Percent Speedup would be `100 * (old-new)/new` OR you can say "X times faster" by simply doing `old/new`

stumpy/aampi.py
```diff
@@ -247,26 +247,9 @@ def _update_egress(self, t):
...
-        # Calculate the (top-k) matrix profile values/indices for the last subsequence
-        # by using its corresponding distance profile `D`
-        self._P[-1] = np.inf
-        self._I[-1] = -1
-        for i, d in enumerate(D):
-            if d < self._P[-1, -1]:
-                idx = np.searchsorted(self._P[-1], d, side="right")
-                core._shift_insert_at_index(self._P[-1], idx, d)
-                core._shift_insert_at_index(self._I[-1], idx, i + self._n_appended)
+        core._update_incremental_PI(
```
seanlaw, line 250: Maybe it's just me but it feels like `_update_incremental_PI` is doing too much? Maybe we should be splitting into two separate functions rather than one giant one?

NimaSarajpoor:
@seanlaw 
> I think perhaps the wording should be "Percent Reduction" (i.e., 100 * (new-old)/old).

Right.... that is how I calculated the numbers. 

> OR you can say "X times faster" by simply doing old/new. I prefer "X times faster" for our comparison.

Noted. I like this more as it is clearer. To avoid confusion for future readers who follow the comments, I am going to provide the tables with the new numbers below:

| n = 1000         | running time <br> (current version) | running time <br> (PR's version) | X times faster |
| -------------------- | --------------- | ------------ | ------------------------------- |
| stumpi(egress=True)  | 0.00035         | 0.00014      | 2.5                              |
| stumpi(egress=False) | 0.00033         | 0.00015      | 2.2                    |
| aampi(egress=True)   | 0.00026         | 0.00004      | 6.5                     |
| aampi(egress=False)  | 0.00022         | 0.00006      | 3.7                    |



| n = 10_000       | running time <br> (current version) | running time <br> (PR's version) | X times faster |
| -------------------- | --------------- | ------------ | ------------------------------- |
| stumpi(egress=True)  | 0.00138         | 0.00022      | 6.27                    |
| stumpi(egress=False) | 0.00114         | 0.00023      | 4.9                      |
| aampi(egress=True)   | 0.00147         | 0.00032      | 4.6                     |
| aampi(egress=False)  | 0.00121         | 0.00032      | 3.8                     |


| n =  100_000      |  running time <br> (current version) | running time <br> (PR's version)  | X times faster |
| -------------------- | --------------- | ------------ | ------------------------------- |
| stumpi(egress=True)  | 0.01162         | 0.00079      | 14.7                     |
| stumpi(egress=False) | 0.00943         | 0.00088      | 10.7                     |
| aampi(egress=True)   | 0.01387         | 0.00282      | 4.9                    |
| aampi(egress=False)  | 0.01154         | 0.00285      | 4.0                     |

stumpy/aampi.py
```diff
@@ -322,30 +305,17 @@ def _update(self, t):
         if np.any(~self._T_isfinite[-self._m :]):
             D[:] = np.inf
 
-        core.apply_exclusion_zone(D, D.shape[0] - 1, self._excl_zone, np.inf)
-
-        update_idx = np.argwhere(D[:l] < self._P[:l, -1]).flatten()
```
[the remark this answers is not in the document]
NimaSarajpoor: This is indeed considered in `core._update_incremental_PI`. 

Note that `D[:l]` is the same as `D[:-1]` where D is the distance profile of the last subsequence. So, previously we had this:
```
# compute distance profile for last subsequence
# ignore last element and update not-yet-extended matrix profile 
# compute the new (top-k) profile value and append that to not-yet-extended matrix profile
```

And now we have this:
```
# extend matrix profile and set its last value to `np.inf`
Note: now everything else is similar to _update_egress

# update matrix profile using D
```

stumpy/aampi.py
NimaSarajpoor: I feel these two parts are related. This is how I think about it: I say to myself if there is a new data point, and hence a new distance profile that corresponds to the last subsequence, what changes should be made? One, the NN of previous subsgeuqneces might be changed (which is first for-loop) AND the NN of the last subsequence must be determined (the second for-loop). 

Having said that, we might still say that the second for-loop is just finding top-k in a 1D array. So, we can create one function that just does that. But, how about the first for-loop? I am trying to see how that (independent) function should look like. Maybe we just create one function for the second for-loop, `_find_top_k`, and then call it from `_update_incremental_PI`

seanlaw:
@NimaSarajpoor Considering that all of the existing tests are passing and the performance is improved, I feel pretty good about merging this. Do you think it's ready?

NimaSarajpoor:
@seanlaw 

>  It looks like there's refactor of the code and then njit-ing that code

Right. That's it!

> Was there anything that you had doubts about?

My first concern is whether the added test function is clear.  My second concern is regarding the comment I added for case 2 in the test function, i.e.
```
    # case 2: For a given time series `T`, obtain the matrix profile `P` and
    # matrix profile indices `I` of `T[1:]` based on the matrix profile and
    # matrix profile indices of `T[:-1]`.
    # In the following test: n_appended = 1
```

I think the comment above is slightly wrong. I think I need to make it clear that the updated profile is different than just doing `stumpy.stump(T[1:], m)`. So, I think it should have been something like this:

```
    # case 2: For a given time series `T`, obtain the matrix profile `P` and
    # matrix profile indices `I` of `T[1:]` based on the matrix profile and
    # matrix profile indices of `T[:-1]`, WITHOUT DISREGARDING THE NEAREST
    # NEIGHBOURS IN THE PROFILE THAT REFERS TO ALREADY-REMOVED DATA.
    # In the following test: n_appended = 1
```

seanlaw:
> My first concern is whether the added test function is clear. My second concern is regarding the comment I added for case 2 in the test function

Okay, I will take a closer look.

stumpy/core.py
```diff
@@ -4368,3 +4368,60 @@ def get_ray_nworkers(ray_client):
...
+        A 2D array representing the matrix profile index of T,
+        with shape (len(T) - m + 1, k), where `m` is the window size
+        I[-1, :] should be set to -1.
+
+    excl_zone : int
+        Size of the exclusion zone. That is, after finding the next-best-match
+        located at index `idx`, we ignore subsequences with start index in range
+        (idx -  excl_zone, idx + excl_zone + 1).
+
+    n_appended : int
```
seanlaw, line 4401: I think you should add a comment about how this parameter is associated with "egress=True" and "egress=False" situations?

NimaSarajpoor: I think that should make it clearer 👍

NimaSarajpoor:
@seanlaw 
>  I can't understand case 2. There seems to be too much happening and your intent isn't clear even with the comment(s)

I have the same feeling regarding case 2, which represents `egress=False` case. 

> >WITHOUT DISREGARDING THE NEAREST NEIGHBOURS IN THE PROFILE THAT REFERS TO ALREADY-REMOVED DATA.
> 
> I think this is probably the most important thing to highlight. It sounds like this test is trying to make sure that your newly added function respects this point and does not ignore it. Is that right? And all of this is associated with n_appended?
Right. 

My main point was to help future me remember why I used a particular approach for calculating `P_ref`  and `I_ref`. Now that you mention it, I think it would be good to have a test to just check that specific statement.

> I think it would make sense to split case 1 and case 2 into two separate tests with more specific names (i.e., the name of the first case can be kept but it seems like you are testing something more nuanced in the second case).

Noted. Please allow me to separate the cases, and revise the code.

seanlaw:
> My main point was to help future me remember why I used a particular approach for calculating P_ref and I_ref. Now that you mention it, I think it would be good to have a test to just check that specific statement.

Exactly! Thank you for persisting

stumpy/core.py
NimaSarajpoor: I added a note at the bottom of docstring.

NimaSarajpoor:
@seanlaw 
I am making some minor changes. Will let you know once I am done so that you can provide with me your comments.

NimaSarajpoor:
@seanlaw 
I think it is ready for your review. You may want to pay closer attention to the docstring of the new function in `core.py` and the test functions, particularly `test_update_incremental_PI_egressTrue_MemoryCheck`. This test function still looks a bit weird but let's see what you think.

seanlaw:
@NimaSarajpoor I will take a look

tests/test_core.py
```diff
@@ -1,6 +1,7 @@
 import functools
 import math
 import os
+import random
```
seanlaw, line 4: Instead of importing `random`, why not just do `np.random.rand()`? That will give you a single random number and without needing to import `random` when `np` is available?

seanlaw (changes requested):
@NimaSarajpoor I think the docstrings are fine. Do you think we are ready to merge?

tests/test_core.py
NimaSarajpoor: Right! Noticed it now after carefully reading the doc! Going to fix it now!

NimaSarajpoor:
I replaced `random` with `np.random`. 

@seanlaw 
> I think the docstrings are fine. Do you think we are ready to merge?

Thanks for checking that! So, feel free to merge it once all tests pass

seanlaw:
@NimaSarajpoor Thanks again for the wonderful contribution!

PR #1004, opened 2024-07-12
Investigate Github Workflow RunTime

NimaSarajpoor:
See: https://github.com/TDAmeritrade/stumpy/issues/982#issuecomment-2224137251

### DIFF
--- a/tests/test_gpu_aamp_ostinato.py
+++ b/tests/test_gpu_aamp_ostinato.py
@@ -80,13 +80,13 @@ def test_input_not_overwritten():
 def test_extract_several_consensus():
     # This test is to further ensure that the function `gpu_aamp_ostinato`
     # does not tamper with the original data.
-    Ts = [np.random.rand(n) for n in [256, 512, 1024]]
+    Ts = [np.random.rand(n) for n in [64, 128]]
     Ts_ref = [T.copy() for T in Ts]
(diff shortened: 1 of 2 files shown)


seanlaw (commented):
@NimaSarajpoor The changes look good to me. Can you describe what is actually taking a long time or causing the spike in time?

If this is sufficient, is there a reason why we don't change the tests in `test_ostinato` and `test_aamp_ostinato`?

NimaSarajpoor:
@seanlaw 
The spike is coming from calling `gpu_ostinato`. 

I think the only difference between  `test_ostinato` and `test_gpu_ostinato` is the part that we compute the consensus. The former uses `ostinato` and the latter uses `gpu_ostinato`. These two functions themselves are similar and the only difference is that the former calls `stump` and the latter calls `gpu_stump`.  When we turn off CUDA and JIT, `stump` algo should be faster. For example, on my Mac, for the window size 20 and the input list `Ts` that contains time series of lengths 512 and 1024, the `gpu_ostinato` takes 60 sec while `ostinato` takes 3 sec. I did not dig further.

seanlaw:
Shall I merge this now?

NimaSarajpoor:
@seanlaw 

> Shall I merge this now?

sure 👍

Issue #1011, opened 2024-07-21
passing-input-with-finite-value to `core._mass` is being violated


NimaSarajpoor:
The function `core.mass(Q, T, ...)` calls `core.preprocess(T, m, ...)`, which return the following values:
* T
* M_T
* Σ_T
* T_subseq_isconstant

We then pass these values to `core._mass`. However, these `M_T` and `Σ_T` can contain non-finite value if the original `T` has non-finite value. In fact, if you just follow the function `core._mass` and go the callee function and then continue till you reach the very end of the road, you can see that at the end, we use that in the function `core._calculate_squared_distance` to determine whether a distance should be infinite or not.

https://github.com/TDAmeritrade/stumpy/blob/3077d0ddfb315464321dc86f8ec3bf2cab9ce3b1/stumpy/core.py#L1093-L1094

In the function `core._calculate_squared_distance`, the flag is correctly set. Was wondering if should do the same for `core._mass`?

seanlaw:
What exactly are you proposing? To set `fastmath={"nsz", "arcp", "contract", "afn", "reassoc"}` for `core._mass`? Maybe... I don't know.

NimaSarajpoor:
Right! 

I am now thinking what if we just avoid setting fastmath here for core._mass because this function does not do any arithmetic operations. I can try it out and check its impact on performance if there is any.

NimaSarajpoor:
In the following, `A --> B` means function `A` calls function `B`.

```
# in stumpy.core
mass --> _mass
_mass --> calculate_distance_profile
calculate_distance_profile -->  _calculate_squared_distance_profile
_calculate_squared_distance_profile --> _calculate_squared_distance
```

I removed the `fastmath` flag for all expect the last one, which comes with `fastmath={"nsz", "arcp", "contract", "afn", "reassoc"}`.

I then compared the performance of `core.mass` for the following input:
```
seed = 0
np.random.seed(seed)
Q = np.random.rand(100)
T = np.random.rand(1_000_000)
```

I ran the following script:
```
import numpy as np
import time
import stumpy

def check_mass_performance():
    n_iter = 100

    seed = 0
    np.random.seed(seed)
    Q = np.random.rand(100)
    T = np.random.rand(1000000)

    t_lst = []
    for _ in range(n_iter):
        start = time.time()
        stumpy.mass(Q, T)
        t_lst.append(time.time() - start)
    
[...]

PR #1012, opened 2024-07-21
Fixed #1006 RuntimeWarning `divide by zero encountered in divide`

### LINKED ISSUE #1006: RuntimeWarning: divide by zero encountered in divide
Running the following script

```
import numpy as np
import stumpy
import time

def test_extract_several_consensus():
    Ts = [np.random.rand(n) for n in [64, 128]]
    Ts_ref = [T.copy() for T in Ts]
    Ts_comp = [T.copy() for T in Ts]

    m = 20

    k = 2  # Get the first `k` consensus motifs
    for _ in range(k):
        print(f'===== {_} =====')
        radius, Ts_idx, subseq_idx = stumpy.ostinato(Ts_comp, m)
    
        consensus_motif = Ts_comp[Ts_idx][subseq_idx : subseq_idx + m].copy()
        for i in range(len(Ts_comp)):
            if i == Ts_idx:
                query_idx = subseq_idx
            else:
                query_idx = None

            idx = np.argmin(stumpy.core.mass(consensus_motif, Ts_comp[i], query_idx=query_idx))

            Ts_comp[i][idx : idx + m] = np.nan
            Ts_ref[i][idx : idx + m] = np.nan

            np.testing.assert_almost_equal(
                Ts_ref[i][np.isfinite(Ts_ref[i])], Ts_comp[i][np.isfinite(Ts_comp[i])]
            )


if __name__ == "__main__":
    test_extract_several_consensus()
```

Gives:
```
===== 0 =====
===== 1 =====
/Users/nimasarajpoor/miniconda3/envs/py310/lib/python3.10/site-packages/stumpy/core.py:2257: 
RuntimeWarning: divide by zero encountered in divide
```

I took a look at the code:

https://github.com/TDAmeritrade/stumpy/blob/e7bb2b85ee8761683480ac093437e71989252d15/stumpy/core.py#L2248-L2260

and tried to swap lines 2252 and 2253, but that did not work.

----

Also:
If I disable JIT, I get an extra warning, saying:
```
python3.10/site-packages/numpy/lib/nanfunctions.py:1879: RuntimeWarning: Degrees of freedom <= 0 for slice.
```

NimaSarajpoor:
See #1006 

This PR addresses part1, 2, and 3 as discussed here: https://github.com/TDAmeritrade/stumpy/issues/1006#issuecomment-2229551395


seanlaw:
@NimaSarajpoor It looks like we have a failing unit test

stumpy/core.py
```diff
@@ -2254,6 +2254,7 @@ def preprocess_diagonal(
 
     M_T, Σ_T = compute_mean_std(T, m)
     Σ_T[T_subseq_isconstant] = 1.0  # Avoid divide by zero in next inversion step
+    Σ_T[~T_subseq_isfinite] = 1.0
```
NimaSarajpoor, line 2257: @seanlaw 

I was thinking more about this and noticed that `Σ_T[~T_subseq_isfinite] = 1.0` sets the std of ANY non-finite subsequence to `1.0`. For example, the std of subsequence `[np.nan, 1, 2]` will be set to `1.0`. This however does not affect the computation of matrix profile.

tests/test_core.py
```diff
@@ -1753,3 +1756,16 @@ def test_process_isconstant_2d():
...
     npt.assert_array_equal(T_subseq_isconstant_ref, T_subseq_isconstant_comp)
+
+
+def test_preprocess_diagonal_std_inverse():
+    T = np.random.rand(64)
+    m = 3
+    T[:m] = np.nan
+
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
```
NimaSarajpoor, line 1767: Is there a better way to catch warning for the sake of unit testing?

stumpy/core.py
seanlaw: @NimaSarajpoor I don't think I am getting your point. Is doing `Σ_T[~T_subseq_isfinite] = 1.0` a problem? To some extent, I kind of view ANY subsequence with a single non-finite value to be interpreted as a subsequence with ALL `np.nan`. Then, by that logic (maybe you can disagree?), that entire subsequence is "constant". Of course, it's not really constant! In other words, a single `np.inf`/`np.nan` within the subsequence renders the entire subsequence useless for computing distances and a "bad" subsequence cannot actually have a `Σ_T` value. What do you think?

Hmmm, now that I think about it, what happens if we set `Σ_T[~T_subseq_isfinite] = np.nan` instead?? This would avoid any divide-by-zero since it would be divide-by-nan, which is fine, and it would be separate/independent from `T_subseq_isconstant` and we keep things explicit?

tests/test_core.py
seanlaw: What is it that you are trying to do?

stumpy/core.py
NimaSarajpoor: > Is doing `Σ_T[~T_subseq_isfinite] = 1.0` a problem?

Not really. I think we can do `Σ_T[~T_subseq_isfinite] = v`, where `v` is any non-zero value and we are still fine. In fact, we can do  `Σ_T[T_subseq_isconstant] = v` as well where `v` is just any non-zero value.

> what happens if we set Σ_T[~T_subseq_isfinite] = np.nan instead?? This would avoid any divide-by-zero since it would be divide-by-nan, which is fine, and it would be separate/independent from T_subseq_isconstant and we keep things explicit?

Right!! I think 

```
Σ_T[T_subseq_isconstant] = 1.0
Σ_T[~T_subseq_isfinite] = np.nan
```
 is better as it separates the "subsequence with at least one `np.nan`" case from the "constant, finite subsequence" case.

tests/test_core.py
NimaSarajpoor: I am trying to add a test function to look for warnings and raise an error if it catches one that says "divide by zero encountered in divide"

stumpy/core.py
seanlaw: Cool! Please let me know if this breaks any tests

tests/test_core.py
seanlaw: I think it might be more convenient to do this. So, here's an example test:

```
import pytest
import numpy as np

@pytest.mark.filterwarnings("error:divide by zero encountered in divide")
def test_something():
    a = np.ones(5, dtype=np.float64)
    b = np.zeros(5, dtype=np.float64)
    c = a / b
```

Notice that there are two parts:
1. `error` - turns whatever the ensuing warning is into an error 
2. the warning text to match for raising an error - notice that if you chaged `divide` to `multiply` then this does NOT raise an error and instead only warns us

NimaSarajpoor: Thank you for the lesson!! Will revise the test function.

stumpy/ostinato.py
```diff
@@ -58,10 +58,9 @@ def _across_series_nearest_neighbors(
     nns_subseq_idx = np.zeros(k, dtype=np.int64)
 
     for i in range(k):
-        QT = core.sliding_dot_product(Ts[Ts_idx][subseq_idx : subseq_idx + m], Ts[i])
-        distance_profile = core._mass(
-            Q,
-            Ts[i],
+        QT = core.sliding_dot_product(Q, Ts[i])
```
NimaSarajpoor, line 61: Since `Q` and `T` are UNPROCESSED, `QT[idx]` can get meaningless value for a non-finite subsequence that starts at `idx`. Better to just replace them with `np.nan`. So, for example, we can do something like this after computing `QT` (`QT = core.sliding_dot_product(Q, Ts[i])`):

```
# post-processing QT
if np.any(~np.isfinite(Q)):
     QT[:] = np.nan

QT[~np.isfinite(M_Ts[i])] = np.nan
``` 


Note that `QT[idx]` will be ignored though eventually once it comes to computing its corresponding value in  the distance profile via `core.calculate_distance_profile` (see a few lines below).  In other words, we do not need the "post-processing of QT" but I think it makes code clearer.

Any concerns?

NimaSarajpoor (commented):
@seanlaw 
I have one comment. Can you please take a look and let me know what you think?

stumpy/ostinato.py
seanlaw: I understand the need to replace the ugly `Ts[Ts_idx][subseq_idx : subseq_idx + m]` with `Q`. That's fine but

```
QT = core.sliding_dot_product(Q, Ts[i])
```

doesn't seem to require any post-processing when either `Q` or `T` contain `np.nan`.

NimaSarajpoor: May bad! You are right. No need for post-processing `QT`

NimaSarajpoor: @seanlaw 
> maybe I should better stop here and try to address https://github.com/TDAmeritrade/stumpy/issues/1011 first. Then come back and resume the work on this PR.

I did an experiment and shared the result in https://github.com/TDAmeritrade/stumpy/issues/1011. We can either work on that and then come back here. Or, on the second thought, we can still merge this (after I resolve the merge conflicts), and then work on https://github.com/TDAmeritrade/stumpy/issues/1011.

What do you think?

seanlaw: Since we have a merge conflict here anyways, let's fix #1011 first and then come back here and strikethrough the comment:

> Note: Since we pass inputs with non-finite values to core.calculate_distance_profile (Not only QT but also M_T), maybe I should better stop here and try to address https://github.com/TDAmeritrade/stumpy/issues/1011 first. Then come back and resume the work on this PR.

so that we are focused on this issue/PR instead of being distracted by jumping back and forth. I am getting confused by having so many things open and so it feels disorganized

PR #1015, opened 2024-07-23
Fixed #1014 Fix max_matches=None bug

### LINKED ISSUE #1014: _motifs max_matches=None bug
When finding motifs, the function encounters a bug when `max_matches` is set to `None` to not restrict the number of matches. If the `max_matches` argument is set to `None` for `stumpy.motifs`, it assigns `max_matches = np.inf` before passing to `_motifs` which then tries to slice the `query_matches` with `np.inf` as seen in the error below.

Example:
```
>>> import numpy as np
>>> import stumpy
>>> arr = np.array([0.,1,2,2,0,1,2,2,0,0,0,0])
>>> mp = stumpy.stump(arr, m=3)
>>> stumpy.motifs(arr, mp[:,0], max_matches=None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/user0/code/venv/lib/python3.10/site-packages/stumpy/core.py", line 138, in inner_wrapper
    return norm(*args, **kwargs)
  File "/Users/user0/code/venv/lib/python3.10/site-packages/stumpy/motifs.py", line 361, in motifs
    motif_distances, motif_indices = _motifs(
  File "/Users/user0/code/venv/lib/python3.10/site-packages/stumpy/motifs.py", line 134, in _motifs
    motif_distances.append(query_matches[:max_matches, 0])
TypeError: slice indices must be integers or None or have an __index__ method
```

ejorgensen-wl:
This fix addresses issue #1014 by not slicing the `query_matches` if `max_matches` is set to `None` for `motifs` (which in turn calls `_motifs` with `max_matches=np.inf`.

I made an initial effort to add testing, but I'm sure there are better ways to do so.

# Pull Request Checklist

- [x] Create a new branch

### DIFF
--- a/tests/test_motifs.py
+++ b/tests/test_motifs.py
@@ -679,7 +679,7 @@ def test_motifs_with_max_matches_none():
         max_motifs=max_motifs,
     )
 
-    ref_len = len(T) / m
+    ref_len = len(T) - m + 1
 
-    npt.assert_(ref_len <= comp_distance.shape[1])
-    npt.assert_(ref_len <= comp_indices.shape[1])
+    npt.assert_(ref_len >= comp_distance.shape[1])
+    npt.assert_(ref_len >= comp_indices.shape[1])
--- a/tests/test_motifs.py
+++ b/tests/test_motifs.py
@@ -656,3 +656,30 @@ def test_motifs_with_isconstant():
 
     npt.assert_almost_equal(ref_distances, comp_distance)
     npt.assert_almost_equal(ref_indices, comp_indices)
+
+
+def test_motifs_with_max_matches_none():
+    T = np.random.rand(16)
+    m = 3
+
+    max_motifs = 1
+    max_matches = None
+    max_distance = np.inf
+    cutoff = np.inf
+
+    # performant
+    mp = naive.stump(T, m, row_wise=True)
+    comp_distance, comp_indices = motifs(
+        T,
+        mp[:, 0].astype(np.float64),
+        min_neighbors=1,
+        max_distance=max_distance,
+        cutoff=cutoff,
+        max_matches=max_matches,
+        max_motifs=max_motifs,
+    )
+
+    ref_len = len(T) / m
+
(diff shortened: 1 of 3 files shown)


seanlaw:
@ejorgensen-wl Thanks again for identifying this bug. The decision to do:

https://github.com/TDAmeritrade/stumpy/blob/3077d0ddfb315464321dc86f8ec3bf2cab9ce3b1/stumpy/motifs.py#L336-L337

was decided a long time ago and, in hindsight, that was a poor choice. Instead, I think we could've/should've been more specific and done:

```
if max_matches is None:  # pragma: no cover
    max_matches = P.shape[-1]
```

`P.shape[-1]` would be the same as the maximum number of subsequences in `T` that are possible and is a more explicit number than `np.inf`, which isn't even an integer (it's a floating point value). 

What do you think about this remedy instead? Would this work?

tests/test_motifs.py
```diff
@@ -656,3 +656,30 @@ def test_motifs_with_isconstant():
...
+        T,
+        mp[:, 0].astype(np.float64),
+        min_neighbors=1,
+        max_distance=max_distance,
+        cutoff=cutoff,
+        max_matches=max_matches,
+        max_motifs=max_motifs,
+    )
+
+    ref_len = len(T) / m
```
seanlaw, line 682:
```suggestion
    ref_len = len(T) - m + 1
```

seanlaw:
> That's an excellent and much cleaner solution than what I was thinking, thanks! I implemented your suggestions in my latest commit.

@ejorgensen-wl Thank you for your patience and for working through this with us! We really appreciate the PR. Once the tests all pass, I'll merge the PR (unless you had anything further to add?).

ejorgensen-wl:
I like it as is - thanks!

seanlaw:
@ejorgensen-wl Thank you again for the excellent PR! I hope you will consider contributing more!

PR #1060, opened 2025-01-07
Fix #1059 missing fastmath

### LINKED ISSUE #1059: Missing fastmath flag in some njit functions
As mentioned in PR #1025 (See [this comment](https://github.com/TDAmeritrade/stumpy/pull/1025#issuecomment-2571498719)), all njit functions must have fastmath flag. This issue is created so that this can be addressed separately. After that, we can resume PR #1048 and finish it. Then, we can resume the work on #1025. ( I am trying to stay on the current track as much as possible while addressing "necessary" changes.)

I've created the script `check_fastmath.py` in the root directory and executed it via `./test.sh`. The tests fail which is expected. So, the next step is to add fastmath flags to those njit functions.

@seanlaw 
Do you think creating this new issue complicates our flow? If not, I can push the changes from my local branch to  remote branch.

NimaSarajpoor:
See #1059 

The functions with missing fastmath flags are:

* [x] core._get_array_ranges
* [x] core._get_ranges
* [x] core._merge_topk_PI
* [x] core._merge_topk_ρI
* [x] core._shift_insert_at_index
* [x] core._update_incremental_PI


fastmath.py
```diff
@@ -0,0 +1,47 @@
...
+            missing_fastmath.append(f"{module_name}.{func_name}")
+
+    if len(missing_fastmath) > 0:
+        msg = "Found one or more functions that are missing the `fastmath` flag. "
+        msg += f"The function(s) are:\n {missing_fastmath}\n"
+        raise ValueError(msg)
+
+    return
+
+
```
seanlaw, line 35: This can be simplified and should be indented within:

```
if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument('--check', action="store_true")  # args.check will be `False` if this flag is not provided
    args = parser.parse_args()

    if args.check:
        check_fastmath()

File: test.sh [hunk 2]
```diff
@@ -93,6 +93,13 @@ check_print()
     fi
 }
 
+fastmath()
+{
+    echo "Checking Missing fastmath flags in njit functions"
+    ./fastmath.py --perform check
```
seanlaw, line 99: Based on my suggestion above, this would be

```
./fastmath.py --check
```

test.sh
```diff
@@ -146,14 +153,14 @@ set_ray_coveragerc()
 show_coverage_report()
 {
     set_ray_coveragerc
-    coverage report -m --fail-under=100 --skip-covered --omit=docstring.py,min_versions.py,ray_python_version.py,stumpy/cache.py $fcoveragerc
+    coverage report -m --fail-under=100 --skip-covered --omit=check_fastmath.py,docstring.py,min_versions.py,ray_python_version.py,stumpy/cache.py $fcoveragerc
```
seanlaw, line 156: `omit=fastmath.py` (not `check_fastmath.py`)

test.sh
```diff
@@ -146,14 +153,14 @@ set_ray_coveragerc()
...
 }
 
 gen_coverage_xml_report()
 {
     # This function saves the coverage report in Cobertura XML format, which is compatible with codecov
     set_ray_coveragerc
-    coverage xml -o $fcoveragexml --fail-under=100 --omit=docstring.py,min_versions.py,ray_python_version.py,stumpy/cache.py $fcoveragerc
+    coverage xml -o $fcoveragexml --fail-under=100 --omit=check_fastmath.py,docstring.py,min_versions.py,ray_python_version.py,stumpy/cache.py $fcoveragerc
```
seanlaw, line 163: `omit=fastmath.py`

test.sh
```diff
@@ -93,6 +93,13 @@ check_print()
     fi
 }
 
+fastmath()
```
seanlaw, line 96: I think this functionshould be called `check_fastmath`

test.sh
```diff
@@ -333,6 +340,12 @@ check_print
 check_naive
 check_ray
 
+
+if [[ ! -z $NUMBA_DISABLE_JIT && ${NUMBA_DISABLE_JIT} -eq "0" ]]; then
+  fastmath
```
seanlaw, line 345: Based on my comment about, this would be `check_fastmath`

seanlaw (changes requested):
I did not look at the `def check_fastmath` function yet and I assume it is doing the right thing?

test.sh
NimaSarajpoor: Done

test.sh
NimaSarajpoor: Makes sense! Thanks for pointing that out.

fastmath.py
NimaSarajpoor: THANK YOU!! Revised accordingly.

fastmath.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+
+import argparse
+import importlib
+
+from stumpy.cache import get_njit_funcs
```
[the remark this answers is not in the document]
NimaSarajpoor: I noticed that `docstring.py` is standalone but didn't know what to do here. I think I was afraid of violating the DRY rule. You are correct. This should not depend on stumpy's module. I can copy/paste. As an alternative solution, I am now trying to see if I can use regex to catch all njit-decorated functions.

fastmath.py
```diff
@@ -0,0 +1,78 @@
...
+    for filepath in filepaths:
+        fname = filepath.name
+        if fname not in ignore and fname.endswith(".py"):
+            file_contents = ""
+            with open(filepath, encoding="utf8") as f:
+                file_contents = f.read()
+
+            matches = re.findall(pattern, file_contents, re.DOTALL)
+            for match in matches:
+                func_name = match.split("def ")[-1].split("(")[0]
```
NimaSarajpoor, line 39: **[Note to self]**
There should be a better way to find the function's name that appears between `def ` and `(`. We can try to use another regex pattern here, or we can see if there is a way to modify the initial pattern, i.e. `r"@njit.*?def\s+\w+\("`, to address this.

stumpy/cache.py
```diff
@@ -43,11 +43,16 @@ def get_njit_funcs():
...
-                            njit_funcs.append((module_name, func_name))
+                    scenario_1 = isinstance(decorator, ast.Name) and (
+                        decorator.id == "njit"
+                    )
+                    scenario_2 = (
+                        isinstance(decorator, ast.Call)
+                        and isinstance(decorator.func, ast.Name)
+                        and decorator.func.id == "njit"
+                    )
+                    if scenario_1 or scenario_2:
```
NimaSarajpoor, line 54: @seanlaw 
Are you okay with the names `scenario_1` & `scenario_2`? Alternatively, we can say `cond1` & `cond2`.

NimaSarajpoor: okay.... actually I do not like the way I did it. Going to revise it to `if/elif/else`.

fastmath.py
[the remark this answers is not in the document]
seanlaw: > I noticed that docstring.py is standalone but didn't know what to do here. I think I was afraid of violating the DRY rule. You are correct. This should not depend on stumpy's module. I can copy/paste.

It's a bit funky but somebody might run `./fastmath.py --check` without having installed `stumpy` and so there is no `stumpy.cache` to import.

>  As an alternative solution, I am now trying to see if I can use regex to catch all njit-decorated functions.

If you're having problems, I think this variation would work:

```
def get_njit_funcs():
    """
    Identify all njit functions

    Parameters
    ----------
    None

    Returns
    -------
    out : list
        A list of (`module_name`, `func_name`) pairs
    """
    pkg_dir = pathlib.Path(__file__).parent / "stumpy"
    module_names = [
        fname.stem
        for fname in pathlib.Path(pkg_dir).iterdir()
        if not fname.stem.startswith(".")
    ]

    njit_funcs = []

    for module_name in module_names:
        filepath = pkg_dir / f"{module_name}.py"
        file_contents = ""
            file_contents = f.read()
        for node in module.body:
            if isinstance(node, ast.FunctionDef):
                func_name = node.name
                for decorator in node.decorator_list:
                    if isinstance(decorator, ast.Call) and isinstance(
                        decorator.func, ast.Name
                    ):
                        if decorator.func.id == "njit":
                            njit_funcs.append((module_name, func_name))

    return njit_funcs
```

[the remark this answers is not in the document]
seanlaw: For consistency, we should update `cache.get_njit_funcs` to look the same. We don't really need to import `pkgutil`

fastmath.py
seanlaw: See above, I think the original version still works and avoids regex

fastmath.py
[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 

> As an alternative solution, I am now trying to see if I can use regex to catch all njit-decorated functions.
> > If you're having problems, I think this variation would work:

No problems... the original version did work. I just wanted to try out regex. And another reason for using regex was that I noticed we did not catch `core._update_incremental_PI` initially.  Also, I was thinking that maybe we should treat its output as the ref. Then, move `get_njit_funcs` to `stumpy/core.py` and write a test function for it.

Too much?

fastmath.py
NimaSarajpoor: > May I ask "why are you doing it this way?" Why is there a need to deviate from the original?

My main reason was that I noticed we did not catch `core._update_incremental_PI` with the current version of `cache.get_njit_functions` (and that is why I changed it in this PR to fix that). I thought the ouput that comes from "regex" should not have this issue (and the chance of missing a njit function is less)

fastmath.py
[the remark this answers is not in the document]
seanlaw: > There was no problem actually... the original version did work. I just wanted to try out regex.

You may have heard the old saying

> Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

Sometimes, regular expressions can be useful. However, we should always try to avoid regex when a clearer/obvious implementation is available. I try to use it sparingly as they can be hard to read/maintain. If we find that it's the only way then I can accept that reality.

> And another reason for using regex was that I noticed we did not catch core._update_incremental_PI initially

You're right! I think the issue is that `core._update_incremental_PI` has a bare `@njit` and was not detected as a callable "decorator function". We can resolve this with a little bit more code (that should get inserted back into `cache.py`):

```
def get_njit_funcs():
    """
    Identify all njit functions

    Parameters
    ----------
    None

    Returns
    -------
    out : list
        A list of (`module_name`, `func_name`) pairs
    """
    pkg_dir = pathlib.Path(__file__).parent / "stumpy"
    module_names = [
        fname.stem
        for fname in pathlib.Path(pkg_dir).iterdir()
        if not fname.stem.startswith(".")
    ]

    njit_funcs = []

    for module_name in module_names:
        filepath = pkg_dir / f"{module_name}.py"
        file_contents = ""
            file_contents = f.read()
        for node in module.body:
            if isinstance(node, ast.FunctionDef):
                func_name = node.name
                for decorator in node.decorator_list:
                    decorator_name = None

                    if isinstance(decorator, ast.Name):
                        # Bare @njit decorator (i.e., no parantheses, no parameters)
                        decorator_name = decorator.id
                    if isinstance(decorator, ast.Call) and isinstance(
                        decorator.func, ast.Name
                    ):
                        # @njit decorator is a function (i.e., with parantheses)
                        decorator_name = decorator.func.id

                    if decorator_name == "njit":
                        njit_funcs.append((module_name, func_name))

    return njit_funcs
```

> Also, I was thinking that maybe we should treat its output as the ref. Then, move cache.get_njit_funcs to stumpy/core.py and write a test function for it.

I think we should leave it in `cache.py` as `cache.py` doesn't get tested. I mean, if you want to move it and add tests then I will not stop you either.

fastmath.py
seanlaw: Please see my [response above](https://github.com/TDAmeritrade/stumpy/pull/1060#discussion_r1907389337)

fastmath.py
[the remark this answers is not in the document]
NimaSarajpoor: > Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

> Sometimes, regular expressions can be useful. However, we should always try to avoid regex when a clearer/obvious implementation is available.  I try to use it sparingly as they can be hard to read/maintain. If we find that it's the only way then I can accept that reality.

This is actually the first time I am hearing this saying and I realized that this is a famous saying! 😅 And I had to spend some time to use ChatGPT & regex101 to get that short regex pattern. My knowledge in regex is limited but, still, can see your point.


>  I think the issue is that core._update_incremental_PI has a bare @njit and was not detected as a callable "decorator function". 

Correct! and the change you proposed is cleaner than the change I made to `cache.get_njit_funcs`. So, I will revise that.

> I think we should leave it in cache.py as cache.py doesn't get tested.

Yeah...noticed it... that's why I suggested to move it to `core` and add test for it. BUT 👇 

> One of the bad things about regex is that it is "greedy" and so you may end up with MORE unexpected matches

This is my concern too! So, we may end up with changing the ref in our test function which may not make sense that much. I mean... the ref should be trustworthy.

stumpy/core.py
```diff
@@ -2347,6 +2347,7 @@ def _count_diagonal_ndist(diags, m, n_A, n_B):
 
 @njit(
     # "i8[:, :](i8[:], i8, b1)"
+    fastmath={"nsz", "arcp", "contract", "afn", "reassoc"}
```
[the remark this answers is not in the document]
NimaSarajpoor: > May I ask why this isn't fastmath=True? Doesn't this get called after ...

**TLDR:**
Will change it to `fastmath=True`

**Detail:**
According to the callers of this private function, the input array a in `_get_array_ranges(a, ...)` always has finite values. So, it can be `fastmath=True`. Then, I noticed that it is called only once and the heavy part of computation is related to the computation of matrix profile. So, I thought: maybe I should ignore its callers (I mean... just consider the function itself) and use a safer(?) fastmath flag for it. But, I just realized that the output of this function is not useful when input array a contains at least one `np.inf` and/or `np.nan`. AND, more importantly, this is a PRIVATE function! So, I am going to change it to `fastmath=True`.

fastmath.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Returns
+    -------
+    njit_funcs : list
+        A list of all njit functions, where each element is a tuple of the form
+        (module_name, func_name)
+    """
+    ignore_py_files = ["__init__", "__pycache__"]
+
+    pkg_dir = pathlib.Path(__file__).parent / "stumpy"
```
seanlaw, line 25: Tell me what you think. I almost feel like `pkg_dir` should be an input parameter. So, in `__main__`, we pass our desired directory into this function or possibly make it a required command line input parameter. This will help ensure that this function can be generalized in the future (if we need to use it for another package).

fastmath.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    missing_fastmath = []  # list of njit functions with missing fastmath flags
+    for module_name, func_name in get_njit_funcs():
+        module = importlib.import_module(f".{module_name}", package="stumpy")
+        func = getattr(module, func_name)
+        if "fastmath" not in func.targetoptions.keys():
+            missing_fastmath.append(f"{module_name}.{func_name}")
+
+    if len(missing_fastmath) > 0:
+        msg = "Found one or more functions that are missing the `fastmath` flag. "
```
seanlaw, line 79: "one or more `@njit` functions"

fastmath.py
NimaSarajpoor: I agree!

This is similar to the challenge I am trying to handle at work. There is a module that contains not just the process but also some values that are related to the input data. There was no problem initially. But then the scope of project became bigger and the same process was needed to be applied to different data that are somewhat different. So, now, I have to decouple the process from those values!

So... I agree. Although we may not use it again later, it is still better to have just the "process" in the function

NimaSarajpoor (commented):
@seanlaw 
I tried to make some changes according to our recent discussion.  I left some comments to bring your attention to a few things.

fastmath.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+
+import argparse
+import ast
+import importlib
+import pathlib
+
+
+def get_njit_funcs(dir_path):
```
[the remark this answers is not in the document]
seanlaw: I agree 100%

fastmath.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--check", action="store_true")
+    args = parser.parse_args()
+
+    if args.check:
+        pkg_dir = pathlib.Path(__file__).parent / "stumpy"
+        check_fastmath(pkg_dir, package="stumpy")
```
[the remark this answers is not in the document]
seanlaw: > Should I try to pass these values via command line?

Yes, I think you should so it will be `./fastmath.py --check stumpy`

```
if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--check", dest="pkg_dir")
    args = parser.parse_args()

    if args.pkg_dir:
        pkg_dir = pathlib.Path(args.pkg_dir).resolve()  # Absolute path
        pkg_name = pkg_dir.name
        check_fastmath(pkg_dir, pkg_name)
```

Even if you do `./fastmath.py --check ./stumpy` or some variation, the above `argparse` should still work.

fastmath.py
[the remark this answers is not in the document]
seanlaw: Why isn't it called `pkg_dir` (instead of `dir_path`) to keep it simple/consistent?

fastmath.py
```diff
@@ -0,0 +1,96 @@
...
+    return
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--check", action="store_true")
+    args = parser.parse_args()
+
+    if args.check:
+        pkg_dir = pathlib.Path(__file__).parent / "stumpy"
```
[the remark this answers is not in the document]
seanlaw: See my suggestion above

fastmath.py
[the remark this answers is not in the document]
NimaSarajpoor: > Why isn't it called pkg_dir (instead of dir_path) to keep it simple/consistent?

Wanted to reflect that the value of the param needs be the path of a directory. I think I should have considered the purpose of the function, which is checking .py files of a package's directory.  So, I am going to revise it. I will go with `pkg_dir`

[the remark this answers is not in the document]
seanlaw: `dir` and `path` are too similar in meaning and are almost redundant in this case

fastmath.py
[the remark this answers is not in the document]
NimaSarajpoor: Thanks for `parser.add_argument("--check", dest="pkg_dir")`  👍  I am learning new things here!!

fastmath.py
```diff
@@ -0,0 +1,102 @@
...
+import pathlib
+
+
+def get_njit_funcs(pkg_dir):
+    """
+    Identify all njit functions
+
+    Parameters
+    ----------
+    pkg_dir : An instance of pathlib.Path
```
NimaSarajpoor, line 15: @seanlaw 
Do you think the type of parameter `pkg_dir` should be `str`? Then, in the function, we can do: 
```
pkg_dir = pathlib.Path(pkg_dir)
```

----

On one hand, we run the script fastmath.py via terminal and that's all we need. On the other hand, I am thinking how easily one can use the function `get_njit_funcs`. Should I even bother to think about it? If yes, then I think passing path as `str` is easier than passing an instance of `pathlib.Path(...)`. 

What do you think?

fastmath.py
```diff
@@ -0,0 +1,102 @@
...
+
+    return
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--check", dest="pkg_dir")
+    args = parser.parse_args()
+
+    if args.pkg_dir:
```
NimaSarajpoor, line 99: I was wondering whether the `if condition` should be changed to: `if args.pkg_dir is not None:` just for the sake of readability.

fastmath.py
seanlaw: I agree. It should just be type `str` as it is the easiest to manage

NimaSarajpoor:
@seanlaw 
Can you please do a final check and merge if there is no concern and all tests are passing?

seanlaw:
> @seanlaw Can you please do a final check and merge if there is no concern and all tests are passing?

Will do! Thanks for pulling this together

seanlaw:
Thanks @NimaSarajpoor. This has now been merged!

NimaSarajpoor:
@seanlaw 
Great! That was one good PR [for me]! Thanks for all of your inputs and suggestions.

PR #1070, opened 2025-01-28
Fixed #1069 Added numba cache dir for pytest

### LINKED ISSUE #1069: Wrong Numba Cache Directory
@NimaSarajpoor  [Had asked](https://github.com/TDAmeritrade/stumpy/issues/1066#issuecomment-2617797555):

> I think we should open an issue for https://github.com/TDAmeritrade/stumpy/pull/1067#discussion_r1930017203. This can help us to track it separately. What do you think?

This is the corresponding ticket in response to his question.

From what I understand, once a package is installed, the `nbc`/`nbi` files are typically written to the `site-packages/stumpy/__pycache__` subdirectory if a user imports `stumpy`. However, specifically for `pytest`, cache files are written locally in `stumpy/__pycache__` because `pytest` isn't using the installed package and, instead, are using the files from the local `stumpy/*py`! So, when a user uses `stumpy`, caching occurs in site-packages and all is well (since `cache._clear()` targets `site-packages/stumpy/__pycache__`. However, for unit testing, we get a different behavior because `pytest` is basically running things locally (i.e., it is creating the bytecode locally and not from `site-packages`) and accessing the bytecode/cache from `stumpy/__pycache__` rather than from `site-packages/stumpy/__pycache__`. In other words, `pytest` isn't using the installed stumpy and, instead, `pytest` is using/importing local `stumpy` and creating the bytecode locally (independent of the installed package in site-packages).

So, it's a bit misleading when we do `./setup.sh && ./test.sh` 😢

Note, that this (cache directory) behavior should NOT be observed from an end user's standpoint (i.e., when `stumpy` is installed and accessed from site-packages) and should only affect unit testing.

seanlaw:
See #1069 

# Pull Request Checklist

- [x] Create a new branch

### DIFF
--- a/stumpy/cache.py
+++ b/stumpy/cache.py
@@ -168,7 +168,12 @@ def _get_cache(cache_dir=None):
     else:  # pragma: no cover
         site_pkg_dir = site.getsitepackages()[0]
         numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
-    return [f.name for f in pathlib.Path(numba_cache_dir).glob("*nb*") if f.is_file()]
+
+    return [
+        f"{numba_cache_dir}/{f.name}"
+        for f in pathlib.Path(numba_cache_dir).glob("*nb*")
+        if f.is_file()
+    ]
 
 
 def _recompile():
--- a/stumpy/cache.py
+++ b/stumpy/cache.py
@@ -133,7 +133,7 @@ def clear(cache_dir=None):
 
     Parameters
     ----------
-    cache_dir : str
+    cache_dir : str, default None
         The path to the numba cache directory
 
     Returns
@@ -163,7 +163,7 @@ def _get_cache(cache_dir=None):
     warnings.warn(CACHE_WARNING)
     if cache_dir is not None:  # pragma: no cover
         numba_cache_dir = str(cache_dir)
-    if "PYTEST_CURRENT_TEST" in os.environ:
+    elif "PYTEST_CURRENT_TEST" in os.environ:
         numba_cache_dir = "stumpy/__pycache__"
     else:  # pragma: no cover
         site_pkg_dir = site.getsitepackages()[0]
@@ -205,33 +205,35 @@ def _recompile():
     return
 
 
-def _save():
+def _save(cache_dir):
     """
     Save all njit functions
 
     Parameters
     ----------
-    None
+    cache_dir : str
+        The path to the numba cache directory
 
     Returns
     -------
     None
     """
     _enable()
-    _clear()
+    _clear(cache_dir)
     _recompile()
 
     return
 
 
-def save():
+def save(cache_dir=None):
     """
     Save/overwrite all the cache data files of
     all-so-far compiled njit functions.
 
     Parameters
     ----------
-    None
+    cache_dir : str, default None
+        The path to the numba cache directory
 
     Returns
     -------
@@ -243,6 +245,6 @@ def save():
     else:  # pragma: no cover
         warnings.warn(CACHE_WARNING)
 
-    _save()
+    _save(cache_dir)
 
     return
--- a/stumpy/cache.py
+++ b/stumpy/cache.py
@@ -218,6 +218,7 @@ def _save():
     None
     """
     _enable()
+    _clear()
     _recompile()
 
     return
--- a/stumpy/cache.py
+++ b/stumpy/cache.py
@@ -5,6 +5,7 @@
 import ast
 import importlib
 import inspect
+import os
 import pathlib
 import site
 import warnings
@@ -102,48 +103,57 @@ def _enable():
                 raise
 
 
-def _clear():
+def _clear(cache_dir=None):
     """
     Clear numba cache
 
     Parameters
     ----------
-    None
+    cache_dir : str
+        The path to the numba cache directory
 
     Returns
     -------
     None
     """
-    site_pkg_dir = site.getsitepackages()[0]
-    numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
+    if cache_dir is not None:  # pragma: no cover
+        numba_cache_dir = str(cache_dir)
+    elif "PYTEST_CURRENT_TEST" in os.environ:
+        numba_cache_dir = "stumpy/__pycache__"
+    else:  # pragma: no cover
+        site_pkg_dir = site.getsitepackages()[0]
+        numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
+
     [f.unlink() for f in pathlib.Path(numba_cache_dir).glob("*nb*") if f.is_file()]
 
 
-def clear():
+def clear(cache_dir=None):
     """
     Clear numba cache directory
 
     Parameters
     ----------
-    None
+    cache_dir : str
+        The path to the numba cache directory
 
     Returns
     -------
     None
     """
     warnings.warn(CACHE_WARNING)
-    _clear()
+    _clear(cache_dir)
 
     return
 
 
-def _get_cache():
+def _get_cache(cache_dir=None):
     """
     Retrieve a list of cached numba functions
 
     Parameters
     ----------
-    None
+    cache_dir : str
+        The path to the numba cache directory
 
     Returns
     -------
@@ -151,8 +161,13 @@ def _get_cache():
         A list of cached numba functions
     """
     warnings.warn(CACHE_WARNING)
-    site_pkg_dir = site.getsitepackages()[0]
-    numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
+    if cache_dir is not None:  # pragma: no cover
+        numba_cache_dir = str(cache_dir)
+    if "PYTEST_CURRENT_TEST" in os.environ:
+        numba_cache_dir = "stumpy/__pycache__"
+    else:  # pragma: no cover
+        site_pkg_dir = site.getsitepackages()[0]
+        numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
     return [f.name for f in pathlib.Path(numba_cache_dir).glob("*nb*") if f.is_file()]
 
 
--- a/stumpy/cache.py
+++ b/stumpy/cache.py
@@ -253,14 +253,24 @@ def save():
     Notes
     -----
     The cache is never cleared before saving/overwriting and may be explicitly cleared
-    by calling `cache.clear()` before saving. If `cache.save()` is called for the first
-    time (before any `njit` function is called) then only the `.nbi` files (i.e., the
-    "cache index") for all `njit` functions are saved. As each `njit` function (and
+    by calling `cache.clear()` before saving. It is best practice to call `cache.save()`
+    only after calling all of your `njit` functions. If `cache.save()` is called for the
+    first time (before any `njit` function is called) then only the `.nbi` files (i.e.,
+    the "cache index") for all `njit` functions are saved. As each `njit` function (and
     sub-functions) is called then their corresponding `.nbc` file (i.e., "object code")
     is saved. Each `.nbc` file will only be saved after its `njit` function is called
-    once. However, subsequent calls to `cache.save()` (after clearing the cache via
-    `cache.clear()`) will automatically save BOTH the `.nbi` files as well as the `.nbc`
-    files as long as their `njit` function has been called at least once.
+    at least once. However, subsequent calls to `cache.save()` (after clearing the cache
+    via `cache.clear()`) will automatically save BOTH the `.nbi` files as well as the
+    `.nbc` files as long as their `njit` function has been called at least once.
+
+    Examples
+    --------
+    >>> import stumpy
+    >>> from stumpy import cache
+    >>> import numpy as np
+    >>> cache.clear()
+    >>> mp = stumpy.stump(np.array([584., -11., 23., 79., 1001., 0., -19.]), m=3)
+    >>> cache.save()
     """
     if numba.config.DISABLE_JIT:
         msg = "Could not save/cache function because NUMBA JIT is disabled"
--- a/stumpy/cache.py
+++ b/stumpy/cache.py
@@ -205,39 +205,40 @@ def _recompile():
     return
 
 
-def _save(cache_dir):
+def _save():
     """
     Save all njit functions
 
     Parameters
     ----------
-    cache_dir : str
-        The path to the numba cache directory
+    None
 
     Returns
     -------
     None
     """
     _enable()
-    _clear(cache_dir)
     _recompile()
 
     return
 
 
-def save(cache_dir=None):
+def save():
     """
-    Save/overwrite all the cache data files of
-    all-so-far compiled njit functions.
+    Save/overwrite all of the cached njit functions.
 
     Parameters
     ----------
-    cache_dir : str, default None
-        The path to the numba cache directory
+    None
 
     Returns
     -------
     None
+
+    Notes
+    -----
+    The cache is never cleared before saving/overwriting and may be explicitly
+    cleared by calling `cache.clear()` before saving.
     """
     if numba.config.DISABLE_JIT:
         msg = "Could not save/cache function because NUMBA JIT is disabled"
@@ -250,6 +251,6 @@ def save(cache_dir=None):
         msg += "The `stumpy` cache files may not be saved/cleared correctly!"
         warnings.warn(msg)
 
-    _save(cache_dir)
+    _save()
 
     return
--- a/stumpy/cache.py
+++ b/stumpy/cache.py
@@ -5,7 +5,6 @@
 import ast
 import importlib
 import inspect
-import os
 import pathlib
 import site
 import warnings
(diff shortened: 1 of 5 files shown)


seanlaw:
@NimaSarajpoor [Commented](https://github.com/TDAmeritrade/stumpy/issues/1069#issuecomment-2620666362):

> I feel the underlying challenge is "how can we make sure we are reading data from the correct location to which data was written?" So, can we try to track that numba cache directory by defining a config var like:
config.NUMBA_CACHE_DIR which can be set to None as default. Then, if cache._save() is called, we can write the directory there. We can even raise warning if the user provided numba cache directory is different than an already-set config. config.NUMBA_CACHE_DIR

According to [the numba docs](https://numba.pydata.org/numba-doc/dev/reference/envvars.html#envvar-NUMBA_CACHE_DIR):

> Numba picks the cache directory in the following order:
>
> 1. In-tree cache. Put the cache next to the corresponding source file under a __pycache__ directory following how .pyc files are stored.
> 2. User-wide cache. Put the cache in the user’s application directory using appdirs.user_cache_dir from the [Appdirs package](https://github.com/ActiveState/appdirs).
> 2. IPython cache. Put the cache in an IPython specific application directory. Stores are made under the numba_cache in the directory returned by IPython.paths.get_ipython_cache_dir().

So, by default, the cache files are written next the source files under a `__pycache__` directory. This is precisely the behavior that we are seeing (for 1) above). For users, 99% of the time the cache files will be written to `site-packages/stumpy/__pycache__`. For `pytest` unit testing, 100% of the time the cache files will be written to `stumpy.git/stumpy/__pycache__`. My code currently differentiates between the "users 99%" and our "100% pytest" scenario via the `if/else` condition:

```
    if cache_dir is not None:  # pragma: no cover
        numba_cache_dir = str(cache_dir)
    elif "PYTEST_CURRENT_TEST" in os.environ:
        numba_cache_dir = "stumpy/__pycache__"
    else:  # pragma: no cover
        site_pkg_dir = site.getsitepackages()[0]
        numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
```

I haven't tested this but I added the first `if` condition above to **potentially** handle 2) and 3) OR when the user defines the cache directory via `NUMBA_CACHE_DIR` (i.e., `numba.config.CACHE_DIR`). Although, I would say that these are beyond the scope of `stumpy`. When `stumpy` is installed as a pacakge, the `stumpy` cache files should **always** be expected in `site-packages/stumpy/__pycache__`. The only way that this would change is if the user does something unexpected (i.e., set the `NUMBA_CACHE_DIR`. In that case, that's no longer our problem to solve.

We can possibly check `if numba.config.CACHE_DIR != '':` and then spit out a warning that `cache` may have unexpected behavior.

tests/test_cache.py
```diff
@@ -11,17 +12,23 @@ def test_cache_get_njit_funcs():
 def test_cache_save_after_clear():
     T = np.random.rand(10)
     m = 3
-    stump(T, m)
 
     cache.save()
+    stump(T, m)
     ref_cache = cache._get_cache()
 
+    if numba.config.DISABLE_JIT:
```
NimaSarajpoor, line 20: @seanlaw 
If you are okay with doing `numba_cache_dir = "stumpy/__pycache__"` inside test function and pass it to cache functions, then we can remove this if-else block.

stumpy/cache.py
```diff
@@ -102,58 +103,77 @@ def _enable():
...
     -------
     out : list
         A list of cached numba functions
     """
     warnings.warn(CACHE_WARNING)
-    site_pkg_dir = site.getsitepackages()[0]
-    numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
-    return [f.name for f in pathlib.Path(numba_cache_dir).glob("*nb*") if f.is_file()]
+    if cache_dir is not None:  # pragma: no cover
+        numba_cache_dir = str(cache_dir)
```
NimaSarajpoor, line 165:
```suggestion
        numba_cache_dir = cache_dir
```

NimaSarajpoor (commented):
@seanlaw 
I added some comments for your consideration..

stumpy/cache.py
```diff
@@ -102,58 +103,77 @@ def _enable():
...
 
     Returns
     -------
     None
     """
-    site_pkg_dir = site.getsitepackages()[0]
-    numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
+    if cache_dir is not None:  # pragma: no cover
+        numba_cache_dir = str(cache_dir)
+    elif "PYTEST_CURRENT_TEST" in os.environ:
```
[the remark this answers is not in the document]
NimaSarajpoor: Btw, after moving the PYTEST part to `tests/test_cache.py`, we can simplify the if-else condition and say:
```
if cache_dir is None:  
    site_pkg_dir = site.getsitepackages()[0]
    cache_dir = site_pkg_dir + "/stumpy/__pycache__"

# and then use `cache_dir` later in the code
```

stumpy/cache.py
```diff
@@ -102,58 +103,77 @@ def _enable():
...
+        The path to the numba cache directory
 
     Returns
     -------
     None
     """
-    site_pkg_dir = site.getsitepackages()[0]
-    numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
+    if cache_dir is not None:  # pragma: no cover
+        numba_cache_dir = str(cache_dir)
```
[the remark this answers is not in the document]
seanlaw: If it's suppose to be a string, why not cast it as a string?

stumpy/cache.py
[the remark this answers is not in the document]
seanlaw: Excellent!

stumpy/fastmath.py
```diff
@@ -55,11 +56,13 @@ def _set(module_name, func_name, flag):
...
+        msg = "One or more fastmath flags have been set/reset. "
+        msg += "Please call `cache._recompile()` to ensure that all njit functions "
+        msg += "are properly recompiled."
+        warnings.warn(msg)
     except AttributeError as e:
         if numba.config.DISABLE_JIT and (
             str(e) == "'function' object has no attribute 'targetoptions'"
-            or str(e) == "'function' object has no attribute 'recompile'"
         ):
             pass
```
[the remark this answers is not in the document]
seanlaw: Sure!

tests/test_cache.py
seanlaw: Is that true? What about during coverage testing?

seanlaw: Or do you mean that we don't need to check `len(ref_cache)` at all?

seanlaw:
@NimaSarajpoor I made most of the changes but kept some. Let me know what you think

stumpy/cache.py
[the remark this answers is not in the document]
NimaSarajpoor: Since its type is mentioned in the docstirng, casting it as `str` seems to be [almost] redundant. If a user makes a mistake and pass an object like `Pathlib.path()` to this function, then casting as `str` helps the program to work as usual without throwing error. Is that what you are trying to achieve here?

stumpy/cache.py
NimaSarajpoor: Going to resolve this one for now as there is a discussion about this in another chain of comments.

tests/test_cache.py
NimaSarajpoor: @seanlaw 
> Or do you mean that we don't need to check len(ref_cache) at all?

Initially, I thought that it should be okay to not check it as we clearly set that path in the test function now and pass it to cache functions.  I've been thinking more about this and I've started to doubt my initial logic. Please let me know what you think/feel about the following:

We are not setting the path for the location to which the cache data is going to be written. I am okay with that because, as you pointed out, we do not want to play with NUMBA_CACHE_DIR, and according to official Numba doc, we know where (by default) it is going to write the cache data into. Since WE do not CONTROL where we want to save those cache files, it might be a bit strange to trust that everything will work out. So, maybe we should keep this if-else here.

stumpy/cache.py
```diff
@@ -190,32 +205,35 @@ def _recompile():
...
-    None
+    cache_dir : str
+        The path to the numba cache directory
 
     Returns
     -------
     None
     """
     _enable()
+    _clear(cache_dir)
```
NimaSarajpoor, line 222: I was wondering if we should remove `_clear` from here. Not because it is a bad idea to clear before saving again but because it results in adding param `cache_dir` to the function `_save`, and it might become misleading. A user may think it can actually save to that directory. The function `cache.save` does have warning but what if user does not even change the default value of `numba.config.CACHE_DIR`, and just pass a path to `cache.save`?

NimaSarajpoor (commented):
@seanlaw 
I just left one new comment.

stumpy/cache.py
seanlaw: > it might become misleading. A user may think it can actually save to that directory. 

I think that is a fair point. I accept. This means that we must remind the user to always `cache.clear()` first

stumpy/cache.py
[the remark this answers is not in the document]
seanlaw: I don't think we need to overthink this one too much. If you don't see any possible harm then we should leave it.

tests/test_cache.py
seanlaw: I think it will be good to keep it. It's a nice sanity check

seanlaw: I will keep tie if/else

stumpy/cache.py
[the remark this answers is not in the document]
NimaSarajpoor: All good. Going to resolve it.

seanlaw:
I've also added a simple/stupid "check" to see if `cache._clear()` has been called before `cache._save()`.

seanlaw:
I've also added a simple/stupid "check" to see if `cache._clear()` has been called before `cache._save()`.

seanlaw:
I've also added a simple/stupid "check" to see if `cache._clear()` has been called before `cache._save()`.

stumpy/cache.py
```diff
@@ -220,13 +248,25 @@ def save():
     Returns
     -------
     None
+
+    Notes
+    -----
+    The cache is never cleared before saving/overwriting and may be explicitly
+    cleared by calling `cache.clear()` before saving.
     """
+    global CACHE_CLEARED
```
NimaSarajpoor, line 257: We can remove it if we are not using it here. I can see that this is used in the child function `cache._save`

tests/test_cache.py
```diff
@@ -11,17 +12,27 @@ def test_cache_get_njit_funcs():
...
+    else:  # pragma: no cover
+        assert len(ref_cache) > 0
 
+    cache.clear(cache_dir)
+    assert len(cache._get_cache(cache_dir)) == 0
     cache.save()
-    comp_cache = cache._get_cache()
 
-    # testing cache._save() after cache._clear()
+    stump(T, m)
```
NimaSarajpoor, line 33: Do we want to test if `cache.save()` results in saving all-so-far compiled functions into `cache_dir`? If yes, then we may remove `stump` from here.

stumpy/cache.py
```diff
@@ -102,58 +103,77 @@ def _enable():
...
 
-def clear():
+def clear(cache_dir=None):
     """
     Clear numba cache directory
 
     Parameters
     ----------
-    None
+    cache_dir : str, default None
```
NimaSarajpoor, line 138: You may want to add `default None` to other functions that accept the parameter `cache_dir`. Btw, should we let users know to what it will be set when not provided?

stumpy/cache.py
```diff
@@ -102,58 +103,77 @@ def _enable():
...
 
     Returns
     -------
     None
     """
-    site_pkg_dir = site.getsitepackages()[0]
-    numba_cache_dir = site_pkg_dir + "/stumpy/__pycache__"
+    global CACHE_CLEARED
+
+    if cache_dir is not None:  # pragma: no cover
```
NimaSarajpoor, line 121: I think we can remove `# pragma: no cover` now

NimaSarajpoor (commented):
@seanlaw 
Just left some minor comments. Other than those, I think it is all good!

NimaSarajpoor:
@seanlaw 
> I've also added a simple/stupid "check" to see if `cache._clear()` has been called before `cache._save()`.

My thought was to just add a note to docstring. It is nice to see that you tried to create some level of protection for users by tracking the calls to those functions.

tests/test_cache.py
seanlaw: I don't understand your question. If we can `cache.save()` after `cache.clear()`, then nothing happens, right? `cache.save()` doesn't actually save anything yet, right?

If we remove the second call to `stump` then the cache directory should still be empty, right?

NimaSarajpoor: > If we remove the second call to stump then the cache directory should still be empty, right?

No. In fact, we just do:

```
cache.clear()
cache.save()  
```
we will have files in cache directory. Those are the index `.nbi` files of ALL njit functions.

And if I do:
```
cache.clear()
stump(T, m)
cache.clear()
cache.save()
```

we not only have all those `.nbi` files again but also `.nbc` files of all those njit functions that were compiled because of running `stump(T, m)`. So, `cache.clear()` clears the on-disk cached files, and `cache.save()` looks at the in-momery cache data and write those `.nbc` files into cache directory.

seanlaw: I don't know, I was mostly just trying to keep your original. What is the point of calling `stump(T, m)? I think we should write the unit test so that it tests the typical process/steps that the user should/would do in real life. I'm confused now.

seanlaw: I'm wondering if we are trying to test too many things in a single test and it's starting to become convoluted? Each test should be obvious and it is not obvious to me what we need to test.

seanlaw: Okay, I played around with it a bit more and I think I understand the behavior now. As a result I added a note to `cache.clear()`. Let me know if it is hard to understand

NimaSarajpoor: @seanlaw 

> I think we should write the unit test so that it tests the typical process/steps that the user should/would do in real life. I'm confused now.

> Each test should be obvious and it is not obvious to me what we need to test.

That is a valid point and it forced me to think more about the `cache.save()` and why it was created in the first place. Originally, we had the following process: If cache is enabled and the user does `cache.recompile()`, it will write cache data into cache directory. The decision was to use `save` as it is more common/meaningful and less technical than "recompile".  

So,  what to test?

(1) In a very basic scenario, a user may just do: 
```
cache.clear() # always the first line

cache.save()
stump(T, m)  # and/or some other stuff
```

OR 
```
cache.clear()  # always the first line

stump(T, m)  #and/or some other stuff
cache.save() 

```

The first scenario is very similar to `cache.enable(); stump(T, m)`. So, going to skip it as it seems trivial. We may test the second scenario though where `cache.save()` is called after calling the function `stump`

(2) Another case is that if user keeps changing the njit function and tries to recompile everytime. In this scenario, they can call `cache.save` everytime but they shouldn't unless they have a valid reason (?) for keep (over)writing cashe files into cache directory. This does not make sense. If they want to change njit function, they can just do `cache.recompile()` and then at the end, when they want to write to cache, just do `cache.save`. This means that this case is similar to the scenario provided in the previous case

----

So, do you think it should suffice if we just test the following process?
```
cache.clear()  # always the first line

stump(T, m)  #and/or some other stuff
cache.save() 
```

seanlaw: > So, do you think it should suffice if we just test the following process?

Yes, I think this is the preferred scenario to show.

NimaSarajpoor: @seanlaw 
> I think we should write the unit test so that it tests the typical process/steps that the user should/would do in real life

I am reading this sentence from your comment again and again, and thinking: "is there a reason for user to clear cache in the middle of process and write again? Unfortunately, I did not consider "the typical process/steps that the user should/would do in real life" when I was writing the test initially. Sorry for the confusion it has created. 

I proposed the following process
```
cache.clear()  # always the first line

stump(T, m)  #and/or some other stuff
cache.save() 
```
in my previous comment. I think this process should be okay. We already tested `cache.recompile` in other tests. Having said that, one may want to check if `cache.recompile` really results in WRITING the up-to-date version of njit function to the cache directory. We haven't tested its "WRITING" part yet. So, I would like to know your opinion on the following test function:

```
def test_cache_save():
    def save_cache(cache_dir):
        cache.clear(cache_dir)
        fastmath._add_assoc(0.0, np.inf)  # 0.0 for fastmath flag True (default)
        fastmath._set("fastmath", "_add_assoc", False)
        cache.save()
        return

    def load_cache(cache_dir):
        cache._enable()
        out = fastmath._add_assoc(0.0, np.inf)
        return out

    cache_dir = "stumpy/__pycache__"
    save_cache(cache_dir)
    out = load_cache(cache_dir)
    assert np.isnan(out)
```

seanlaw: I don't see how this test "proves" that the most up-to-date version of the test is being written. You may call the function `load_cache` all you want but I'm not sure your unit test is "loading"  anything if it's still the same in-memory process. I think the only way to "prove" that the cache is being loaded would be to save, leave the Python session, spin up a new Python session and see if the compile time still exists the second time around. This may be too much

NimaSarajpoor: I noticed the comments but missed the comment you provided here as update. Thought you are still trying to make a decision. I should have added those comments in the first place when I was writing the test as it is not obvious how Numba operates behind the scene. Going to resolve it now.

seanlaw:
@NimaSarajpoor Thank you for your comments and review. I think we ended up in a good place with this PR!

NimaSarajpoor:
@seanlaw 
Do you think this is ready? 

I took a look at the changes again. It seems good to me. Thanks for explaining the behaviour of cache.save() in the section "notes" of its docstring.

PR #1078, opened 2025-03-27
Fixed #1077 Add extra check for window size

### LINKED ISSUE #1077: Raise warning when it is not feasible to get at least one pair of non-trivial motif
As explained [here](https://github.com/TDAmeritrade/stumpy/discussions/1076#discussioncomment-12432542), it is possible to have `T` and `m` that result in 0 subsequence that satisfy the exclusion rule, i.e. `|i - j| > excl_zone`, where `i` and `j` are the start index of two subsequences `T[i, i + m]` and `T[j : j + m]`.

See Example below:
```
import numpy as np
import stumpy

T = np.random.rand(10)
m = 8

out = stumpy.stump(T, m)

# out
[[inf -1 -1 -1]
 [inf -1 -1 -1]
 [inf -1 -1 -1]]
```

In this example, `len(T)` is 10, and `m = 8`, and the start index of subsequences are `0, 1, 2`. However, since the [default] `excl_zone` is `2`, then it means there is no `i,j` in `{0, 1, 2}` that satisfies the exclusion rule. It would be a good idea to raise warning in such scenario. As suggested [here](https://github.com/TDAmeritrade/stumpy/discussions/1076#discussioncomment-12432645), `core.check_window_size` seems to be a good place for raising that warning.

NimaSarajpoor:
See #1077

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -554,11 +554,12 @@ def get_max_window_size(n):
     return max_m
 
 
-def check_window_size(m, max_size=None):
+def check_window_size(m, max_size=None, excl_zone=None, last_start_index=None):
     """
     Check the window size and ensure that it is greater than or equal to 3 and, if
     `max_size` is provided, ensure that the window size is less than or equal to the
-    `max_size`
+    `max_size`. Furthermore, if `excl_zone` is provided, then it will also check if the
+    window size is too large and could lead to meaningless results.
 
     Parameters
     ----------
@@ -567,6 +568,13 @@ def check_window_size(m, max_size=None):
 
     max_size : int, default None
         The maximum window size allowed
+
+    excl_zone : int, default None
+        The exclusion zone. If provided, then the `last_start_index` must also be
+        provided.
+
+    last_start_index : int, default None
+        The start index of last subsequence.
 
     Returns
     -------
@@ -588,6 +596,37 @@ def check_window_size(m, max_size=None):
 
     if max_size is not None and m > max_size:
         raise ValueError(f"The window size must be less than or equal to {max_size}")
+
+    if excl_zone is not None:
+        if last_start_index is None:
+            raise ValueError(
+                "last_start_index must be provided when excl_zone is not None"
+            )
+
+        # Check if subsequneces have non-trivial neighbours
+
+        # Case 1:
+        # There is at least one subsequence with non-trivial neighbour
+        # i.e. For AN `i`, there exists at least one `j` such that |i - j| > excl_zone
+        # In this case, we just need to consider the two subsequences that are furthest
+        # apart from each other.
+        # In other words: |0 - last_start_index| > excl_zone
+        cond_1 = last_start_index <= excl_zone
+
+        # Case 2:
+        # Check if each single subsequence has at least one non-trivial neighbor
+        # i. e. For ANY `i`, there exists at least one `j` such that |i - j| > excl_zone
+        # In this case, we need to consider the subseuqence whose furthest neighbour is
+        # the shortest compared to other subsequences.
+        # In other words: |0 - ceil(last_start_index / 2)| > excl_zone
+        cond_2 = math.ceil(last_start_index / 2) <= excl_zone
+
+        if cond_1 or cond_2:
+            msg = (
+                f"The window size, 'm = {m}', may be too large and could lead to "
+                + "meaningless results. Consider reducing 'm' where necessary"
+            )
+            warnings.warn(msg)
 
 
 @njit(fastmath=config.STUMPY_FASTMATH_TRUE)
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -595,34 +595,45 @@ def check_window_size(m, max_size=None, n=None):
         raise ValueError(f"The window size must be less than or equal to {max_size}")
 
     if n is not None:
-        # Raise warning if there is at least one subsequence with no
-        # non-trivial neighbour in the case of a self-join.
+        # Raise warning if there is at least one subsequence with no eligible
+        # (non-trivial) neighbor in the case of a self-join.
 
         # For any time series `T`, an "eligible nearest neighbor" subsequence for
         # the central-most subsequence must be located outside the `excl_zone`,
-        # and the central-most subsequence will ALWAYS have the smallest index-wise
-        # distance to its furthest neighbour amongs all other subsequences. Therefore,
-        # we only need to check whether the `excl_zone` eliminates all "neighbors" for
-        # the central-most subsequence in `T`. In fact, we just need to verify whether
-        # the `excl_zone` eliminates the "neighbor" that is furthest away (index-wise)
+        # and the central-most subsequence will ALWAYS have the smallest relative 
+        # (index-wise) distance to its farthest neighbor amongst all other subsequences. 
+        # Therefore, we only need to check whether the `excl_zone` eliminates all "neighbors"
+        # for the central-most subsequence in `T`. In fact, we just need to verify whether
+        # the `excl_zone` eliminates the "neighbor" that is farthest away (index-wise)
         # from the central-most subsequence. If it does not, this implies that all
         # subsequences in `T` will have at least one "eligible nearest neighbors"
         # outside their respective `excl_zone`
 
         excl_zone = int(math.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
 
-        l = n - m + 1
+        l = n - m + 1 
+        # The start index of subsequences are: 0, 1, ..., l-1
 
-        # If `l` is odd (`l == 2k+1`):
-        # The central subsequence is at index `k`, with furthest neighbors at `0`
-        # and `2k`, both `k == l // 2` indices away from the central-most subsequence.
+        # If `l` is odd (hence, `l == 2c+1`):
+        # The central subsequence is located at index `c == l // 2`, 
+        # with two farthest neighbors, one located at index `0`(to the left of `c`)
+        # and the other located at index `l - 1 == 2c` (to the right of `c`). 
+        # Both indices `0` and `2c` are exactly `c == l // 2` index positions away
+        # from the central subsequence located at index `c`.
 
-        # If `l` is even (`l == 2k`):
-        # The central subsequences are at `k-1` and `k`. The furthest neighbor is
-        # at `2k-1` for `k-1`, and `0` for `k`. In both cases, the subsequence
-        # and its furthest neighbor are `k == l // 2` indices away.
-        max_gap = l // 2
-        if max_gap <= excl_zone:
+        # If `l` is even (hence, `l == 2c`):
+        # There are two central-most subsequences located at index locations 
+        # `c` and `c-1`. For subsequence at `c`, its farthest neighbor will be
+        # located at index `0` (to the left of `c`) and, for `c-1`, its farthest
+        # neighbor is located at index `l - 1 == 2c - 1` (to the right of `c-1`).
+        # In both cases, each central subsequence and its farthest neighbor are 
+        # `c == l // 2` indices away.
+
+        # Therefore, regardless if `l` is even or odd, for the central
+        # subsequence for any time series, the index location of its 
+        # farthest neighbor will always be `l // 2` index positions away.
+        diff_to_farthest_idx = l // 2
+        if diff_to_farthest_idx <= excl_zone:
             msg = (
                 f"The window size, 'm = {m}', may be too large and could lead to "
                 + "meaningless results. Consider reducing 'm' where necessary"
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -614,14 +614,14 @@ def check_window_size(m, max_size=None, n=None):
         l = n - m + 1
         # The start index of subsequences are: 0, 1, ..., l-1
 
-        # If `l` is odd (hence, `l == 2c+1`):
+        # If `l` is odd (`l` can be expressed as `l == 2c+1`):
         # The central subsequence is located at index `c == l // 2`,
         # with two farthest neighbors, one located at index `0`(to the left of `c`)
         # and the other located at index `l - 1 == 2c` (to the right of `c`).
         # Both indices `0` and `2c` are exactly `c == l // 2` index positions away
         # from the central subsequence located at index `c`.
 
-        # If `l` is even (hence, `l == 2c`):
+        # If `l` is even (`l` can be expressed as `l == 2c`):
         # There are two central-most subsequences located at index locations
         # `c` and `c-1`. For subsequence at `c`, its farthest neighbor will be
         # located at index `0` (to the left of `c`) and, for `c-1`, its farthest
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -611,7 +611,7 @@ def check_window_size(m, max_size=None, n=None):
         # For any `i` and its corresponding `jmax`, we have:
         # |i - jmax| >= d > excl_zone
 
-        # The minimum |i - jmax| is achieved when `S_i` is the middle ubsequence,
+        # The minimum |i - jmax| is achieved when `S_i` is the middle subsequence,
         # i.e. i == int(ceil((n - m) / 2)), and its corresponding jmax is 0. Hence,
         # we just need to make sure the following inequity is satisfied:
         # |int(ceil((n - m) / 2)) - 0| > excl_zone`
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -1412,7 +1412,7 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0, query_idx=None):
         raise ValueError(f"`Q` is {Q.ndim}-dimensional and must be 1-dimensional. ")
     Q_isfinite = np.isfinite(Q)
 
-    check_window_size(m, max_size=Q.shape[-1])
+    check_window_size(m, max_size=Q.shape[0])
 
     if query_idx is not None:  # pragma: no cover
         query_idx = int(query_idx)
@@ -1759,7 +1759,7 @@ def mass(
         raise ValueError(f"Q is {Q.ndim}-dimensional and must be 1-dimensional. ")
     Q_isfinite = np.isfinite(Q)
 
-    check_window_size(m, max_size=Q.shape[-1])
+    check_window_size(m, max_size=Q.shape[0])
 
     if query_idx is not None:
         query_idx = int(query_idx)
@@ -1984,7 +1984,7 @@ def mass_distance_matrix(
         T_subseq_isconstant=T_subseq_isconstant,
     )
 
-    check_window_size(m, max_size=min(Q.shape[-1], T.shape[-1]))
+    check_window_size(m, max_size=min(Q.shape[0], T.shape[0]))
 
     return _mass_distance_matrix(
         Q,
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -598,7 +598,7 @@ def check_window_size(m, max_size=None, excl_zone=None, last_start_index=None):
         raise ValueError(f"The window size must be less than or equal to {max_size}")
 
     if excl_zone is not None:
-        if last_start_index is None:
+        if last_start_index is None:  # pragma: no cover
             raise ValueError(
                 "last_start_index must be provided when excl_zone is not None"
             )
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -558,8 +558,8 @@ def check_window_size(m, max_size=None, n=None):
     """
     Check the window size and ensure that it is greater than or equal to 3 and, if
     `max_size` is provided, ensure that the window size is less than or equal to the
-    `max_size`. Furthermore, if `n` is provided, then it checks whether all
-    subsequences have at least one non-trivial neighbor in a self-join.
+    `max_size`. Furthermore, if `n` is provided, then a self-join is assumed and it
+    checks whether all subsequences have at least one non-trivial neighbor.
 
     Parameters
     ----------
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -570,8 +570,8 @@ def check_window_size(m, max_size=None, excl_zone=None, last_start_index=None):
         The maximum window size allowed
 
     excl_zone : int, default None
-        Size of the exclusion zone. If provided, then the `last_start_index` must also be
-        provided.
+        Size of the exclusion zone. If provided, then the `last_start_index`
+        must also be provided.
 
     last_start_index : int, default None
         Start index of the last subsequence
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -570,7 +570,7 @@ def check_window_size(m, max_size=None, n=None):
         The maximum window size allowed
 
     n : int, default None
-        The length of the time series.
+        The length of the time series in a self-join case
 
     Returns
     -------
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -611,9 +611,12 @@ def check_window_size(m, max_size=None, n=None):
         # For any `i` and its corresponding `jmax`, we have:
         # |i - jmax| >= d > excl_zone
 
-        # The minimum |i - jmax| is achieved when `S_i` is the middle subsequence,
-        # i.e. i == int(ceil((n - m) / 2)), and its corresponding jmax is 0. Hence,
-        # we just need to make sure the following inequity is satisfied:
+        # Hence, as long as the `S_i` that corresponds to `d` has one non-trivial
+        # neighbour, any other subsequence has one non-trivial neighbour as well.
+
+        # The minimum |i - jmax| is achieved when `S_i` is the middle ubsequence,
+        # i.e. i == int(ceil((n - m) / 2)). Its corresponding `jmax` is 0. Hence,
+        # we just need to make sure the following inequality is satisfied:
         # |int(ceil((n - m) / 2)) - 0| > excl_zone`
 
         excl_zone = int(math.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -615,14 +615,13 @@ def check_window_size(m, max_size=None, n=None):
 
         # If `l` is odd (`l == 2k+1`):
         # The central subsequence is at index `k`, with furthest neighbors at `0`
-        # and `2k`, both `k == l // 2` indices away.
+        # and `2k`, both `k == l // 2` indices away from the central-most subsequence.
 
         # If `l` is even (`l == 2k`):
         # The central subsequences are at `k-1` and `k`. The furthest neighbor is
         # at `2k-1` for `k-1`, and `0` for `k`. In both cases, the subsequence
         # and its furthest neighbor are `k == l // 2` indices away.
         max_gap = l // 2
-
         if max_gap <= excl_zone:
             msg = (
                 f"The window size, 'm = {m}', may be too large and could lead to "
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -594,33 +594,20 @@ def check_window_size(m, max_size=None, n=None):
         raise ValueError(f"The window size must be less than or equal to {max_size}")
 
     if n is not None:
-        # The following code raises warning if there is at least one subsequence
-        # with no non-trivial neighbor. The following logic does not check if
-        # a subsequence has a non-finite value.
-
-        # Logic: For each subsequnece `S_i = T[i : i + m]`, its neighbor `S_j`
-        # is non-trivial if |i - j| > excl_zone. Let's denote `S_jmax` as
-        # the neighbor that is furthest away from `S_i` (index-wise). So:
-        # |i - jmax| >= |i - j|
-        # Therefore, if `S_i` has at least one non-trivial neighbor, then `S_jmax` is
-        # definitely a non-trivial neighbor. Because:
-        # |i - jmax| >= |i - j| > excl_zone
-        # To ensure ALL subsequences have at least one non-trivial neighbor, we can just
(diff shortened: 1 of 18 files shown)


seanlaw (changes requested):
Can you please provide a few test cases? It's hard to know what the parameters are related to

NimaSarajpoor (commented):
@seanlaw 
I've added two tests. I left some comments for your consideration.

stumpy/core.py
```diff
@@ -554,11 +554,12 @@ def get_max_window_size(n):
     return max_m
 
 
-def check_window_size(m, max_size=None):
+def check_window_size(m, max_size=None, excl_zone=None, last_start_index=None):
```
seanlaw, line 557: I feel like you're making this overly complicated. I think that you can get by changing the parameters of to:

```
def check_window_size(m, max_size=None, n=None):
    if max_size is not None and m > max_size:
        raise ValueError(f"The window size must be less than or equal to {max_size}")

    if n is not None:
        if excl_zone + m > n or math.ceil(last_start_index / 2) < excl_zone:
            msg = (
                f"The window size, 'm = {m}', may be too large and could lead to "
                + "meaningless results. Consider reducing 'm' where necessary"
            )
            warnings.warn(msg)
```

Please check the logic of the final `if` condition. I would then add a comment to explain both conditions

seanlaw: `last_start_index` doesn't read well

stumpy/core.py
```diff
@@ -589,6 +597,37 @@ def check_window_size(m, max_size=None):
...
+
+        # Case 2:
+        # Check if each single subsequence has at least one non-trivial neighbor
+        # i. e. For ANY `i`, there exists at least one `j` such that |i - j| > excl_zone
+        # In this case, we need to consider the subseuqence whose furthest neighbour is
+        # the shortest compared to other subsequences.
+        # In other words: |ceil(last_start_index / 2) - 0| > excl_zone
+        cond_2 = (math.ceil(last_start_index / 2) - 0) > excl_zone
+
+        if not cond_1 or not cond_2:
```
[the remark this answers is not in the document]
seanlaw: > However, if cond_2 is satisfied, it means cond_1 is satisfied as well. 

If you are sure about this then, yes. Please explain the logic. Even after reading your comment, I still don't fully understand it.

stumpy/stump.py
```diff
@@ -711,20 +711,25 @@ def stump(
...
         diags = np.arange(excl_zone + 1, n_A - m + 1, dtype=np.int64)
     else:
+        excl_zone = None
         diags = np.arange(-(n_A - m + 1) + 1, n_B - m + 1, dtype=np.int64)
 
+    core.check_window_size(
+        m,
+        max_size=min(T_A.shape[0], T_B.shape[0]),
+        excl_zone=excl_zone,
+        last_start_index=l - 1,
```
seanlaw, line 730: Is the issue that we are really only concerned with `excl_zone` when it is a self-join and not an AB-join?

seanlaw: As a bit of feedback for you to consider, the code that you've proposed feels like you are trying to combine too many things together and it is becoming unreadable. Even if your logic is correct, It's impossible for others to grasp your logic by looking at the code and that's not good for long term maintenance.

seanlaw (changes requested):

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: > If you are sure about this then, yes.

Cool 👍 

> Please explain the logic. Even after reading your comment, I still don't fully understand it.

Noted. I will work on its clarity.

stumpy/core.py
NimaSarajpoor: Agree! 

I can see `excl_zone` is computed inside the function which is better because it results in less input params (It also helps reader to quickly see its relationship with the window size `m`)

> `last_start_index` doesn't read well. By adding `n` as a parameter instead, it feels more natural

Right... It makes it easier to keep in mind what the param represents.

stumpy/stump.py
NimaSarajpoor: > Is the issue that we are really only concerned with `excl_zone` when it is a self-join and not an AB-join?

Correct. In AB-join, the neighbours of a sequence exist in another time series. So, as long as the length of each time series is at least `m`, a subsequence in one has a neighbour in the other.

>  feels like you are trying to combine too many things together

I was trying to do exactly that!! I thought it might be better to keep the if-else simple, and then call  `core.check_window_size` only once. I can see that calling it twice in the if-else block actually makes it easier to understand!


>  Even if your logic is correct, It's impossible for others to grasp your logic by looking at the code and that's not good for long term maintenance.

Totally true! Thanks for the feedback!!

tests/test_core.py
```diff
@@ -192,6 +192,26 @@ def test_check_max_window_size():
             core.check_window_size(m, max_size=3)
 
 
+def test_check_window_size_excl_zone_case1():
```
NimaSarajpoor, line 195: This test is redundant. The ` test_check_window_size_excl_zone_case2` should suffice.

stumpy/core.py
```diff
@@ -589,6 +593,37 @@ def check_window_size(m, max_size=None):
     if max_size is not None and m > max_size:
         raise ValueError(f"The window size must be less than or equal to {max_size}")
 
+    if n is not None:
```
NimaSarajpoor, line 596: @seanlaw 
Can you please read the comments and let me know if the logic is clear?

stumpy/core.py
```diff
@@ -611,7 +611,7 @@ def check_window_size(m, max_size=None, n=None):
         # For any `i` and its corresponding `jmax`, we have:
         # |i - jmax| >= d > excl_zone
 
-        # The minimum |i - jmax| is achieved when `S_i` is the middle ubsequence,
+        # The minimum |i - jmax| is achieved when `S_i` is the middle subsequence,
         # i.e. i == int(ceil((n - m) / 2)), and its corresponding jmax is 0. Hence,
         # we just need to make sure the following inequity is satisfied:
```
seanlaw, line 616: "inequality"

NimaSarajpoor:
@seanlaw 
IIRC, you mentioned before that we can simply check the condition for all subsequences. After going through the long explanation of my logic, I am thinking of changing the approach.

```
l = n - m + 1
# leftmost neighbor is at index `0`
# rightmost neighbor is at index `l-1`
for i in range(l):
     max_neighbor_gap = max(i - 0, (l-1)-i)
     if max_neighbor_gap <= excl_zone:
         msg = ....
         warnings.warn(msg)
         break
```

What is your opinion?

seanlaw:
I don't mind it if the time is negligible but I wonder if we can either replace the for loop with numpy vectorized functions?

NimaSarajpoor:
> but I wonder if we can either replace the for loop with numpy vectorized functions?

This should work.

```
l = n - m + 1
indices = np.arange(l)

# leftmost neighbor is at index `0`
# rightmost neighbor is at index `l-1`
max_gaps = np.maximum(indices - 0, (l - 1) - indices)

if np.any(max_gaps <= excl_zone):    # ALTERNATIVELY: np.min(max_gaps) <= excl_zone
    msg = ...   #  warning message
    warnings.warn(msg)
```

> I don't mind it if the time is negligible

Cool. Will check it.

stumpy/core.py
NimaSarajpoor: Please ignore for now as I've changed the approach.

seanlaw:
> This should work.

Hmmm, it feels like there's likely a simple equation that can be used rather than checking/generating the full set of `indices`

NimaSarajpoor:
> For any time series, T, an "eligible nearest neighbor" subsequence for the central-most subsequence (located at index i = l // 2) must be located outside of the excl_zone and this central-most subsequence will ALWAYS have the smallest number of "eligible nearest neighbors" amongst all other subsequences. Thus, all we need to do is check whether the excl_zone eliminates all "eligible nearest neighbors" for the central-most subsequence in T. If it doesn't, then that implies that all other subsequences within T will have at least one or more eligible nearest neighbor(s) that will lie outside of their respective excl_zone

That was the logic I was trying to explain in that very long comment. You have a strong skill for delivering a concise message while avoiding verbosity!!!


> Thus, all we need to do is check whether the excl_zone eliminates all "eligible nearest neighbors" for the central-most subsequence in T

YES! EXACTLY!! And, in fact, we just need to check if its furthest neighbour is eliminated. For the central-most subsequence that starts at `math.ceil((l-1) / 2))`, its furthest neighbour is located at 0.


> I think it's simply if l // 2 <= excl_zone

> It looks like there might be a caveat if l has an even length?

I think `math.ceil((l - 1) / 2))` should work.

```
l = n - m + 1
minmax_gap = math.ceil((l-1) / 2)) - 0   # 0 is the leftmost neighbour

# `l` is even. Indices are  {0, 1, 2, 3, 4, 5}: minmax_gap = 3
# `l` is odd. Indices are  {0, 1, 2, 3, 4}: minmax_gap = 2
```

NimaSarajpoor:
@seanlaw 
Okay... Now that we have a more concise, less verbose explanation, I will switch back to my previous approach and will check for that central-most subsequence. I will replace my comment with the one you shared.

seanlaw:
> Okay... Now that we have a more concise, less verbose explanation, I will switch back to my previous approach and will check for that central-most subsequence. I will replace my old comment with the one you shared.

Sounds good!

> You have a strong skill for delivering a concise message while avoiding verbosity!!!

I think that it requires that you step into the shoes of a not-so-technical user and think about what it is that you want to teach them (while providing enough orientation and background). In this case, you read the "final" draft of my explanation but what you didn't get to see was how many times I wrote, deleted, edited, re-edited, started over from scratch, and then eventually worked out a reasonably acceptable explanation. So, please don't think that "it just comes out". In addition to writing, a writer must also learn the more important skill of self-editing, which I still continue to hone (and you can too!).

stumpy/stump.py
```diff
@@ -711,18 +711,17 @@ def stump(
...
     n_A = T_A.shape[0]
     n_B = T_B.shape[0]
     l = n_A - m + 1
 
+    ignore_trivial = core.check_ignore_trivial(T_A, T_B, ignore_trivial)
     excl_zone = int(np.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
-
-    if ignore_trivial:
+    if ignore_trivial:  # self-join
+        core.check_window_size(m, max_size=n_A, n=n_A)
```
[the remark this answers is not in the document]
seanlaw: but it's a self-join so `n_A == n_B`.  `n_B` really has little to no meaning for a self-join

stumpy/core.py
```diff
@@ -568,6 +569,9 @@ def check_window_size(m, max_size=None):
     max_size : int, default None
         The maximum window size allowed
 
+    n : int, default None
+        The length of the time series in a self-join case
```
seanlaw, line 573: "The length of the time series in the case of a self-join. `n` should be set to `None` in the case of an AB-join.

stumpy/core.py
```diff
@@ -589,6 +594,41 @@ def check_window_size(m, max_size=None):
...
         raise ValueError(f"The window size must be less than or equal to {max_size}")
 
+    if n is not None:
+        # Raise warning if there is at least one subsequence with no
+        # non-trivial neighbour in the case of a self-join.
+
+        # For any time series `T`, an "eligible nearest neighbor" subsequence for
+        # the central-most subsequence must be located outside the `excl_zone`,
+        # and the central-most subsequence will ALWAYS have the smallest index-wise
+        # distance to its furthest neighbour amongs all other subsequences. Therefore,
```
NimaSarajpoor, line 604: @seanlaw 

Regarding: 

> this central-most subsequence will ALWAYS have the smallest number of "eligible nearest neighbors" amongst all other subsequences. 

All subsequences with a start index that is at least `excl_zone` indices away from index `0` and `len(T) - m`  have the same number of neighbours. The central-most subsequence, however, has the smallest value for `max(n_left_neighbours, n_right_neighbours)`.

To address this, I've changed `the smallest number of "eligible nearest neighbors"` to `the smallest index-wise distance to its furthest neighbour`

---

Do you know a better/clearer phrase instead of `index-wise distance`?

stumpy/core.py
```diff
@@ -589,6 +594,41 @@ def check_window_size(m, max_size=None):
...
+
+        # If `l` is odd (`l == 2k+1`):
+        # The central subsequence is at index `k`, with furthest neighbors at `0`
+        # and `2k`, both `k == l // 2` indices away from the central-most subsequence.
+
+        # If `l` is even (`l == 2k`):
+        # The central subsequences are at `k-1` and `k`. The furthest neighbor is
+        # at `2k-1` for `k-1`, and `0` for `k`. In both cases, the subsequence
+        # and its furthest neighbor are `k == l // 2` indices away.
+        max_gap = l // 2
```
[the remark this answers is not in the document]
seanlaw: I think we should be a bit more verbose as you are packing too much information in one sentence.

```
        # The central subsequence is located at index `c == l // 2`, with its farthest eligible neighbors located at index `0`(to the left of `c`)
        # and at index `l == 2c` (to the right of `c`). Both indices `0` and `l == 2c` are exactly `l // 2` index positions away from the central-most subsequence located at index `c`.

        # If `l` is even (`l == 2c`):
        # There are two central subsequences located at index locations `c` and `c-1`. For `c`, its farthest eligible neighbor will be located at index `0` (to the left of `c`) and, for `c-1`, its farthest eligible neighbor is located 
        # at index `l == 2c-1` (to the right of `c-1`). In both cases, each central subsequence
        # and its farthest neighbor are `c == l // 2` indices away.
        # Therefore, regardless if `l` is even or add, for the central
        # subsequence for any time series, the index location of its 
        # farthest eligible neighbor will always be `l // 2` index
        # positions away
        farthest_idx = l // 2
```

Notice that I use "farthest" (rather than "furthest") since we are talking about the literal distance. And I've renamed `max_gap` to the more appropriate `farthest_idx`

stumpy/core.py
```diff
@@ -589,6 +594,41 @@ def check_window_size(m, max_size=None):
...
+        if max_gap <= excl_zone:
```
seanlaw, line 625: `if farthest_idx <= excl_zone`

I think that communicates the idea very well now!

seanlaw (changes requested):

stumpy/core.py
```diff
@@ -589,6 +594,41 @@ def check_window_size(m, max_size=None):
...
+        # the "neighbor" that is furthest away (index-wise) from the central-most
+        # subsequence. If it does not, this implies that all subsequences in `T`
+        # will have at least one or more "eligible nearest neighbors" outside
+        # their respective `excl_zone`
+
+        excl_zone = int(math.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
+
+        l = n - m + 1
+
+        # If `l` is odd (`l == 2k+1`):
```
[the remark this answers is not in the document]
NimaSarajpoor: > It's probably best NOT to use k and to choose a different variable since k is reserved for top-K. 

👍 

> What is `k` suppose to be?

Simply `l // 2`

> Maybe use c?

Sounds good. My main point is to help reader understand why the central-most subsequence is at `l // 2`. So, when I say an odd number `l` can be represented as `2 * c + 1`, I assume the reader can quickly picture the following in their mind when we talk about central-most subsequence.

```
< c number of elements> <one element> <c number of elements>
```

 This shows that the central-most subsequence is at index `c`, which is `l // 2`. 


> [From another comment] I think we should be a bit more verbose 

I get it now after reading my explanation above 😅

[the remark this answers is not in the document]
seanlaw: > So, when I say an odd number l can be represented as 2 * c + 1, I assume the reader can quickly picture the following in their mind when we talk about central-most subsequence.

Honestly/embarrassingly, even now that I understand it, I do not think that my future self can quickly picture it in my mind. So, it better to lay it out clearly. I noticed that you tend to state the facts. However, you fail to explain the relevance of the facts clearly and then to connect the facts together in an obvious/coherent way. Having worked with you, I know that you've gained the intuition in your brain and things "make sense" but now we must describe all of that intuition in an easy-to-follow way by making all of the "connections" explicit (no matter how obvious it may seem to you).

stumpy/core.py
seanlaw: Hmmm, I guess, technically, it's not really the "farthest index"? Saying "distance" doesn't feel right either because it gets confused with Euclidean distance.

seanlaw: Ahh, I see that you've come to a similar conclusion in https://github.com/TDAmeritrade/stumpy/pull/1078#discussion_r2023928170 with "index-wise distance". Hmmm...

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: > you are packing too much information in one sentence.

Need to work on finding that balance! 

---

I think what you proposed is much clearer!! Shared a few notes below. Please let me know what you think.

(1) Maybe we should avoid using the term "eligible" here (?) Because we haven't checked yet if `excl_zone` eliminates that or not. So, it is just a neighbor.

(2) Going to apply a small fix: 

> and at index `l == 2c` (to the right of `c`)

at the index `l - 1`, which is `2c` (to the right of `c`)

> at index `l == 2c-1` (to the right of `c-1`).

at the index `l - 1`, which is `2c - 1` (to the right of `c`)


(3) Regarding:

> farthest_idx = l // 2

I have a tiny feeling that one may make a mistake in the future and think `farthest_idx` is about the INDEX that is far away from another index. However, it is more about the difference between two indices. Do you think we should use something like `farthest_diff_idx` instead? I can see the comment though just above that line. So, if reader pays attention to the comment (and they should), we are fine (??)

stumpy/core.py
NimaSarajpoor: Or `farthest_idx_diff`?

seanlaw: I think I like `diff_to_farthest_idx`. It's long but, to your point, we've described it in the comment and we'll only use this variable once so it's not a big deal.

stumpy/core.py
seanlaw: I think "the smallest relative (index-wise) distance to its farthest neighbor"

Also, "amongst" is missing a "t"

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: >  noticed that you tend to state the facts. However, you fail to explain the relevance of the facts clearly and then to connect the facts together in an obvious/coherent way. 

My research's supervisor was right!

> we must describe all of that intuition in an easy-to-follow way by making all of the "connections" explicit (no matter how obvious it may seem to you).

Correct.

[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
> So, when I say an odd number l can be represented as 2 * c + 1, I assume the reader can quickly picture the following in their mind when we talk about central-most subsequence.

> `< c number of elements> <one element> <c number of elements>`

After reading your comment, I added more explanation regarding this locally...but then I decided to remove that part as I realized it does not add a value.

stumpy/core.py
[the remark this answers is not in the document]
NimaSarajpoor: > 1) Maybe we should avoid using the term "eligible" here (?) Because we haven't checked yet if excl_zone eliminates that or not. So, it is just a neighbor.

I've eliminated the term `eligible` in some parts of the comments. If we notice the explanation becomes unclear, I will add them again.

stumpy/core.py
```diff
@@ -554,11 +554,12 @@ def get_max_window_size(n):
...
 
 
-def check_window_size(m, max_size=None):
+def check_window_size(m, max_size=None, n=None):
     """
     Check the window size and ensure that it is greater than or equal to 3 and, if
     `max_size` is provided, ensure that the window size is less than or equal to the
-    `max_size`
+    `max_size`. Furthermore, if `n` is provided, then it checks whether all
+    subsequences have at least one non-trivial neighbor in a self-join.
```
seanlaw, line 562: Furthermore, if `n` is provided, then a self-join is assumed and it checks whether all subsequences have at least one non-trivial neighbor.

stumpy/core.py
```diff
@@ -589,6 +594,52 @@ def check_window_size(m, max_size=None):
...
+
+        # For any time series `T`, an "eligible nearest neighbor" subsequence for
+        # the central-most subsequence must be located outside the `excl_zone`,
+        # and the central-most subsequence will ALWAYS have the smallest relative
+        # (index-wise) distance to its farthest neighbor amongst all other subsequences.
+        # Therefore, we only need to check whether the `excl_zone` eliminates all
+        # "neighbors" for the central-most subsequence in `T`. In fact, we just need to
+        # verify whether the `excl_zone` eliminates the "neighbor" that is farthest
+        # away (index-wise) from the central-most subsequence. If it does not, this
+        # implies that all subsequences in `T` will have at least one "eligible nearest
```
seanlaw, line 609: implies that all subsequences in `T` will have at least one "eligible nearest neighbor" that is located outside of their respective `excl_zone`

stumpy/core.py
```diff
@@ -589,6 +594,52 @@ def check_window_size(m, max_size=None):
...
+        # neighbors" outside their respective `excl_zone`
+
+        excl_zone = int(math.ceil(m / config.STUMPY_EXCL_ZONE_DENOM))
+
+        l = n - m + 1
+        # The start index of subsequences are: 0, 1, ..., l-1
+
+        # If `l` is odd (`l` can be expressed as `l == 2c+1`):
```
seanlaw, line 617: Here, you refer to `c` but nobody knows what it is until the next line

stumpy/mstump.py
```diff
@@ -1228,7 +1228,9 @@ def mstump(
         err = f"T is {T_A.ndim}-dimensional and must be at least 1-dimensional"
         raise ValueError(f"{err}")
 
-    core.check_window_size(m, max_size=min(T_A.shape[1], T_B.shape[1]))
+    # mstump currently supports self-join. Therefore, the argument `n` should be
+    # passed to the function `core.check_window_size`.
+    core.check_window_size(m, max_size=min(T_A.shape[1], T_B.shape[1]), n=T_A.shape[1])
```
seanlaw, line 1233: I'm noticing an inconsistency where sometimes we use `T.shape[1]` and sometimes we use `T.shape[-1]`. Given that `T` can only be 2 dimensional, I wonder if it is better/safer to choose `T.shape[1]`?

stumpy/mstump.py
```diff
@@ -1228,7 +1228,9 @@ def mstump(
         err = f"T is {T_A.ndim}-dimensional and must be at least 1-dimensional"
         raise ValueError(f"{err}")
 
-    core.check_window_size(m, max_size=min(T_A.shape[1], T_B.shape[1]))
+    # mstump currently supports self-join. Therefore, the argument `n` should be
```
seanlaw, line 1231: What is this comment for?

seanlaw (changes requested):
Everything looks pretty good. I left a few minor comments

stumpy/core.py
NimaSarajpoor: Good point! Was not sure about adding something like `, where c is l // 2`. I think I need to add that. And I think that should suffice. Right? 

I mean...at this point, I am not talking about any subsequence's index. I just want to share that an odd value can be expressed (represented / written) as `2 * c + 1`

stumpy/mstump.py
NimaSarajpoor: I noticed it too. Do you want to address this (and any similar cases in other modules) here?

stumpy/mstump.py
NimaSarajpoor: IIRC, `mstump` only supports self-join and there was no study for a AB-join. IMHO, it does not even make sense to have AB-join for `mstump`. But, I might be wrong, and a paper might be published in the future to show `mstump` for AB-join. So, we may need to update this module to add such support. In that case, we probably need to avoid passing the argument `n` to the function `core.check_window_size` . 

The comment is to help our future selves to not forget that.

stumpy/mstump.py
seanlaw: Yes, I think it is okay to address it here in this PR.

stumpy/core.py
seanlaw: > I just want to let user know that an odd value can be expressed (represented / written) as 2 * c + 1

Why is this point important? Why is it important to state this here? Maybe a concrete example with `l = 5` would help to explain?

stumpy/mstump.py
seanlaw: Maybe "the argument `n=T_A.shape[1]` should be passed..."

stumpy/core.py
NimaSarajpoor: > Why is this point important? 

To help reader follow the logic (more easily?) I could have avoided using it, and just used `l // 2` instead. However, I think that makes it harder for the readers to realize that `(l - 1) - (l // 2)` is equal to `l // 2` when `l` is odd. 

> Why is it important to state this here? 

So, if user knows / understands in advance that an odd value `l` can be expressed as `2 * c + 1`, then they can quickly see: `l - 1` is `2c`, and `(l - 1) - (l // 2)` gives `c`.

> Maybe a concrete example with l = 5 would help to explain?

I can add that example and then conclude the logic. Do you have something like this in mind?

```
# If `l` is odd
# Suppose `l == 5`. So, the start index of the subsequences are: 0, 1, 2, 3, 4
# The central subsequence is located at index 2, with two farthest neighbors, one
# located at index 0, and the other is located at index 4. In both cases, the relative 
# (index-wise) distance is 2, which is `5 // 2`. In general, it can be shown that the distance from
# the central subsequence to its furthest neighbor is `l // 2`.
```

I can do the same for the case ` # If l is even`

seanlaw: The example is very easy to follow. I think it is okay to say "The central subsequence is located at index position `c = 2`"

stumpy/mstump.py
```diff
@@ -1228,8 +1228,8 @@ def mstump(
         err = f"T is {T_A.ndim}-dimensional and must be at least 1-dimensional"
         raise ValueError(f"{err}")
 
-    # mstump currently supports self-join. Therefore, the argument `n` should be
-    # passed to the function `core.check_window_size`.
+    # mstump currently supports self-join. Therefore, the argument `n=T_A.shape[1]`
```
seanlaw, line 1231: "currently only supports self-joins"

stumpy/core.py
```diff
@@ -568,6 +569,10 @@ def check_window_size(m, max_size=None):
     max_size : int, default None
         The maximum window size allowed
 
+    n : int, default None
+        The length of the time series in the case of a self-join.
+        `n` should be set to `None` in the case of an `AB-join`.
```
seanlaw, line 574: "``n`` should not be supplied (or set to ``None``) in the case of an ``AB-join``"

Note that for docstrings, all variables MUST be enclosed in TWO backticks

stumpy/mstumped.py
```diff
@@ -505,7 +505,9 @@ def mstumped(
         err = f"T is {T_A.ndim}-dimensional and must be at least 1-dimensional"
         raise ValueError(f"{err}")
 
-    core.check_window_size(m, max_size=min(T_A.shape[1], T_B.shape[1]))
+    # mstump currently supports self-join. Therefore, the argument `n` should be
```
seanlaw, line 508: "currently only supports self-joins"

stumpy/mstump.py
```diff
@@ -1228,7 +1228,9 @@ def mstump(
         err = f"T is {T_A.ndim}-dimensional and must be at least 1-dimensional"
         raise ValueError(f"{err}")
 
-    core.check_window_size(m, max_size=min(T_A.shape[1], T_B.shape[1]))
+    # mstump currently supports self-join. Therefore, the argument `n=T_A.shape[1]`
+    # should be passed to the function `core.check_window_size`.
```
seanlaw, line 1232: "must be passed"   

"should be" implies uncertainty and too passive

seanlaw (changes requested):
@NimaSarajpoor Everything looks good on my end aside from a few minor changes. Please let me know if this is ready to be merged.

stumpy/core.py
NimaSarajpoor: > "`n` should not be supplied (or set to `None`) in the case of an `AB-join`"

👍 

> Note that for API docstrings (that are shown to users in our readthedocs), all variables MUST be enclosed in TWO backticks instead of SINGLE backticks.

Didn't know that! Thanks for the info. Will fix it.

stumpy/mstump.py
NimaSarajpoor: Cool note! Understood 👍

stumpy/core.py
NimaSarajpoor: @seanlaw 

> > Note that for API docstrings (that are shown to users in our readthedocs), all variables MUST be enclosed in TWO backticks instead of SINGLE backticks.
> 
> Didn't know that! Thanks for the info. Will fix it.

I just noticed that the function `core.check_window_size` is not shown in [STUMPY's readthedocs](https://stumpy.readthedocs.io/en/latest/index.html). Having said that, I can follow the two-backtick rule regardless if the function is shown to the user in the readthedocs or not.

NimaSarajpoor: @seanlaw 
Okay. I pushed the changed and enclosed variables with TWO backticks. I've assumed you are okay with this:

> I can follow the two-backtick rule regardless if the function is shown to the user in the readthedocs or not.

If not, please let me know and I can revert the change.

NimaSarajpoor:
@seanlaw 
> Everything looks good on my end aside from a few minor changes. Please let me know if this is ready to be merged.

I addressed the comments. If you have no concern regarding [this comment](https://github.com/TDAmeritrade/stumpy/pull/1078#discussion_r2029910249), it should be ready to be merged now.

seanlaw:
Thank you for contributing this @NimaSarajpoor!

PR #1080, opened 2025-04-09
Fixed #1061 failure in snippet unit tests due to the instability of `np.sum` for array with many small floating point numbers

### LINKED ISSUE #1061: unit testing failure for snippets

NimaSarajpoor:
See issue #1061


tests/test_precision.py
```diff
@@ -115,15 +115,54 @@ def test_calculate_squared_distance():
...
+    m = 3
+
+    i, j = 2, 10
+    # M_T, Σ_T = core.compute_mean_std(T, m)
+    # Σ_T[i] is `650.912209452633`
+    # Σ_T[j] is `722.0717285148525`
+
+    # This test raises an error if arithmetic operation in ...
+    # ... `gpu_stump._compute_and_update_PI_kernel` does not
+    # generates the same result if values of variable for mean and std
```
seanlaw, line 138: "generate"

seanlaw: "same values AS"

seanlaw (changes requested):
@NimaSarajpoor Should the unit tests be failing?

NimaSarajpoor:
>  How is np.sum in the performant version different from np.sum being used in naive.py?

The array that is passed as argument to `np.sum`, is different. And, depending on that array, `np.sum` sometimes shows its bad behaviour sometime it does not. 

**A SIMPLE EXAMPLE TO BETTER UNDERSTAND THE PROBLEM**
Suppose that, in naive version, there are two arrays, `A` and `A'`.  We want to compare "SUM(A)" and "SUM(A')", and based on that choose the snippet index and its corresponding subsequence. For the sake of simplicity, let's say there is the list `lst = [A, A']`, and I want to choose the index that corresponds to array with lowest sum. In our particular rare case, the arrays `A` and `A'` have the same elements but in different order. So, the expectation is to get `index 0` as the result. But, we recently realized that `np.sum` can be unstable, and we may get `np.sum(A) > np.sum(A')`, and that gives index `1` as the output.  As you pointed out, We do not need to be worried as long as the performant version shows the same behaviour. Here is the tricky part... the arrays `A` and `A'` in performant version can be different than their naive version. Those two arrays might have the same elements though, and `np.sum` might give correct output there for those arrays, i.e. we might get `np.sum(A) == np.sum(A')` in the performant version.


**A REAL EXAMPLE TO SEE THE ISSUE**
To better understand this, we can take a look at this example:

```
# input
seed = 1615
np.random.seed(seed)
T = np.random.uniform(-1000.0, 1000.0, 64)

m = 10
s = 3
k = 3

# macOS
# env2: python 3.13, numpy 2.1.3, numba 0.61.0
```

I would like to check the 2D array `np.minimum(D, Q)` in the last iteration of [this for-loop in snippet algo](https://github.com/TDAmeritrade/stumpy/blob/main/stumpy/snippets.py#L307-L316). I would like to check the rows at indices 1 and 3 (let's call them arrays `a` and `b`) and answer the following questions:
```
(1) Do `a` and `b` have the same elements?
(2) With `math.fsum`: SUM(a)==SUM(b)?
(3) With `np.sum`: SUM(a)==SUM(b)?
```

And I want to answer those questions for three cases `naive`, `performant`, and `performant with disabled JIT`. You can see the result below for two different environments, followed by the code.

```
# macOS
# env: python 3.9, numpy 2.0.2, numba 0.60.0

==================================================
naive
Do a and b have the same elements?  True
With math.fsum: SUM(a)==SUM(b)?  True
With np.sum: SUM(a)==SUM(b)?  False
==================================================
performant
Do a and b have the same elements?  True
With math.fsum: SUM(a)==SUM(b)?  True
With np.sum: SUM(a)==SUM(b)?  False
==================================================
performant with disabled JIT
Do a and b have the same elements?  True
With math.fsum: SUM(a)==SUM(b)?  True
With np.sum: SUM(a)==SUM(b)?  False
```

NimaSarajpoor:
You are correct. This should resolve the MAIN root cause.

One approach is to break down the snippet function and creates a callee that accepts "D" as input (as a side note: this logic can also mean a user can provide their own D to compute snippet??). Then, we can compute D in both naive and performant, and if the values are all close , then we pass only one of them to the callees in both naive and performant versions. What do you suggest?

seanlaw:
I think that the correct thing to do is to understand why the `naive.get_all_mpdist_profiles` and `_get_all_profiles` produce different results and then to fix the naive version.

NimaSarajpoor:
@seanlaw 

> I think that the correct thing to do is to understand why the `naive.get_all_mpdist_profiles` and `_get_all_profiles` produce different results and then to fix the naive version. Then it shouldn't be a precision issue.

Let `D_ref` and `D_comp` denote, respectively, the output of `naive.get_all_mpdist_profiles`, and the output of `_get_all_profiles`.

IIUC the goal is to understand the main cause of having different values in`D_ref[i, j]` and `D_comp[i, j]`, and by resolving that, we can get EXACTLY the same output. Since the performant version not only has a different algo (implementation) but also is njit-decorated, I think I should follow the two following steps:

(1) First, disable JIT and see if the values are matching. If not, we can try to resolve it.
(2) Second, after resolving (1), enable JIT and see if the values are matching. If not, we can try to resolve it.

Please let me know if I misunderstood your point.

seanlaw:
> Please let me know if I misunderstood your point.

I think you've captured it. If you recall, when we use to compare our performant `stump` algo with the `naive` version and the order of indices were different because, fundamentally, the order of operations was different. Nonetheless, we still implemented a naive version that then followed and ensured the same traversal order. 

My suspicion is that the difference between `naive.get_all_mpdist_profiles` and `_get_all_profiles` may be due to something similar. Both functions also call an underlying `mpdist_vect` function, which may also produced different results and so we should fix the discrepancies at the root of it.

NimaSarajpoor:
> Does that make sense?

YES!! 

> My suspicion is that the difference between naive.get_all_mpdist_profiles and _get_all_profiles may be due to something similar

Right. Let me dig into both and report back.

seanlaw:
> As the next step, I think I need to compare the elements of those matrices closely and see if it can help with narrowing down the issue.

Sounds good! It's great to see your comments again 👍

seanlaw:
Fixed in commit https://github.com/stumpy-dev/stumpy/commit/7c0cff5c0909a7b4a832f0937a11a5ae7721a4bd

Issue #929, opened 2023-10-24
Ignore_trivial does not work as expected


frankl1:
Hi, 

I am trying stumpy, I really love the library by the way.

I have the impression that the parameter `ignore_trivial` does not work as expected. Here is a minimum example.

```
import stumpy
steam_df = pd.read_csv("https://zenodo.org/record/4273921/files/STUMPY_Basics_steamgen.csv?download=1")
stumpy.stump(steam_df['water level'], m, ignore_trivial=True)
```

I expect this code not to exclude trivial matches, and therefore the mp should be 0 everywhere as this is a self-join. However, the result is the following:
![image](https://github.com/TDAmeritrade/stumpy/assets/23366578/0b722588-d6f4-4425-971e-b0eba4f43128)

Even more suprising, the result is the same if I set ignore trivial to False.
![image](https://github.com/TDAmeritrade/stumpy/assets/23366578/854336b5-abff-4040-b869-e676f9dc83e2)

Clearly, something is not working as expected.

I am using Stumpy 1.9.2

Thanks

seanlaw:
@frankl1 Thank you for your question and for your kind words. Here are a few observations as I work through the example:

1. By default, `ignore_trivial` is automatically set to `True` for self-joins (i.e., if you don't supply a second time series, `T_B`, as an input parameter)
2. For self-joins, `ignore_trivial` should ALWAYS be `True`. Otherwise, every subsequence will find itself, called the "self-match"/"trivial-match" (or some local neighboring subsequence next to the self-match) as its "one-nearest-neighbor".
3. When `ignore_trivial=True` this does **NOT** mean "not to exclude trivial matches". In fact, `ignore_trivial=True` means to "disallow the nearest neighbor to exist within each subsequence's "exclusion zone" (usually within +/- `m/2` surrounding the subsequence of interest). 
4. Overriding or disabling `ignore_trivial` for self-joins is not advised and you are running into dangerous territory that is not supported (i.e., you are on your own if you decide to go down this path). According to the original authors of the matrix profile, the "self-match"/"trivial-match" is meaningless/obvious since you know that a subsequence should match itself and you don't need to compute a matrix profile to know this. Currently, there is no way to get around this for a self-join (not even by setting `from stumpy import config;  config.STUMPY_EXCL_ZONE_DENOM=np.inf`;  as this will allow any subsequence to match except for the self-match)
[...]

PR #1110, opened 2025-12-31
Fixed #930 Improve `ignore_trivial` Docstring and Warnings

### LINKED ISSUE #930: Improve `ignore_trivial` docstring
It has [come to our attention](https://github.com/TDAmeritrade/stumpy/issues/929) that the docstring for `ignore_trivial` may be too vague. We should consider improving the docs for this

seanlaw:
Fixed #930 and #929

# Pull Request Checklist

- [x] Create a new branch

### DIFF
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -4541,7 +4541,7 @@ def check_self_join(ignore_trivial):
             stumpy.stump(T, m, ignore_trivial=False)
     ```
     """
-    if not ignore_trivial:
+    if ignore_trivial is False:
         msg = "`ignore_trivial` cannot be `False` for a self-join and "
         msg += "has been automatically overridden and set to `True`."
         warnings.warn(msg)


stumpy/core.py
```diff
@@ -4509,3 +4509,24 @@ def _update_incremental_PI(D, P, I, excl_zone, n_appended=0):
             _shift_insert_at_index(I[-1], idx, i + n_appended)
 
     return
+
+
+def check_self_join(ignore_trivial):
```
seanlaw: I don't love this function name. Note that there is already another function called `core.check_ignore_trivial`, which

> Check inputs and verify the appropriateness for self-joins vs AB-joins and provides relevant warnings.

The difference is that `core.check_self_join` only applies to self-joins (i.e., it only uses `ignore_trivial`) while `core.check_ignore_trivial` applies to `T_A`, `T_B`, and `ignore_trivial`.

stumpy/core.py
NimaSarajpoor: I was wondering if we should remove the last sentence:

> Default is `None` which corresponds to a self-join.

stumpy/core.py
NimaSarajpoor: I think the underlying reason that resulted in this new function is that we first need to know if it is really self-join and then call this new function to MODIFY `ignore_trivial`. Because, the following if-block in `core.check_ignore_trivial`  does NOT modify `ignore_trivial`. This makes sense because we can have two arrays with same values and the case might not be self-join.  

```
if ignore_trivial is False and are_arrays_equal(T_A, T_B):
   msg = "Arrays T_A, T_B are equal, which implies a self-join. "
    msg += "Try setting `ignore_trivial = True`."
    warnings.warn(msg)
```

----

So, what if we just create a flag `is_self_join` and re-use the function `core.check_ignore_trivial` (with extra param `is_self_join`)  to modify `ignore_trivial`? I tried that but I feel it complicates the function `core.check_ignore_trivial`.

stumpy/core.py
```diff
@@ -4509,3 +4509,40 @@ def _update_incremental_PI(D, P, I, excl_zone, n_appended=0):
...
+    m = 50
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="`ignore_trivial` cannot be `False`")
+        for _ in range(5):
+            stumpy.stump(T, m, ignore_trivial=False)
+    ```
+    """
+    if not ignore_trivial:
+        msg = "`ignore_trivial` cannot be `False` for a self-join and "
+        msg += "has been automatically overridden and set to `True`."
```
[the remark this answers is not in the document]
seanlaw: Yeah, the trouble is that the function itself never actually sets the value of `ignore_trivial` so there's a bit of a disconnect as to "when" the warning occurs and the value of `ignore_trivial` is actually set.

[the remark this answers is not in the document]
seanlaw: > Right, and I don't think we are going to visit that function anytime soon in the future.

Right. And we likely won't use this `core.check_self_join` function for anything else and then somehow forget to set `ignore_trivial=True` in the next line after the function is called.

seanlaw:
@NimaSarajpoor I've made a few small changes based on your review. I think it's ready to be merged unless anything stands out to you?

NimaSarajpoor:
Nothing stands out to me. All good! 👍

seanlaw:
Thanks @NimaSarajpoor!

PR #1118, opened 2026-01-22
Fixed #1111 Add module for sliding dot product; Include pyfftw as (soft) dependency

NimaSarajpoor:
See #1111 

# Pull Request Checklist

Below is a simple checklist but please do not hesitate to ask for assistance!

- [ ] Fork, clone, and checkout the newest version of the code
- [ ] Create a new branch
- [ ] Make necessary code changes
- [ ] Install `black` (i.e., `python -m pip install black` or `conda install -c conda-forge black`)
- [ ] Install `flake8` (i.e., `python -m pip install flake8` or `conda install -c conda-forge flake8`)
- [ ] Install `pytest-cov` (i.e., `python -m pip install pytest-cov` or `conda install -c conda-forge pytest-cov`)
- [ ] Run `black --exclude=".*\.ipynb" --extend-exclude=".venv" --diff ./` in the root stumpy directory
- [ ] Run `flake8 --extend-exclude=.venv ./` in the root stumpy directory
- [ ] Run `./setup.sh dev && ./test.sh` in the root stumpy directory
- [ ] Reference a Github issue (and create one if one doesn't already exist)

PR #1124, opened 2026-01-27
Fixed #1123 Add missing Returns section in docstrings

### LINKED ISSUE #1123: `docstring.py` assumes that a method of class should NOT return a value
If a method (of a class) returns a value, then the `else` part in the following lines from  [docstring.py](https://github.com/stumpy-dev/stumpy/blob/ce05903e728fe095be18f3798b9b4a25eff0065d/docstring.py):

https://github.com/stumpy-dev/stumpy/blob/ce05903e728fe095be18f3798b9b4a25eff0065d/docstring.py#L26-L33

captures that output variable name from its docstring and includes it in the `args`. This issue was initially exposed when a class with method `__call__` was added in PR #1118.  One suggestion provided in [this comment](https://github.com/stumpy-dev/stumpy/pull/1118#issuecomment-3804947149) was:

> Given that this is an extreme case, it almost feels like it merits its on elif condition that specifically handles the specific \__call\__ case? 

I think the real question we should ask ourselves  is:  "Should we treat methods like regular functions, meaning they can return value?

I understand that all methods of all classes in STUMPY `main` branch currently return nothing ("None"). However, I think it is not unreasonable to assume that, in future, a class might be added that has a method that returns a value. So, maybe we should revise the `else` part in the code above ??

seanlaw:
See #1123 

# Pull Request Checklist

- [ ] Create a new branch
- [ ] Run `black --exclude=".*\.ipynb" --extend-exclude=".venv" --diff ./` in the root stumpy directory
- [ ] Run `flake8 --extend-exclude=.venv ./` in the root stumpy directory
- [ ] Run `./setup.sh dev && ./test.sh` in the root stumpy directory

### DIFF
--- a/docstring.py
+++ b/docstring.py
@@ -56,7 +56,7 @@ def check_args(doc_args, sig_args, file_name, func_name, class_name=None):
         msg += f"parameter(s): {diff_args}\n"
         raise RuntimeError(msg)
 
-    diff_args = docstring_args.difference(signature_args)
+    diff_args = doc_args.difference(sig_args)
     if len(diff_args) > 0:
         msg = "Found one or more unsupported arguments/parameters with docstring in \n"
         msg += f"file: {file_name}\n"


docstring.py
NimaSarajpoor: I know it is not relevant to this PR but I was wondering if we could fix the name of parameters here. Once  the changes are pulled into my branch in PR #1118, that PR becomes cleaner.

NimaSarajpoor: I think we should add a condition to check for cases when `docstring` is None...like when we completely forget to add a docstring for a function/method. Otherwise, an error will occur at line 13 (IIRC) and there will be no message there to tell us what function caused that error. It should be OK to just mix it with the current condition, so:

```
if len(docstring) is None or len(re.findall(r"Parameters", docstring)) != 1:

```

NimaSarajpoor (commented):
@seanlaw 
Thanks for the PR!! I left two comments. Not relevant directly to this PR but they are related to `docstring.py`.

docstring.py
seanlaw: Done! I changed the variables inside of the function instead.

stumpy/scraamp.py
```diff
@@ -798,7 +802,11 @@ def P_(self):
 
         Returns
         -------
-        None
+        out1 : numpy.ndarray
```
NimaSarajpoor:
```suggestion
        out : numpy.ndarray
```

stumpy/scraamp.py
```diff
@@ -819,7 +827,11 @@ def I_(self):
 
         Returns
         -------
-        None
+        out2 : numpy.ndarray
```
NimaSarajpoor:
```suggestion
        out : numpy.ndarray
```

stumpy/scrump.py
```diff
@@ -1084,7 +1088,8 @@ def P_(self):
 
         Returns
         -------
-        None
+        out1 : numpy.ndarray
```
NimaSarajpoor:
```suggestion
        out : numpy.ndarray
```

docstring.py
NimaSarajpoor:
```suggestion
    diff_args = doc_args.difference(sig_args)
```

NimaSarajpoor: @seanlaw 
Apologies for not checking your changes after my initial comments. That was on me! I committed the change right from here.

PR #1126, opened 2026-02-07
Refactor/sdp structure cleanup

NimaSarajpoor:
This is to address `PR 1` described in [this comment](https://github.com/stumpy-dev/stumpy/pull/1118#issuecomment-3854039523). Have copied the corresponding notes below:

> 1. Create a new sdp.py file
> 2. Move _sliding_dot_product to sdp.py and call it _njit_sliding_dot_product
> 3. Removing the core._sliding_dot_product altogether (currently, you are keeping both and calling one from the other. This is a bad idea)
> 4. Replace all (three) references to core._sliding_dot_product with sdp._njit_sliding_dot_product
> 5. Update move tests from tests/test_core.py to tests/test_sdp.py
> 6. Add an sdp._sliding_dot_product that simply calls _njit_sliding_dot_product (complexity gets added in future PRs)
> 7. Make core.sliding_dot_product = sdp._sliding_dot_product

### DIFF
--- a/tests/test_sdp.py
+++ b/tests/test_sdp.py
@@ -1,17 +1,9 @@
+import naive
 import numpy as np
 import pytest
 from numpy import testing as npt
 
 from stumpy import sdp
-
-
-def naive_rolling_window_dot_product(Q, T):
-    window = len(Q)
-    result = np.zeros(len(T) - window + 1)
-    for i in range(len(result)):
-        result[i] = np.dot(T[i : i + window], Q)
-    return result
-
 
 test_data = [
     (np.array([-1, 1, 2], dtype=np.float64), np.array(range(5), dtype=np.float64)),
@@ -25,20 +17,20 @@ test_data = [
 
 @pytest.mark.parametrize("Q, T", test_data)
 def test_njit_sliding_dot_product(Q, T):
-    ref_mp = naive_rolling_window_dot_product(Q, T)
+    ref_mp = naive.rolling_window_dot_product(Q, T)
     comp_mp = sdp._njit_sliding_dot_product(Q, T)
     npt.assert_almost_equal(ref_mp, comp_mp)
 
 
 @pytest.mark.parametrize("Q, T", test_data)
 def test_convolve_sliding_dot_product(Q, T):
-    ref_mp = naive_rolling_window_dot_product(Q, T)
+    ref_mp = naive.rolling_window_dot_product(Q, T)
     comp_mp = sdp._convolve_sliding_dot_product(Q, T)
     npt.assert_almost_equal(ref_mp, comp_mp)
 
 
 @pytest.mark.parametrize("Q, T", test_data)
 def test_sliding_dot_product(Q, T):
-    ref_mp = naive_rolling_window_dot_product(Q, T)
+    ref_mp = naive.rolling_window_dot_product(Q, T)
     comp_mp = sdp._sliding_dot_product(Q, T)
     npt.assert_almost_equal(ref_mp, comp_mp)
--- a/tests/naive.py
+++ b/tests/naive.py
@@ -2379,3 +2379,11 @@ def mpdist_custom_func(P_ABBA, m, percentage, n_A, n_B):
         MPdist = P_ABBA[k]
 
     return MPdist
+
+
+def rolling_window_dot_product(Q, T):
+    window = len(Q)
+    result = np.zeros(len(T) - window + 1)
+    for i in range(len(result)):
+        result[i] = np.dot(T[i : i + window], Q)
+    return result
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -35,14 +35,6 @@ except ModuleNotFoundError:
 TEST_THREADS_PER_BLOCK = 10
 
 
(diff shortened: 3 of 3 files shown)


stumpy/core.py
```diff
@@ -692,27 +664,8 @@ def sliding_dot_product(Q, T):
...
-
-    Padding is done automatically in fftconvolve step
     """
-    n = T.shape[0]
-    m = Q.shape[0]
-    Qr = np.flipud(Q)  # Reverse/flip Q
-    QT = convolve(Qr, T)
-
-    return QT.real[m - 1 : n]
+    return sdp._sliding_dot_product(Q, T)
```
NimaSarajpoor: @seanlaw 
Is this what you meant when you said [in this comment (see notes on PR 1)](https://github.com/stumpy-dev/stumpy/pull/1118#issuecomment-3854039523):

> 7. Make core.sliding_dot_product = sdp._sliding_dot_product

?

---

I think we should handle `sdp._convolve_sliding_dot_product` ([see PR 2 in this comment](https://github.com/stumpy-dev/stumpy/pull/1118)) in this PR. The function `core.sliding_dot_product` in the branch `main` can be copied to the `sdp` module, and renamed to `_convolve_sliding_dot_product`. `sdp._sliding_dot_product` should point to `sdp._convolve_sliding_dot_product`. This is not a serious issue if the plan is to handle PR 2 quickly right after PR 1.

NimaSarajpoor:
We also need to modify [Tutorial_DiscordMERLIN.ipynb](https://github.com/stumpy-dev/stumpy/blob/main/docs/WIP/Tutorial_DiscordMERLIN.ipynb) as it uses `core._sliding_dot_product`, and need to run the whole notebook again. However, the problem is that the notebook loads data from directories that do not exist. Also, the data provided by [MERLIN webpage](https://sites.google.com/view/merlin-find-anomalies/datasets) is not accessible anymore.

stumpy/core.py
seanlaw: Yes, that makes perfect sense. It's not actually new code or new SDP methods. It's simply a refactoring of existing code

stumpy/core.py
```diff
@@ -1327,7 +1280,7 @@ def _p_norm_distance_profile(Q, T, p=2.0):
             T_squared[i] = (
                 T_squared[i - 1] - T[i - 1] * T[i - 1] + T[i + m - 1] * T[i + m - 1]
             )
-        QT = _sliding_dot_product(Q, T)
+        QT = sdp._njit_sliding_dot_product(Q, T)
```
seanlaw: With the name being `sdp._njit_sliding_dot_product`, it actually makes logical sense. This `_p_norm_distance_profile` is an NJIT decorated function and, therefore, it MUST use an NJIT capable SDP function!  I like that this distinction is clear and obvious

While `core.sliding_dot_product` is NOT necessarily an NJIT function (it could be but doesn't have to be).

seanlaw (changes requested):
@NimaSarajpoor Looking good! This is much easier to track

stumpy/core.py
NimaSarajpoor: Correct 👍

tests/test_sdp.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+import pytest
+from numpy import testing as npt
+
+from stumpy import sdp
+
+
+def naive_rolling_window_dot_product(Q, T):
```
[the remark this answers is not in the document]
NimaSarajpoor: > The only time I would move a function to naive.py is when it was used across multiple test files.

Noted. Since I am now using it in `tests/test_sdp.py` too, it makes sense to just move it to `naive.py`. Is that correct?

NimaSarajpoor:
> We also need to modify [Tutorial_DiscordMERLIN.ipynb](https://github.com/stumpy-dev/stumpy/blob/main/docs/WIP/Tutorial_DiscordMERLIN.ipynb) as it uses core._sliding_dot_product, and need to run the whole notebook again. However, the problem is that the notebook loads data from directories that do not exist. Also, the data provided by [MERLIN webpage](https://sites.google.com/view/merlin-find-anomalies/datasets) is not accessible anymore.

I've also noticed that the code in the notebook requires some updates. For instance, in the notebook, I see:
```
T, M_T, Σ_T = core.preprocess(ts, m)
```

However, [this function in branch main](https://github.com/stumpy-dev/stumpy/blob/f7af69ab873b4e3c9b9a7cec17bcc75092821c7d/stumpy/core.py#L2149-L2219) now returns four values instead of three.

@seanlaw 
This notebook is outdated. I think we should submit an issue for this, and address it later and separately.

stumpy/sdp.py
```diff
@@ -0,0 +1,77 @@
...
+        Time series or sequence
+
+    Returns
+    -------
+    out : numpy.ndarray
+        Sliding dot product between `Q` and `T`.
+    """
+    m = Q.shape[0]
+    l = T.shape[0] - m + 1
+    out = np.empty(l)
```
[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
I will revise the functions in subsequent PRs.

PR #1131, opened 2026-02-14
[Failed] Fixed #1112 Migrate to New NumPy Random Number Generator (RNG) API

### LINKED ISSUE #1112: Migrate to New NumPy Random Number Generator (RNG) API
See [NumPy Reference](https://numpy.org/doc/1.22/reference/random/index.html#quick-start)

Currently, STUMPY still uses the legacy way of drawing random numbers. We should migrate to the new Generator.

This should include updating:

1. All code
2. All unit tests
3. All docstrings
4. All examples
5. All tutorials
6. All README

seanlaw:
See #1112 

# Pull Request Checklist

- [x] Create a new branch

# To Do List
1. Search for `np.random`
2. Replace `np.random.rand(x)` with `pytest.RNG.random(x)`
3. Replace `np.random.rand(x*y).reshape(x,y)` with `pytest.RNG.random(size=(x, y))`
4. Replace `np.random.uniform(x, y, [z])` with `pytest.RNG.uniform(x, y, size=z)`
5. Replace `np.random.permutation([x, y, z])` with `pytest.RNG.permutation([x, y, z])`
6. Replace `np.random.randint(x, y, z)` with `pytest.RNG.integers(x, y, z)`
7. Replace `np.random.choice([x, y], l, replace=True)` with `pytest.RNG.choice([x, y], l, replace=True)`
8. Replace `np.random.normal(size=x)` with` pytest.RNG.normal(size=x)`
9. Replace all `np.random` from docstrings, README, tutorials, etc

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,7 +9,8 @@ import json
 import numpy as np
 import pytest
 
-pytest.RNG = np.random.default_rng()
+bit_gen = np.random.PCG64()
+pytest.RNG = np.random.Generator(bit_gen)
 # pytest.RNG.bit_generator.state = {
 #     Set RNG state here
 # }
--- a/conftest.py
+++ b/conftest.py
@@ -46,12 +46,7 @@ def _get_rng_state():
 def _set_rng_state(state):
     """
     Store existing RNG state and set RNG state
-
-    If no state is provided, a hardcoded, safe state will be used
     """
-    if state is None:
-        state = pytest.SEED_ZERO.copy()
-
     pytest.LAST_RNG_STATE = pytest.RNG.bit_generator.state.copy()
     pytest.RNG.bit_generator.state = state
 
--- a/conftest.py
+++ b/conftest.py
@@ -4,3 +4,21 @@
 # to fix eventual module import errors that can arise, for example when
 # running tests from inside VS code.
 # See https://stackoverflow.com/a/34520971
+import json
+
+import numpy as np
+import pytest
+
+pytest.RNG = np.random.default_rng()
+# pytest.RNG.bit_generator.state = {
+#     Set RNG state here
+# }
+
+
+def pytest_configure(config):
+    """
+    Called after command line options have been parsed
+    and all plugins and initial conftest files been loaded.
+    """
+    state = json.dumps(pytest.RNG.bit_generator.state, indent=4)
+    print(f"conftest.py: pytest.RNG.bit_generator.state = {state}")
--- a/conftest.py
+++ b/conftest.py
@@ -14,6 +14,18 @@ pytest.RNG = np.random.default_rng()
 #     Set RNG state here
 # }
 
+pytest.LAST_RNG_STATE = None
+pytest.FIXED_RNG_STATE = {
+    # DO NOT CHANGE/ALTER!!
+    "bit_generator": "PCG64",
+    "state": {
+        "state": 195349167630453735115769518810051464980,
+        "inc": 247589055400886363559049235690497450585,
+    },
+    "has_uint32": 0,
+    "uinteger": 0,
+}
+
 
 def pytest_configure(config):
     """
@@ -22,3 +34,44 @@ def pytest_configure(config):
     """
     state = json.dumps(pytest.RNG.bit_generator.state, indent=4)
     print(f"conftest.py: pytest.RNG.bit_generator.state = {state}")
+
+
+def _get_rng_state():
+    """
+    Get a copy of the current RNG state
+    """
+    return pytest.RNG.bit_generator.state.copy()
+
+
+def _set_rng_state(state):
+    """
+    Store existing RNG state and set RNG state
+
+    If no state is provided, a hardcoded, safe state will be used
+    """
+    if state is None:
+        state = pytest.SEED_ZERO.copy()
+
+    pytest.LAST_RNG_STATE = pytest.RNG.bit_generator.state.copy()
+    pytest.RNG.bit_generator.state = state
+
+
+def _reset_rng_state():
+    """
+    Restore the RNG state to the last recorded RNG state
+    """
+    pytest.RNG.bit_generator.state = pytest.LAST_RNG_STATE.copy()
+    pytest.LAST_RNG_STATE = None
+
+
+def _fix_rng_state():
+    """
+    Set the RNG state to a fixed, hardcoded, safe state
+    """
+    _set_rng_state(pytest.FIXED_RNG_STATE)
+
+
+pytest.get_rng_state = _get_rng_state
+pytest.set_rng_state = _set_rng_state
+pytest.fix_rng_state = _fix_rng_state
+pytest.unfix_rng_state = _reset_rng_state
--- a/test.sh
+++ b/test.sh
@@ -175,7 +175,7 @@ set_ray_coveragerc()
 show_coverage_report()
 {
     set_ray_coveragerc
-    coverage report --show-missing --fail-under=100 --skip-covered --omit=fastmath.py,docstring.py,versions.py $fcoveragerc
+    coverage report --show-missing --fail-under=100 --skip-covered --omit=fastmath.py,docstring.py,versions.py,conftest.py $fcoveragerc
     check_errs $?
 }
 
--- a/test.sh
+++ b/test.sh
@@ -83,7 +83,7 @@ check_flake()
 check_print()
 {
     if [[ $print_mode == "verbose" ]]; then
-        if [[ `grep print --exclude="conftext.py" */*.py | wc -l` -gt "0" ]]; then
+        if [[ `grep print */*.py | wc -l` -gt "0" ]]; then
             echo "Error: print statement found in code"
             grep print */*.py
             exit 1
--- a/test.sh
+++ b/test.sh
@@ -83,7 +83,7 @@ check_flake()
 check_print()
 {
     if [[ $print_mode == "verbose" ]]; then
-        if [[ `grep print */*.py | wc -l` -gt "0" ]]; then
+        if [[ `grep print --exclude="conftext.py" */*.py | wc -l` -gt "0" ]]; then
             echo "Error: print statement found in code"
             grep print */*.py
             exit 1
--- a/tests/test_aamp_mmotifs.py
+++ b/tests/test_aamp_mmotifs.py
@@ -51,17 +51,15 @@ test_data = [
 
 
 def test_aamp_mmotifs_default_parameters():
-    motif_distances_ref = np.array(
-        [[0.0, 0.06315749, 0.25275899, 0.34087884, 0.3452315]]
-    )
-    motif_indices_ref = np.array([[19, 77, 63, 52, 71]])
-    motif_subspaces_ref = [np.array([2])]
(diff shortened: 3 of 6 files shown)


test.sh
```diff
@@ -83,7 +83,7 @@ check_flake()
 check_print()
 {
     if [[ $print_mode == "verbose" ]]; then
-        if [[ `grep print */*.py | wc -l` -gt "0" ]]; then
+        if [[ `grep print --exclude="conftext.py" */*.py | wc -l` -gt "0" ]]; then
```
NimaSarajpoor:
```suggestion
        if [[ `grep print --exclude="conftest.py" */*.py | wc -l` -gt "0" ]]; then
```
?

But... tests are passing now with `--exclude="conftext.py"` ??

conftest.py
```diff
@@ -4,3 +4,21 @@
...
+import numpy as np
+import pytest
+
+pytest.RNG = np.random.default_rng()
+# pytest.RNG.bit_generator.state = {
+#     Set RNG state here
+# }
+
+
+def pytest_configure(config):
```
NimaSarajpoor:
```suggestion
def pytest_configure():
```

NimaSarajpoor (commented):
@seanlaw 
I've left just one comment for your consideration.

test.sh
seanlaw: > I just learned */*.py checks only the .py files that are located in one-level deep directory.

This is definitely where "future Sean" thanks "past Sean" 😆

conftest.py
```diff
@@ -4,3 +4,74 @@
...
+
+
+def _set_rng_state(state):
+    """
+    Store existing RNG state and set RNG state
+
+    If no state is provided, a hardcoded, safe state will be used
+    """
+    if state is None:
+        state = pytest.SEED_ZERO.copy()
```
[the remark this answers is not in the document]
seanlaw: This should be removed

seanlaw:
@NimaSarajpoor I added three new files for you to review

tests/test_aamp_motifs.py
```diff
@@ -71,12 +71,12 @@ def test_aamp_motifs_one_motif():
...
     # Fix seed, because in some case motifs can be off by an index resulting in test
     # fails, which is caused since one of the motifs is not repeated perfectly in T.
-    np.random.seed(1234)
+    pytest.fix_rng_state()
 
     # The time series is random noise with two motifs for m=10:
     # * (almost) identical step functions at indices 10, 110 and 210
     # * identical linear slopes at indices 70 and 170
-    T = np.random.normal(size=300)
+    T = pytest.RNG.normal(size=300)
```
NimaSarajpoor: To be consistent with other test functions, I was wondering if we should go with `random` rather than normal distribution.

tests/test_aamp_ostinato.py
```diff
@@ -20,78 +20,81 @@ def dask_cluster():
...
-                dask_client, Ts, m, p=p
+                dask_client, Ts, m
             )
 
             npt.assert_almost_equal(ref_radius, comp_radius)
             npt.assert_almost_equal(ref_Ts_idx, comp_Ts_idx)
             npt.assert_almost_equal(ref_subseq_idx, comp_subseq_idx)
 
 
+@pytest.mark.parametrize("seed", [41, 88, 290, 292, 310, 328, 538, 556, 563, 570])
```
NimaSarajpoor: I think you missed removing this one.

tests/test_aamp_ostinato.py
```diff
@@ -20,78 +20,81 @@ def dask_cluster():
...
+            npt.assert_almost_equal(ref_Ts_idx, comp_Ts_idx)
+            npt.assert_almost_equal(ref_subseq_idx, comp_subseq_idx)
 
+    pytest.unfix_rng_state()
 
-@pytest.mark.parametrize("seed", [41, 88, 290, 292, 310, 328, 538, 556, 563, 570])
-def test_deterministic_ostinatoed(seed, dask_cluster):
+
+def test_random_ostinatoed(dask_cluster):
     with Client(dask_cluster) as dask_client:
```
NimaSarajpoor: Like some of other test functions, I think we should add `pytest.fix_rng_state()` above this line, and add `pytest.unfix_rng_state()` at the end.

tests/test_aamp_motifs.py
NimaSarajpoor:
```suggestion
    # naive.distance(T[70:90], T[170:190]) = 0.0
```

tests/test_aamp_ostinato.py
```diff
@@ -20,78 +20,81 @@ def dask_cluster():
...
             )
 
             npt.assert_almost_equal(ref_radius, comp_radius)
             npt.assert_almost_equal(ref_Ts_idx, comp_Ts_idx)
             npt.assert_almost_equal(ref_subseq_idx, comp_subseq_idx)
 
 
+@pytest.mark.parametrize("seed", [41, 88, 290, 292, 310, 328, 538, 556, 563, 570])
+def test_deterministic_ostinatoed(seed, dask_cluster):
+    pytest.fix_rng_state()
```
NimaSarajpoor: Do we need to fix state here? I can see we do not fix state for `test_random_ostinato`.

tests/test_aamp_ostinato.py
```diff
@@ -20,78 +20,81 @@ def dask_cluster():
...
         npt.assert_almost_equal(ref_subseq_idx, comp_subseq_idx)
 
 
-@pytest.mark.parametrize(
-    "seed", np.random.choice(np.arange(10000), size=25, replace=False)
-)
-def test_random_ostinatoed(seed, dask_cluster):
-    with Client(dask_cluster) as dask_client:
+def test_deterministic_ostinato():
+    pytest.fix_rng_state()
```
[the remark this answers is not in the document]
seanlaw: Since the test says "deterministic", I was inclined to keep ensure that the inputs would be identical over time. If that weren't the case then it seems like the deterministic tests could be removed.

seanlaw:
I discovered yesterday that this migration will require a lot more work since, unlike the traditional `np.random` that shares a global state, the new random number generator is only local to, say, a module. In order to share that random state across `naive.py`, `conftest.py`, all of the unit tests, and all of the `stumpy` modules, I'll have to add a global `RNG = np.random.default_rng()` variable to `core.py` and then share it across everything.

tests/test_aamp_motifs.py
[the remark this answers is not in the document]
NimaSarajpoor: I just recalculated the numbers for non-normalized distance.

```
  # naive.distance(T[10:30], T[110:130]) = 16.65
  # naive.distance(T[10:30], T[210:230]) = 0.1
  # naive.distance(T[110:130], T[210:230]) =   16.68
  # Hence T[10:30] is the motif representative for this motif
```

tests/test_aamp_mmotifs.py
```diff
@@ -51,17 +51,15 @@
 
 
 def test_aamp_mmotifs_default_parameters():
-    motif_distances_ref = np.array(
-        [[0.0, 0.06315749, 0.25275899, 0.34087884, 0.3452315]]
-    )
-    motif_indices_ref = np.array([[19, 77, 63, 52, 71]])
-    motif_subspaces_ref = [np.array([2])]
+    motif_distances_ref = np.array([[0.0, 0.10660435, 0.27927693]])
```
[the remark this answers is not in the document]
NimaSarajpoor: > While we may be cheating (this is true), nonetheless, if somebody breaks aamp_mmotifs (to produce incorrect results) then this test should still fail

Got it! Makes sense 👍

NimaSarajpoor:
@seanlaw 
> In order to share that random state across `naive.py`....

Can you please provide an example for this so that I can better understand the challenge? Do we need to share random state in certain cases? Because I think we create a randomly-generated input data `T` and then we simply pass that input `T` to both naive and performant functions. So, sharing the random state does not seem to be required. Am I missing something?

seanlaw:
> Can you please provide an example for this so that I can better understand the challenge? Do we need to share random state in certain cases? Because I think we create a randomly-generated input data T and then we simply pass that input T to both naive and performant functions. So, sharing the random state does not seem to be required. Am I missing something?

Here's a simple illustrative example.

```
# naive.py
import numpy as np

def some_func(T, choices):
    c = np.random.choice(choices)
    out = np.empty(len(T))
    for i in range(len(T)):
        out[i] = 0.0
        # This next for-loop acts like a naive implementation of `T[i] * c`
        for _ in range(c):
            out[i] += T[i]
    return out
```

```
# some_module.py
import numpy as np

def some_func(T, choices):
    c = np.random.choice(choices)
    out = T * c
    return out
```

```
# test.py
import numpy as np
import numpy.testing as npt
import naive
import some_module

def test_some_func():
    T = np.random.rand(10)
    choices = np.random.randint(50, size=(4))
    
    np.random.seed(10)
    ref = naive.some_func(T, choices)
    
    np.random.seed(10)
    cmp = some_module.some_func(T, choices)
    
    npt.assert_almost_equal(ref, cmp)
```

In this case, we have a dummy function (naive and performant) and all it does is take an array `T`, an array of integers `choices`, and randomly chooses one integer to multiply against each element of `T`. This code works because:

1. `T` is the same in both cases
2. `choices` is the same in both cases
3. We make sure to set the random seed BEFORE we randomly select an integer from `choices`. In doing so, we ensure that BOTH functions will end up choosing the same value from `choices`!

Note that we have exactly this case in some of our `stumpy` code!

Now, if we transition this "legacy" numpy code over to the modern way of implementing random number generators, we'd have to do something like (this is untested!):

```
# core.py
import numpy as np

RNG = np.random.default_rng()  # A shared RNG

def get_rng_state():
    return RNG.bit_generator.state

def set_rng_state(state):
    RNG.bit_generator.state = state
```

```
# naive.py
import numpy as np
import core

def some_func(T, choices):
    c = core.RNG.choice(choices)
    out = np.empty(len(T))
    for i in range(len(T)):
        out[i] = 0.0
        for _ in range(c):
            out[i] += T[i]
    return out
```

```
# some_module.py
import numpy as np
import core

def some_func(T, choices):
    c = core.RNG.choice(choices)
    out = T * c
    return out
```

```
# test.py
import numpy as np
import numpy.testing as npt
import naive
import some_module
import core

def test_some_func():
    # Technically, `T`, and `choices` does NOT need to depend on `core.RNG` and, instead,
    # this could've been generated from a local rng but this simplifies things
    T = core.RNG.random(10)
    choices = core.RNG.integers(0, 50, 4)
    
    state = core.get_rng_state()
    ref = naive.some_func(T, choices)
    
    core.set_rng_state(state)
    cmp = some_module.some_func(T, choices)
    
    npt.assert_almost_equal(ref, cmp)
```

@NimaSarajpoor at least this is what I am thinking. Please see if you can get my point and feel free to ask clarifying questions!

NimaSarajpoor:
@seanlaw 
> Note that we have exactly this case in some of our stumpy code!

Got it 👍  

And since we have a case like the following in some modules in `/stumpy/`:
```
# some_module.py
import numpy as np
import core

def some_func(T, choices):
    c = core.RNG.choice(choices)
    out = T * c
    return out
```

then we cannot set the random state in `conftest.py`, and we should use a stumpy's module (like `core.py`) instead.

PR #1132, opened 2026-02-17
Add pyfftw sdp

NimaSarajpoor:
This is to address `PR 3` described in https://github.com/stumpy-dev/stumpy/pull/1118#issuecomment-3854039523. Have copied the corresponding notes below:

> 1. Add `_PYFFTW_SLIDING_DOT_PRODUCT` to `sdp.py`
> 2. Add unit tests that demonstrate that `_PYFFTW_SLIDING_DOT_PRODUCT` matches the results of `naive_rolling_window_dot_product`
> 3. Handle the `test.sh` `check_fftw_pyfftw` stuff here too

### DIFF
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -267,29 +267,34 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
             self.rfft_objects[key] = rfft_obj
             self.irfft_objects[key] = irfft_obj
         else:
+            # Update the input and output arrays of the cached FFTW objects
+            # in case their original input and output arrays were reallocated
+            # in a previous call
             rfft_obj.update_arrays(real_arr, complex_arr)
             irfft_obj.update_arrays(complex_arr, real_arr)
 
         # Compute RFFT of T
         real_arr[:n] = T
         real_arr[n:] = 0.0
-        rfft_obj.execute()  # output is stored in complex_arr
+        rfft_obj.execute()
+        rfft_of_T = rfft_obj.output_array.copy()
+        # output array is stored in complex_arr, so make a copy
+        # to avoid losing it when it is overwritten when computing
+        # the RFFT of Q
 
-        # need to make a copy since the array will be
-        # overwritten later during the RFFT(Q) step
-        rfft_of_T = complex_arr.copy()
-
-        # Compute RFFT of Q (reversed and scaled by 1/next_fast_n)
+        # Compute RFFT of Q (reversed, scaled, and zero-padded)
+        # Note: scaling is needed since the thin wrapper `execute`
+        # is called which does not perform normalization
         np.multiply(Q[::-1], 1.0 / next_fast_n, out=real_arr[:m])
         real_arr[m:] = 0.0
-        rfft_obj.execute()  # output is stored in complex_arr
-        rfft_of_Q = complex_arr
+        rfft_obj.execute()
+        rfft_of_Q = rfft_obj.output_array
 
-        # Compute IRFFT of the element-wise product of the RFFTs
-        np.multiply(rfft_of_Q, rfft_of_T, out=complex_arr)
-        irfft_obj.execute()  # output is stored in real_arr
+        # Convert back to time domain by taking the inverse RFFT
+        np.multiply(rfft_of_T, rfft_of_Q, out=irfft_obj.input_array)
+        irfft_obj.execute()
 
-        return real_arr[m - 1 : n]
+        return irfft_obj.output_array[m - 1 : n]
 
 
 if PYFFTW_IS_AVAILABLE:  # pragma: no cover
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -148,10 +148,19 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
         Cache of FFTW inverse transform objects with
         (next_fast_n, n_threads, planning_flag) as lookup keys.
 
+    Methods
+    -------
+    __call__(Q, T, n_threads=1, planning_flag="FFTW_ESTIMATE")
+        Compute the sliding dot product between `Q` and `T` using FFTW
+        via pyfftw, and cache FFTW objects if not already cached.
+
     Notes
     -----
     The class maintains internal caches of FFTW objects to avoid redundant planning
     operations when called multiple times with similar-sized inputs and parameters.
+    When `planning_flag=="FFTW_ESTIMATE"`, there will be no planning operation.
+    However, caching FFTW objects is still beneficial as the overhead of creating
+    those objects can be avoided in subsequent calls.
 
     Examples
     --------
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -282,7 +282,7 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
             rfft_obj.update_arrays(real_arr, complex_arr)
             irfft_obj.update_arrays(complex_arr, real_arr)
 
-        # Compute the (circular) convolution between T and Q[::-1], 
+        # Compute the (circular) convolution between T and Q[::-1],
         # each zero-padded to length `next_fast_n` by performing
         # the following three steps:
 
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -294,7 +294,7 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
         return real_arr[m - 1 : n]
 
 
-if FFTW_IS_AVAILABLE:
+if FFTW_IS_AVAILABLE:  # pragma: no cover
     _pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)
 
 
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -293,7 +293,7 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
         rfft_obj.input_array[:n] = T
         rfft_obj.input_array[n:] = 0.0
         rfft_obj.execute()
-        rfft_of_T = rfft_obj.output_array.copy()
+        rfft_T = rfft_obj.output_array.copy()
 
         # Step 2
         # Compute RFFT of Q (reversed, scaled, and zero-padded)
@@ -302,11 +302,11 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
         np.multiply(Q[::-1], 1.0 / next_fast_n, out=rfft_obj.input_array[:m])
         rfft_obj.input_array[m:] = 0.0
         rfft_obj.execute()
-        rfft_of_Q = rfft_obj.output_array
+        rfft_Q = rfft_obj.output_array
 
         # Step 3
         # Convert back to time domain by taking the inverse RFFT
-        np.multiply(rfft_of_T, rfft_of_Q, out=irfft_obj.input_array)
+        np.multiply(rfft_T, rfft_Q, out=irfft_obj.input_array)
         irfft_obj.execute()
 
         return irfft_obj.output_array[m - 1 : n]  # valid portion
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -282,7 +282,7 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
             rfft_obj.update_arrays(real_arr, complex_arr)
             irfft_obj.update_arrays(complex_arr, real_arr)
 
-        # Compute the circular convolution between T and Q[::-1]
+        # Compute the (circular) convolution between T and Q[::-1]
         # using FFT, and then take the relevant slice of the output
 
         # Step 1
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -282,28 +282,33 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
             rfft_obj.update_arrays(real_arr, complex_arr)
             irfft_obj.update_arrays(complex_arr, real_arr)
 
-        # Compute RFFT of T
-        real_arr[:n] = T
-        real_arr[n:] = 0.0
+        # Compute the circular convolution between T and Q[::-1]
+        # using FFT, and then take the relevant slice of the output
+
+        # Step 1
+        # Compute RFFT of T (zero-padded)
+        # Must make a copy of output to avoid losing it when the array is
+        # overwritten when computing the RFFT of Q in the next step
+        rfft_obj.input_array[:n] = T
+        rfft_obj.input_array[n:] = 0.0
         rfft_obj.execute()
         rfft_of_T = rfft_obj.output_array.copy()
-        # output array is stored in complex_arr, so make a copy
-        # to avoid losing it when it is overwritten when computing
-        # the RFFT of Q
 
+        # Step 2
         # Compute RFFT of Q (reversed, scaled, and zero-padded)
-        # Note: scaling is needed since the thin wrapper `execute`
-        # is called which does not perform normalization
-        np.multiply(Q[::-1], 1.0 / next_fast_n, out=real_arr[:m])
-        real_arr[m:] = 0.0
+        # Scaling is required because the thin wrapper `execute`
+        # that will be called below does not perform normalization
+        np.multiply(Q[::-1], 1.0 / next_fast_n, out=rfft_obj.input_array[:m])
+        rfft_obj.input_array[m:] = 0.0
         rfft_obj.execute()
         rfft_of_Q = rfft_obj.output_array
 
+        # Step 3
         # Convert back to time domain by taking the inverse RFFT
         np.multiply(rfft_of_T, rfft_of_Q, out=irfft_obj.input_array)
         irfft_obj.execute()
 
-        return irfft_obj.output_array[m - 1 : n]
+        return irfft_obj.output_array[m - 1 : n]  # valid portion
 
 
 if PYFFTW_IS_AVAILABLE:  # pragma: no cover
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -6,6 +6,13 @@ from scipy.signal import convolve, oaconvolve
 
 from . import config
 
+try:  # pragma: no cover
+    import pyfftw
+
+    FFTW_IS_AVAILABLE = True
+except ImportError:  # pragma: no cover
+    FFTW_IS_AVAILABLE = False
+
 
 @njit(fastmath=config.STUMPY_FASTMATH_TRUE)
 def _njit_sliding_dot_product(Q, T):
@@ -107,6 +114,188 @@ def _pocketfft_sliding_dot_product(Q, T):
     fft_2d = r2c(True, tmp, axis=-1)
 
     return c2r(False, np.multiply(fft_2d[0], fft_2d[1]), n=next_fast_n)[m - 1 : n]
+
+
+class _PYFFTW_SLIDING_DOT_PRODUCT:
+    """
+    A class to compute the sliding dot product using FFTW via pyfftw.
+
+    This class uses FFTW (via pyfftw) to efficiently compute the sliding dot product
+    between a query sequence Q and a time series T. It preallocates arrays and caches
+    FFTW objects to optimize repeated computations with similar-sized inputs.
+
+    Parameters
+    ----------
+    max_n : int, default=2**20
+        Maximum length to preallocate arrays for. This will be the size of the
+        real-valued array. A complex-valued array of size `1 + (max_n // 2)`
+        will also be preallocated. If inputs exceed this size, arrays will be
+        reallocated to accommodate larger sizes.
+
+    Attributes
+    ----------
+    real_arr : pyfftw.empty_aligned
+        Preallocated real-valued array for FFTW computations.
+
+    complex_arr : pyfftw.empty_aligned
+        Preallocated complex-valued array for FFTW computations.
+
+    rfft_objects : dict
+        Cache of FFTW forward transform objects, keyed by
+        (next_fast_n, n_threads, planning_flag).
+
+    irfft_objects : dict
+        Cache of FFTW inverse transform objects, keyed by
+        (next_fast_n, n_threads, planning_flag).
+
+    Notes
+    -----
+    The class maintains internal caches of FFTW objects to avoid redundant planning
+    operations when called multiple times with similar-sized inputs and parameters.
+
+    Examples
+    --------
+    >>> sdp_obj = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=1000)
+    >>> Q = np.array([1, 2, 3])
+    >>> T = np.array([4, 5, 6, 7, 8])
+    >>> result = sdp_obj(Q, T)
+
+    References
+    ----------
+    `FFTW documentation <http://www.fftw.org/>`__
+
+    `pyfftw documentation <https://pyfftw.readthedocs.io/>`__
+    """
+
+    def __init__(self, max_n=2**20):
+        """
+        Initialize the `_PYFFTW_SLIDING_DOT_PRODUCT` object, which can be called
+        to compute the sliding dot product using FFTW via pyfftw.
+
+        Parameters
+        ----------
+        max_n : int, default=2**20
+            Maximum length to preallocate arrays for. This will be the size of the
+            real-valued array. A complex-valued array of size `1 + (max_n // 2)`
+            will also be preallocated.
+
+        Returns
+        -------
+        None
+        """
+        # Preallocate arrays
+        self.real_arr = pyfftw.empty_aligned(max_n, dtype="float64")
+        self.complex_arr = pyfftw.empty_aligned(1 + (max_n // 2), dtype="complex128")
+
+        # Store FFTW objects, keyed by (next_fast_n, n_threads, planning_flag)
+        self.rfft_objects = {}
+        self.irfft_objects = {}
+
+    def __call__(self, Q, T, n_threads=1, planning_flag="FFTW_ESTIMATE"):
+        """
+        Compute the sliding dot product between `Q` and `T` using FFTW via pyfftw,
+        and cache FFTW objects if not already cached.
+
+        Parameters
+        ----------
+        Q : numpy.ndarray
+            Query array or subsequence.
+
+        T : numpy.ndarray
+            Time series or sequence.
+
+        n_threads : int, default=1
+            Number of threads to use for FFTW computations.
+
+        planning_flag : str, default="FFTW_ESTIMATE"
+            The planning flag that will be used in FFTW for planning.
+            See pyfftw documentation for details. Current options, ordered
+            ascendingly by the level of aggressiveness in planning, are:
+            "FFTW_ESTIMATE", "FFTW_MEASURE", "FFTW_PATIENT", and "FFTW_EXHAUSTIVE".
+            The more aggressive the planning, the longer the planning time, but
+            the faster the execution time.
+
+        Returns
+        -------
+        out : numpy.ndarray
+            Sliding dot product between `Q` and `T`.
+
+        Notes
+        -----
+        The planning_flag is defaulted to "FFTW_ESTIMATE" to be aligned with
+        MATLAB's FFTW usage (as of version R2025b)
+        See: https://www.mathworks.com/help/matlab/ref/fftw.html
+
+        This implementation is inspired by the answer on StackOverflow:
+        https://stackoverflow.com/a/30615425/2955541
+        """
+        m = Q.shape[0]
+        n = T.shape[0]
+        next_fast_n = pyfftw.next_fast_len(n)
+
+        # Update preallocated arrays if needed
+        if next_fast_n > len(self.real_arr):
+            self.real_arr = pyfftw.empty_aligned(next_fast_n, dtype="float64")
+            self.complex_arr = pyfftw.empty_aligned(
+                1 + (next_fast_n // 2), dtype="complex128"
+            )
+
+        real_arr = self.real_arr[:next_fast_n]
+        complex_arr = self.complex_arr[: 1 + (next_fast_n // 2)]
+
+        # Get or create FFTW objects
+        key = (next_fast_n, n_threads, planning_flag)
+
+        rfft_obj = self.rfft_objects.get(key, None)
+        if rfft_obj is None:
+            rfft_obj = pyfftw.FFTW(
+                input_array=real_arr,
+                output_array=complex_arr,
+                direction="FFTW_FORWARD",
+                flags=(planning_flag,),
+                threads=n_threads,
+            )
+            self.rfft_objects[key] = rfft_obj
+        else:
+            rfft_obj.update_arrays(real_arr, complex_arr)
+
+        irfft_obj = self.irfft_objects.get(key, None)
+        if irfft_obj is None:
+            irfft_obj = pyfftw.FFTW(
+                input_array=complex_arr,
+                output_array=real_arr,
+                direction="FFTW_BACKWARD",
+                flags=(planning_flag, "FFTW_DESTROY_INPUT"),
+                threads=n_threads,
+            )
+            self.irfft_objects[key] = irfft_obj
+        else:
+            irfft_obj.update_arrays(complex_arr, real_arr)
+
+        # RFFT(T)
+        real_arr[:n] = T
+        real_arr[n:] = 0.0
+        rfft_obj.execute()  # output is in complex_arr
+        complex_arr_T = complex_arr.copy()
+
+        # RFFT(Q)
+        # Scale by 1/next_fast_n to account for
+        # FFTW's unnormalized inverse FFT via execute()
+        np.multiply(Q[::-1], 1.0 / next_fast_n, out=real_arr[:m])
+        real_arr[m:] = 0.0
+        rfft_obj.execute()  # output is in complex_arr
+
+        # RFFT(T) * RFFT(Q)
+        np.multiply(complex_arr, complex_arr_T, out=complex_arr)
+
+        # IRFFT (input is in complex_arr)
+        irfft_obj.execute()  # output is in real_arr
+
+        return real_arr[m - 1 : n]
+
+
+if FFTW_IS_AVAILABLE:
+    _pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)
 
 
 def _sliding_dot_product(Q, T):
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -282,8 +282,9 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
             rfft_obj.update_arrays(real_arr, complex_arr)
             irfft_obj.update_arrays(complex_arr, real_arr)
 
-        # Compute the (circular) convolution between T and Q[::-1]
-        # using FFT, and then take the relevant slice of the output
+        # Compute the (circular) convolution between T and Q[::-1], 
+        # each zero-padded to length `next_fast_n` by performing
+        # the following three steps:
 
         # Step 1
         # Compute RFFT of T (zero-padded)
--- a/stumpy/sdp.py
+++ b/stumpy/sdp.py
@@ -9,9 +9,9 @@ from . import config
 try:  # pragma: no cover
     import pyfftw
 
-    FFTW_IS_AVAILABLE = True
+    PYFFTW_IS_AVAILABLE = True
 except ImportError:  # pragma: no cover
-    FFTW_IS_AVAILABLE = False
+    PYFFTW_IS_AVAILABLE = False
 
 
 @njit(fastmath=config.STUMPY_FASTMATH_TRUE)
@@ -121,7 +121,7 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
     A class to compute the sliding dot product using FFTW via pyfftw.
 
     This class uses FFTW (via pyfftw) to efficiently compute the sliding dot product
-    between a query sequence Q and a time series T. It preallocates arrays and caches
+    between a query sequence, Q, and a time series, T. It preallocates arrays and caches
     FFTW objects to optimize repeated computations with similar-sized inputs.
 
     Parameters
@@ -141,12 +141,12 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
         Preallocated complex-valued array for FFTW computations.
 
     rfft_objects : dict
-        Cache of FFTW forward transform objects, keyed by
-        (next_fast_n, n_threads, planning_flag).
+        Cache of FFTW forward transform objects with
+        (next_fast_n, n_threads, planning_flag) as lookup keys.
 
     irfft_objects : dict
-        Cache of FFTW inverse transform objects, keyed by
-        (next_fast_n, n_threads, planning_flag).
+        Cache of FFTW inverse transform objects with
+        (next_fast_n, n_threads, planning_flag) as lookup keys.
 
     Notes
     -----
@@ -247,7 +247,9 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
         key = (next_fast_n, n_threads, planning_flag)
 
         rfft_obj = self.rfft_objects.get(key, None)
-        if rfft_obj is None:
+        irfft_obj = self.irfft_objects.get(key, None)
+
+        if rfft_obj is None or irfft_obj is None:
             rfft_obj = pyfftw.FFTW(
                 input_array=real_arr,
                 output_array=complex_arr,
@@ -255,12 +257,6 @@ class _PYFFTW_SLIDING_DOT_PRODUCT:
                 flags=(planning_flag,),
                 threads=n_threads,
             )
-            self.rfft_objects[key] = rfft_obj
-        else:
-            rfft_obj.update_arrays(real_arr, complex_arr)
-
-        irfft_obj = self.irfft_objects.get(key, None)
-        if irfft_obj is None:
             irfft_obj = pyfftw.FFTW(
                 input_array=complex_arr,
                 output_array=real_arr,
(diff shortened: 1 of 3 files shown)


stumpy/sdp.py
```diff
@@ -109,6 +116,188 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+        (next_fast_n, n_threads, planning_flag).
+
+    irfft_objects : dict
+        Cache of FFTW inverse transform objects, keyed by
+        (next_fast_n, n_threads, planning_flag).
+
+    Notes
+    -----
+    The class maintains internal caches of FFTW objects to avoid redundant planning
+    operations when called multiple times with similar-sized inputs and parameters.
```
NimaSarajpoor: Maybe add:

```
When `planning_flag=="FFTW_ESTIMATE"`, there will be no planning operation. 
However, caching FFTW objects is still beneficial as it avoids the overhead of 
creating those objects. 
```

tests/test_sdp.py
```diff
@@ -74,6 +74,9 @@ def get_sdp_function_names():
         if func_name.endswith("sliding_dot_product"):
             out.append(func_name)
 
+    if sdp.FFTW_IS_AVAILABLE:  # pragma: no cover
```
NimaSarajpoor: Here, in `test_sdp.py`, I am checking `sdp.FFTW_IS_AVAILABLE` instead of adding 

```
try:
    import pyfftw

    FFTW_IS_AVAILABLE = True
except ImportError:  
    FFTW_IS_AVAILABLE = False
```

---

Note that if I use the try-except block here, `flake8` will complain that `pyfftw` is imported but not used.

stumpy/sdp.py
```diff
@@ -109,6 +116,188 @@ def _pocketfft_sliding_dot_product(Q, T):
     return c2r(False, np.multiply(fft_2d[0], fft_2d[1]), n=next_fast_n)[m - 1 : n]
 
 
+class _PYFFTW_SLIDING_DOT_PRODUCT:
+    """
+    A class to compute the sliding dot product using FFTW via pyfftw.
+
+    This class uses FFTW (via pyfftw) to efficiently compute the sliding dot product
+    between a query sequence Q and a time series T. It preallocates arrays and caches
```
seanlaw: "a query sequence, Q, and a time series, T"

stumpy/sdp.py
```diff
@@ -109,6 +116,188 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+                1 + (next_fast_n // 2), dtype="complex128"
+            )
+
+        real_arr = self.real_arr[:next_fast_n]
+        complex_arr = self.complex_arr[: 1 + (next_fast_n // 2)]
+
+        # Get or create FFTW objects
+        key = (next_fast_n, n_threads, planning_flag)
+
+        rfft_obj = self.rfft_objects.get(key, None)
```
seanlaw: Would it make sense to combine both:

```
key = (next_fast_n, n_threads, planning_flag)

rfft_obj = self.rfft_objects.get(key, None)
irfft_obj = self.irfft_objects.get(key, None)

if rfft_obj is None or irfft_obj is None:
    rfft_obj = pyfftw.FFTW(
        input_array=real_arr,
        output_array=complex_arr,
        direction="FFTW_FORWARD",
        flags=(planning_flag,),
        threads=n_threads,
    )
    irfft_obj = pyfftw.FFTW(
        input_array=complex_arr,
        output_array=real_arr,
        direction="FFTW_BACKWARD",
        flags=(planning_flag, "FFTW_DESTROY_INPUT"),
        threads=n_threads,
    )
    self.rfft_objects[key] = rfft_obj
    self.irfft_objects[key] = irfft_obj
else:
    rfft_obj.update_arrays(real_arr, complex_arr)
    irfft_obj.update_arrays(complex_arr, real_arr)
```

This also assumes that things MUST be done in pairs. There shouldn't be a case where `rfft_obj` is defined but `irfft_obj` is not defined, right? That would be suspect

seanlaw (changes requested):
@NimaSarajpoor I've left some comments for you to address. I think we can afford to be clearer since all of this `pyfftw` stuff will be hard to maintain. We should probably be as verbose (and add more comments cross referencing their docs as possible). I'll do another few passes after you've responded and made modifications. I think this `pyfftw` stuff needs to be crystal clear

stumpy/sdp.py
```diff
@@ -109,6 +116,188 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+            rfft_obj = pyfftw.FFTW(
+                input_array=real_arr,
+                output_array=complex_arr,
+                direction="FFTW_FORWARD",
+                flags=(planning_flag,),
+                threads=n_threads,
+            )
+            self.rfft_objects[key] = rfft_obj
+        else:
+            rfft_obj.update_arrays(real_arr, complex_arr)
```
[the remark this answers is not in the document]
NimaSarajpoor: Correct. The [update_array](https://pyfftw.readthedocs.io/en/latest/source/pyfftw/pyfftw.html#pyfftw.FFTW.update_arrays) method basically tells the cached object to use the data buffer that are associated with the current `real_arr` and `complex_arr`.  The code would not work if the arrays used originally had already been destroyed.

stumpy/sdp.py
```diff
@@ -109,6 +116,188 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+            )
+            self.irfft_objects[key] = irfft_obj
+        else:
+            irfft_obj.update_arrays(complex_arr, real_arr)
+
+        # RFFT(T)
+        real_arr[:n] = T
+        real_arr[n:] = 0.0
+        rfft_obj.execute()  # output is in complex_arr
+        complex_arr_T = complex_arr.copy()
```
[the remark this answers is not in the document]
NimaSarajpoor: I think I need to improve the comment and the variable name...

```
rfft_obj.execute()  
# The output, i.e. RFFT(T), is now stored in `complex_arr`
# Need to make a copy as the content of this array 
# will be overwritten 

RFFT_T = complex_arr.copy()
```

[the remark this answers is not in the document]
NimaSarajpoor: Okay..I've made some changes. And tried to use proper attributes of the objects to improve readability.

test.sh
```diff
@@ -154,27 +154,58 @@ gen_ray_coveragerc()
...
     echo "    def ray_.*" >> .coveragerc_override
     echo "    def test_.*_ray*" >> .coveragerc_override
 }
 
-set_ray_coveragerc()
+check_fftw_pyfftw()
 {
-    # If `ray` command is not found then generate a .coveragerc_override file
-    if ! command -v ray &> /dev/null
+    if ! command -v fftw-wisdom &> /dev/null \
```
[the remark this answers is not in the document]
seanlaw: So, what should we do?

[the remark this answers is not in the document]
NimaSarajpoor: I think we should remove the check `command -v fftw-wisdom` from `./test.sh`. 

**There is still something that I cannot figure out though.** If `pyfftw` can be imported with no error, does that mean it can compute fft for any dtype listed in [the Schema table in pyfftw doc](https://pyfftw.readthedocs.io/en/latest/source/pyfftw/pyfftw.html#pyfftw.FFTW)? I think the answer is "No". Because I noticed the issue https://github.com/pyFFTW/pyFFTW/issues/70 says:

> In some Linux distributions it is common to have only the fftw3 library and not fftw3f. PyFFTW should detect that it is not available and work anyways (just not accept float32/single-precision inputs). 

In addition, what I infer from  [these lines of code in pyfftw](https://github.com/pyFFTW/pyFFTW/blob/a03697866ce38d61f0baf2ff72741e8a6980b92d/pyfftw/pyfftw.pyx#L71-L124) is that `pyfftw` might be installed successfully based on some available `fftw*` but it may not work for all supported dtypes. In our code, `dtype` is set to `np.float64`. One way to check is to use `wisdom` returned by ` pyfftw.export_wisdom()`:

> Return the FFTW wisdom as a tuple of strings.
> The first string in the tuple is the string for the double precision wisdom, the second is for single precision, and the third for long double precision. If any of the precisions is not supported in the build, the string is empty.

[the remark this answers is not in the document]
seanlaw: Can you post an issue to the `pyfftw` repo?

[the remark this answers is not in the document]
NimaSarajpoor: [No response yet](https://github.com/pyFFTW/pyFFTW/issues/430) but I am inclined to believe that the following two items should suffice:

(1) Just check `python -c "import pyfftw"` If this works, it means pyfftw (and its underlying FFTW) should work!

(2) Creates a simple python script that uses small data to quickly checks for 6 cases: 3 precisions, 2 mode (single-threading, multi-threading), and throws warning. Then, we can call it in shell script. Sounds Hacky?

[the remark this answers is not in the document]
seanlaw: (1) Is fine

(2) I think we should only allow for 64bit/double precision (or better) but never 32bit/single precision. Instead of a python script, why can't it be in a unit test?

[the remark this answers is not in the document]
NimaSarajpoor: >    I think we should only allow for 64bit/double precision (or better) but never 32bit/single precision.

👍 

>  why can't it be in a unit test?

My bad. In fact, having that python script is redundant as they need to be tested anyway if we want to provide support for a certain set of precisions.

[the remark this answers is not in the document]
NimaSarajpoor: @seanlaw 
I've made two changes to the code:
(1) Added param `real_dtype` to allow a user to change that if needed
(2) Added new test functions to check multi-threaded mode

stumpy/sdp.py
```diff
@@ -109,6 +116,221 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+        ----------
+        max_n : int, default=2**20
+            Maximum length to preallocate arrays for. This will be the size of the
+            real-valued array. A complex-valued array of size `1 + (max_n // 2)`
+            will also be preallocated.
+
+        real_dtype : str, default="float64"
+            The real data type to use for the preallocated arrays. Must be either
+            "float64" or "longdouble". The complex data type will be set to
+            "complex128" or "clongdouble" respectively.
```
seanlaw: Missing comma:

> "complex128" or "clongdouble", respectively.

seanlaw (changes requested):

test.sh
```diff
@@ -154,27 +154,57 @@ gen_ray_coveragerc()
...
+check_fftw_pyfftw()
 {
-    # If `ray` command is not found then generate a .coveragerc_override file
-    if ! command -v ray &> /dev/null
+    if ! python -c "import pyfftw" &> /dev/null;
     then
-        echo "Ray Not Installed"
+        echo "pyFFTW cannot be imported."
+    else
+        echo "pyFFTW can be imported!"
```
seanlaw: > pyFFTW was successfully imported

NimaSarajpoor:
@seanlaw 
This PR should be ready to be merged if there are no concerns.

stumpy/sdp.py
```diff
@@ -109,6 +116,221 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+        # Step 3
+        # Convert back to time domain by taking the inverse RFFT
+        np.multiply(rfft_T, rfft_Q, out=irfft_obj.input_array)
+        irfft_obj.execute()
+
+        return irfft_obj.output_array[m - 1 : n]  # valid portion
+
+
+if PYFFTW_IS_AVAILABLE:  # pragma: no cover
+    _pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)
```
seanlaw: What is the point of this? It seems like you want to be able to access `sdp._pyfftw_sliding_dot_product` from another module. 

But, later, in `tests/test_sdp.py`, you use `sdp._PYFFTW_SLIDING_DOT_PRODUCT(max_n)`. So, when should you use `sdp._PYFFTW_SLIDING_DOT_PRODUCT` vs `sdp._pyfftw_sliding_dot_product`?

NimaSarajpoor: > So, when should you use sdp._PYFFTW_SLIDING_DOT_PRODUCT vs sdp._pyfftw_sliding_dot_product?

My intention was to give users access to the **callable** object `sdp._pyfftw_sliding_dot_produc`. I now think that I should just remove `sdp._pyfftw_sliding_dot_product`.  No need to be worried about the users as that is a private object. 

Then, in `tests/test_py`, we can include the callable object in `get_sdp_function_names` as follows:

```
     if sdp.PYFFTW_IS_AVAILABLE:  # pragma: no cover
         _pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**10)
         setattr(sdp, '_pyfftw_sliding_dot_product', _pyfftw_sliding_dot_product)
         out.append("_pyfftw_sliding_dot_product")
```

NimaSarajpoor: @seanlaw 
I pushed a commit in which I removed `sdp._pyfftw_sliding_dot_product` from `stumpy/sdp.py`. I also revised `tests/test_sdp.py` accordingly.

seanlaw: I feel like this is even worse. Why would we do `setattr` inside of a test?

All of the other `_x_sliding_dot_product` functions inside of `sdp` are all named using the same template name. Really, after you do:

```
if PYFFTW_IS_AVAILABLE:  # pragma: no cover
    _pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)
```

then inside of the test function, wouldn't this be enough?

```
def get_sdp_function_names():
    out = []
    for func_name, _ in inspect.getmembers(sdp, inspect.isfunction):
        if func_name.endswith("sliding_dot_product"):
            out.append(func_name)
```

NimaSarajpoor: `inspect.getmembers(sdp, inspect.isfunction)` does not catch that but I think I was using a wrong solution. I should have used:

`inspect.getmembers(sdp, callable)`

This catches both `_njit_...` and `_pyfftw_..` callable objects. Will apply the change. Need to add `_pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)` to `sdp` module.

---

> Really, after you do:

> if PYFFTW_IS_AVAILABLE:  # pragma: no cover
    _pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)


Is it ~necessary~ ok to create an object when it is private and has not been used (yet) in the `sdp` module in this PR?

seanlaw: > This catches both _njit_... and _pyfftw_.. callable objects.

Great!

> Is it necessary ok to create an object when it is private and has not been used (yet) in the sdp module in this PR?

I get your point. However, I feel that the point of `_pyfftw_sliding_dot_product = _PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)` is to ensure that all "valid"/"available"/callable sliding dot product functions inside of the `sdp` product should be consistently named according `_x_sliding_dot_product`. Originally, your `sdp` module had tests that were creating an `_PYFFTW_SLIDING_DOT_PRODUCT(max_n=2**20)` private object and patching it into the `sdp` module just so it could be discoverable and this felt like an antipattern.

stumpy/sdp.py
```diff
@@ -6,6 +6,13 @@
 
 from . import config
 
+try:  # pragma: no cover
+    import pyfftw
+
+    PYFFTW_IS_AVAILABLE = True
+except ImportError:  # pragma: no cover
+    PYFFTW_IS_AVAILABLE = False
+
```
NimaSarajpoor, line 26: Maybe add a comment here that says: "The name of any callable object that computes the sliding dot product should end with `sliding_dot_product`"

tests/test_sdp.py
```diff
@@ -152,3 +152,80 @@ def test_sdp_power2():
...
+
+    return
+
+
+def test_pyfftw_sdp_multithreaded_longdouble():
+    if not sdp.PYFFTW_IS_AVAILABLE:  # pragma: no cover
+        pytest.skip("Skipping Test pyFFTW Not Installed")
+
+    # This test checks that the pyfftw_sdp can be initialized
+    # with multiple threads and longdouble data type
```
NimaSarajpoor:
```suggestion
    # longdouble data type, and then called in a multithreaded mode.
```

Note that the number of threads is not set when the instance is created

tests/test_sdp.py
```diff
@@ -152,3 +152,80 @@ def test_sdp_power2():
...
+    np.testing.assert_allclose(comp, ref)
+
+    return
+
+
+def test_pyfftw_sdp_multithreaded():
+    if not sdp.PYFFTW_IS_AVAILABLE:  # pragma: no cover
+        pytest.skip("Skipping Test pyFFTW Not Installed")
+
+    # This test checks that the pyfftw_sdp can be initialized
```
NimaSarajpoor:
```suggestion
    # This test checks that the pyfftw_sdp instantiated object
    can be called in a multithreaded mode.
```

tests/test_sdp.py
```diff
@@ -152,3 +152,80 @@ def test_sdp_power2():
...
+    # with multiple threads
```
NimaSarajpoor:
```suggestion
```

stumpy/sdp.py
```diff
@@ -109,6 +116,223 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+    """
+
+    def __init__(self, max_n=2**20, real_dtype="float64"):
+        """
+        Initialize the `_PYFFTW_SLIDING_DOT_PRODUCT` object, which can be called
+        to compute the sliding dot product using FFTW via pyfftw.
+
+        Parameters
+        ----------
+        max_n : int, default=2**20
```
NimaSarajpoor:
```suggestion
        max_n : int, default 2**20
```

stumpy/sdp.py
```diff
@@ -109,6 +116,223 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+            Maximum length to preallocate arrays for. This will be the size of the
+            real-valued array. A complex-valued array of size `1 + (max_n // 2)`
+            will also be preallocated.
+
+        real_dtype : str, default="float64"
```
NimaSarajpoor:
```suggestion
        real_dtype : str, default "float64"
```

seanlaw:
@NimaSarajpoor This might be a distraction but a (very, very small) part of me was wondering if we should be using a closure instead of a class:

https://realpython.com/python-closure/

I don't know if it makes things easier/harder to read. Given all of the array allocations, maybe a class is the right thing to use. I just wanted to bring it to your attention since I'm not sure we're getting the full benefits of a class since:

1. None of the class attributes are actually being accessed externally
2. We don't really have an external method (except `__call__`) but we do want to make it a function

Just something to consider. No pressure though.

NimaSarajpoor:
> We don't really have an external method (except __call__) but we do want to make it a function

This is something that bothered me too but thought the reason for being bothered is just my lack of experience and, also, I didn't know/think about an alternative option.

> but a (very, very small) part of me was wondering if we should be using a closure instead of a class:
> https://realpython.com/python-closure/

Thanks for sharing this! I will go through this and try to modify the script. Let's see how it will look like.

NimaSarajpoor:
@seanlaw 
The code is modified to use closure. I pushed the changes so that you can see it too. It looks readable to me. Also, this way the user cannot change the attributes that are used internally. I think that should be a good thing. Right?

seanlaw:
> The code is modified to use closure. I pushed the changes so that you can see it too. It looks readable to me. 

Yeah, it feels pretty natural! There is one thing that still bothers me and that is "how do you change/reset max_n after the function has been created?"

The natural thing to do is to overwrite the existing `sdp._pyfftw_sliding_dot_product` with a new call to `_make_pyfftw_sliding_dot_proudct` along with your desired `max_n` but feels a bit cumbersome (or hard to remember). Did you have any thoughts about this?

It feels like we're very close.

> Also, this way the user cannot change the attributes that are used internally. I think that should be a good thing. Right?

Indeed. This is partially why I didn't like using the class approach!

NimaSarajpoor:
@seanlaw 
> There is one thing that still bothers me and that is "how do you change/reset max_n after the function has been created?"

>  Did you have any thoughts about this?

I did not think about this scenario. I am now thinking in what scenario(s) a user might want to change `max_n`. Maybe they need to process much larger arrays at some point and they want to increase `max_n` to a certain value once to cover those lengths (?) or maybe they already processed some large arrays and now there is no need to have large preallocated arrays.

We can define a new function inside `_make_pyfftw_sliding_dot_product` and then return that object too. Then users can leverage that function to change values. Not sure if that is a good approach though as now users need to track two different objects returned by `_make_pyfftw_sliding_dot_product`. The option `class` might be better if we are going to end up with having more than one object returned by `_make_pyfftw_sliding_dot_product`.

NimaSarajpoor:
> @seanlaw
> 
> > There is one thing that still bothers me and that is "how do you change/reset max_n after the function has been created?"
> 
> > Did you have any thoughts about this?
> 
> I did not think about this scenario. I am now thinking in what scenario(s) a user might want to change `max_n`. Maybe they need to process much larger arrays at some point and they want to increase `max_n` to a certain value once to cover those lengths (?) or maybe they already processed some large arrays and now there is no need to have large preallocated arrays.
> 
> We can define a new function inside `_make_pyfftw_sliding_dot_product` and then return that object too. Then users can leverage that function to change values. Not sure if that is a good approach though as now users need to track two different objects returned by `_make_pyfftw_sliding_dot_product`. The option `class` might be better if we are going to end up with having more than one object returned by `_make_pyfftw_sliding_dot_product`.

Okay... on second thought, I think it should be okay to allow the returned callable object has a "callable feature" for resetting `max_n`. I think most users do not need to reset the pre-allocated arrays, and this design should be okay. See the changes in the last commit.

stumpy/sdp.py
NimaSarajpoor: will add comment at the top of the script

seanlaw:
> I think it should be okay to allow the returned callable object has a "callable feature" for resetting max_n

Can you please show me a complete example of how this works (i.e., from instantiating a new object to changing the `max_n`)?

NimaSarajpoor:
@seanlaw 

> > I think it should be okay to allow the returned callable object has a "callable feature" for resetting max_n
> 
> Can you please show me a complete example of how this works (i.e., from instantiating a new object to changing the `max_n`)?

```
# import
import numpy as np

from stumpy import sdp

# naive function to assert output
def naive_sdp(Q, T):
    out = np.zeros(len(T) - len(Q) + 1, dtype=np.float64)
    for i in range(len(out)):
        out[i] = np.dot(Q, T[i : i + len(Q)])
    return out


# pyfftw (default max_n: 2**20)
pyfftw_sdp = sdp._pyfftw_sliding_dot_product 

# case 1
Q = np.random.rand(2**2)
T = np.random.rand(2**15)
ref = naive_sdp(Q, T)
comp = pyfftw_sdp(Q, T)
np.testing.assert_allclose(comp, ref)

# set pre-allocated arrays' size to 2^10
pyfftw_sdp.reset_arr(2**10)

# case 2 
# len(T) < length of preallocated real-valued array
Q = np.random.rand(2**2)
T = np.random.rand(2**5)
ref = naive_sdp(Q, T)
comp = sdp._pyfftw_sliding_dot_product(Q, T)
np.testing.assert_allclose(comp, ref)

# case 3
# len(T) > length of preallocated real-valued array
Q = np.random.rand(2**2)
T = np.random.rand(2**11)
ref = naive_sdp(Q, T)
comp = sdp._pyfftw_sliding_dot_product(Q, T)
np.testing.assert_allclose(comp, ref)
```

Maybe I should've used the name `set_arr` instead of `reset_arr`. Because, as shown in this example, I "set" it to smaller size rather than just resetting the arrays to their original sizes. (Btw, note that user cannot check if the array is resized. So, I am allowing user to do something that they cannot verify. Is that okay??)

stumpy/sdp.py
```diff
@@ -114,6 +126,208 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+        rfft_Q = rfft_obj.output_array
+
+        # Step 3
+        # Convert back to time domain by taking the inverse RFFT
+        np.multiply(rfft_T, rfft_Q, out=irfft_obj.input_array)
+        irfft_obj.execute()
+
+        return irfft_obj.output_array[m - 1 : n]  # valid portion
+
+    def reset_arr(max_n):  # pragma: no cover
```
seanlaw: I feel strongly that `reset_arr` is too vague. I would prefer `set_max_n(max_n)`

seanlaw:
> Btw, note that user cannot check if the array is resized. So, I am allowing user to do something that they cannot verify. Is that okay??

I don't think I understand what you mean by "user cannot check if the array is resized". Can you articulate what could possibly go wrong? What would somebody need/want to verify?

Do the arrays automatically get resized if, later, somebody uses an array that is larger than `max_n`?

NimaSarajpoor:
> Do the arrays automatically get resized if, later, somebody uses an array that is larger than `max_n`?

Yes. Based on the current design, if `n=len(T) > max_n`, the preallocated array will be resized automatically, and its length will be set to `n`. 

> I don't think I understand what you mean by "user cannot check if the array is resized". Can you articulate what could possibly go wrong? What would somebody need/want to verify?

Please ignore. That was stupid! 

---

> Based on the current design, if `n=len(T) > max_n`, the preallocated array will be resized automatically

What do you think about this? Now that we are planning to add `set_max_n`, I was wondering if we should avoid allowing the length of preallocated arrays to be increased automatically. So, if the length of input is greater than `max_n`, we can create temporary preallocated arrays. This way we do not touch the original preallocated array with length `max_n`.

seanlaw:
> This way we do not touch the original preallocated array with length max_n.

I guess I would need to consider what the side effect would be if the preallocated were to be automatically resized (i.e., what would be the problem with this?). Currently, I don't see any issues with automatically resizing the preallocated array

stumpy/sdp.py
```diff
@@ -114,9 +126,210 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+        rfft_Q = rfft_obj.output_array
+
+        # Step 3
+        # Convert back to time domain by taking the inverse RFFT
+        np.multiply(rfft_T, rfft_Q, out=irfft_obj.input_array)
+        irfft_obj.execute()
+
+        return irfft_obj.output_array[m - 1 : n]  # valid portion
+
+    def set_max_n(max_n):  # pragma: no cover
```
NimaSarajpoor: @seanlaw 
Should we do `set_max_n(max_n=None)` here, and then, if not provided (default), set its value to the original `max_n`?

NimaSarajpoor (commented):
@seanlaw 
I just left one comment. I think it should be ready to merge after discussing/addressing that comment.

stumpy/sdp.py
```diff
@@ -114,9 +126,210 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+    """
+    A closure to compute the sliding dot product using FFTW via pyfftw
+
+    This closure returns a callable that computes the sliding dot product between
+    a query array, ``Q``, and a time series, ``T``. It preallocates arrays and caches
+    FFTW objects to optimize repeated computations with similar-sized inputs.
+
+    Parameters
+    ----------
+    max_n : int, default 2**20
```
seanlaw: The more I think about this, the more I feel like `max_n` is the wrong parameter name because "max" implies that the array length cannot exceed this number but we might actually violate that. Perhaps, what we're really talking about is the "initial size/length" and so maybe something like `init_size` or `init_len`.

seanlaw: Additionally, I notice that we are overloading `n` in `n_threads` and `next_fast_n` so the `n` is starting to either losing meaning or we are being inconsistent when we add `max_n`.

I can handle two different meanings for `n` within the same function but probably not more. I almost feel like I might be okay with `nfl = pyfftw.next_fast_len(n)` as long as it is defined clearly (i.e., this assignment isn't hidden/buried and I can quickly lookup what `nfl` is)

stumpy/sdp.py
seanlaw: I think about WHO would call `set_max_n` (or maybe rename to `set_size` - see comment above) and I would hope that it is somebody who knows explicitly what size they want to set. In that case, I would not put a default value and force the user to be specific.

seanlaw (commented):
@NimaSarajpoor I've left some comments for you to consider. Once completed, please let me know as I'd like to go over the code in more detail before you move on (as it's been a while since I've gone over things).

stumpy/sdp.py
NimaSarajpoor: > it is somebody who knows explicitly what size they want to set. In that case, I would not put a default value and force the user to be specific.

Ok... that makes perfect sense.

stumpy/sdp.py
NimaSarajpoor: Currently, `n` appears in the following cases:

1. The len of `T` --> `n`
2. next_fast_len(n) --> `next_fast_n`
3. The (initial) len of (real-valued) preallocated array --> `max_n`
4. number of threads --> `n_threads`

----

1. Good
2. The value returned by `next_fast_len(n)` is strongly tied to `n`. So, I think it should be okay to keep `next_fast_n`.  This is used in several lines. So using a name that reflects its description should help with the readability of the code.
3. As you pointed out, it is better to avoid using `max_n`.  I think `init_len` is a good choice.
4. Can we use `num_threads` instead? I can see `n_threads` is used in other modules though as well.

seanlaw: 2. Okay, keep it as `next_fast_n`
 3. Cool. Let's do `init_len` then
 4. I think `n_threads` is fine. We're already use to seeing it elsewhere and I never get it confused with other code.

I think simply changing `max_n` to `init_len` should be sufficient

stumpy/sdp.py
```diff
@@ -130,17 +130,19 @@ def _make_pyfftw_sliding_dot_product(init_len=2**20, real_dtype="float64"):
...
+    inputs.
 
     Parameters
     ----------
     init_len : int, default 2**20
         Initial length to preallocate arrays for. This will be the size of the
         real-valued array. A complex-valued array of size 1 + (init_len // 2)
-        will also be preallocated. If inputs exceed this size, arrays will be
-        reallocated to accommodate larger sizes.
+        will also be preallocated. If the length of input arrays exceed
```
seanlaw: "if the length of the input arrays exceedS ``init_len``, THEN the preallocated arrays will AUTOMATICALLY be resized..."

stumpy/sdp.py
```diff
@@ -150,11 +152,12 @@ def _make_pyfftw_sliding_dot_product(init_len=2**20, real_dtype="float64"):
...
     sliding_dot_product : callable
-        A callable that computes the sliding dot product between Q and T using FFTW
-        via pyfftw, and caches FFTW objects if not already cached. The callable
-        automatically resizes the preallocated arrays if the inputs exceed
-        the current initial length. In addition, the callable has a method
-        `set_init_len` to set the preallocated arrays to a new initial length.
+        A callable object that computes the sliding dot product between ``Q``
+        and ``T`` using FFTW via pyfftw, and caches FFTW objects if not already
+        cached. The callable object automatically resizes the preallocated arrays
+        if the length of input arrays exceed the current initial length.
```
seanlaw: See above

stumpy/sdp.py
```diff
@@ -272,13 +276,13 @@ def sliding_dot_product(Q, T, n_threads=1, planning_flag="FFTW_ESTIMATE"):
             irfft_obj.update_arrays(complex_view, real_view)
 
         # Compute the (circular) convolution between T and Q[::-1],
-        # each zero-padded to length next_fast_n by performing
+        # each zero-padded to the length `next_fast_n`, by performing
```
seanlaw: "to the length OF `next_fast_n`"

NimaSarajpoor:
@seanlaw 
Ok... I think this should be ready now for another round of review.

stumpy/sdp.py
```diff
@@ -114,9 +126,214 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+        # Must make a copy of output to avoid losing it when the output array
+        # is overwritten when computing the RFFT of Q in the "Step 2" below
+        rfft_obj.input_array[:n] = T
+        rfft_obj.input_array[n:] = 0.0
+        rfft_obj.execute()
+        rfft_T = rfft_obj.output_array.copy()
+
+        # Step 2
+        # Compute RFFT of Q (reversed, scaled, and zero-padded)
+        # Scaling is required because the thin wrapper execute
```
seanlaw: remove "execute" at the end of the line?

stumpy/sdp.py
```diff
@@ -114,9 +126,214 @@ def _pocketfft_sliding_dot_product(Q, T):
...
+        rfft_Q = rfft_obj.output_array
+
+        # Step 3
+        # Convert back to time domain by taking the inverse RFFT
+        np.multiply(rfft_T, rfft_Q, out=irfft_obj.input_array)
+        irfft_obj.execute()
+
+        return irfft_obj.output_array[m - 1 : n]  # valid portion
+
+    def set_init_len(init_len):  # pragma: no cover
```
seanlaw: Maybe it is better to do `update_init_len` and then in the docstring to "Update the size of the preallocated arrays to a new initial length". For some reason `set` sounds too passive

seanlaw (changes requested):
@NimaSarajpoor I think everything looks good. Just a few comments for you to consider

stumpy/sdp.py
NimaSarajpoor: Ok... I've decided to expand the explanation. what do you think?

```
# Step 2
# Compute RFFT of Q (reversed, scaled, and zero-padded)
# Scaling by `1.0 / next_fast_n` needs to be applied
# at some point in the process to ensure that the final result is correct. 
# The thin wrapper ".execute()" used in different steps does not include
# scaling. 
# This scaling can be applied to "zero_padded T" or its RFFT, or to
# "zero_padded reversed Q" or its RFFT. Or, it can be applied to 
# the multiplication of the two RFFTs, or to the final result after the IRFFT!
# Here, we choose to apply the scaling to "reversed Q" 
# to reduce the number of floating point operations for such scaling process.
np.multiply(Q[::-1], 1.0 / next_fast_n, out=rfft_obj.input_array[:m])
rfft_obj.input_array[m:] = 0.0
rfft_obj.execute()
rfft_Q = rfft_obj.output_array
```

seanlaw: > to ensure that the final result is correct

I feel like:

1. This is certainly better
2. Our "future self" will not remember what a "correct" result must entail/include and, therefore, what the PyFFTW result is "missing". In other words, I think the comment should provide a clear explanation of what precisely PyFFTW returns and why that result is "good" but not what we're ultimately looking for. There is missing logic/rationale that needs to be documented so it can be easier for us to orient ourselves in the future

NimaSarajpoor: > I think the comment should provide a clear explanation of what precisely PyFFTW returns and why that result is "good" but not what we're ultimately looking for.

Agree!!! Will revise the comments to reflect that.

NimaSarajpoor: ok... I think it should be okay to expand comments to provide further clarification. Before computing the circular convolution, I added explanation regarding the underlying logic. I pushed the changes

NimaSarajpoor:
**For our future reference:**

(R)FFT and I(R)FFT are implemented in a way that: 
* IFFT(FFT(x)) = x
* IRFFT(RFFT(x)) = x

By convention, (R)FFT is scaled by 1 (i.e. no scaling), and I(R)FFT is scaled by `1 / len(x)`

## Example 1
```
import numpy as np

n_T = 4
T = np.random.rand(n_T)
R = np.fft.rfft(T)
IR = np.fft.irfft(R)
np.testing.assert_allclose(T, IR, rtol=1e-5, atol=1e-8)
```

## Example 2
Let's use pyfftw instead.
```
import numpy as np
import pyfftw

n_T = 4
real_view = pyfftw.empty_aligned(n_T, dtype="float64")
complex_view = pyfftw.empty_aligned(n_T//2 + 1, dtype="complex128")
planning_flag = "FFTW_ESTIMATE"
n_threads = 1

rfft_obj = pyfftw.FFTW(
    input_array=real_view,
    output_array=complex_view,
    direction="FFTW_FORWARD",
    flags=(planning_flag,),
    threads=n_threads,
)

irfft_obj = pyfftw.FFTW(
    input_array=complex_view,
    output_array=real_view,
    direction="FFTW_BACKWARD",
    flags=(planning_flag,),
    threads=n_threads,
)

T = np.random.rand(n_T)
real_view[:] = T
rfft_obj()
# rfft_obj.output_array contains the RFFT of T

complex_view[:] = rfft_obj.output_array.copy()
irfft_obj()
# irfft_obj.output_array contains the IRFFT of the RFFT of T

np.testing.assert_allclose(T, irfft_obj.output_array, rtol=1e-5, atol=1e-8)
```


## Example 3
Let's use pyfftw instead. However, this time, we will use the thin wrapper "execute" that does Fourier transform only without any scaling. 

```
import numpy as np
import pyfftw


n_T = 4
real_view = pyfftw.empty_aligned(n_T, dtype="float64")
complex_view = pyfftw.empty_aligned(n_T//2 + 1, dtype="complex128")
planning_flag = "FFTW_ESTIMATE"
n_threads = 1

rfft_obj = pyfftw.FFTW(
    input_array=real_view,
    output_array=complex_view,
    direction="FFTW_FORWARD",
    flags=(planning_flag,),
    threads=n_threads,
)

irfft_obj = pyfftw.FFTW(
    input_array=complex_view,
    output_array=real_view,
    direction="FFTW_BACKWARD",
    flags=(planning_flag,),
    threads=n_threads,
)

T = np.random.rand(n_T)
real_view[:] = T
rfft_obj.execute()
# rfft_obj.output_array contains the RFFT of T

complex_view[:] = rfft_obj.output_array.copy()
irfft_obj.execute()
# irfft_obj.output_array contains the IRFFT of the RFFT of T

scaling_factor = T / irfft_obj.output_array
print("Scaling factor:", scaling_factor)
```

which gives: `Scaling factor: [0.25 0.25 0.25 0.25]`. As observed, scaling factor is `1/len(T)`.

stumpy/sdp.py
NimaSarajpoor: FYI: I expanded the comment myself. I used LLM though to polish it.

NimaSarajpoor: @seanlaw 
Can you please take a look at the revised comments and let me know what you think?

seanlaw: I think the comments are much better now!

seanlaw:
> For our future reference:

Should this information be recorded in https://github.com/stumpy-dev/sliding_dot_product/blob/main/docs/Tutorial_FFT_Based_SDP.ipynb instead??

stumpy/sdp.py
NimaSarajpoor: This particular comment

> Our "future self" will not remember what a "correct" result must entail/include and, therefore, what the PyFFTW result is "missing".

was really helpful. It encouraged me to read my comments again and again... to understand whether it can be easily digested or not.

NimaSarajpoor:
> > For our future reference:
> 
> Should this information be recorded in https://github.com/stumpy-dev/sliding_dot_product/blob/main/docs/Tutorial_FFT_Based_SDP.ipynb instead??

YES, it should! Will add the info to the notebook (will update this comment)

---

Update: Created PR https://github.com/stumpy-dev/sliding_dot_product/pull/39 to address this.

NimaSarajpoor:
@seanlaw 
I made minor changes in https://github.com/stumpy-dev/stumpy/pull/1132/changes/3cd46d721f9d873541c0e52724823c87471ba8db. How do you feel about this PR now? Please let me know if anything else needs to be addressed in this PR.

seanlaw:
> How do you feel about this PR now? Please let me know if anything else needs to be addressed in this PR.

I think it is good and appears to complete:

> 1. Add _PYFFTW_SLIDING_DOT_PRODUCT to sdp.py
> 2. Add unit tests that demonstrate that _PYFFTW_SLIDING_DOT_PRODUCT matches the results of naive_rolling_window_dot_product
> 3. Handle the test.sh check_fftw_pyfftw stuff here too

If so, then I think it is ready?

NimaSarajpoor:
Perfect! I will check one last time just in case and then merge.

test.sh
```diff
@@ -154,27 +154,57 @@ gen_ray_coveragerc()
...
+        echo "Ray not installed"
         gen_ray_coveragerc
-        fcoveragerc="--rcfile=.coveragerc_override"
     else
-        echo "Ray Installed"
+        echo "Ray installed"
+    fi
+
+    if ! command -v fftw-wisdom &> /dev/null \
+    || ! python -c "import pyfftw" &> /dev/null;
```
NimaSarajpoor: We should remove the condition `! command -v fftw-wisdom`. As we noticed previously, `fftw-wisdom` checks if FFTW is installed in the system. However, as observed before, `pyfftw` can still work if `FFTW` is not installed in the system

test.sh
```diff
@@ -154,27 +154,57 @@ gen_ray_coveragerc()
...
+    then
+        echo "FFTW and/or pyFFTW not Installed"
```
NimaSarajpoor:
```suggestion
        echo "pyFFTW cannot be imported"
```

test.sh
```diff
@@ -154,27 +154,57 @@ gen_ray_coveragerc()
...
+        gen_pyfftw_coveragerc
+    else
+        echo "FFTW and pyFFTW Installed"
```
NimaSarajpoor:
```suggestion
        echo "pyFFTW was successfully imported"
```

test.sh
```diff
@@ -154,27 +154,57 @@ gen_ray_coveragerc()
...
     gen_coveragerc_boilerplate
     echo "    def .*_ray_*" >> .coveragerc_override
-    echo "    def ,*_ray\(*" >> .coveragerc_override
+    echo "    def .*_ray\(*" >> .coveragerc_override
     echo "    def ray_.*" >> .coveragerc_override
     echo "    def test_.*_ray*" >> .coveragerc_override
 }
 
-set_ray_coveragerc()
+check_fftw_pyfftw()
```
NimaSarajpoor:
```suggestion
check_pyfftw()
```

test.sh
```diff
@@ -361,6 +391,7 @@ check_print
 check_pkg_imports
 check_naive
 check_ray
+check_fftw_pyfftw
```
NimaSarajpoor:
```suggestion
check_pyfftw
```

NimaSarajpoor:
@seanlaw 
Thanks for all the inputs! Going to merge this now.

PR #1152, opened 2026-07-10
Fixed #1151 Forward query_idx from aamp_match to core.mass_absolute

### LINKED ISSUE #1151: `aamp_match` does not forward `query_idx` to `core.mass_absolute`, so the self-match is never zeroed
### Summary

`stumpy.aamp_match` — and therefore the public `stumpy.match(..., normalize=False)`, which reroutes to it via `@core.non_normalized` — accepts `query_idx` and forwards it to `core._find_matches`, but never forwards it to `core.mass_absolute`. The normalized twin, `stumpy.motifs.match`, forwards it to **both** `core.mass` and `core._find_matches`.

`core._find_matches` seeds `candidate_idx = query_idx` and then breaks out of its loop when `D[candidate_idx] > atol + max_distance`. That is correct only when `D[query_idx] == 0`, which is exactly what `core.mass_absolute`'s `query_idx` parameter exists to establish — per its own docstring, *"If query_idx is provided, the distance between Q and T[query_idx : query_idx + m] will automatically be set to zero."*

So on the non-normalized path:

* the documented guarantee that *"the self-match will be returned first"* is not honoured, and
* the `Subsequences `Q` and `T[query_idx:query_idx+m]` are different ...` warning can never be emitted.

### Provenance

* #860 (`82ce308`, 2023-05-08) added `query_idx` to **both** `core.mass` and `core.mass_absolute`. `core.mass_absolute`'s two `query_idx` branches were given `# pragma: no cover`, because nothing called them.
* #856 (`f789386`, 2023-05-09) then wired `query_idx` through to `core.mass` from `motifs.match` — a single added line, `query_idx=query_idx,`. `aamp_motifs.py` was not touched.

`core.mass_absolute`'s `query_idx` handling has been unreachable ever since. No call site in the package passes it, and the `# pragma: no cover` markers keep the branch invisible to `coverage report --fail-under=100`.

### Reproducer

stumpy 1.14.1 (latest release) and `main` @ `86145c2`; numpy 2.4.6; Python 3.13.

```python
import warnings
import numpy as np
import stumpy

T = np.cumsum(np.random.default_rng(11).standard_normal(600))
m, qi = 40, 200
Q = T[qi : qi + m] + 0.5  # query_idx deliberately does not describe Q

for fn in (stumpy.match, stumpy.aamp_match):
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter("always")
        out = fn(Q, T, query_idx=qi, max_distance=1.0)
    warned = any("query_idx" in str(x.message) for x in w)
    print(f"{fn.__name__:12} rows={str(out.shape):7} warned={warned}")
```

Actual:

```
match        rows=(1, 2)  warned=True
aamp_match   rows=(0,)    warned=False
```

Expected: `aamp_match` returns the self-match first with distance `0.0`, and warns — as `match` does.

### Severity: low, and worth being precise about why

For **correct** usage (`Q == T[query_idx:query_idx+m]`) the omission is benign. `_mass_absolute` computes distances with `scipy.spatial.distance.cdist`, which returns exactly `0.0` for identical rows, so `D[query_idx]` is already the global minimum and the self-match is already returned first. There is no wrong answer on the happy path — presumably why this went unnoticed for three years.

What is lost is the guard. `mass_absolute`'s warning exists precisely to catch a `query_idx` that does not describe `Q`, and on the non-normalized path it cannot fire.

`aamp_motifs` and `aamp_mmotifs` route through `aamp_match`, but always pass `query_idx=candidate_idx` together with the real subsequence, so they land in the benign branch and are unaffected.

### Fix

I have a PR ready: a one-line forward in `aamp_motifs.py`, removal of the two now-reachable `# pragma: no cover` markers in `core.mass_absolute`, and a `test_aamp_match_query_idx` in `tests/test_aamp_motifs.py`. (Neither `tests/test_motifs.py` nor `tests/test_aamp_motifs.py` currently contains a `query_idx` test.)

Verified locally:

* correct usage: output byte-identical to current, 1-D and multi-dimensional (`np.array_equal` → `True`);
* `aamp_motifs` `motif_distances` / `motif_indices` unchanged across 30 seeds;
* the new test covers `core.py:1375` and `core.py:1410-1411` with no partial branches, so `--fail-under=100` still passes;
* `black`, `isort --profile black`, `flake8` clean.

Happy to open it — will reference this issue.

cagataykenter:
Fixes #1151.

`aamp_match` accepts `query_idx` and forwards it to `core._find_matches`, but never forwards it to `core.mass_absolute`. The normalized twin, `motifs.match`, forwards it to both `core.mass` and `core._find_matches`.

`core._find_matches` seeds `candidate_idx = query_idx` and then breaks when `D[candidate_idx] > atol + max_distance`, so it depends on `D[query_idx] == 0` — which is what `core.mass_absolute`'s `query_idx` parameter exists to establish.

`query_idx` was added to `core.mass` **and** `core.mass_absolute` in #860; #856 wired it into `core.mass` from `motifs.match` the next day, but `aamp_motifs.py` was not updated. `core.mass_absolute`'s `query_idx` branches have been unreachable ever since — which is why they carry `# pragma: no cover`. They are now covered, so the pragmas are removed.

### Changes

* `stumpy/aamp_motifs.py` — forward `query_idx` to `core.mass_absolute`.
* `stumpy/core.py` — drop the two `# pragma: no cover` markers on `mass_absolute`'s now-reachable `query_idx` branches.
* `tests/test_aamp_motifs.py` — add `test_aamp_match_query_idx`. There was previously no `query_idx` test in either `tests/test_motifs.py` or `tests/test_aamp_motifs.py`.

### Behaviour

For a correct self-join this is a **no-op**: `_mass_absolute` uses `scipy.spatial.distance.cdist`, which returns exactly `0.0` for the identical row, so `D[query_idx]` was already the minimum and the self-match was already returned first.

What it restores is the behaviour when `query_idx` does **not** describe `Q` — the case `mass_absolute`'s warning exists to catch:

| | `match` | `aamp_match` before | `aamp_match` after |
| --- | --- | --- | --- |
| rows returned | `(1, 2)` | `(0,)` | `(1, 2)` |
| self-match first | yes | no | yes |
| warns | yes | **no** | yes |

### Verification

* Correct usage: output byte-identical to `main` (`np.array_equal` → `True`), single- and multi-dimensional.
* `aamp_motifs` (the internal caller, which always passes `query_idx=candidate_idx` with the real subsequence): identical `motif_distances` and `motif_indices` across 30 seeds.
* `tests/test_aamp_motifs.py`, `tests/test_motifs.py`, `tests/test_precision.py`: all pass under `NUMBA_DISABLE_JIT=1`.
* The new test alone covers `core.py:1375`, `:1376`, `:1410`, `:1411` with no partial branches, so `coverage report --fail-under=100` still passes with the pragmas gone.
* `black --check`, `isort --profile black --check-only`, `flake8`: clean.

### DIFF
--- a/stumpy/aamp_motifs.py
+++ b/stumpy/aamp_motifs.py
@@ -398,7 +398,9 @@ def aamp_match(
 
     D = np.empty((d, n - m + 1))
     for i in range(d):
-        D[i, :] = core.mass_absolute(Q[i], T[i], T_subseq_isfinite[i], p=p)
+        D[i, :] = core.mass_absolute(
+            Q[i], T[i], T_subseq_isfinite[i], p=p, query_idx=query_idx
+        )
     D = np.mean(D, axis=0)
 
     return core._find_matches(
--- a/stumpy/core.py
+++ b/stumpy/core.py
@@ -1372,7 +1372,7 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0, query_idx=None):
 
     check_window_size(m, max_size=Q.shape[0])
 
-    if query_idx is not None:  # pragma: no cover
+    if query_idx is not None:
         query_idx = int(query_idx)
         T_isfinite_idx = np.isfinite(T[query_idx : query_idx + m])
         if not np.all(Q_isfinite == T_isfinite_idx) or not np.allclose(
@@ -1407,7 +1407,7 @@ def mass_absolute(Q, T, T_subseq_isfinite=None, p=2.0, query_idx=None):
         if T_subseq_isfinite is None:
             T, T_subseq_isfinite = preprocess_non_normalized(T, m)
         distance_profile[:] = _mass_absolute(Q, T, p)
-        if query_idx is not None:  # pragma: no cover
+        if query_idx is not None:
             distance_profile[query_idx] = 0.0
 
         distance_profile[~T_subseq_isfinite] = np.inf
--- a/tests/test_aamp_motifs.py
+++ b/tests/test_aamp_motifs.py
@@ -239,3 +239,22 @@ def test_aamp_match_T_subseq_isfinite(Q, T):
         )
 
         npt.assert_almost_equal(left, right)
+
+
+def test_aamp_match_query_idx():
+    T = np.random.uniform(-1000, 1000, [64])
+    m = 8
+    query_idx = 20
+    Q = T[query_idx : query_idx + m]
+
+    # `mass_absolute` zeroes the self-match distance when told where `Q` lives.
+    D = core.mass_absolute(Q, T, query_idx=query_idx)
+    npt.assert_almost_equal(D[query_idx], 0.0)
+
+    # A `Q` that is not the subsequence at `query_idx` must still return the
+    # self-match first, and must warn, exactly as `stumpy.match` does.
+    with pytest.warns(UserWarning):
+        out = aamp_match(Q + 0.5, T, query_idx=query_idx, max_distance=1.0)
+
+    assert out[0, 1] == query_idx
+    npt.assert_almost_equal(out[0, 0], 0.0)


seanlaw:
@cagataykenter Thank you for the thorough explanation and for discovering this issue. It is greatly appreciated! Please allow me some time to review

seanlaw:
@cagataykenter Thank you again for this excellent contribution! I am curious how you discovered this issue and what you use STUMPY for?

PR #1147, opened 2026-07-07
Fixed #1112 Migrate to New NumPy Random Number Generator (RNG) API

seanlaw:
See issue #1112 

This is a second attempt that follows failed PR #1131 
# Pull Request Checklist

Below is a simple checklist but please do not hesitate to ask for assistance!

- [x] Fork, clone, and checkout the newest version of the code
- [x] Create a new branch
- [x] Make necessary code changes
- [x] Install `black` (i.e., `python -m pip install black` or `conda install -c conda-forge black`)
- [x] Install `flake8` (i.e., `python -m pip install flake8` or `conda install -c conda-forge flake8`)
- [x] Install `pytest-cov` (i.e., `python -m pip install pytest-cov` or `conda install -c conda-forge pytest-cov`)
- [x] Run `black --exclude=".*\.ipynb" --extend-exclude=".venv" --diff ./` in the root stumpy directory
- [x] Run `flake8 --extend-exclude=.venv ./` in the root stumpy directory
- [x] Run `./setup.sh dev && ./test.sh` in the root stumpy directory
- [x] Reference a Github issue (and create one if one doesn't already exist)

# To Do List
1. Search for "np.random" in all `stumpy/*.py` and `tests/test_*py` files
2. Replace `np.random.rand(x)` with `rng.RNG.random(x)`
3. Replace `np.random.rand(x*y).reshape(x,y)` with `rng.RNG.random(size=(x, y))`
4. Replace `np.random.uniform(x, y, [z])` with `rng.RNG.uniform(x, y, size=z)`
5. Replace `np.random.permutation([x, y, z])` with `rng.RNG.permutation([x, y, z])`
6. Replace `np.random.randint(x, y, z)` with `rng.RNG.integers(x, y, z)`
7. Replace `np.random.choice([x, y], l, replace=True)` with `rng.RNG.choice([x, y], l, replace=True)`
8. Replace `np.random.normal(size=x)` with` rng.RNG.normal(size=x)`
9. Replace all `np.random` from docstrings, README, tutorials, etc
10. Fix `tests/test_non_normalized.py`
11. Search for"seed" in all `stumpy/*.py` and `tests/test_*py` files
12. Run unit tests on Google Colab

PR #1154, opened 2026-07-11
Fixed #1153, Fixed #707 Track Random Seed

### LINKED ISSUE #707: Add Ability to Reproduce Failed Unit Tests

### LINKED ISSUE #1153: Track Random Seed
Following issue #1112, we have shifted our (incorrect) goal from migrating to the new random number generator to simply making the existing legacy random number generator more transparent by:

1. Ensuring that all of STUMPY is derived from the same legacy RNG
2. We display the random seed at the start of each test file, which will help ensure better reproducibility

Unfortunately, the NEW NumPy RNG does NOT produce the same results across different NumPy versions

seanlaw:
Fixed #1153 and #707

Also see #1112 #1131 #1147

# Pull Request Checklist

- [x] Left a meaningful comment on the original Github issue to discuss the detailed approach for your contribution
- [x] Forked, cloned, and checkedout the newest version of the code
- [x] Created a new branch
- [x] Ran `./setup.sh dev && ./test.sh` in the root stumpy directory

### DIFF
--- a/stumpy/rng.py
+++ b/stumpy/rng.py
@@ -33,6 +33,11 @@ def fix_seed(seed):
     A context manager for setting the RNG seed to a fixed, hardcoded, safe seed
     and then returning the RNG back to its previous state prior to the seed change
 
+    This is typically used when you want to generate a specific random sequence once.
+    To repeat the same random sequence, use `fix_state` instead. If you are picking
+    a random seed directly before calling `fix_seed` then you probably want to use
+    `fix_state` instead!
+
     Parameters
     ----------
     seed : int
@@ -48,3 +53,27 @@ def fix_seed(seed):
         yield
     finally:
         RNG.set_state(state)
+
+
+@contextmanager
+def fix_state():
+    """
+    A context manager for setting the RNG state to a fixed, hardcoded, safe state
+    and then returning the RNG back to its previous state prior to the state change
+
+    This is typically used when you want to repeat the same random sequence more than
+    once.
+
+    Parameters
+    ----------
+    None
+
+    Returns
+    -------
+    None
+    """
+    curr_state = RNG.get_state()
+    try:
+        yield
+    finally:
+        RNG.set_state(curr_state)
--- a/stumpy/rng.py
+++ b/stumpy/rng.py
@@ -1,0 +1,50 @@
+from contextlib import contextmanager
+
+import numpy as np
+
+# Note that an initial SEED = 0 is disallowed
+# in order to account for unit testing
+SEED = np.random.randint(1, 4_294_967_295)
+RNG = np.random.RandomState(seed=SEED)
+
+
+def set_seed(seed):
+    """
+    Permanently set the RNG seed to a different value
+
+    Parameters
+    ----------
+    seed : int
+        The random seed for (permanently) setting the random number generator to
+
+    Returns
+    -------
+    None
+    """
+    global SEED
+    global RNG
+    SEED = seed
+    RNG = np.random.RandomState(seed=SEED)
+
+
+@contextmanager
+def fix_seed(seed):
+    """
+    A context manager for setting the RNG seed to a fixed, hardcoded, safe seed
+    and then returning the RNG back to its previous state prior to the seed change
+
+    Parameters
+    ----------
+    seed : int
+        The random seed for (temporarily) setting the random number generator to
+
+    Returns
+    -------
+    None
+    """
+    state = RNG.get_state()
+    RNG.seed(seed)
+    try:
+        yield
+    finally:
+        RNG.set_state(state)
--- a/stumpy/rng.py
+++ b/stumpy/rng.py
@@ -1,10 +1,14 @@
+import os
 from contextlib import contextmanager
 
 import numpy as np
 
 # Note that an initial SEED = 0 is disallowed
 # in order to account for unit testing
-SEED = np.random.randint(1, 4_294_967_296, dtype=np.uint32)
+if os.getenv("STUMPY_SEED") is not None:  # pragma: no cover
+    SEED = int(os.getenv("STUMPY_SEED"))
+else:
+    SEED = np.random.randint(1, 4_294_967_296, dtype=np.uint32)
 RNG = np.random.RandomState(seed=SEED)
 
 
--- a/stumpy/rng.py
+++ b/stumpy/rng.py
@@ -4,7 +4,7 @@ import numpy as np
 
 # Note that an initial SEED = 0 is disallowed
 # in order to account for unit testing
-SEED = np.random.randint(1, 4_294_967_295)
+SEED = np.random.randint(1, 4_294_967_295, dtype=np.uint32)
 RNG = np.random.RandomState(seed=SEED)
 
 
--- a/stumpy/rng.py
+++ b/stumpy/rng.py
@@ -4,7 +4,7 @@ import numpy as np
 
 # Note that an initial SEED = 0 is disallowed
 # in order to account for unit testing
-SEED = np.random.randint(1, 4_294_967_295, dtype=np.uint32)
+SEED = np.random.randint(1, 4_294_967_296, dtype=np.uint32)
 RNG = np.random.RandomState(seed=SEED)
 
 
@@ -47,12 +47,12 @@ def fix_seed(seed):
     -------
     None
     """
-    state = RNG.get_state()
+    curr_state = RNG.get_state()
     RNG.seed(seed)
     try:
         yield
     finally:
-        RNG.set_state(state)
+        RNG.set_state(curr_state)
 
 
 @contextmanager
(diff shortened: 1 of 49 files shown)


seanlaw:
Currently, `tests/test_floss.py::test_cac` is failing because the `stumpy/floss.py::_iac` function is modified to use our custom `rng.RNG`:

```
    with rng.fix_seed(seed):
        I = rng.RNG.randint(0, width, size=width, dtype=np.int64)
        if bidirectional is False:  # Idealized 1-dimensional matrix profile index
            I[:-1] = width
            for i in range(width - 1):
                I[i] = rng.RNG.randint(i + 1, width, dtype=np.int64)

        target_AC = _nnmark(I)

        params = np.empty((n_iter, 2), dtype=np.float64)
        for i in range(n_iter):
            hist_dist = scipy.stats.rv_histogram(
                (target_AC, np.append(np.arange(width), width))
            )
            data = hist_dist.rvs(size=n_samples)
            a, b, c, d = scipy.stats.beta.fit(data, floc=0, fscale=width)

            params[i, 0] = a
            params[i, 1] = b
```

However, internally, the `hist_dist` uses the default `np.random` random number generator to sample from the desired histogram distribution rather than using our custom `rng.RNG`. To remedy this, we can force `hist_dist` to use our custom `rng.RNG` by adding ` hist_dist.random_state = rng.RNG` and the result looks like:

```
with rng.fix_seed(seed):
        I = rng.RNG.randint(0, width, size=width, dtype=np.int64)
        if bidirectional is False:  # Idealized 1-dimensional matrix profile index
            I[:-1] = width
            for i in range(width - 1):
                I[i] = rng.RNG.randint(i + 1, width, dtype=np.int64)

        target_AC = _nnmark(I)

        params = np.empty((n_iter, 2), dtype=np.float64)
        for i in range(n_iter):
            hist_dist = scipy.stats.rv_histogram(
                (target_AC, np.append(np.arange(width), width))
            )
            hist_dist.random_state = rng.RNG
            data = hist_dist.rvs(size=n_samples)
            a, b, c, d = scipy.stats.beta.fit(data, floc=0, fscale=width)

            params[i, 0] = a
            params[i, 1] = b
```

seanlaw:
With our new setup for tracking the random seed, at the start of each unit test file, you'll see:

```
conftest.py: rng.set_seed(3032615292)
```

So, to reproduce all of the tests, you simply need to go into the `conftest.py` file and replace the random seed inside of:

```
def pytest_configure(config):
    """
    Called after command line options have been parsed
    and all plugins and initial conftest files been loaded.
    """
    # rng.set_seed(seed)  # Replace this with your failed unit test seed

    print(f"conftest.py: rng.set_seed({rng.SEED})")
```

This should make reproducing failed tests locally a lot easier! And then revert the `conftest.py` file when you're done (i.e., remove the seed).

Note that all of this is still using the old "legacy numpy random number generator" and so all of our tests (that require a specific seed) are unchanged.

seanlaw:
@NimaSarajpoor It is ready to be reviewed at your earliest convenience

NimaSarajpoor:
> @NimaSarajpoor It is ready to be reviewed at your earliest convenience

I will look into this PR within the next few days.

stumpy/rng.py
```diff
@@ -0,0 +1,79 @@
+from contextlib import contextmanager
+
+import numpy as np
+
+# Note that an initial SEED = 0 is disallowed
+# in order to account for unit testing
+SEED = np.random.randint(1, 4_294_967_295, dtype=np.uint32)
```
NimaSarajpoor:
```suggestion
SEED = np.random.randint(1, 4_294_967_296, dtype=np.uint32)
```


Next line shows `np.random.RandomState(seed=SEED)`. The [numpy's doc](https://numpy.org/doc/stable/reference/random/legacy.html#legacy-random-generation) provides the following statement regarding the value of seed:

> Values can be any integer between 0 and 2**32 - 1 **inclusive**.

Since the value of `high` in `np.random.randint` is **excluded** (see [doc](https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html#numpy-random-randint)), the param `high` should be set to `2**32 == 4_294_967_296`

tests/test_rng.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_fix_seed():
+    init_state = rng.RNG.get_state()
+    init_seed = init_state[1][0]  # This returns the seed
+
+    with rng.fix_seed(0):
+        state = rng.RNG.get_state()
+        seed = state[1][0]
+        assert seed != init_seed
```
NimaSarajpoor, line 24: Isn't it better to do `assert seed == 0` instead?

What if `seed=1` and `init_seed = 5`? The current assertion does not fail. BUT, I believe it should fail because `seed` is expected to be zero here because of passing 0 to `rng.fix_seed`.

stumpy/rng.py
```diff
@@ -0,0 +1,79 @@
...
+    Parameters
+    ----------
+    seed : int
+        The random seed for (temporarily) setting the random number generator to
+
+    Returns
+    -------
+    None
+    """
+    state = RNG.get_state()
```
NimaSarajpoor: To be consistent with the context manager `fix_state`:

```suggestion
    curr_state = RNG.get_state()
```

stumpy/rng.py
```diff
@@ -0,0 +1,79 @@
...
+    RNG.seed(seed)
```
NimaSarajpoor, line 55: IIUC, the RNG state will be set to a fixed state in this context manager (similar to what `fix_state` does), and that state is associated with the `seed`. Is that correct?

stumpy/rng.py
```diff
@@ -0,0 +1,79 @@
...
+@contextmanager
+def fix_seed(seed):
+    """
+    A context manager for setting the RNG seed to a fixed, hardcoded, safe seed
+    and then returning the RNG back to its previous state prior to the seed change
+
+    This is typically used when you want to generate a specific random sequence once.
+    To repeat the same random sequence, use `fix_state` instead. If you are picking
+    a random seed directly before calling `fix_seed` then you probably want to use
+    `fix_state` instead!
```
NimaSarajpoor, line 43: Can you please elaborate the last sentence? 

> If you are picking a random seed directly before calling `fix_seed` then you probably want to use `fix_state` instead!

stumpy/rng.py
NimaSarajpoor: When you have a `seed`, it might be easier to remember that one can use `fix_seed` to fix the seed (and state). Is that why you did not refactor `fix_seed` and `fix_state`?

NimaSarajpoor (commented):
@seanlaw 
You made a couple of attempts before and you finally did it! I left a few comments for your consideration.

stumpy/rng.py
seanlaw: Fixed!

stumpy/rng.py
seanlaw: So, if you want to use `fix_seed`, the only time you should use it is if you had a specific seed in mind (from previous experience). For example:

```
seed = 1234  # Because you have a bias from another experiment
fix_seed(seed)
```

There is no point in doing:
```
seed = np.random.randint(1_000_000)
fix_seed(seed)
```
because the point of `fix_seed` is to produce a deterministic state that you know of. If you do `seed = np.random.randint(1_000_000)` first and then pass that seed to `fix_seed(seed)` then that implies that you don't actually care what the starting state is. In that case, you might as well just call `with fix_state():` since it automatically gives you a random state and, in fact, has a wider range of "seeds" than is given by `np.random.randint(1_000_000)`

stumpy/rng.py
seanlaw: Fixed

stumpy/rng.py
seanlaw: > Is that correct?

Yes

> Is that why you did not refactor fix_seed and fix_state?

Yes

tests/test_rng.py
seanlaw: Maybe I got too cute but I changed this to:

```
assert seed == 0
assert seed != init_seed
```

I think both assertions are important

NimaSarajpoor: And those two assertions  together mean: `init_seed != 0`. I believe that is the assumption you had in mind before writing `with rng.fix_seed(0):`. But I guess you may want to explicitly write `assert seed != init_seed` as that is one of the main purposes of that test function.

stumpy/rng.py
NimaSarajpoor: Got the point. Thanks for further explanation.

NimaSarajpoor:
@seanlaw 
Thanks for addressing the comments. I have no further comments from my side. Should be good to merge once all tests pass

seanlaw:
Looking at the original issue, I am inclined to add support for a `STUMPY_SEED` environment variable. This should make it even easier to reproduce tests. Currently, one needs to modify conftest.py and supply the seed and that feels clunky

PR #1159, opened 2026-07-24
Drop redundant k clamp in _mpdist_vect and _aampdist_vect

### LINKED ISSUE #1082: Suggesting a minor enhancement in `mpdist.py:: _mpdist_vect`
I think we can replace 

https://github.com/TDAmeritrade/stumpy/blob/423c679fdfcdd547744d2b9a2aed7565124fd6e1/stumpy/mpdist.py#L100-L106

with 

```
P_ABBA = np.empty(2 * j, dtype=np.float64) 
  
 if k is None: 
     percentage = np.clip(percentage, 0.0, 1.0) 
     k = math.ceil(percentage * (2 * Q.shape[0]))    # BECAUSE `P_ABBA.shape[0] == 2 * j`, see the first line 
  
 k = min(int(k), P_ABBA.shape[0] - 1) 
```

DMZ22:
# Pull Request Checklist

- [x] Left a meaningful comment on the original Github issue to discuss the detailed approach for your contribution
- [x] Forked, cloned, and checkedout the newest version of the code
- [x] Created a new branch
- [ ] Ran `./setup.sh dev && ./test.sh` in the root stumpy directory

On that last item: I'm on Windows, so rather than `./test.sh` I ran the two affected suites directly — `pytest tests/test_mpdist.py tests/test_aampdist.py` → **82 passed**. Happy to run anything further you'd like. `black` reported 99 files unchanged and `flake8` was clean.

(Reopening [#1158](https://github.com/stumpy-dev/stumpy/pull/1158), which the template bot closed because I'd replaced the checklist with a plain description — my mistake.)

---

Closes #1082.

`P_ABBA` is allocated as `np.empty(2 * j, ...)`, so `P_ABBA.shape[0] - 1` **is** `2 * j - 1`. That means the `k = min(int(k), P_ABBA.shape[0] - 1)` on the following line already applies exactly the same upper bound as the `min(..., 2 * j - 1)` inside the `if k is None` branch, so the inner clamp is redundant:

```python
P_ABBA = np.empty(2 * j, dtype=np.float64)

if k is None:
    percentage = np.clip(percentage, 0.0, 1.0)
    k = math.ceil(percentage * (2 * Q.shape[0]))   # inner min(..., 2*j-1) removed

k = min(int(k), P_ABBA.shape[0] - 1)               # already bounds k by 2*j-1
```

Applied to `aampdist` as well, as you noted on the issue.

### Why it's behaviour-preserving
`min(min(a, b), b) == min(a, b)`, and the `int()` is a no-op because `math.ceil` already returns an `int`. I also checked it empirically over 20,000 randomised `(n, m, percentage)` combinations — including `percentage` values outside `[0, 1]` so the clipping path is exercised — and the resulting `k` was identical in every case.

### DIFF
--- a/stumpy/aampdist.py
+++ b/stumpy/aampdist.py
@@ -72,7 +72,7 @@ def _aampdist_vect(
 
     if k is None:
         percentage = np.clip(percentage, 0.0, 1.0)
-        k = min(math.ceil(percentage * (2 * Q.shape[0])), 2 * j - 1)
+        k = math.ceil(percentage * (2 * Q.shape[0]))
 
     k = min(int(k), P_ABBA.shape[0] - 1)
 
--- a/stumpy/mpdist.py
+++ b/stumpy/mpdist.py
@@ -101,7 +101,7 @@ def _mpdist_vect(
 
     if k is None:
         percentage = np.clip(percentage, 0.0, 1.0)
-        k = min(math.ceil(percentage * (2 * Q.shape[0])), 2 * j - 1)
+        k = math.ceil(percentage * (2 * Q.shape[0]))
 
     k = min(int(k), P_ABBA.shape[0] - 1)
 


seanlaw:
@DMZ22 Thank you for your help. I will merge this once the tests all pass

PR #1180, opened 2026-08-12
Updated derivation with improved numerical stability

seanlaw:
# Pull Request Checklist

- [x] Created a new branch


docs/REF/Matrix_Profile_Derivation.ipynb
```diff
@@ -832,15 +842,645 @@
...
+    "# Catastrophic Cancellation\n",
+    "\n",
+    "However, if you look at the recurrence relationship, the covariance, $cov(Q_{i,m},T_{i,m})$, is computed based on the previously calculated covariance, $cov(Q_{i-1,m},T_{i-1,m})$, along with a simple looking \"correction\"/\"update\" term $[(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$. Sadly, this correction term, which involves the subtraction of  two nearby numbers, can lead to [catastrophic cancellation](https://en.wikipedia.org/wiki/Catastrophic_cancellation) and rounding errors introduced during the recurrence can accumulate over subsequent updates, potentially causing significant loss of accuracy. Luckily, Section 3.3 of the [Matrix Profile XIV: Scaling Time Series Motif Discovery with GPUs to Break a Quintillion Pairwise Comparisons a Day and Beyond](https://www.cs.ucr.edu/~eamonn/SCAMP-camera-ready-final1.pdf) paper recommends using a \"centered sum-of-products\" to improve the numerical stability (i.e., we can avoid the much larger catastrophic cancellation but we can't avoid the extrememly small loss of precision in the recurrence)."
```
NimaSarajpoor: > along with a simple looking \"correction\"/\"update\" term $[(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$

The term $\frac{m-1}{m^{2}}$ is missing.

NimaSarajpoor: > Sadly, this correction term, which involves the subtraction of  two nearby numbers...

 It would be great to provide a small explanation regarding this statement as readers might not realize why the two pieces in the update term might result in nearby values.

NimaSarajpoor: > to improve the numerical stability (i.e., we can avoid the much larger catastrophic cancellation but we can't avoid the **extrememly** small loss of precision in the recurrence).

Typo: 'extrememly'

> ... paper recommends using a "centered sum-of-products" to improve the numerical stability (i.e., we can avoid the much larger catastrophic cancellation but we can't avoid the extrememly small loss of precision in the recurrence).

When I read "i.e., we can avoid the much larger catastrophic cancellation but we can't avoid the extrememly small loss of precision in the recurrence", I think it means the recommendation provided by the paper can only help with larger catastrophic cancelation. However, the extremely small loss of precision is inevitable and cannot be fixed. 

Is that correct?

docs/REF/Matrix_Profile_Derivation.ipynb
```diff
@@ -832,15 +842,645 @@
...
+    "    \\overline{Q_{i,m}T_{i,m}}\n",
+    "    &=\n",
+    "        \\sum \\limits_{0 \\leq j \\lt m}\n",
+    "            \\left(\n",
+    "                q_{i+j} - \\mu_{Q_{i,m}}\n",
+    "            \\right)\n",
+    "            \\left(\n",
+    "                t_{i+j} - M_{T_{i,m}}\n",
+    "            \\right)\n",
+    "    \\qquad\\text{and} \\qquad \\therefore\n",
```
NimaSarajpoor, line 874: I see three dots instead of the term "therefore" in the rendered version. If that is intentional, do you think it is a common symbol and easy to understand?

docs/REF/Matrix_Profile_Derivation.ipynb
```diff
@@ -832,15 +842,645 @@
...
+    "    cov(Q_{i,m},T_{i,m}) = \\frac{\\overline{Q_{i,m}T_{i,m}}}{m}\n",
+    "    \\qquad\\text{and} \\qquad\n",
+    "    \\rho(Q_{i,m},T_{i,m}) = \\frac{cov(Q_{i,m},T_{i,m})}{\\sigma_{Q_{i,m}}\\Sigma_{T_{i,m}}}\n",
+    "\\end{align}\n",
+    "\n",
+    "Thus, unlike the Welford approach earlier, if you can calculate $\\overline{Q_{i,m}T_{i,m}}$ in a numerically stable way, then $cov(Q_{i,m},T_{i,m})$ can be trivially computed by dividing $\\overline{Q_{i,m}T_{i,m}}$ by $m$. So, let's expand the product and see where we can get:\n",
+    "\n",
+    "\\begin{align}\n",
+    "    \\overline{Q_{i,m}T_{i,m}}\n",
```
NimaSarajpoor, line 883: Maybe we simply start with (23) and keep the `E` notation to simplify the derivation in (45)-(52). This should make this derivation similar to [the derivation on covariance wikipedia page](https://en.wikipedia.org/wiki/Covariance), i.e $COV(X,Y) = E[(X-E(X))(Y-E(Y))] = ... = E(XY) - E(X)E(Y)$ 

What do you think?

docs/REF/Matrix_Profile_Derivation.ipynb
```diff
@@ -832,15 +842,645 @@
...
+    "The recurrence avoids the potentially severe cancellation associated with forming large raw sum-of-products and subsequently subtracting $m\\mu_{Q_{i,m}}M_{T_{i,m}}$. It does not eliminate floating-point error entirely: the differences $q_{i+m-1}-q_{i-1}$ and $t_{i+m-1}-t_{i-1}$, as well as the recursive accumulation itself can still introduce rounding error.\n",
+    "\n",
+    "Additionally, all of these quantities can be precomputed! If we let:\n",
+    "\n",
+    "1. $df^{T}_{i,m} = \\frac{ t_{i+m-1} - t_{i-1} }{2}$\n",
+    "2. $df^{Q}_{i,m} = \\frac{ q_{i+m-1} - q_{i-1} }{2}$\n",
+    "3. $dg^{T}_{i,m} = \\left( t_{i+m-1} - M_{T_{i,m}} \\right) + \\left( t_{i-1} - M_{T_{i-1,m}} \\right)$\n",
+    "4. $dg^{Q}_{i,m} = \\left( q_{i+m-1} - \\mu_{Q_{i,m}} \\right) + \\left( q_{i-1} - \\mu_{Q_{i-1,m}} \\right)$\n",
+    "\n",
+    "Then our equation simply becomes:\n",
```
NimaSarajpoor, line 1382: It should be worth it to better demonstrate/explain the benefit of achieving this new equation.  We can show that the new equation (91) has LARGER UPDATE and therefore should be more stable as it has less catastrophic cancelation.  

Let's start with the equation (40), i.e. 

$$cov(Q_{i,m}, T_{i,m}) = cov(Q_{i-1,m}, T_{i-1,m}) + update$$ 

Let's multiply both sides by `m`:

$$ m * cov(Q_{i,m}, T_{i,m}) = m * cov(Q_{i-1,m}, T_{i-1,m}) + m * update $$

Recall that `m * cov` is basically the summation part of "centered sum-of-products". So, this means the update term in the equation (91), i.e. $df^{T}.dg^{Q} + df^{Q}.dg^{T}$, is equivalent to `m * update` (NOT `update`). This quickly shows that the new equation has LARGER UPDATE.

NimaSarajpoor: Let's see what else we can get from the relationship between equation (40) and (91). Let's start from equation (40) again:

$$cov(Q_{i,m}, T_{i,m}) = cov(Q_{i-1,m}, T_{i-1,m}) + \frac{m-1}{m^{2}}U$$ 

where

$$ U = [(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$$

---

Let's multiply both sides of equation (40) by $m$:

$$m * cov(Q_{i,m}, T_{i,m}) = m * cov(Q_{i-1,m}, T_{i-1,m}) + \frac{m-1}{m}U$$ 

Again, recall that `m * cov` is basically the summation part of "centered sum-of-products". So, if I compare the equation above with the equation (91), I can see:

$$  \frac{m-1}{m}U = df^{T}.dg^{Q} + df^{Q}.dg^{T} $$ 

Note that $\frac{m-1}{m} < 1$.  If $U$ is small, then $\frac{m-1}{m}U$ is smaller, meaning the value on the right hand side of the equation above is smaller than $U$. Therefore, if we say that $U$ involves the subtraction of  two nearby numbers and that can lead to [catastrophic cancellation](https://en.wikipedia.org/wiki/Catastrophic_cancellation), then, in that scenario, we will have "catastrophic cancellation" on the right hand side too as that is even smaller!

Can we simply use the update term $\frac{m-1}{m}U$ when we use the summation part of "centered sum-of-products"?

NimaSarajpoor (commented):
@seanlaw 
It was great to see all the steps involved in deriving equation (91). Thanks for putting them together so clearly! I have shared a few comments for your consideration.

docs/REF/Matrix_Profile_Derivation.ipynb
seanlaw: Maybe it is not so obvious but this was done on purpose. I felt like $\frac{m-1}{m^2}$ term is merely a normalization term (since it's basically $\frac{1}{m}$) rather than a correction/update term. I wanted to only focus on the catastrophic cancellation that is happening within the square brackets. In other words, I am defining the "correction" part as the square bracket and the normalization part as $\frac{m-1}{m^2}$. Perhaps, together (when you "normalize" the "correction"), you get the "update" term?

seanlaw: Yes, that is correct. Since we are traversing the diagonal and using the previous value to update the next, the small loss of precision in the first iteration will affect the second iteration and so on and so forth up until the nth iteration. However, if the loss of precision is small (without catastrophic cancellation), the accumulated loss is actually imperceptible (i.e., the magnitude is much, much smaller than we care about). When there is catastrophic cancellation, this really gets amplified quickly as the loss of precision is BIG.

docs/REF/Matrix_Profile_Derivation.ipynb
seanlaw: > If that is intentional

Yes, that is intentional. I believe it is a common symbol in math (at least, as long as I've remembered it since high school)

It is closely related to the "\because" (or "since") symbol $\because$, which is the flipped $\therefore$ symbol that are both often used together

docs/REF/Matrix_Profile_Derivation.ipynb
seanlaw: When I cross reference the equation lines that your referenced, I'm not following your suggestion (please [see this "equation referencing" comment](https://github.com/stumpy-dev/stumpy/pull/1180#discussion_r3789401042)). Additionally, maybe my brain is stubborn but I can't seem to think clearly when the equations are written as $COV(X,Y) = E[(X-E(X))(Y-E(Y))] = ... = E(XY) - E(X)E(Y)$. I tend to find this hard to follow but I don't know why... maybe it's because I hate statistics 🥲

docs/REF/Matrix_Profile_Derivation.ipynb
seanlaw: Maybe it's an issue on my end when things are being rendered in my Jupyter notebook but I'm only seeing a single equation number for a block of related/multiline equations:

Screenshot 2026-08-15 at 8 53 58 AM

If you are seeing the same thing, maybe we can refer to the lines as (14.1) - (14.5) (to mean lines 1-5 in equation block 14)? Otherwise, I am counting equations one-by-one each time :(

Or maybe you have a different workflow?

seanlaw: > Can we simply use the update term $\frac{m−1}{m}U$ when we use the summation part of "centered sum-of-products"?

I've tried reading your comments a few times but I am unable to get your point.

> Therefore, if we say that $U$ involves the subtraction of two nearby numbers and that can lead to catastrophic cancellation, then, in that scenario, we will have "catastrophic cancellation" on the right hand side too as that is even smaller!

I feel like I'm only reading this as a statement of fact but I am getting the sense that you have a much stronger point/argument that I am overlooking :(

You started with:
> It should be worth it to better demonstrate/explain the benefit of achieving this new equation. 

Maybe I've been staring at the equations too long but I'm not able to see "how" this is demonstrating the benefit any further (i.e., what is the special insight that "proves" the benefit clearly). I know you so I can feel that there is something (intuition?) there. I'm just not smart enough to infer your intuition but I would like to hear more.

docs/REF/Matrix_Profile_Derivation.ipynb
seanlaw: In other words, we can resolve the BIG catastrophic cancellation but we can (probably?) NEVER resolve the small, small loss of precision since we are dealing with floating point numbers, which must involve loss of precision

NimaSarajpoor: So that was intentional 👍 

If the correction term is not small, is it possible to still have loss of precision after multiplying its value by $\frac{m-1}{m^{2}}$ ?

Because it simply makes the value smaller and therefore the loss of precision can have higher impact on the final outcome.

NimaSarajpoor: Thanks for the explanation! 

Can we try to reflect this explanation in the text? When I was reading the sentence, the terms "recommends" and "improve" gave me a positive feeling. But then, the part ended with "but we can't avoid the extrememly small loss of precision in the recurrence" and that gave me a negative feeling... and that was a bit strange.

docs/REF/Matrix_Profile_Derivation.ipynb
NimaSarajpoor: Fair point! 👍

docs/REF/Matrix_Profile_Derivation.ipynb
seanlaw: If the correction term is not small (i.e., near zero) then it implies that there is no catastrophic cancellation (I think?). Having said that, loss of precision will always exist. I think what matters is the magnitude of that loss of precision. I don't know for certain but when asked Claude to give me an estimate of the accumulated error for a time series with `n = 100_000_000` it showed that (depending on the dynamic range of the time series data), the old way would be precise to $10^{-4}-10^{-8}$ while the df/dg approach would be (MORE) precise to $10^{-8}-10^{-12}$. Note that double float precision is probably good to, at best, $10^{-16}$.

So, yes, there will be loss of precision (partially due to multiplying by the constant $\frac{m-1}{m^2}$) but it probably doesn't account for a 4 order magnitude of accumulated loss of precision all on its own. 

Does that help?

seanlaw: I've added an entire new section to explain this.

seanlaw: I will add some of this to the doc

NimaSarajpoor: @seanlaw 
I think it does to some extent. I am still trying to fill a certain gap in my head. I am going to read the updated file.

docs/REF/Matrix_Profile_Derivation.ipynb
NimaSarajpoor: > Or maybe you have a different workflow?

I didn't download the notebook. I just opened the file on Github from your branch. It renders the file differently and I didn't know that it affects the equations numbers! Apologies for the inconvenience. 

Please allow me to download the updated notebook and read it again to see the added information. I will use the equation numbers from there.

NimaSarajpoor: I downloaded and read the (updated) notebook. I opened the file in notebook this time and cannot see equation numbers. Strange!

image


---


> I've tried reading your comments a few times but I am unable to get your point.

If I understand correctly, the gist of this PR is that: 
* The correction term $[(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$, which involves the subtraction of two nearby numbers, is small and can lead to catastrophic cancellation. 
* The proposal is to use $df^{T}.dg^{Q} + df^{Q}.dg^{T}$ instead (in a new recurrence relation).

HOWEVER, it can be shown that $df^{T}.dg^{Q} + df^{Q}.dg^{T}$ is actually smaller than the original correction term $[(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$. IMO, it means that if the original "correction" term $[(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$ is small and can lead to catastrophic cancelation, then $df^{T}.dg^{Q} + df^{Q}.dg^{T}$ is smaller, meaning $(df^{T}.dg^{Q}) - (-df^{Q}.dg^{T})$ can also lead to catastrophic cancelation and it can be even worse!

So, we might be better off using the original correction term when updating the $\overline{QT}$.

seanlaw: Hmm, I see your point and do not yet have a good answer...

NimaSarajpoor: I have been thinking about conducting an experiment that can demonstrate empirically which option works better when updating $\overline{QT}$. 

If we can compute the actual values directly with higher precision, we can see which option works better. 

What do you think?

seanlaw: > What do you think?

I am certainly not opposed to it. Perhaps, the statement about "catastrophic cancellation" was not correct and that language can be softened. From a PR perspective, at what point is the derivation "correct" and what we need to do is adjust the language (maybe as a separate PR)?

NimaSarajpoor: > Perhaps, the statement about "catastrophic cancellation" was not correct

Probably...If "a - b" has catastrophic cancellation, then the issue might be addressed if we change it to subtraction-free operations (with positive-only elements). However, this is not achieved here. 


> From a PR perspective, at what point is the derivation "correct" 

I think the moment we obtained the original **correction** term, i.e. $[(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$ . We can then show that:

$$\overline{QT_{i}} = \overline{QT_{i-1}} + \frac{m-1}{m} \times correction$$

And that's it (unless we can show the reformulated version is better)

---

But, something is still bothering me. If the equation above is better (i.e., numerically more stable), then what about the following?

$$m * \overline{QT_{i}} = m * \overline{QT_{i-1}} + (m-1) \times correction$$

So, we compute $m * \overline{QT_{i}}$ and keep updating it by using $(m-1) \times correction$ (?!?)

NimaSarajpoor: @seanlaw 
I did an experiment to check which of the following equations results in less numerical error for $\overline{QT}$. No consistent result was observed!

## Recursion Relation Using Original Formula

$$ \overline{QT_{i}} =  \overline{QT_{i-1}} + \frac{m-1}{m} [(q_{i+m-1} - \\mu_{Q_{i,m-1}})(t_{i+m-1} - M_{T_{i,m-1}}) - (q_{i-1} - \\mu_{Q_{i,m-1}}) (t_{i-1} - M_{T_{i,m-1}})]$$

## Recursion Relation Using New Formula

$$ \overline{QT_{i}} =  \overline{QT_{i-1}} + df^{T}.dg^{Q} + df^{Q}.dg^{T} $$


## Ground Truth

$$  \overline{QT_{i}} = (T[i:i+m] - M_{T_{i,m}}).(Q[i:i+m] - \mu_{Q_{i,m}})$$


## Code
```
import matplotlib.pyplot as plt
import numpy as np
from stumpy import core

# Ground Truth
def _mcov_precise(Q, T, M_T, μ_Q, m, i):
    mcov = np.dot(
            (T[i : i + m] - M_T[i]),
            (Q[i : i + m] - μ_Q[i]),
        )

    return mcov   


def mcov_precise(Q, T, m):
    if len(Q) != len(T):
        msg = 'The length of two arrays should be the same'
        raise ValueError(msg)
    M_T, _ = core.compute_mean_std(T, m)
    μ_Q, _ = core.compute_mean_std(Q, m)

    l = len(T) - m + 1
    out = np.empty(l, dtype=np.float64)
    for i in range(l):
        out[i] = _mcov_precise(Q, T, M_T, μ_Q, m, i)

    return out


# original
def _mcov_sliding_original(mcov, constant, cov_a, cov_b, cov_c, cov_d, i):
    mcov_next = mcov + constant * (
        cov_a[i] * cov_b[i]
        - cov_c[i] * cov_d[i]
    )

    return mcov_next


def mcov_sliding_original(Q, T, m):
    if len(Q) != len(T):
        msg = 'The length of two arrays should be the same'
        raise ValueError(msg)

    (
        Q, # T_A
        μ_Q,
        σ_Q_inverse,
        μ_Q_m_1,
        Q_subseq_isfinite,
        Q_subseq_isconstant,
    ) = core.preprocess_diagonal(Q, m, T_subseq_isconstant=None)
    
    (
        T, # T_B
        M_T,
        Σ_T_inverse,
        M_T_m_1,
        T_subseq_isfinite,
        T_subseq_isconstant,
    ) = core.preprocess_diagonal(T, m, T_subseq_isconstant=None)
    
    
    # compute for elements for updating cov or mcov
    cov_a = T[m - 1 :] - M_T_m_1[:-1]
    cov_b = Q[m - 1 :] - μ_Q_m_1[:-1]
    
    cov_c = np.empty(M_T_m_1.shape[0], dtype=np.float64)
    cov_c[1:] = T[: M_T_m_1.shape[0] - 1]
    cov_c[0] = T[-1]
    cov_c[:] = cov_c - M_T_m_1
    
    cov_d = np.empty(μ_Q_m_1.shape[0], dtype=np.float64)
    cov_d[1:] = Q[: μ_Q_m_1.shape[0] - 1]
    cov_d[0] = Q[-1]
    cov_d[:] = cov_d - μ_Q_m_1


    constant = (m-1) / m
    l = len(T) - m + 1
    out = np.empty(l, dtype=np.float64)
    out[0] = _mcov_precise(Q, T, M_T, μ_Q, m, 0)
    for i in range(1, l):
        out[i] = _mcov_sliding_original(
            out[i-1], 
            constant, 
            cov_a, 
            cov_b, 
            cov_c, 
            cov_d, 
            i
        )

    return out


# new
def _mcov_sliding_new(mcov, Q, T, M_T, μ_Q, m, i):
    mcov_next = mcov + 1/2 * (T[i+m-1] - T[i-1]) * ((Q[i+m-1]-μ_Q[i]) + (Q[i-1] - μ_Q[i-1])) + 1/2 * (Q[i+m-1] - Q[i-1]) * ((T[i+m-1]-M_T[i]) + (T[i-1] - M_T[i-1]))

    return mcov_next


def mcov_sliding_new(Q, T, m):
    
    if len(Q) != len(T):
        msg = 'The length of two arrays should be the same'
        raise ValueError(msg)
    M_T, _ = core.compute_mean_std(T, m)
    μ_Q, _ = core.compute_mean_std(Q, m)

    l = len(T) - m + 1
    out = np.empty(l, dtype=np.float64)
    out[0] = _mcov_precise(Q, T, M_T, μ_Q, m, 0)
    for i in range(1, l):
        out[i] = _mcov_sliding_new(
            out[i-1], 
            Q, 
            T, 
            M_T, 
            μ_Q, 
            m,
            i,
        )


    return out
```

## Results

**code to plot**
```
np.random.seed(0)
T = np.random.rand(2 ** 20)
Q = np.random.rand(2 ** 20) 

def plot_figure(Q, T, scale=1):
    Q = Q * scale
    T = T * scale
    
    fig, axs = plt.subplots(2, 2, figsize=(15, 4))
    for i, m in enumerate([10, 1000]):
        # i: figure column
        ref = mcov_precise(Q, T, m)[1:]
        comp_original = mcov_sliding_original(Q, T, m)[1:]
        comp_new = mcov_sliding_new(Q, T, m)[1:]
    
        np.testing.assert_allclose(comp_original, ref, atol=1e-7)
        np.testing.assert_allclose(comp_new, ref, atol=1e-7)
    
        # plot
        # absolute error
        axs[0, i].plot(np.abs(comp_original - ref), color='b', label='original')
        axs[0, i].plot(np.abs(comp_new - ref), color='r', label='new')
        axs[0, i].legend()
        axs[0, i].set_title(f'abs error when m={m}')

        # relative error
        axs[1, i].plot(np.abs(comp_original - ref)/np.abs(ref), color='b', label='original')
        axs[1, i].plot(np.abs(comp_new - ref)/np.abs(ref), color='r', label='new')
        axs[1, i].legend()
        axs[1, i].set_title(f'relative error when m={m}')
    
    plt.tight_layout()
    fig.savefig(f'error_{scale}.png')
    plt.show()

    return
```

**plot_figure(Q, T, scale=0.001)**
error_0 001

**plot_figure(Q, T, scale=1.0)**
error_1 0

**plot_figure(Q, T, scale=1000.0)**
error_1000 0

**plot_figure(Q, T, scale=10000.0)**
Note: The assertion fails for one element when using the "new" formula. The assertion part in `plot_figure` was commented out for plotting this figure.

error_10000

seanlaw: Okay, maybe there is nothing to do aside from adding the derivation to the notebook and removing the comments about catastrophic cancellation

NimaSarajpoor: 👍

PR #1181, opened 2026-08-13
Fixed #1175 Migrate tests/test_aamp_motifs.py to npt.assert_allclose

### LINKED ISSUE #1175: Migrate to `npt.assert_allclose`

viknesh-ai:
Fixed #1175

Continuing the file-by-file split from #1178 — this one covers `tests/test_aamp_motifs.py`. Will open the next file's PR once this one merges.

`assert_almost_equal` only checks a fixed absolute tolerance and NumPy's docs recommend `assert_allclose` instead. Eight call sites in this file, three different wrinkles:

- Two calls compare the `left`/`right` argument order backwards relative to the `actual, desired` convention (`right`/`out` is always the stumpy-computed value in this file, `left` is either a hand-derived expected value or the naive helper's output) — flipped both.
- Four calls compare against `aamp_match`/`naive_aamp_match` output, which mixes a float distance with an int index and so comes back `dtype=object` — same issue as `test_aamp.py`, same fix: cast both sides to `float64` before comparing, since the values are always numeric.
- The original `decimal=` values weren't all the same (`4`, `6`, and the implicit default of `7`), so I kept each site's original tolerance rather than flattening everything to `1.5e-07` — translated `decimal=4`/`decimal=6` to `atol=1.5e-04`/`atol=1.5e-06` respectively, and the two unlabeled ones to `atol=1.5e-07` per the usual convention.

Also ran the file across five random seeds (JIT disabled, CUDA simulated, matching `test.sh coverage`) since `test_data` in this file draws from unseeded `rng.RNG` — wanted to make sure the tolerances hold up beyond just the one seed I happened to run with.

**Update:** per review, also renamed every `left`/`right` variable to `ref_`/`cmp_` (naive output = `ref`, stumpy-computed output = `cmp`), and flipped the `npt.assert_array_equal` argument order to match the same `actual, desired` convention as `assert_allclose`.

## To Do List

Standing checklist from @seanlaw for this migration, applies to every file in the split - status below is for `test_aamp_motifs.py`:

- [x] Use `atol=1.5e-07` rather than `atol=1.5*10**-07`
- [ ] Eventually, replace the ugly `1.5*10**-config.STUMPY_TEST_PRECISION` with `config.STUMPY_TEST_PRECISION = 1.5e-07` (in `config.py`) - tracked as a follow-up, not part of this per-file migration
- [x] Use `cmp` instead of `comp` - n/a for this file, it never used `comp` naming to begin with
- [x] Rename naive outputs to `ref_` and stumpy-computed outputs to `cmp_` (applies to `npt.assert_array_equal` too, not just `npt.assert_allclose`) - this file originally used `left`/`right`, all renamed

# Pull Request Checklist

- [x] Created a new branch

### DIFF
--- a/tests/test_aamp_motifs.py
+++ b/tests/test_aamp_motifs.py
@@ -51,12 +51,12 @@ def test_aamp_motifs_one_motif():
     m = 3
     max_motifs = 1
 
-    left_indices = [[0, 5]]
-    left_profile_values = [[0.0, 0.0]]
+    ref_indices = [[0, 5]]
+    ref_profile_values = [[0.0, 0.0]]
 
     for p in [1.0, 2.0, 3.0]:
         mp = naive.aamp(T, m, p=p)
-        right_distance_values, right_indices = aamp_motifs(
+        cmp_distance_values, cmp_indices = aamp_motifs(
             T,
             mp[:, 0],
             max_motifs=max_motifs,
@@ -65,8 +65,8 @@ def test_aamp_motifs_one_motif():
             p=p,
         )
 
-        npt.assert_array_equal(left_indices, right_indices)
-        npt.assert_allclose(right_distance_values, left_profile_values, atol=1.5e-04)
+        npt.assert_array_equal(cmp_indices, ref_indices)
+        npt.assert_allclose(cmp_distance_values, ref_profile_values, atol=1.5e-04)
 
 
 def test_aamp_motifs_two_motifs():
@@ -105,8 +105,8 @@ def test_aamp_motifs_two_motifs():
 
         mp = naive.aamp(T, m)
 
-        # left_indices = [[70, 170], [10, 210]]
-        left_profile_values = [
+        # ref_indices = [[70, 170], [10, 210]]
+        ref_profile_values = [
             [0.0, 0.0],
             [
                 0.0,
@@ -114,7 +114,7 @@ def test_aamp_motifs_two_motifs():
             ],
         ]
 
-        right_distance_values, right_indices = aamp_motifs(
+        cmp_distance_values, cmp_indices = aamp_motifs(
             T,
             mp[:, 0],
             max_motifs=max_motifs,
@@ -124,7 +124,7 @@ def test_aamp_motifs_two_motifs():
 
         # We ignore indices because of sorting ambiguities for equal distances.
         # As long as the distances are correct, the indices will be too.
-        npt.assert_allclose(right_distance_values, left_profile_values, atol=1.5e-06)
+        npt.assert_allclose(cmp_distance_values, ref_profile_values, atol=1.5e-06)
 
 
 def test_aamp_naive_match_exact():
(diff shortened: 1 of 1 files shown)


tests/test_aamp_motifs.py
seanlaw: To be consistent, let's also change all "naive" outputs to `ref_` (instead of `right`) and all `stumpy` generated outputs to `cmp_` (instead of `left`).

For example, we would have:
```
ref_indices = [[0, 5]]
ref_profile_values = [[0.0, 0.0]]
```
and
```
cmp_distance_values, cmp_indices = aamp_motifs(
```

Please add this to our To Do list

seanlaw: This (converting to `ref_` and `cmp_` should also apply to `npt.assert_array_equal` as well

seanlaw (changes requested):

tests/test_aamp_motifs.py
viknesh-ai: Sure will do the changes!

viknesh-ai: Done — renamed left/right to ref_/cmp_ throughout, and flipped assert_array_equal to cmp_indices, ref_indices to match the actual, desired convention. 11/11 still passing. Added it to the To Do list in this PR's description; I don't have write access to edit issue #1175 directly so feel free to fold it into the master checklist there if you'd like it tracked at that level too.

seanlaw (commented):
@viknesh-ai Looks good to me. Ready to merge?

PR #1188, opened 2026-08-16
Migrate tests/test_core.py to npt.assert_allclose

viknesh-ai:
Related to #1175

Continuing the file-by-file split — this one covers `tests/test_core.py`. Largest and messiest file in the split so far: 131 call sites, and unlike the other files, the naming here was inconsistent throughout (some pairs `ref`/`comp`, some `ref`/`cmp`, some `ref_X`/`comp_X`, and a few reversed as suffix forms like `P_ref`/`P_comp`, `I_ref`/`I_comp`).

`assert_almost_equal` only checks a fixed absolute tolerance and NumPy's docs recommend `assert_allclose` instead. What changed:

- Normalized every pair to the `ref_X`/`cmp_X` prefix convention, including the reversed suffix forms (`P_ref`/`P_comp` -> `ref_P`/`cmp_P`, `I_ref`/`I_comp` -> `ref_I`/`cmp_I`, `mp_ref` -> `ref_mp`, etc.) and a couple of `comp`-not-`cmp` stragglers. Did a token-level rename (not a blind substring replace) to avoid clobbering unrelated words like `compare`/`compute`/`comparison` in comments.
- All 131 original calls were already in `(ref, comp)` order, so every one needed flipping to `(cmp_side, ref_side)`.
- None of the 131 sites used `decimal=` or referenced `config.STUMPY_TEST_PRECISION` — all used numpy's implicit `decimal=7` default, so all 131 use the literal `atol=1.5e-07`.
- Checked the actual runtime dtype of every distinct `naive.*` source feeding a `ref_` side before touching anything. Only `naive.find_matches`/`core._find_matches` (2 call sites, in `test_find_matches_all`/`test_find_matches_maxmatch`) come back `dtype=object` (a `[distance, index]` pair array, same root cause as the earlier files) — cast both sides to `float64` there. Everywhere else the naive source (`compute_mean_std`, `rolling_nanstd`, `distance_matrix`, `rolling_isconstant`, `get_array_ranges`, `rolling_window_dot_product`, `merge_topk_PI`/`merge_topk_ρI`, and this file's own local `naive_idx_to_mp`/`naive_bfs_indices` helpers) is already clean `float64`/`int64`, so no unnecessary casting was added.
- Also applied the same `ref_`/`cmp_` naming fix to a handful of `npt.assert_array_equal` calls that shared the same inconsistent naming (`test_apply_exclusion_zone*`, `test_get_mask_slices`, `test_gpu_searchsorted`, `test_process_isconstant_*`), per the standing rule that this convention applies to `assert_array_equal` too, not just `assert_allclose`.
- One non-`naive.*` pair (`test_array_to_temp_file`, a value round-tripped through `core.array_to_temp_file` + `np.load`) got the same treatment: the original value is `ref_val`, the round-tripped value is `cmp_val`. Left the unrelated `left`/`right` pairs in `test_jagged_list_to_array*`/`test_replace_distance` untouched since those aren't naive-vs-stumpy comparisons and are out of scope for this migration.
- A few commented-out `# npt.assert_almost_equal(...)` lines got their syntax updated too for consistency, staying commented.

Ran the full file twice locally (normal mode and JIT-disabled/CUDA-simulated, matching `test.sh coverage`) — 244 passed + 1 skipped (GPU test, no CUDA available locally) in normal mode, 245 passed in CI-equivalent mode (the GPU test runs under the CUDA simulator there).

## To Do List

Standing checklist from @seanlaw for this migration, applies to every file in the split - status below is for `test_core.py`:

- [x] Use `atol=1.5e-07` rather than `atol=1.5*10**-07` - applies to all 131 sites here (none referenced `config.STUMPY_TEST_PRECISION`)
- [ ] Eventually, replace the ugly `1.5*10**-config.STUMPY_TEST_PRECISION` with `config.STUMPY_TEST_PRECISION = 1.5e-07` (in `config.py`) - tracked as a follow-up, not part of this per-file migration (n/a for this specific file, but keeping the item open since it's still pending overall)
- [x] Use `cmp` instead of `comp` - this file had several `comp`/`comp_X` stragglers, all renamed
- [x] Rename naive outputs to `ref_` and stumpy-computed outputs to `cmp_` (applies to `npt.assert_array_equal` too) - this file had the most inconsistent naming of the series (including reversed suffix forms like `P_ref`/`P_comp`), fully normalized

# Pull Request Checklist

- [x] Created a new branch

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -772,7 +772,7 @@ def test_apply_exclusion_zone():
 
         naive.replace_inf(ref)
         naive.replace_inf(cmp)
-        npt.assert_array_equal(ref, cmp)
+        npt.assert_array_equal(cmp, ref)
 
 
 def test_apply_exclusion_zone_int():
@@ -790,7 +790,7 @@ def test_apply_exclusion_zone_int():
 
         naive.replace_inf(ref)
         naive.replace_inf(cmp)
-        npt.assert_array_equal(ref, cmp)
+        npt.assert_array_equal(cmp, ref)
 
 
 def test_apply_exclusion_zone_bool():
@@ -808,7 +808,7 @@ def test_apply_exclusion_zone_bool():
 
         naive.replace_inf(ref)
         naive.replace_inf(cmp)
-        npt.assert_array_equal(ref, cmp)
+        npt.assert_array_equal(cmp, ref)
 
 
 def test_apply_exclusion_zone_multidimensional():
@@ -829,7 +829,7 @@ def test_apply_exclusion_zone_multidimensional():
 
         naive.replace_inf(ref)
         naive.replace_inf(cmp)
-        npt.assert_array_equal(ref, cmp)
+        npt.assert_array_equal(cmp, ref)
 
 
 def test_preprocess():
@@ -1065,28 +1065,28 @@ def test_compare_parameters():
 def test_jagged_list_to_array():
     arr = [np.array([0, 1]), np.array([0]), np.array([0, 1, 2, 3])]
 
-    left = np.array([[0, 1, -1, -1], [0, -1, -1, -1], [0, 1, 2, 3]], dtype="int64")
-    right = core._jagged_list_to_array(arr, fill_value=-1, dtype="int64")
-    npt.assert_array_equal(left, right)
+    ref = np.array([[0, 1, -1, -1], [0, -1, -1, -1], [0, 1, 2, 3]], dtype="int64")
+    cmp = core._jagged_list_to_array(arr, fill_value=-1, dtype="int64")
+    npt.assert_array_equal(cmp, ref)
 
-    left = np.array(
+    ref = np.array(
         [[0, 1, np.nan, np.nan], [0, np.nan, np.nan, np.nan], [0, 1, 2, 3]],
         dtype="float64",
     )
-    right = core._jagged_list_to_array(arr, fill_value=np.nan, dtype="float64")
-    npt.assert_array_equal(left, right)
+    cmp = core._jagged_list_to_array(arr, fill_value=np.nan, dtype="float64")
+    npt.assert_array_equal(cmp, ref)
 
 
 def test_jagged_list_to_array_empty():
     arr = []
 
-    left = np.array([[]], dtype="int64")
-    right = core._jagged_list_to_array(arr, fill_value=-1, dtype="int64")
-    npt.assert_array_equal(left, right)
+    ref = np.array([[]], dtype="int64")
+    cmp = core._jagged_list_to_array(arr, fill_value=-1, dtype="int64")
+    npt.assert_array_equal(cmp, ref)
 
-    left = np.array([[]], dtype="float64")
-    right = core._jagged_list_to_array(arr, fill_value=np.nan, dtype="float64")
-    npt.assert_array_equal(left, right)
+    ref = np.array([[]], dtype="float64")
+    cmp = core._jagged_list_to_array(arr, fill_value=np.nan, dtype="float64")
+    npt.assert_array_equal(cmp, ref)
 
 
 def test_get_mask_slices():
@@ -1102,7 +1102,7 @@ def test_get_mask_slices():
     for mask in mask_cases:
         ref_slices = naive._get_mask_slices(mask)
         cmp_slices = core._get_mask_slices(mask)
-        npt.assert_array_equal(ref_slices, cmp_slices)
+        npt.assert_array_equal(cmp_slices, ref_slices)
 
 
 def test_idx_to_mp():
@@ -1532,7 +1532,7 @@ def test_gpu_searchsorted():
             device_A, device_V, device_bfs, nlevel, is_left, device_cmp_IDX
         )
         cmp_IDX = device_cmp_IDX.copy_to_host()
-        npt.assert_array_equal(ref_IDX, cmp_IDX)
+        npt.assert_array_equal(cmp_IDX, ref_IDX)
 
         is_left = False  # test case
         ref_IDX = [np.searchsorted(A[i], V[i], side="right") for i in range(n)]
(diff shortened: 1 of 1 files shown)


tests/test_core.py
```diff
@@ -1768,20 +1772,20 @@ def test_process_isconstant_1d_default():
...
     T[:m] = 0.5  # constant subsequence
     T[-m:] = np.nan  # non-finite subsequence
 
-    T_subseq_isconstant_ref = naive.rolling_isconstant(T, m, a_subseq_isconstant=None)
-    T_subseq_isconstant_comp = core.process_isconstant(T, m, T_subseq_isconstant=None)
+    ref_T_subseq_isconstant = naive.rolling_isconstant(T, m, a_subseq_isconstant=None)
+    cmp_T_subseq_isconstant = core.process_isconstant(T, m, T_subseq_isconstant=None)
 
-    npt.assert_array_equal(T_subseq_isconstant_ref, T_subseq_isconstant_comp)
+    npt.assert_array_equal(ref_T_subseq_isconstant, cmp_T_subseq_isconstant)
```
seanlaw: `npt.assert_allclose` and switch ref/cmp order

tests/test_core.py
```diff
@@ -751,64 +751,64 @@ def test_mass_absolute_distance_matrix(T_A, T_B):
...
-        comp[:] = T[:]
-        core.apply_exclusion_zone(comp, i, exclusion_zone, np.inf)
+        cmp[:] = T[:]
+        core.apply_exclusion_zone(cmp, i, exclusion_zone, np.inf)
 
         naive.replace_inf(ref)
-        naive.replace_inf(comp)
-        npt.assert_array_equal(ref, comp)
+        naive.replace_inf(cmp)
+        npt.assert_array_equal(ref, cmp)
```
seanlaw: switch ref/cmp order

tests/test_core.py
```diff
@@ -751,64 +751,64 @@ def test_mass_absolute_distance_matrix(T_A, T_B):
...
-        comp[:] = T[:]
-        core.apply_exclusion_zone(comp, i, exclusion_zone, -1)
+        cmp[:] = T[:]
+        core.apply_exclusion_zone(cmp, i, exclusion_zone, -1)
 
         naive.replace_inf(ref)
-        naive.replace_inf(comp)
-        npt.assert_array_equal(ref, comp)
+        naive.replace_inf(cmp)
+        npt.assert_array_equal(ref, cmp)
```
seanlaw: switch ref/cmp order

tests/test_core.py
```diff
@@ -751,64 +751,64 @@ def test_mass_absolute_distance_matrix(T_A, T_B):
...
-        comp[:] = T[:]
-        core.apply_exclusion_zone(comp, i, exclusion_zone, False)
+        cmp[:] = T[:]
+        core.apply_exclusion_zone(cmp, i, exclusion_zone, False)
 
         naive.replace_inf(ref)
-        naive.replace_inf(comp)
-        npt.assert_array_equal(ref, comp)
+        naive.replace_inf(cmp)
+        npt.assert_array_equal(ref, cmp)
```
seanlaw: switch ref/cmp order

tests/test_core.py
```diff
@@ -817,19 +817,19 @@ def test_apply_exclusion_zone_multidimensional():
...
-        comp[:, :] = T[:, :]
-        core.apply_exclusion_zone(comp, i, exclusion_zone, np.inf)
+        cmp[:, :] = T[:, :]
+        core.apply_exclusion_zone(cmp, i, exclusion_zone, np.inf)
 
         naive.replace_inf(ref)
-        naive.replace_inf(comp)
-        npt.assert_array_equal(ref, comp)
+        naive.replace_inf(cmp)
+        npt.assert_array_equal(ref, cmp)
```
seanlaw: switch ref/cmp order

tests/test_core.py
```diff
@@ -1101,8 +1101,8 @@ def test_get_mask_slices():
 
     for mask in mask_cases:
         ref_slices = naive._get_mask_slices(mask)
-        comp_slices = core._get_mask_slices(mask)
-        npt.assert_array_equal(ref_slices, comp_slices)
+        cmp_slices = core._get_mask_slices(mask)
+        npt.assert_array_equal(ref_slices, cmp_slices)
```
seanlaw: switch ref/cmp order

tests/test_core.py
```diff
@@ -1522,25 +1526,25 @@ def test_gpu_searchsorted():
...
+        cmp_IDX = np.full(n, -1, dtype=np.int64)
+        device_cmp_IDX = cuda.to_device(cmp_IDX)
         _gpu_searchsorted_kernel[blocks_per_grid, threads_per_block](
-            device_A, device_V, device_bfs, nlevel, is_left, device_comp_IDX
+            device_A, device_V, device_bfs, nlevel, is_left, device_cmp_IDX
         )
-        comp_IDX = device_comp_IDX.copy_to_host()
-        npt.assert_array_equal(ref_IDX, comp_IDX)
+        cmp_IDX = device_cmp_IDX.copy_to_host()
+        npt.assert_array_equal(ref_IDX, cmp_IDX)
```
seanlaw: switch ref/cmp order

seanlaw (changes requested):
@viknesh-ai For `assert_array_equal`, please also ensure that switch `cmp` is first and `ref` is second

tests/test_core.py
viknesh-ai: Done — converted to assert_allclose (bool arrays, atol=1.5e-07) with cmp first.

tests/test_core.py
viknesh-ai: Done.

tests/test_core.py
viknesh-ai: Done.

tests/test_core.py
viknesh-ai: Done.

tests/test_core.py
viknesh-ai: Done.

tests/test_core.py
viknesh-ai: Done.

tests/test_core.py
viknesh-ai: Done — fixed this one and the same pattern a few lines below (is_left=False case) since it was identical.

PR #1206, opened 2026-09-12
Fix precision loss for identical subsequences

### LINKED ISSUE #1160: Investigate Failed Test in test_stump.py
```
STUMPY_SEED=2633420849 NUMBA_DISABLE_JIT=1 NUMBA_ENABLE_CUDASIM=1 pixi run tests custom tests/test_stump.py
```

This test is failing [here](https://github.com/stumpy-dev/stumpy/actions/runs/30075773821/job/89475692489?pr=1159)

It would be good to understand why this is happening.

For some reason the naive way of computing the matrix profile is different from the performant way

iAnjaneySingh:
## Description

Fixes #1160.

When two identical subsequences are compared, the rolling covariance
recurrence can accumulate a small floating-point error. This can produce
a Pearson correlation slightly below 1.0 and therefore a small non-zero
distance for exactly identical subsequences.

This change detects cases where the computed Pearson correlation is
extremely close to 1.0 and the two raw subsequences are exactly equal,
then sets the Pearson correlation to 1.0.

## Validation

- Reproduced the failure from #1160.
- Confirmed the discrepancy came from floating-point error in the
  rolling covariance recurrence.
- `tests/test_stump.py`: 29 passed.
- Full test suite: passed.
- Test coverage: 100%.

# Pull Request Checklist

- [ ] Left a meaningful comment on the original Github issue to discuss the detailed approach for your contribution and received confirmation from the maintainers before proceeding with this pull request
- [x] Created a new branch
- [x] Check this box if AI code assistance was used to generate
  - [x] less than 25% of the code in this pull request
  - [ ] 25-50% of the code in this pull request
  - [ ] more than 50% of the code in this pull request


stumpy/stump.py
```diff
@@ -206,6 +206,11 @@ def _compute_diagonal(
                     pearson = cov * Σ_T_inverse[uint64_j] * σ_Q_inverse[uint64_i]
                     pearson = min(1.0, pearson)
 
+                    if pearson > 1.0 - 1e-8 and np.array_equal(
+                        T_A[uint64_i : uint64_i + uint64_m],
```
seanlaw, line 210: The `_compute_diagonal` function is the most important function in our entire package and needs to be highly performant (i.e., without bloat or unneeded code).

Adding `np.array_equal` is bad because:

1. it would dramatically slow down the computation, say, for a long time series with mostly zero values
2. This does not take into account time series with identical subsequences that are only different by a constant vertical shift

Additionally, what's missing is that you should clearly demonstrate that the originally failing test is passing after this fix is applied. That is the most important criteria (in addition to the regular test suite passing)

seanlaw (changes requested):
@iAnjaneySingh I've left a comment for you to address and consider

stumpy/stump.py
iAnjaneySingh: Thanks for the detailed feedback, Sean. That makes sense, especially since _compute_diagonal is performance-critical. I’ll remove the np.array_equal check and look into a more efficient numerical fix that also handles subsequences that differ by a constant shift. I will also make sure to clearly show the original failing test passing after the fix, along with the full test suite results.

seanlaw: @iAnjaneySingh Another approach that we had discussed before was adding a (optional) post-processing step. So, after the matrix profile is computed, could we take the subsequence pairs that have:

1. a distance/pearson value below a certain threshold
2. (nearly) identical standard deviations
3. select a handful of data points (e.g., 4 data points) from each pair of subsequences and see if their pairwise differences are the same
4. Maybe some other fast criteria that leverages numbers that we've already computed

If so, then we recompute the dot product for these two subsequences.

